PR #4, opened 2018-11-01
tls: fix opaque struct definitions

RReverser:
Using uninhabited enums for opaque types is actually a bad idea, since compiler can rely on them being, well, uninhabited, and it's easy to run into UB.

There are various other approaches to represent them, and I chose to wrap newtype around `c_void`.


ghedo:
Thanks.

PR #7, opened 2019-01-18
ffi: export setter functions for config params

LPardue:
There were transport parameters that were not able to be set :(

This fixes that

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -102,6 +102,11 @@ pub extern fn quiche_config_load_priv_key_from_pem_file(config: &mut Config,
 }
 
 #[no_mangle]
+pub extern fn quiche_config_verify_peer(config: &mut Config, v: bool) {
+    config.verify_peer(v);
+}
+
+#[no_mangle]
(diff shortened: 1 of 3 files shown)


src/ffi.rs
```diff
@@ -131,6 +136,26 @@ pub extern fn quiche_config_set_initial_max_streams_bidi(config: &mut Config, v:
     config.set_initial_max_streams_bidi(v);
 }
 
+#[no_mangle]
+pub extern fn quiche_config_set_initial_max_streams_uni(config: &mut Config, v: u64) {
+    config.set_initial_max_streams_uni(v);
+}
+
+#[no_mangle]
+pub extern fn set_ack_delay_exponent(config: &mut Config, v: u64) {
```
RReverser: Shouldn't these be prefixed with `quiche_config_` for consistency as well?

LPardue: Yes thanks. Fixed in later pushes

ghedo:
:tada:

PR #12, opened 2019-01-23
stream: stream_recv writes data to a provided buffer

nickfajones:
The buffer is provided from 'c' as a pointer to an array of uint8_t and
a length thereof.

In addition, from 'c', a reference to a bool is provided so that the fin
status of the read can be returned to the caller.

The return value for the 'c' function is the number of bytes written
into the provided buffer.

The rust interface is similar, in that it also provides a buffer, in
the form of a mutable slice of 'u8' type.  The slice self reports its
own size.

The number of bytes written to the slice, and the fin status of the
read is returned as a tuple.

The important fix is in the RecvBuf class, where the pop method now
takes a slice as a parameter and copies queued pages of data into the
slice.

The existing RangeBuf interface is removed from RecvBuf::pop and
Stream::recv_pop and as a result, RecvBuf no longer needs to be
exposed through the ffi interface.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -180,16 +180,20 @@ fn main() {
 
         let streams: Vec<u64> = conn.readable().collect();
         for s in streams {
-            let data = conn.stream_recv(s, std::usize::MAX).unwrap();
+            while let Ok((read, fin)) = conn.stream_recv(s, &mut buf) {
+                debug!("{} received {} bytes", conn.trace_id(), read);
 
-            debug!("{} stream {} has {} bytes (fin? {})",
-                   conn.trace_id(), s, data.len(), data.fin());
+                let stream_buf = &buf[..read];
 
-            print!("{}", unsafe { std::str::from_utf8_unchecked(&data) });
+                debug!("{} stream {} has {} bytes (fin? {})",
+                       conn.trace_id(), s, stream_buf.len(), fin);
 
-            if s == HTTP_REQ_STREAM_ID && data.fin() {
-                info!("{} response received, closing..,", conn.trace_id());
-                conn.close(true, 0x00, b"kthxbye").unwrap();
+                print!("{}", unsafe { std::str::from_utf8_unchecked(&stream_buf) });
+
+                if s == HTTP_REQ_STREAM_ID && fin {
+                    info!("{} response received, closing..,", conn.trace_id());
+                    conn.close(true, 0x00, b"kthxbye").unwrap();
+                }
             }
         }
 
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -123,9 +123,9 @@ fn main() {
 
             debug!("got {} bytes", len);
 
-            let buf = &mut buf[..len];
+            let pkt_buf = &mut buf[..len];
 
-            let hdr = match quiche::Header::from_slice(buf, LOCAL_CONN_ID_LEN) {
+            let hdr = match quiche::Header::from_slice(pkt_buf, LOCAL_CONN_ID_LEN) {
                 Ok(v) => v,
 
                 Err(e) => {
@@ -198,7 +198,7 @@ fn main() {
             };
 
             // Process potentially coalesced packets.
-            let read = match conn.recv(buf) {
+            let read = match conn.recv(pkt_buf) {
                 Ok(v)  => v,
 
                 Err(quiche::Error::Done) => {
@@ -217,7 +217,16 @@ fn main() {
 
             let streams: Vec<u64> = conn.readable().collect();
             for s in streams {
-                handle_stream(conn, s, args.get_str("--root"));
+                while let Ok((read, fin)) = conn.stream_recv(s, &mut buf) {
+                    debug!("{} received {} bytes", conn.trace_id(), read);
+
+                    let stream_buf = &buf[..read];
+
+                    debug!("{} stream {} has {} bytes (fin? {})",
+                           conn.trace_id(), s, stream_buf.len(), fin);
+
+                    handle_stream(conn, s, stream_buf, args.get_str("--root"));
+                }
             }
         }
 
@@ -258,21 +267,9 @@ fn main() {
     }
 }
 
-fn handle_stream(conn: &mut quiche::Connection, stream: u64, root: &str) {
-    let stream_data = match conn.stream_recv(stream, std::usize::MAX) {
-        Ok(v) => v,
-
-        Err(quiche::Error::Done) => return,
-
-        Err(e) => panic!("{} stream recv failed {:?}",
-                         conn.trace_id(), e),
-    };
-
-    info!("{} stream {} has {} bytes (fin? {})", conn.trace_id(),
-          stream, stream_data.len(), stream_data.fin());
-
-    if stream_data.len() > 4 && &stream_data[..4] == b"GET " {
-        let uri = &stream_data[4..stream_data.len()];
+fn handle_stream(conn: &mut quiche::Connection, stream: u64, buf: &[u8], root: &str) {
+    if buf.len() > 4 && &buf[..4] == b"GET " {
+        let uri = &buf[4..buf.len()];
         let uri = String::from_utf8(uri.to_vec()).unwrap();
         let uri = String::from(uri.lines().next().unwrap());
         let uri = std::path::Path::new(&uri);
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -356,12 +356,20 @@ pub extern fn quiche_conn_send(conn: &mut Connection, out: *mut u8,
 
 #[no_mangle]
 pub extern fn quiche_conn_stream_recv(conn: &mut Connection, stream_id: u64,
-                                      max_len: usize) -> *const RangeBuf {
-    match conn.stream_recv(stream_id, max_len) {
-        Ok(b) => Box::into_raw(Box::new(b)),
+                                      out: *mut u8, out_len: usize, fin: &mut bool)
+                                                            -> ssize_t {
+    // TODO: limit out_len to MAX_SSIZE to allow the result to fit in ssize_t
+    let out = unsafe { slice::from_raw_parts_mut(out, out_len) };
 
-        Err(_) => ptr::null_mut(),
-    }
+    let (out_len, out_fin) = match conn.stream_recv(stream_id, out) {
+        Ok(v) => v,
+
+        Err(e) => return e.to_c(),
+    };
+
+    *fin = out_fin;
+
+    out_len as ssize_t
 }
 
 #[no_mangle]
@@ -375,26 +383,6 @@ pub extern fn quiche_conn_stream_send(conn: &mut Connection, stream_id: u64,
 
         Err(e) => e.to_c(),
     }
-}
-
-#[no_mangle]
-pub extern fn quiche_rangebuf_data(b: &mut RangeBuf) -> *const u8 {
-    (&b).as_ptr()
-}
-
-#[no_mangle]
-pub extern fn quiche_rangebuf_len(b: &mut RangeBuf) -> usize {
-    b.len()
-}
-
-#[no_mangle]
-pub extern fn quiche_rangebuf_fin(b: &mut RangeBuf) -> bool {
-    b.fin()
-}
-
-#[no_mangle]
-pub extern fn quiche_rangebuf_free(b: *mut RangeBuf) {
-    unsafe { Box::from_raw(b) };
 }
 
 #[no_mangle]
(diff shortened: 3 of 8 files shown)


examples/client.rs
```diff
@@ -179,15 +179,32 @@ fn main() {
...
+                Err(e) => {
+                    error!("{} recv failed: {:?}", conn.trace_id(), e);
+                    conn.close(false, e.to_wire(), b"fail").unwrap();
+                    break 'recv;
+                },
+            };
+
+            debug!("{} received {} bytes", conn.trace_id(), read);
+
+            let rec_buf = &buf[..read];
```
ghedo: Maybe call this "stream_buf"?

examples/server.rs
```diff
@@ -216,8 +216,27 @@ fn main() {
...
+                        break 'recv;
+                    },
+
+                    Err(e) => panic!("{} stream recv failed {:?}",
+                                     conn.trace_id(), e),
+                };
+
+                debug!("{} received {} bytes", conn.trace_id(), read);
+
+                let rec_buf = &buf[..read];
```
ghedo: Same, "stream_buf" seems better.

src/ffi.rs
```diff
@@ -356,12 +356,21 @@ pub extern fn quiche_conn_send(conn: &mut Connection, out: *mut u8,
 
 #[no_mangle]
 pub extern fn quiche_conn_stream_recv(conn: &mut Connection, stream_id: u64,
-                                      max_len: usize) -> *const RangeBuf {
-    match conn.stream_recv(stream_id, max_len) {
-        Ok(b) => Box::into_raw(Box::new(b)),
+                                      out: *mut u8, out_len: usize, fin: *mut bool)
```
ghedo: You can actually make `fin` a reference directly, .i.e.:
```rust
                                       out: *mut u8, out_len: usize, fin: &mut bool)
```
so you don't need the `.as_mut().unwrap()` later.

src/ffi.rs
```diff
@@ -356,12 +356,21 @@ pub extern fn quiche_conn_send(conn: &mut Connection, out: *mut u8,
...
+                                                            -> ssize_t {
+    // TODO: limit out_len to MAX_SSIZE to allow the result to fit in ssize_t
+    let out = unsafe { slice::from_raw_parts_mut(out, out_len) };
+    let fin = unsafe { fin.as_mut().unwrap() };
 
-        Err(_) => ptr::null_mut(),
-    }
+    let (out_len, out_fin) = match conn.stream_recv(stream_id, out) {
+        Ok((out_len, out_fin)) => (out_len, out_fin),
```
ghedo: You don't need to name the tuple values here, can just do:
```rust
         Ok(v) => v,
```

src/lib.rs
```diff
@@ -907,7 +907,10 @@ impl Connection {
                     // Feed crypto data to the TLS state, if there's data
                     // available at the expected offset.
                     if space.crypto_stream.readable() {
-                        let buf = space.crypto_stream.recv_pop(std::usize::MAX)?;
+                        let mut buf =
+                            vec![0; space.crypto_stream.recv_size()];
+                        // TODO: what if fin is true?
```
ghedo: CRYPTO frames don't have a `fin` flag, so there's nothing we need to do (i.e. this comment can be removed).

src/lib.rs
```diff
@@ -1491,11 +1497,11 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        let buf = stream.recv_pop(max_len)?;
+        let (pop_len, fin) = stream.recv_pop(out)?;
```
ghedo: I don't quite like the `pop_len` name, just `len` should be fine.

src/stream.rs
```diff
@@ -72,13 +76,13 @@ impl Stream {
         self.recv.push(buf)
     }
 
-    pub fn recv_pop(&mut self, max_len: usize) -> Result<RangeBuf> {
-        let buf = self.recv.pop(max_len)?;
+    pub fn recv_pop(&mut self, out: &mut [u8]) -> Result<(usize, bool)> {
+        let (pop_len, fin) = self.recv.pop(out)?;
```
ghedo: Same re: `pop_len`, `len` should be fine here as well.

src/stream.rs
```diff
@@ -176,9 +180,10 @@ impl RecvBuf {
         Ok(())
     }
 
-    fn pop(&mut self, max_len: usize) -> Result<RangeBuf> {
-        let mut out = RangeBuf::default();
-        let mut out_len = max_len;
+    fn pop(&mut self, out: &mut [u8]) -> Result<(usize, bool)> {
+        let mut fin = false;
+        let mut pop_len = 0;
+        let mut out_len = out.len();
```
ghedo: This variable was misnamed before, so we can probably just rename both `pop_len` and `out_len` as:

`pop_len` => `len`
`out_len` => `cap`

(so it's clear out_len is actually the capacity of the buffer).

src/stream.rs
```diff
@@ -411,326 +416,363 @@ mod tests {
...
+        let mut recv = RecvBuf::default();
+        assert_eq!(recv.len(), 0);
+
+        let mut buf = vec![0; 32];
+        assert_eq!(buf.len(), 32);
 
-        let read = buf.pop(std::usize::MAX).unwrap();
-        assert_eq!(read.len(), 0);
-        assert_eq!(read.fin(), false);
+        let (pop_len, fin) = recv.pop(buf.as_mut_slice()).unwrap();
```
ghedo: s/pop_len/len/

Same below in the other tests.

src/stream.rs
```diff
@@ -741,8 +783,8 @@ mod tests {
 
         assert_eq!(stream.recv_push(third), Err(Error::FlowControl));
 
-        assert_eq!(stream.recv_pop(std::usize::MAX),
-                   Ok(RangeBuf::from(b"helloworld", 0, false)));
+        let (pop_len, _) = stream.recv_pop(buf.as_mut_slice()).unwrap();
```
ghedo: The previous test checked `fin` as well, should probably keep that check here.

examples/client.rs
```diff
@@ -179,15 +179,32 @@ fn main() {
         }
 
         let streams: Vec<u64> = conn.readable().collect();
-        for s in streams {
-            let data = conn.stream_recv(s, std::usize::MAX).unwrap();
+        'recv: for s in streams {
+            let (read, fin) = match conn.stream_recv(s, &mut buf) {
```
ghedo: I think it would be better to rewrite this whole thing as:
```rust
while let Ok((read, fin)) = conn.stream_recv(s, &mut buf) {
    debug!("{} received {} bytes", conn.trace_id(), read);

    let stream_buf = &buf[..read];
    ....
}
```
This way we read all of the data that is available, even if the output buffer is not big enough (and you get error handling wthout having to handle the different cases), and you wouldn't need to add the `'recv` label to the loop.

examples/server.rs
```diff
@@ -216,8 +216,27 @@ fn main() {
             debug!("{} processed {} bytes", conn.trace_id(), read);
 
             let streams: Vec<u64> = conn.readable().collect();
-            for s in streams {
-                handle_stream(conn, s, args.get_str("--root"));
+            'recv: for s in streams {
+                let (read, fin) = match conn.stream_recv(s, &mut buf) {
```
ghedo: Same re: reading data in a loop.

src/lib.rs
```diff
@@ -907,7 +907,10 @@ impl Connection {
                     // Feed crypto data to the TLS state, if there's data
                     // available at the expected offset.
                     if space.crypto_stream.readable() {
-                        let buf = space.crypto_stream.recv_pop(std::usize::MAX)?;
+                        let mut buf =
+                            vec![0; space.crypto_stream.recv_size()];
```
ghedo: I'm not sure about this. `recv_size()` returns the total amount of data that is in the buffer not what is going to be read, e.g. if there are gaps you may just end up allocating a big buffer and then read a lot less. I think we should use a while loop here as well:

```rust
let mut crypto_buf: [u8; 512] = [0; 512];

while let Ok((read, _)) = space.crypto_stream.recv_pop(&mut crypto_buf) {
    let level = space.crypto_level;

     self.tls_state.provide_data(level, &buf[..read]);
}
```

src/stream.rs
```diff
@@ -411,326 +416,363 @@ mod tests {
 
     #[test]
     fn empty_read() {
-        let mut buf = RecvBuf::default();
-        assert_eq!(buf.len(), 0);
+        let mut recv = RecvBuf::default();
+        assert_eq!(recv.len(), 0);
+
+        let mut buf = vec![0; 32];
+        assert_eq!(buf.len(), 32);
```
ghedo: No need to test that `vec![]` works, we can just assume that the compiler is not broken :)

Same below in the other tests.

nickfajones:
Doh, completely forgot to update the 'c' example server. brb

src/stream.rs
```diff
@@ -411,326 +416,363 @@ mod tests {
...
+        assert_eq!(recv.len(), 4);
+        assert_eq!(recv.off(), 15);
 
-        let read = buf.pop(10).unwrap();
-        assert_eq!(read.off(), 15);
-        assert_eq!(read.len(), 4);
-        assert_eq!(read.fin(), true);
-        assert_eq!(&read[..], b"orld");
-        assert_eq!(buf.len(), 0);
+        let (pop_len, fin) = recv.pop(&mut buf[..4]).unwrap();
```
ghedo: This should be `&mut buf[..10]` as in the previous test.

src/lib.rs
nickfajones: Yeah, I had a good look at this too.  Would be nice if we could detect the different continguous blocks and gaps in the RecvBuf.  Maybe some thing in the Hashmap ordering functions so it can be detected during insertion

src/stream.rs
nickfajones: I think it's ok, there should only be 4 bytes left in the stream on the last read?

examples/server.rs
ghedo: I don't think we need to. Right now `stream_recv()` only returns `Done` as error. In the future it should return `InvalidStreamState` if the application tries to read from a local uni stream, but that can be prevented on the application side as well.

It's probably best to avoid the panic on error anyway.

src/stream.rs
ghedo: Well, the test was checking that trying to read more bytes than there are in the `RecvBuf` works, so that seems like a useful test to keep.

src/stream.rs
```diff
@@ -417,326 +422,357 @@ mod tests {
...
-        assert_eq!(buf.len(), 0);
+        let mut recv = RecvBuf::default();
+        assert_eq!(recv.len(), 0);
+
+        let mut buf = vec![0; 32];
 
-        let read = buf.pop(std::usize::MAX).unwrap();
-        assert_eq!(read.len(), 0);
-        assert_eq!(read.fin(), false);
+        let (len, fin) = recv.pop(buf.as_mut_slice()).unwrap_or((0, false));
```
ghedo: To check that `Done` is returned you can also just do:
```rust
assert_eq!(recv.pop(buf.as_mut_slice()), Err(Error::Done));
```
Same below in a few other tests.

src/lib.rs
```diff
@@ -913,11 +913,15 @@ impl Connection {
 
                     // Feed crypto data to the TLS state, if there's data
                     // available at the expected offset.
-                    if space.crypto_stream.readable() {
-                        let buf = space.crypto_stream.recv_pop(std::usize::MAX)?;
-                        let level = space.crypto_level;
+                    let mut crypto_buf: [u8; 512] = [0; 512];
 
-                        self.tls_state.provide_data(level, &buf)
+                    let level = space.crypto_level;
```
nickfajones: Just a note on this change: I had trouble using the `while let` semantics and checking the `readable` state of the stream.  Basically, this doesn't work:
```
while space.crypto_stream.readable() &&
         let Ok((read, _)) = space.crypto_stream.recv_pop(&mut crypto_buf) {
...
```
But the compiler won't let me as the `let` stops me using other boolean operations in the `while`, very rude!

So what I did was add a `return Err(Error::Done)` in `RecvBuf::pop`, and this is something that wasn't done before.

Given that the whole `recv_pop` -> `RecvBuf::pop` stack was already equipped to return `Result` types, I felt this was a good way to deal with this `while let` problem.

ghedo: Yeah, that seems like a good change. The documentation said that we would return `Done` if there was nothing to read, but that was kind of a lie, so this is better now.

ghedo: Actually nvm, the public `stream_recv()` checks that `readable()` is true, so the documentation was accurate. Still, this is a good change for internal use.

PR #15, opened 2019-01-31
Reuse constant

RReverser:
As far as I can tell, these are actually the same constant and are supposed to always stay in sync.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -184,7 +184,7 @@ use std::collections::HashMap;
 pub const VERSION_DRAFT17: u32 = 0xff00_0011;
 
 /// The maximum length of a connection ID.
-pub const MAX_CONN_ID_LEN: usize = 18;
+pub const MAX_CONN_ID_LEN: usize = crate::packet::MAX_CID_LEN as usize;
 
 const CLIENT_INITIAL_MIN_LEN: usize = 1200;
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -184,7 +184,7 @@ use std::collections::HashMap;
 pub const VERSION_DRAFT17: u32 = 0xff00_0011;
 
 /// The maximum length of a connection ID.
-pub const MAX_CONN_ID_LEN: usize = 18;
+pub const MAX_CONN_ID_LEN: u8 = 18;
```
ghedo: The public const should remain `usize` though, so it can be used for sizing arrays without having to convert it.

src/packet.rs
```diff
@@ -37,15 +37,15 @@ use crate::ranges;
 use crate::recovery;
 use crate::stream;
 
+use crate::MAX_CONN_ID_LEN as MAX_CID_LEN;
```
ghedo: Since this is only used in 3 places, it's probably better to just inline `crate::MAX_CONN_ID_LEN` in those places.

RReverser: Yeah but we still need a separate constant with `u8` type to pass typecheck. I'll adjust code a bit.

RReverser:
Updated PR to go the other way around.

ghedo:
Please also rebase and do `cargo +nightly fmt` (don't know if it will change anything, but might as well check).

RReverser:
Done.

ghedo:
Thanks!

PR #18, opened 2019-02-04
Add Cargo cache to Travis

RReverser:
This allows to avoid rebuilding all deps on each CI run, thus significantly speeding up compilations.


ghedo:
Thanks!

PR #19, opened 2019-02-04
examples: clients and servers upgraded to http1.1

nickfajones:


ghedo:
Yeah, we need HTTP/0.9 to interop with the other implementations (some of which are not going to implement HTTP/3). When I looked into this there wasn't any readily available HTTP implementation that supported both HTTP/1.* and 0.9.

Though some of the changes in this PR (e.g. proper options parsing for C examples) look good, so we can probably review and merge them separately.

nickfajones:
Added flags to clients to set HTTP/0.9 request format.  Servers now support both formats.

Seems that I am failing the formatting check.  Should I apply the suggestions of the rust formatter or is there a pragma thingy I should use to disable it?

ghedo:
@nickfajones you need to run `cargo fmt` with nightly cargo and commit the changes.

examples/client.c
```diff
@@ -247,6 +358,9 @@ int main(int argc, char *argv[]) {
     quiche_config_set_initial_max_streams_bidi(config, 100);
     quiche_config_set_initial_max_streams_uni(config, 100);
     quiche_config_set_disable_migration(config, true);
+    if (no_verify) {
+        quiche_config_verify_peer(config, false);
```
ghedo: Could write this as `quiche_config_verify_peer(config, !no_verify);`

examples/client.rs
```diff
@@ -41,7 +41,8 @@ const USAGE: &str = "Usage:
   client -h | --help
 
 Options:
-  --http1                 Send HTTP/1.1 request instead of HTTP/0.9.
+  -h --help               Show this screen.
```
ghedo: This option is already present.

examples/server.c
```diff
@@ -389,9 +552,90 @@ static void timeout_cb(EV_P_ ev_timer *w, int revents) {
...
+            return 1;
+        }
+    }
+
+    if (argc == 0) {
+        usage(argv[0]);
+        return 1;
+    }
+
+    (void)name;
```
ghedo: Why have the `--name` option if it's not used?

examples/server.c
```diff
@@ -429,18 +674,23 @@ int main(int argc, char *argv[]) {
         return -1;
     }
 
-    quiche_config_load_cert_chain_from_pem_file(config, "examples/cert.crt");
-    quiche_config_load_priv_key_from_pem_file(config, "examples/cert.key");
+    quiche_config_load_cert_chain_from_pem_file(config,
+                                                cert?cert:"examples/cert.crt");
```
ghedo: Missing spaces around ?.

examples/server.rs
```diff
@@ -280,43 +280,128 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
+            path.push(v)
         }
+    }
+
+    let data = std::fs::read_to_string(path);
+    if data.is_err() {
+        handle_error(conn, stream, 404, "Not Found", use_http09);
+        return;
+    }
+    let data = data.unwrap();
```
ghedo: This whole thing is better written as:
```rust
let data = match std::fs::read_to_string(path) {
    Ok(data) => data,

    Err(_) => {
         handle_error(conn, stream, 404, "Not Found", use_http09);
        return;
    },
}
```

examples/client.rs
```diff
@@ -83,7 +84,11 @@ fn main() -> Result<(), Box<std::error::Error>> {
 
     config.verify_peer(true);
 
-    config.set_application_protos(&[b"hq-17", b"http/0.9"])?;
+    if args.get_bool("--http09") {
+        config.set_application_protos(&[b"hq-17", b"http/0.9"])?;
```
LPardue:
```suggestion
        config.set_application_protos(&[b"hq-18", b"http/0.9"])?;
```

examples/client.rs
```diff
@@ -83,7 +84,11 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
+    } else {
+        config.set_application_protos(&[b"hq-17", b"http/1.1"])?;
```
LPardue:
```suggestion
        config.set_application_protos(&[b"hq-18", b"http/1.1"])?;
```

ghedo:
Now that we have HTTP/3 support we don't need HTTP/1.1 anymore (we do still need HTTP/0.9 to interop with other implementations though), and I've removed the only remaining HTTP/1.1 support we had in 3b7f5427c1277e9634a298191f21c612b13abcee. Also see #29 .

This said, some of the changes in this PR are good and can probably be salvaged (e.g. the option parsing for the C examples), but that needs to go in a separate PR.

In any case thanks for looking at this.

PR #22, opened 2019-02-06
Initial implementation of HTTP/3

LPardue:


src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+        }
+    }
+
+    #[test]
+    fn headers() {
+        let mut d: [u8; 128] = [42; 128];
+
+        // TODO test QPACK'd headers
+        let frame = H3Frame::Headers {
+            header_block: vec![71, 69, 84, 32, 47],
```
RReverser: Nit: for ASCII constants it's better to use a bytestring literal for readability. E.g. here it would be `b"GET /".to_vec()`.

LPardue: You've drawn my attention to a TODO that really needs addressing. `header_block` should contain a set of QPACK instructions but these tests were written before we had a QPACK codec. 

Right now, the tests create HEADERS and PUSH_PROMISE frames containing garbage values (as far as a strict protocol check would determine), which just so happen to also be an ASCII string that has a striking resemblance to HTTP/1.x syntax. 

I'm inclined to say that the unit test should not use QPACK nor something that can be interpreted as an ASCII string. Thoughts @ghedo?

src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+
+fn parse_settings_frame(
+    payload_length: u64, b: &mut octets::Octets,
+) -> Result<H3Frame> {
+    let mut num_placeholders = None;
+    let mut max_header_list_size = None;
+    let mut qpack_max_table_capacity = None;
+    let mut qpack_blocked_streams = None;
+
+    while b.off() < payload_length as usize {
```
ghedo: This only really works if the SETTINGS frame is at the start of the buffer. Though it would be fixed for free if you do the `b.get_bytes()` trick I suggested above.

src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+
+        assert_eq!(wire_len, 5);
+
+        {
+            let mut b = octets::Octets::with_slice(&mut d);
+            assert_eq!(H3Frame::from_bytes(&mut b).unwrap(), frame);
+        }
+    }
+
+    fn priority_some_stream_to_placeholder() {
```
ghedo: Is this a test? It's missing `#[test]`.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
 
 //! HTTP/3 client and server.
 
+pub mod frame;
```
ghedo: Does this need to be public?

Also, mod declarations should go at the end of the file (see e.g. `src/lib.rs`).

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
 pub mod qpack;
+
+use super::stream;
+use super::Result;
+use crate::octets;
+
+// extern crate http;
```
ghedo: Remove.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+};
+
+const H3_CONTROL_STREAM_TYPE_ID: u8 = 0x43;
+const H3_PUSH_STREAM_TYPE_ID: u8 = 0x50;
+const QPACK_ENCODER_STREAM_TYPE_ID: u8 = 0x48;
+const QPACK_DECODER_STREAM_TYPE_ID: u8 = 0x68;
+
+/// An HTTP/3  error.
+#[derive(Clone, Debug, PartialEq)]
+pub enum H3Error {
```
ghedo: Just `Error`.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+    // There is no error, just stream or connection close
```
ghedo: Nit: comments should end with a dot.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+
+    qpack_encoder_stream_open: bool,
+    peer_qpack_encoder_stream_open: bool,
+    qpack_decoder_stream_open: bool,
+    peer_qpack_decoder_stream_open: bool,
+}
+
+struct H3Util {}
+impl H3Util {
+    pub fn req_hdrs_to_qpack(
```
ghedo: Why are these function part of the `H3Util` struct instead of just normal functions? Also, do they need to be public?

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+        ));
+
+        for (key, value) in request.headers().iter() {
+            headers
+                .push(qpack::Header::new(key.as_str(), value.to_str().unwrap()));
+        }
+
+        let len = encoder.encode(&headers, &mut vec);
+
+        vec.truncate(len.unwrap());
```
ghedo: Must not use `unwrap()` as part of library code unless you are sure it's not going to crash (this applies to many other cases as well).

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+        resp
+    }
+}
+
+impl H3Connection {
+    #[allow(clippy::new_ret_no_self)]
+    fn new(config: &mut H3Config, is_server: bool) -> Result<Box<H3Connection>> {
+        let root = String::clone(&config.root_dir); // TODO shouldn't need to clone here
+
+        Ok(Box::new(H3Connection {
```
ghedo: We don't need to box the h3 connection (we do this for the transport connection because it's required in that specific case).

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+        *resp.status_mut() = StatusCode::from_bytes(status.as_bytes()).unwrap();
+        *resp.version_mut() = http::Version::HTTP_2;
+
+        // debug!("Prepared response {:?}", resp);
+
+        resp
+    }
+}
+
+impl H3Connection {
```
ghedo: Just `Connection`.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+    pub fn is_established(&mut self) -> bool {
+        self.control_stream_open &&
+            self.qpack_encoder_stream_open &&
+            self.qpack_decoder_stream_open
+    }
+
+    pub fn open_control_stream(&mut self, quic_conn: &mut super::Connection) {
+        if !self.control_stream_open {
+            let mut d = [42; 128];
+            let mut b = octets::Octets::with_slice(&mut d);
```
ghedo: There's no need to use octets here, you can just turn `H3_CONTROL_STREAM_TYPE_ID` into a slice use `to_be_bytes()` method https://doc.rust-lang.org/std/primitive.u8.html#method.to_be_bytes

ghedo: Same below.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+            debug!("Amount of priority bytes to send is {}", off);
+            let stream_id = self.get_control_stream_id();
+
+            quic_conn
+                .stream_send(stream_id, &mut d[..off], false)
+                .unwrap();
+        }
+    }
+
+    // Send SETTINGS frame based on H3 config
```
ghedo: Public documentation needs 3 `/`.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+        quic_conn
+            .stream_send(stream_id, &mut d[..off], false)
+            .unwrap();
+    }
+
+    // Send a request
+    pub fn send_request(
+        &mut self, request: http::Request<()>, quic_conn: &mut super::Connection,
+    ) {
+        let mut d = [42; 65535];
```
ghedo: We should find a better way to do this....

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+            fin
+        );
+
+        // H3 unidirectional streams have types as first byte
+        if !stream::is_bidi(stream_id) {
+            let mut o = octets::Octets::with_slice(&mut stream_buf);
+            while o.off() < stream_buf_len {
+                // trace!("loop de loop");
+                // dbg!(&stream_buf);
+                if o.off() == 0 {
```
ghedo: This has the same problem of needing to be at the start of the buffer. But the stream type might be read first, then another stream read is done because data arrives later, and you end up parsing a frame as stream type.

src/h3/mod.rs
```diff
@@ -26,4 +26,830 @@
...
+        info!(
+            "{} stream {} has {} bytes (fin? {})",
+            quic_conn.trace_id(),
+            stream_id,
+            stream_buf_len,
+            fin
+        );
+
+        // H3 unidirectional streams have types as first byte
+        if !stream::is_bidi(stream_id) {
```
ghedo: This massive is pretty ugly. Maybe split into functions and/or return early in the various cases, so you don't need indentation.

src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+
+use crate::octets;
+
+use std::mem;
+
+const ELEM_DEPENDENCY_TYPE_MASK: u8 = 0x30;
+
+/// H3 Prioritized Element type.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum PrioritizedElemType {
```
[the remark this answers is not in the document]
LPardue: It gets used in h3 mod.rs:create_placeholder_tree(), where it creates priorities with well-defined types. Some other way to achieve that?

src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+            0x02 => PrioritizedElemType::Placeholder,
+            0x03 => PrioritizedElemType::CurrentStream,
+            _ => PrioritizedElemType::Error,
+        }
+    }
+}
+
+/// H3 Element Dependency type.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ElemDependencyType {
```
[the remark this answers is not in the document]
LPardue: Same answer

src/h3/frame.rs
```diff
@@ -0,0 +1,856 @@
...
+    DuplicatePush {
+        push_id: u64,
+    },
+}
+
+impl H3Frame {
+    pub fn from_bytes(b: &mut octets::Octets) -> Result<H3Frame> {
+        let payload_length = b.get_varint()?;
+        let frame_type = b.get_u8()?;
+
```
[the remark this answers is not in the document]
LPardue: The frame parsing needs to be refactored to accomodate the fact that we might not have read enough stream data from the QUIC transport, and we have a partial stream.

Some with most other parsing, so I'm tempted to address that in one separate commit.

src/h3/mod.rs
LPardue: Yup, need a full sweep of error handling approach

LPardue:
Addressed most comments. Next up will be the meatier issues

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
+// Copyright (C) 2019, Cloudflare, Inc.
```
ghedo: This should be called `http3-client.rs`.

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    if std::env::var_os("SSLKEYLOGFILE").is_some() {
+        quiche_config.log_keys();
+    }
+
+    let mut quic_conn =
+        quiche::connect(Some(uri_host), &scid, &mut quiche_config).unwrap();
+
+    let mut h3_config = quiche::h3::Config::new().unwrap();
+    let mut h3conn = quiche::h3::connect(&mut h3_config).unwrap();
```
ghedo: Can we create this once the transport handshake succeeded?

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+            break;
+        }
+
+        if quic_conn.is_established() && !req_sent {
+            debug!(
+                "{} QUIC handshake completed, now trying HTTP/3",
+                quic_conn.trace_id()
+            );
+
+            if b"h3-17" != quic_conn.application_proto() {
```
ghedo: We usually do the inverse check <fuction> != <static value>.

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+                // TODO a better error code?
+                quic_conn
+                    .close(false, 0x0, b"I don't support your ALPNs")
+                    .unwrap();
+                break;
+            }
+
+            // TODO make opening control streams saner
+            if !h3conn.is_established() {
+                h3conn.send_settings(&mut *quic_conn);
```
ghedo: `&mut *quic_conn` wat? That shouldn't be needed...

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+                .version(http::Version::HTTP_2)
+                .header("User-Agent", "quiche-http/3")
+                .body(())
+                .unwrap();
+
+            info!("Sending HTTP request {:?}", req);
+
+            if let Err(e) = h3conn.send_request(&mut *quic_conn, req, false) {
+                error!("{} stream send failed {:?}", quic_conn.trace_id(), e);
+            } else {
```
ghedo: If you are handling both `Ok()` and `Err()`, you should just use `match`.

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+                req_sent = true;
+            }
+        }
+
+        let streams: Vec<u64> = quic_conn.readable().collect();
+        for s in streams {
+            info!("{} stream id {} is readable", quic_conn.trace_id(), s);
+            let mut h3_frames: Vec<quiche::h3::frame::Frame> = Vec::new();
+            loop {
```
ghedo: I think the per-stream loop should be handled by the library. There's no point in making every application duplicate this loop.

examples/h3client.rs
```diff
@@ -0,0 +1,285 @@
...
+
+                    Err(e) => {
+                        error!("{} handling stream id {} failed: {:?}", quic_conn.trace_id(), s, e);
+                        quic_conn.close(false, e.to_wire(), b"HTTP/3 fail").unwrap();
+                        break;
+                    },
+                };
+            }
+
+            for mut f in h3_frames {
```
ghedo: This should also be done inside the library.

examples/h3server.rs
```diff
@@ -0,0 +1,396 @@
+// Copyright (C) 2019, Cloudflare, Inc.
```
ghedo: `http3-server.rs`

examples/h3server.rs
```diff
@@ -0,0 +1,396 @@
...
+
+Options:
+  --listen <addr>   Listen on the given IP:port [default: 127.0.0.1:4433]
+  --cert <file>     TLS certificate path [default: examples/cert.crt]
+  --key <file>      TLS certificate key path [default: examples/cert.key]
+  --root <dir>      Root directory [default: examples/root/]
+  --name <str>      Name of the server [default: quic.tech]
+  -h --help         Show this screen.
+";
+type QuicH3Conn = (Box<quiche::Connection>, Option<quiche::h3::Connection>);
```
ghedo: Nit: need empty line before this.

ghedo: Also, it's probably better to make this a proper `struct` and name the fields. Reading the code where they are used is pretty confusing.

examples/h3server.rs
```diff
@@ -0,0 +1,396 @@
...
+                    error!("{} recv failed: {:?}", quich3_conn.0.trace_id(), e);
+                    quich3_conn.0.close(false, e.to_wire(), b"fail").unwrap();
+                    break 'read;
+                },
+            };
+
+            debug!("{} processed {} bytes", quich3_conn.0.trace_id(), read);
+
+            if quich3_conn.0.is_established() {
+                if b"h3-17" != quich3_conn.0.application_proto() {
```
ghedo: Same.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use crate::Error;
+use crate::Result;
+
+use crate::octets;
+
+use std::mem;
```
ghedo: Nit: this should go at the top. The order for imports should be:

1. Built-in (`std`, `core`, ...).
2. External crates.
3. `crate::Error`, `crate::Result`.
4. Internal modules.

Ideally each of the groups above should be seaprated by an empty line. This applies to the other files in this PR as well.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+
+const ELEM_DEPENDENCY_TYPE_MASK: u8 = 0x30;
+
+#[allow(dead_code)]
```
ghedo: Are these constants really dead? They seem to be used below.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+        // debug!("GOT FRAME {:x}, payload_len= {}", frame_type, payload_length);
+
+        // TODO: handling of 0-length frames
+        let frame = match frame_type {
+            DATA_FRAME_TYPE_ID => Frame::Data {
+                payload: b.get_bytes(payload_length as usize)?.to_vec(),
+            },
+
+            HEADERS_FRAME_TYPE_ID => Frame::Headers {
+                header_block: b.get_bytes(payload_length as usize)?.to_vec(),
```
ghedo: Why not just decode QPACK here and put the `Vec<Header>` in the frame? `to_vec()` copies the data, so it's better to avoid it if possible.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+
+            MAX_PUSH_FRAME_TYPE_ID => Frame::MaxPushId {
+                push_id: b.get_varint()?,
+            },
+
+            DUPLICATE_PUSH_FRAME_TYPE_ID => Frame::DuplicatePush {
+                push_id: b.get_varint()?,
+            },
+
+            _ => return Err(Error::InvalidFrame),
```
ghedo: Should probably have an HTTP/3-specific error for this.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+        };
+
+        Ok(frame)
+    }
+
+    pub fn from_bytes2(
```
ghedo: Why do we need 2? Just seems like a pointless code duplication.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+
+    pub fn to_bytes(&self, b: &mut octets::Octets) -> Result<usize> {
+        let before = b.cap();
+
+        trace!("Serializing frame type {:?}", self);
+
+        match self {
+            Frame::Data { payload } => {
+                b.put_varint(payload.len() as u64)?;
+                b.put_varint(0x00)?;
```
ghedo: Should probably use the frame ID constants from above since you went to the trouble of defining them :)

src/h3/frame.rs
ghedo: This can just be `while b.cap() > 0 {` now.

src/h3/frame.rs
```diff
@@ -0,0 +1,879 @@
...
+        weight,
+    })
+}
+
+fn parse_push_promise(
+    payload_length: u64, b: &mut octets::Octets,
+) -> Result<Frame> {
+    let push_id = b.get_varint()?;
+    let header_block_length = payload_length - octets::varint_len(push_id) as u64;
+    let header_block = b.get_bytes(header_block_length as usize)?.to_vec();
```
ghedo: Same as headers regarding doing QPACK decode instead of copying data.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+pub struct Config {
+    pub root_dir: String,
+    pub num_placeholders: u64,
+    pub max_header_list_size: u64,
+    pub qpack_max_table_capacity: u64,
+    pub qpack_blocked_streams: u64,
+}
+
+impl Config {
+    #[allow(clippy::new_ret_no_self)]
```
ghedo: This is not needed anymore.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+    pub fn new() -> Result<Config> {
+        Ok(Config {
+            root_dir: String::new(),
```
ghedo: AFAICT this field is never used, can we just remove it? I don't think it's a good idea to have the library handle this anyway.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+    qpack_decoder: qpack::Decoder,
+
+    qpack_encoder_stream_open: bool,
+    peer_qpack_encoder_stream_open: bool,
+    qpack_decoder_stream_open: bool,
+    peer_qpack_decoder_stream_open: bool,
+}
+
+impl Connection {
+    #[allow(clippy::new_ret_no_self)]
```
ghedo: Not needed.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+    fn new(config: &mut Config, is_server: bool) -> Result<Connection> {
+        let root = String::clone(&config.root_dir); // TODO shouldn't need to clone here
+
+        Ok(Connection {
+            root_dir: root,
```
ghedo: Same re: removing this field.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+            is_server,
+            requests: RequestMap::new(),
+            streams: StreamMap::new(),
+            uni_stream_ledger: Vec::new(),
```
ghedo: It doesn't seem like this needs to be a vector, can't it just be the highest stream ID allocated? Also, we might want to add an API to the transport layer to do this (e.g. `open_stream_bidi/uni()` that returns the lowest available stream ID).

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+            if id < &std::u64::MAX {
+                let id = id + 4;
+                Ok(id)
+            } else {
+                Err(Error::LimitExceeded)
+            }
+        }
+    }
+
+    pub fn is_established(&self) -> bool {
```
ghedo: This name seems confusing. "established" to me means that the handshake is complete, but this only checks if the local streams have been opened, can it be renamed to something else? e.g. `is_open()`.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+            frame::Frame::Settings {
+                num_placeholders,
+                max_header_list_size,
+                qpack_max_table_capacity,
+                qpack_blocked_streams,
+            } =>
+                if self.is_server && num_placeholders.is_some() {
+                    error!("SETTINGS frame with placeholders received at server");
+                    return Err(Error::WrongSettingDirection);
+                } else {
```
ghedo: Don't need to have an else branch since the other one returns early.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+                debug!("CANCEL_PUSH frame received but not doing anything.");
+            },
+            frame::Frame::MaxPushId { .. } => {
+                debug!("MAX_PUSH_ID frame received but not doing anything.");
+            },
+            frame::Frame::GoAway { .. } =>
+                if self.is_server {
+                    error!("GOAWAY frame received at server.");
+                    return Err(Error::UnexpectedFrame);
+                } else {
```
ghedo: Same re: else branch.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+                    let varint_bytes =
+                        stream.buf_bytes(varint_len as usize)?;
+                    let varint = varint_bytes[0];
+
+                    let ty = stream::StreamType::deserialize(varint);
+
+                    if ty.is_none() {
+                        return Err(Error::UnknownStreamType);
+                    } else {
+                        info!("Stream id {} is of type {:?}", stream_id, ty.unwrap());
```
ghedo: Doesn't need to be in else.

src/h3/mod.rs
```diff
@@ -26,4 +26,940 @@
...
+}
+
+/// Creates a new server-side connection.
+pub fn accept(config: &mut Config) -> Result<Connection> {
+    let conn = Connection::new(config, true)?;
+
+    Ok(conn)
+}
+
+pub mod frame;
```
ghedo: This shouldn't be public.

src/h3/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+    FramePayloadLen,
+    FramePayload,
+    PushIdLen,
+    PushId,
+    QpackInstruction,
+    Invalid,
+    Done,
+}
+
+impl std::fmt::Debug for StreamState {
```
ghedo: This whole thing seems to be equivalent to just deriving `Debug` for `StreamState` (i.e. can just add `Debug` to the `#[derive()]` list, note that it should be in alphabetical order for style).

src/h3/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+            StreamType::Control => Some(HTTP3_CONTROL_STREAM_TYPE_ID),
+            StreamType::Push => Some(HTTP3_PUSH_STREAM_TYPE_ID),
+            StreamType::QpackEncoder => Some(QPACK_ENCODER_STREAM_TYPE_ID),
+            StreamType::QpackDecoder => Some(QPACK_DECODER_STREAM_TYPE_ID),
+            _ => None,
+        }
+    }
+}
+
+impl std::fmt::Debug for StreamType {
```
ghedo: This also doesn't seem to be adding a whole lot of information compared to the derived `Debug`.

src/h3/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+    buf: Vec<u8>,
+    buf_read_off: u64,
+    buf_end_pos: u64,
+    next_varint_len: usize,
+    frame_payload_len: u64,
+    frame_type: Option<u8>,
+}
+
+impl Stream {
+    #[allow(clippy::new_ret_no_self)]
```
ghedo: Not needed.

src/h3/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+            frame_payload_len: 0,
+            frame_type: None,
+        })
+    }
+
+    pub fn get_stream_type(&self) -> &Option<StreamType> {
+        &self.ty
+    }
+
+    pub fn get_stream_state(&mut self) -> &StreamState {
```
ghedo: `state()`

LPardue:
Ack your comments. The major contention point at this stage I think is
around responsibility for stream and frame checking, and signalling back to
app.

Let's continue to refine that  while doing all the easy fix ups.

src/h3/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+                              * approach to buffer management */
+            buf_read_off: 0,
+            buf_end_pos: 0,
+            next_varint_len: 0,
+            frame_payload_len: 0,
+            frame_type: None,
+        })
+    }
+
+    pub fn get_stream_type(&self) -> &Option<StreamType> {
```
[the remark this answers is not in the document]
LPardue: I tried that first but `type` is a keyword causing rustc to barf. So I used this version that then got propogated. How about `ty`?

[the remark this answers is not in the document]
ghedo: Ah, yeah, I forgot about that. `ty()` is fine.

src/h3/frame.rs
LPardue: There are a few weak arguments against doing that. 

First, so far in `quiche::h3:frame`, the structure of frames closely aligns with the HTTP/3 wire format specification. That states that HEADERS frames carry a Header Block i.e. the compressed representation of a header list. Changing that to be a Vec<Header> adds some complexity into this area of code for it to call QPACK. I think things will get worse when we want to use dynamic compression and need to also generate and emit enocoder and decoder instructions.

The other argument against embedding the compression inside, is that it gives us greater flexibility to change header compression. I'm not convinced we need that flexibility right now.



1) It couples the HTTP/3 wire format serilization and deserialization to one QPACK li

LPardue:
Latest changes address most low hanging fruit and pull stream/frame handling into the lib. Next will be to return events to the app to inform it that things are taking place e.g. Request and Response headers

examples/h3client.rs
LPardue: I've got two new functions that take the burden of this. Right now they are dumb and read all readables. In future, we should be smarter and selective on what streams to read when, possibly based on QPACK and where the thing is a client or server.

examples/h3client.rs
LPardue: Addressed now, leaving this open to prompt review

src/h3/frame.rs
LPardue: All constants are now undead :skull:

src/h3/mod.rs
LPardue: Removed and now used by example http3-server to actually serve files!

src/h3/mod.rs
LPardue: The function now also checks if the peer opened streams and you received/processed them e.g. Control stream SETTINGS. This is not perfect so leaving open for further discussion.

ghedo:
So close to green build :smile: In any case this looks like a huge step forward, nice work :tada:

LPardue:
Yeah the failures were weird, I couldn't reproduce the clippy errors locally even if I switched cargo versions until I figured out I needed `cargo clippy --examples -- -D warnings`. Nightly fails with `cognitive-complexity-threshold` error, which is why I guess it doesn't need to pass..

LPardue:
`error: the function has a cyclomatic complexity of 44` 🥇

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                    Some(x) => x.to_string(),
+                    None => String::from("None"),
+                };
+
+                let edid = match element_dependency_id {
+                    Some(x) => x.to_string(),
+                    None => String::from("None"),
+                };
+
+                write!(f, "PRIORITY priority element type={:?} element dependency type={:?} prioritized element id={:?} element dependency id={:?} weight={:?}", priority_elem, elem_dependency, peid, edid, weight)?;
```
ghedo:
```suggestion
                write!(f, "PRIORITY pri_type={:?} dep_type={:?} pri_id={:?} dep_id={:?} weight={:?}", priority_elem, elem_dependency, peid, edid, weight)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                write!(f, "CANCEL_PUSH push id={}", push_id)?;
+            },
+
+            Frame::Settings {
+                num_placeholders,
+                max_header_list_size,
+                qpack_max_table_capacity,
+                qpack_blocked_streams,
+            } => {
+                write!(f, "SETTINGS num placeholders={:?}, max header list size={:?}, qpack max table capacity={:?}, qpack blocked streams={:?} ", num_placeholders, max_header_list_size, qpack_max_table_capacity, qpack_blocked_streams)?;
```
ghedo:
```suggestion
                write!(f, "SETTINGS placeholders={:?}, max_headers={:?}, qpack_max_table={:?}, qpack_blocked={:?} ", num_placeholders, max_header_list_size, qpack_max_table_capacity, qpack_blocked_streams)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                let edid = match element_dependency_id {
+                    Some(x) => x.to_string(),
+                    None => String::from("None"),
+                };
+
+                write!(f, "PRIORITY priority element type={:?} element dependency type={:?} prioritized element id={:?} element dependency id={:?} weight={:?}", priority_elem, elem_dependency, peid, edid, weight)?;
+            },
+
+            Frame::CancelPush { push_id } => {
+                write!(f, "CANCEL_PUSH push id={}", push_id)?;
```
ghedo:
```suggestion
                write!(f, "CANCEL_PUSH push_id={}", push_id)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                write!(f, "SETTINGS num placeholders={:?}, max header list size={:?}, qpack max table capacity={:?}, qpack blocked streams={:?} ", num_placeholders, max_header_list_size, qpack_max_table_capacity, qpack_blocked_streams)?;
+            },
+
+            Frame::PushPromise {
+                push_id,
+                header_block,
+            } => {
+                write!(
+                    f,
+                    "PUSH_PROMISE push id={} len={}",
```
ghedo:
```suggestion
                    "PUSH_PROMISE push_id={} len={}",
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                    push_id,
+                    header_block.len()
+                )?;
+            },
+
+            Frame::GoAway { stream_id } => {
+                write!(f, "GOAWAY stream id={}", stream_id)?;
```
ghedo:
```suggestion
                write!(f, "GOAWAY stream_id={}", stream_id)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            },
+
+            Frame::MaxPushId { push_id } => {
+                write!(f, "MAX_PUSH_ID push id={}", push_id)?;
```
ghedo:
```suggestion
                write!(f, "MAX_PUSH_ID push_id={}", push_id)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            },
+
+            Frame::DuplicatePush { push_id } => {
+                write!(f, "DUPLICATE_PUSH push id={}", push_id)?;
```
ghedo:
```suggestion
                write!(f, "DUPLICATE_PUSH push_id={}", push_id)?;
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+    let mut qpack_max_table_capacity = None;
+    let mut qpack_blocked_streams = None;
+
+    while b.off() < settings_length {
+        let setting = b.get_u16()?;
+
+        match setting {
+            SETTINGS_QPACK_MAX_TABLE_CAPACITY => {
+                qpack_max_table_capacity = Some(b.get_varint()?);
+            },
```
ghedo:
```suggestion
            },

```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            SETTINGS_MAX_HEADER_LIST_SIZE => {
+                max_header_list_size = Some(b.get_varint()?);
+            },
```
ghedo:
```suggestion
            },

```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            SETTINGS_QPACK_BLOCKED_STREAMS => {
+                qpack_blocked_streams = Some(b.get_varint()?);
+            },
```
ghedo:
```suggestion
            },

```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            SETTINGS_NUM_PLACEHOLDERS => {
+                num_placeholders = Some(b.get_varint()?);
+            },
```
ghedo:
```suggestion
            },

```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+            _ => {
+                // Unknown Settings parameters must be ignored
```
ghedo:
```suggestion
                // Unknown Settings parameters must be ignored.
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+                trace!("Setting parameter type {} not implemented!", setting);
```
ghedo:
```suggestion
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+        element_dependency_id = Some(b.get_varint()?);
+    }
+
+    let weight = b.get_u8()?;
+    Ok(Frame::Priority {
+        priority_elem,
+        elem_dependency,
+        prioritized_element_id,
+        element_dependency_id,
+        weight,
```
ghedo:
```suggestion
        b.get_u8()?,
```

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+
+    if !priority_elem.is_peid_absent() {
+        prioritized_element_id = Some(b.get_varint()?);
+    }
+
+    if !elem_dependency.is_edid_absent() {
+        element_dependency_id = Some(b.get_varint()?);
+    }
+
+    let weight = b.get_u8()?;
```
ghedo:
```suggestion
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+pub struct Config {
+    num_placeholders: u64,
+    max_header_list_size: u64,
+    qpack_max_table_capacity: u64,
+    qpack_blocked_streams: u64,
+}
+
+impl Config {
+    pub fn new(
+        num_placeholders: u64, max_header_list_size: u64,
```
ghedo: If the parameters are passed in the `new()` method, then you don't need all the `set_*()` ones, or vice versa If you want to keep the set_* ones then the fields of the struct should probably just be `Option`s.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+    ) {
+        self.qpack_max_table_capacity = qpack_max_table_capacity;
+    }
+
+    pub fn set_qpacked_blocked_streams(&mut self, qpack_blocked_streams: u64) {
+        self.qpack_blocked_streams = qpack_blocked_streams;
+    }
+}
+
+type StreamMap = BTreeMap<u64, stream::Stream>;
```
ghedo: This type is pretty short and it's only used once, so you can just inline this in the struct definition and then just use `BTreeMap::new()` when initializing it.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+            self.local_qpack_streams.encoder_stream_id.is_some() &&
+            self.local_qpack_streams.decoder_stream_id.is_some() &&
+            self.peer_control_stream_id.is_some() &&
+            self.peer_qpack_streams.decoder_stream_id.is_some() &&
+            self.peer_qpack_streams.encoder_stream_id.is_some()
+    }
+
+    /// Prepare a request in HTTP/3 wire format, allocate a stream ID and send
+    /// it.
+    pub fn send_request_headers(
```
```suggestion
    pub fn send_request(
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+        &mut self, quic_conn: &mut super::Connection, headers: &[qpack::Header],
+        has_body: bool,
```
ghedo: The `has_body` flag should probably be renamed to `fin`, so you can pass it directly to `stream_send()`.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+        let off = b.off();
+
+        trace!(
+            "{} sending request of size {} on stream {}",
+            quic_conn.trace_id(),
+            off,
+            stream_id
+        );
+
+        if let Err(e) = quic_conn.stream_send(stream_id, &d[..off], true) {
```
```suggestion
        quic_conn.stream_send(stream_id, &d[..off], fin)?;
```

and then remove the next 3 lines as well.

src/h3/mod.rs
ghedo: Thinking about this some more, I think that using the `to_bytes()` method for serializing the frames isn't very good for this use-case.

So we should instead have a function that only write the frame header, call `stream_send()` with it, then call `stream_send()` again with encoded QPACK header block.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+            return Err(Error::from(e));
+        }
+
+        Ok(stream_id)
+    }
+
+    /// Prepare a response in HTTP/3 wire format and send it.
+    pub fn send_response(
+        &mut self, quic_conn: &mut super::Connection, stream_id: u64,
+        headers: &[qpack::Header], has_body: bool,
```
ghedo: Same, re: has_body -> fin.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+
+        quic_conn.stream_send(stream_id, &stream_d[..off], !has_body)?;
+
+        Ok(())
+    }
+
+    /// Prepare paload body in HTTP/3 wire format and send it.
+    pub fn send_body(
+        &mut self, quic_conn: &mut super::Connection, stream_id: u64,
+        body: Vec<u8>, final_data: bool,
```
```suggestion
        body: Vec<u8>, fin: bool,
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+
+        let data = frame::Frame::Data { payload: body };
+
+        let mut stream_b = octets::Octets::with_slice(&mut stream_d);
+        data.to_bytes(&mut stream_b).unwrap();
+
+        let off = stream_b.off();
+
+        trace!(
+            "{} sending response DATA frame of size {} on stream {}",
```
```suggestion
            "{} sending DATA frame of size {} on stream {}",
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+            self.peer_qpack_streams.decoder_stream_id.is_some() &&
+            self.peer_qpack_streams.encoder_stream_id.is_some()
+    }
+
+    /// Prepare a request in HTTP/3 wire format, allocate a stream ID and send
+    /// it.
+    pub fn send_request_headers(
+        &mut self, quic_conn: &mut super::Connection, headers: &[qpack::Header],
+        has_body: bool,
+    ) -> Result<(u64)> {
```
```suggestion
    ) -> Result<u64> {
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+            off,
+            stream_id
+        );
+
+        quic_conn.stream_send(stream_id, &stream_d[..off], final_data)?;
+
+        Ok(())
+    }
+
+    /// Read QUIC streams, assign HTTP/3 meaning to them and process HTTP/3
```
ghedo: This comment is very confusing Should be something like
```
Processes HTTP/3 data received from the peer.

On success it returns an [`Event`] as well as the event's source stream ID. The stream ID can be used when calling [`send_response()`] and [`send_data()`] when responding to incoming requests.
```

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+    /// frames
+    pub fn process(
+        &mut self, quic_conn: &mut super::Connection,
+    ) -> Result<(u64, Event)> {
+        // Read streams and assign HTTP/3 meaning to them.
```
ghedo: This "assign meaning" business is pretty confusing, can just sat "Process HTTP/3 data from readable streams." or something.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+        let streams: Vec<u64> = quic_conn.readable().collect();
+
+        for s in streams {
```
ghedo: You might be able to just do:

```suggestion
        for s in quic_conn.readable() {
```

so you don't need the `collect()` above.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+                    Err(Error::Done) => {
+                        trace!(
+                            "{} done handling stream id {}",
+                            quic_conn.trace_id(),
+                            s
+                        );
+                        break;
+                    },
+
+                    Err(Error::BufferTooShort) => {
```
ghedo: Can remove.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+    }
+
+    /// Open HTTP/3 control stream. Nothing happens if stream
+    /// is already open
+    fn open_control_stream(
+        &mut self, quic_conn: &mut super::Connection,
+    ) -> Result<()> {
+        if self.control_stream_id.is_none() {
+            let stream_id = self.get_available_uni_stream()?;
+            info!(
```
ghedo: Can probably remove, doesn't seem very useful.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+    }
+
+    /// Open QPACK encoder and decoder streams. Nothing happens
+    /// if streams are already open
+    fn open_qpack_streams(
+        &mut self, quic_conn: &mut super::Connection,
+    ) -> Result<()> {
+        if self.local_qpack_streams.encoder_stream_id.is_none() {
+            let stream_id = self.get_available_uni_stream()?;
+            info!(
```
ghedo: Same, re: remove.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+                false,
+            )?;
+
+            // TODO await ACK of stream open?
+            self.local_qpack_streams.encoder_stream_id = Some(stream_id);
+        }
+
+        if self.local_qpack_streams.decoder_stream_id.is_none() {
+            let stream_id = self.get_available_uni_stream()?;
+            info!(
```
ghedo: Remove?

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+
+    /// Send SETTINGS frame based on HTTP/3 configuration
+    fn send_settings(&mut self, quic_conn: &mut super::Connection) -> Result<()> {
+        self.open_control_stream(quic_conn)?;
+
+        let mut d = [42; 128];
+
+        // Client cannot send placeholders, so validate here
+        let num_placeholders = if self.is_server {
+            self.local_settings.num_placeholders
```
ghedo: Isn't this an Option already? So you can just use it directly. It shouldn't be Some for clients.

src/h3/mod.rs
```diff
@@ -26,4 +26,778 @@
...
+        let mut b = octets::Octets::with_slice(&mut d);
+
+        frame.to_bytes(&mut b)?;
+        let off = b.off();
+
+        match self.control_stream_id {
+            Some(id) => {
+                quic_conn.stream_send(id, &d[..off], false)?;
+            },
+            None => {
```
ghedo: You should be returning way earlier if we don't have `control_stream_id`.

src/h3/stream.rs
```diff
@@ -0,0 +1,380 @@
...
+        let mut state = State::StreamTypeLen;
+
+        if crate::stream::is_bidi(id) {
+            ty = Some(Type::Request);
+            initialised = true;
+            // TODO draft 18+ will mean first state is not FramePayloadLenLen
+            state = State::FramePayloadLenLen;
+        };
+
+        trace!(
```
ghedo: Remove.

examples/http3-client.rs
```diff
@@ -0,0 +1,287 @@
...
+                        break;
+                    },
+                }
+            }
+
+            loop {
+                match http3_conn.process(&mut quic_conn) {
+                    Ok((stream_id, quiche::h3::Event::Headers(value))) => {
+                        info!(
+                            "{} Got response headers {:?} on stream id {}",
```
```suggestion
                            "{} got response headers {:?} on stream id {}",
```

examples/http3-client.rs
```diff
@@ -0,0 +1,287 @@
...
+                            quic_conn.trace_id(),
+                            value,
+                            stream_id
+                        );
+                    },
+                    Ok((stream_id, quiche::h3::Event::Data(value))) => {
+                        info!(
+                            "{} Got response data of length {} in stream id {}",
```
```suggestion
                            "{} got response data of length {} in stream id {}",
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+use std::net;
+
+use std::collections::HashMap;
+
+use quiche::h3::qpack::Header;
+use ring::rand::*;
+
+const LOCAL_CONN_ID_LEN: usize = 16;
+
+const MAX_DATAGRAM_SIZE: usize = 1452;
```
ghedo:
```suggestion
const MAX_DATAGRAM_SIZE: usize = 1350;
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+  --root <dir>      Root directory [default: examples/root/]
+  --name <str>      Name of the server [default: quic.tech]
+  -h --help         Show this screen.
+";
+struct HTTP3Client {
+    quiche_conn: Box<quiche::Connection>,
+    http3_conn: Option<quiche::h3::Connection>,
+}
+
+type HTTP3ClientMap = HashMap<Vec<u8>, (net::SocketAddr, HTTP3Client)>;
```
ghedo:
```suggestion
type ClientMap = HashMap<Vec<u8>, (net::SocketAddr, Client)>;
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    // adding it to the collection
+                    client.http3_conn = Some(h3_conn);
+                }
+            }
+
+            if let Some(http3_conn) = &mut client.http3_conn {
+                match http3_conn.process(client.quiche_conn.as_mut()) {
+                    Ok((stream_id, quiche::h3::Event::Headers(value))) => {
+                        info!(
+                            "{} Got request {:?} on stream id {}",
```
ghedo:
```suggestion
                            "{} got request {:?} on stream id {}",
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                            client.quiche_conn.trace_id(),
+                            value,
+                            stream_id
+                        );
+
+                        let resp = build_response(root_dir, &value);
```
ghedo:
```suggestion
                        let (headers, body) = build_response(root_dir, &value);
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+
+                        if resp.0.is_empty() {
```
ghedo: `build_response()` should just return an error if it failed.

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                            client
+                            .quiche_conn
+                            .close(false, 0x0, b"HTTP/3 Failed")
+                            .unwrap();
+                        }
+
+                        if let Err(e) = http3_conn.send_response(
+                            client.quiche_conn.as_mut(),
```
ghedo:
```suggestion
                            &mut client.quiche_conn,
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                        ) {
+                            error!(
+                                "{} stream send failed {:?}",
+                                client.quiche_conn.trace_id(),
+                                e
+                            );
+                        }
+
+                        if let Err(e) = http3_conn.send_body(
+                            client.quiche_conn.as_mut(),
```
ghedo:
```suggestion
                            &mut client.quiche_conn,
```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                            .unwrap();
+                        }
+
+                        if let Err(e) = http3_conn.send_response(
+                            client.quiche_conn.as_mut(),
+                            stream_id,
+                            &resp.0,
+                            true,
+                        ) {
+                            error!(
```
ghedo: If this fails you should stop sending the response and probably just close the connection.

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                            resp.1,
+                            true,
+                        ) {
+                            error!(
+                                "{} stream send failed {:?}",
+                                client.quiche_conn.trace_id(),
+                                e
+                            );
+                        }
+                    },
```
ghedo:
```suggestion
                    },

```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    Ok((stream_id, quiche::h3::Event::Data(value))) => {
+                        info!(
+                            "Got request data of length {} in stream id {}",
+                            value.len(),
+                            stream_id
+                        );
+                    },
```
ghedo:
```suggestion
                    },

```

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    Err(quiche::h3::Error::Done) => {
+                        // TODO
```
ghedo: TODO what? There is nothing to do here, no?

examples/http3-server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    },
```
ghedo:
```suggestion
                    },

```

examples/http3-server.rs
LPardue: I agree, this was a hacky to get around figuring out using native Result in the application. Tips welcome.

src/h3/frame.rs
```diff
@@ -0,0 +1,957 @@
...
+
+            MAX_PUSH_FRAME_TYPE_ID => Frame::MaxPushId {
+                push_id: b.get_varint()?,
+            },
+
+            DUPLICATE_PUSH_FRAME_TYPE_ID => Frame::DuplicatePush {
+                push_id: b.get_varint()?,
+            },
+            _ => {
+                error!("Invalid frame type {}", frame_type);
```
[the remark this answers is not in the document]
LPardue: This is actually a bug. We MUST ignore unknown HTTP/3 frame types so I think what should happen is:

* the h3::Stream state machine should cleanly terminate before we attempt to parse the frame payload.
* h3::Frame::from_bytes should return Error::Done for unknown types.

src/h3/mod.rs
LPardue: Yeah I realised this when using the FFI in C. These parameters must be set and there are only 4 of them (perhaps going down to 3 soon). Since they are simple types I think removing the set_* functions is the correct way to go. We can add new functions for other things in the future as needed, and not modify the new signature.

examples/http3-client.rs
```diff
@@ -0,0 +1,287 @@
...
+                break;
+            }
+
+            let h3_config = quiche::h3::Config::new(0, 1024, 0, 0).unwrap();
+            http3_conn =
+                Some(quiche::h3::connect(&mut quic_conn, &h3_config).unwrap());
+        }
+
+        if let Some(http3_conn) = &mut http3_conn {
+            if !req_sent {
```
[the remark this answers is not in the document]
LPardue: I've removed the flag, leaving this open to double check we are happy with how the code looks now.

src/h3/mod.rs
LPardue: In handle_stream we only read a paltry 124 bytes out of the transport stream,. BufferTooShort in this case means more data should be read from the transport stream, so we catch this error to make sure to keep reading data from the stream. If we read all data out of transport stream, Done is returned and we move on

src/h3/mod.rs
LPardue: I reworked these functions. They are 90% the same code anyway so that's been extracted out.

ghedo:
I rebased the branch on master (which required some changes due to change in ALPN API) and added some changes on top of that.

The "fixup" commit ended-up being bigger than I anticipated, so it's kind of hard to review. Most changes in the http3 examples are due to some variables being renamed, so they are not as scary as they seem.

LPardue:
LGTM

ghedo:
:tada:

PR #26, opened 2019-03-01
Fix incorrect value of wire version for draft-18

LPardue:


ghedo:
Plase don't reuse commit messages, it should say that it's fixing a missing instance of version. Otherwise looks good.

LPardue:
Updated commit message and PR title^^

ghedo:
Tweaked commit message some more.

PR #28, opened 2019-03-24
h3: Add support for GREASE frames

LPardue:
This fixes the bug that we were rejecting unknown frame types.

In order to make this more testable in the future, I added a new utility function in the library to send two stream types (the first two in the reserved range); one has a 0-length payload, the second has some 'random' bytes. These frames are sent at the start of request or response streams conditionally, that boolean is passed new parameter to `send_request` or `send_response`. This kind sucks and it means the ffi has to change. A different approach would be to set a HTTP/3 config option and then check that within the functions. I've left it as is for now to discuss.

I added the `--grease` command line argument to rust examples. We should discuss if this should be more fine grain i.e. controlling grease of transport and HTTP/3: streams, frames, SETTINGS etc.

I've now realised that I forgot to modify the C examples. So those will probably be broken until I fix up. But that will depend on some of the choices above.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -107,6 +107,11 @@ pub extern fn quiche_config_verify_peer(config: &mut Config, v: bool) {
 }
 
 #[no_mangle]
+pub extern fn quiche_config_grease(config: &mut Config, v: bool) {
+    config.grease(v);
+}
+
+#[no_mangle]
 pub extern fn quiche_config_log_keys(config: &mut Config) {
     config.log_keys();
 }
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -389,6 +389,10 @@ impl Connection {
 
         let off = b.off();
 
+        if conn.grease {
+            self.send_grease_frames(conn, stream_id)?;
+        }
+
         trace!(
             "{} sending HEADERS of size {} on stream {}",
             conn.trace_id(),
@@ -579,6 +583,36 @@ impl Connection {
 
             self.local_qpack_streams.decoder_stream_id = Some(stream_id);
         }
+
+        Ok(())
+    }
+
+    /// Send GREASE frames on the provided stream ID.
+    fn send_grease_frames(
+        &mut self, conn: &mut super::Connection, stream_id: u64,
+    ) -> Result<()> {
+        let mut d = [42; 128];
+
+        let mut b = octets::Octets::with_slice(&mut d);
+
+        // Empty GREASE frame.
+        b.put_varint(0)?;
+        b.put_u8(0xb)?;
+
+        // GREASE frame with payload.
+        b.put_varint(18)?;
+        b.put_u8(0x2a)?;
+
+        trace!(
+            "{} sending GREASE frames on stream {}",
+            conn.trace_id(),
+            stream_id
+        );
+
+        let off = b.off();
+        conn.stream_send(stream_id, &d[..off], false)?;
+
+        conn.stream_send(stream_id, b"GREASE is the word", false)?;
 
         Ok(())
     }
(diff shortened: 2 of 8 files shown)


LPardue:
I pushed a change that moves the GREASE bool into the QUIC connection config, with the default behaviour of true. This avoids the horrible impact of parameters on everything. Putting this in the QUIC config means we can, in a single place, control whether to send GREASE types in the handshake, in the QUIC connection and/or the application mapping. Although, for now we only send some HTTP/3 GREASE frames.

src/lib.rs
```diff
@@ -343,6 +346,11 @@ impl Config {
         self.tls_ctx.set_verify(verify);
     }
 
+    /// Configures whether to send GREASE values.
+    pub fn send_grease(&mut self, grease: bool) {
```
ghedo: Should call this `enable_grease()`, or just `grease()`.

src/h3/mod.rs
```diff
@@ -582,6 +589,43 @@ impl Connection {
...
+            "{} sending GREASE frame of type {:x} on stream {}",
+            conn.trace_id(),
+            0xb,
+            stream_id
+        );
+
+        let off = b.off();
+        conn.stream_send(stream_id, &d[..off], false)?;
+
+        let mut b = octets::Octets::with_slice(&mut d);
```
ghedo: Why do you need to recreate this? Can just encode both frames and write them together.

src/h3/mod.rs
```diff
@@ -582,6 +589,43 @@ impl Connection {
...
     }
 
+    fn send_grease_frames(
+        &mut self, conn: &mut super::Connection, stream_id: u64,
+    ) -> Result<()> {
+        let mut d = [42; 128];
+
+        let mut b = octets::Octets::with_slice(&mut d);
+        b.put_varint(0)?;
+        b.put_u8(0xb)?;
```
ghedo: It's not really greasing if you hardcode the values...

src/ffi.rs
```diff
@@ -106,6 +106,11 @@ pub extern fn quiche_config_verify_peer(config: &mut Config, v: bool) {
     config.verify_peer(v);
 }
 
+#[no_mangle]
+pub extern fn quiche_config_send_grease(config: &mut Config, v: bool) {
```
ghedo: You also need to add this to `include/quiche.h`.

src/h3/mod.rs
ghedo: But also, you don't really need to write into the hardcoded buffer (except for the varints), and can just write the single values directly to the stream (like for HEADERS and DATA).

src/h3/mod.rs
```diff
@@ -351,6 +351,10 @@ impl Connection {
         self.streams
             .insert(stream_id, stream::Stream::new(stream_id, true));
 
+        if conn.grease {
```
ghedo: Why not just do this from inside `send_headers()` so you don't duplicate the code?

src/h3/mod.rs
```diff
@@ -361,6 +365,9 @@ impl Connection {
         &mut self, conn: &mut super::Connection, stream_id: u64,
         headers: &[Header], fin: bool,
     ) -> Result<()> {
+        if conn.grease {
+            self.send_grease_frames(conn, stream_id)?;
+        }
```
ghedo: Need empty line after this.

src/h3/mod.rs
LPardue: Yup, I have protyped a better version in the draft 19 branch because the grease values are different and common across all greasing.

LPardue: So I propose landing this change, or waiting to incorporate this change into the draft 19 /20 update.

src/lib.rs
LPardue: done, new name: `grease()`

src/h3/mod.rs

src/h3/mod.rs
LPardue: good observation, done

src/h3/mod.rs

PR #35, opened 2019-04-11
Fix TODO in srs/ffi.rs

jabedude:
This patch fixes a TODO to use MAX_SSIZE to limit out_len parameter in quiche_conn_stream_recv.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -365,6 +365,10 @@ pub extern fn quiche_conn_new_with_tls(
 pub extern fn quiche_conn_recv(
     conn: &mut Connection, buf: *mut u8, buf_len: usize,
 ) -> ssize_t {
+    if buf_len > <ssize_t>::max_value() as usize {
+        panic!("The provided buffer is too large");
+    }
+
     let buf = unsafe { slice::from_raw_parts_mut(buf, buf_len) };
 
     match conn.recv(buf) {
@@ -378,6 +382,10 @@ pub extern fn quiche_conn_recv(
 pub extern fn quiche_conn_send(
     conn: &mut Connection, out: *mut u8, out_len: usize,
 ) -> ssize_t {
+    if out_len > <ssize_t>::max_value() as usize {
+        panic!("The provided buffer is too large");
+    }
+
     let out = unsafe { slice::from_raw_parts_mut(out, out_len) };
 
     match conn.send(out) {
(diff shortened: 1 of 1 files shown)


src/ffi.rs
```diff
@@ -392,7 +392,10 @@ pub extern fn quiche_conn_stream_recv(
     conn: &mut Connection, stream_id: u64, out: *mut u8, out_len: usize,
     fin: &mut bool,
 ) -> ssize_t {
-    // TODO: limit out_len to MAX_SSIZE to allow the result to fit in ssize_t
+    if out_len > <ssize_t>::max_value() as usize {
```
RReverser: A better way could've probably been to use `ssize_t` in arguments in the first place.

@ghedo Also it's worth noting that currently `.h` maps Rust's `usize` to `size_t`, but technically it's not the same even if holds on most platforms - you should either change Rust FFI functions to use `ssize_t` and `size_t` or use `uintptr_t` on the other side.

ghedo: Yeah, using `ssize_t` for the argument itself seems better, @jabedude can you update the PR to do that instead please?

As for not using `usize` directly, I can do that change separately.

jabedude: @ghedo `std::slice::from_raw_parts` takes `usize` as a parameter. Do you want me to recase out_len as `usize`?

ghedo: @jabedude so, on second thought, the POSIX socket API also uses `size_t` as argument and `ssize_t` as return, so it kind of makes to follow that here since we are somewhat trying to mimic that API. Using `ssize_t` as argument also means that we would have to check that it is > 0 so we would still have to do a check.

All in all, keeping the argument as-is and doing the max_value check like you did seems actually not that bad, though instead of returning an error (and BufferTooShort doesn't seem particularly appropriate) we probably just want to `panic!("the provided buffer is too long")`. This case is unlikely to come up in practice, but if it does happen this will provide pretty clear message.

Also, I just realized the same problem also applies to `quiche_conn_recv()`, `quiche_conn_send()` and `quiche_conn_stream_send()` could you add the check for those as well?

ghedo:
Merged (with a small tweak to the commit message) as b9c50420c616253a1f125075a5c3e6d0a4c09359, thanks!

ghedo:
And additional fix in 3ceee44c1f055680ee2153aa14b2579a65944169 (due to the fact that I accidentally included a broken change in that commit to also update a function in `src/h3/ffi.rs` :man_facepalming: ).

PR #37, opened 2019-04-16
Make it easier to use Quiche in projects that already link OpenSSL

bdzzz:
## Overview
Currently Quiche builds a static library that includes BoringSSL. Unfortunately this can't be linked into projects that already link OpenSSL because there will be symbol conflicts. To solve this issue we can also build a dynamic library alongside the static one. The change is pretty simple, just add `cdylib` as a crate type in Cargo.toml and fix the `Makefile`  in examples/ so that it still picks up the static library. I guess that other people will also encounter this problem, so I've added a small paragraph in the Readme explaining how to use Quiche in projects that already link OpenSSl.
I've also added a `clean` target to the Makefile in examples/.

## Possible problems
The only problem I can think of is that if you have a static and dynamic library in the same directory `gcc`  will pick the dynamic one. If Quiche starts building a dynamic library alongside the static one some existing C/C++ projects that link Quiche might experience problems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,4 +44,4 @@ env_logger = "0"
 debug = true
 
 [lib]
-crate-type = ["lib", "staticlib"]
+crate-type = ["lib", "staticlib", "cdylib"]
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -16,7 +16,7 @@ LIBSSL_DIR = $(dir $(shell find ${BUILD_DIR} -name libssl.a))
 
 LDFLAGS = -L$(LIBCRYPTO_DIR) -L$(LIBSSL_DIR) -L$(LIB_DIR)
 
-LIBS = -lcrypto -lssl -lquiche -lev -ldl -pthread
+LIBS = -lcrypto -lssl -l:libquiche.a -lev -ldl -pthread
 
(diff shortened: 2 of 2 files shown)


ghedo:
Thanks! Maybe we can go a step further and just build BoringSSL with `-fvisibility=hidden` by default. So we wouldn't need the note in the README and it would just work out of the box, what do you think?

bdzzz:
> Thanks! Maybe we can go a step further and just build BoringSSL with `-fvisibility=hidden` by default. So we wouldn't need the note in the README and it would just work out of the box, what do you think?

Good call-out. Turns out that BoringSSL  by default is compiled with `-fvisibility-hidden` under `GCC` or `Clang`, and `MSVC` does not export symbols anyway. - https://github.com/google/boringssl/blob/master/CMakeLists.txt#L126

I'll remove the instructions from the `Readme`.

ghedo:
Merged as 57e2936bdf49c3b9239f468470109d5689066d77, thanks!

It turned out that BoringSSL also needs to be built with `-fPIC` in release mode (but Travis CI didn't catch this because the library was only built in debug mode there), so I added that as well. Also slightly tweaked the examples Makefile since we already had a `clean` target.

PR #38, opened 2019-04-22
Fix C examples, adding uthash to submodules

junhochoi:
Tested with MacOS X (Apple LLVM version 10.0.1 (clang-1001.0.46.3) and debian 9.8 (cc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516)


junhochoi:
Will reopen after fixing conflicts.

PR #39, opened 2019-04-23
Fix C examples build

junhochoi:
- Add uthash as submodule
- Fix build error (MacOS clang, gcc)

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -5,6 +5,8 @@ BUILD_DIR = $(CURDIR)/build
 LIB_DIR = $(BUILD_DIR)/debug
 INCLUDE_DIR = ../include
 
+# added for uthash
+INCS = -I$(INCLUDE_DIR) -I../deps/uthash/include
 CFLAGS = -I. -Wall -Werror -pedantic -fsanitize=address -g
 
 ifeq ($(OS), Darwin)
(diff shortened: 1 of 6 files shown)


junhochoi:
> I don't think uthash is a hard dependency for the quiche library itself. Would a documentation update for building C examples be enough?

Right. Also C examples requires libev... so there is more missing dependencies but it's not critical to build Rust main sources. Probably good to document how to build C sources in a separate doc.

LPardue:
AFAIK quiche depends on BoringSSL, not OpenSSL. Unless the examples do something special.

junhochoi:
> AFAIK quiche depends on BoringSSL, not OpenSSL. Unless the examples do something special.

That's because there was `-lcrypto -lssl` in Makefile but actually it's not needed. I updated Makefile and readme. I believe libquiche embeds boringssl so no need to link with it externally.

examples/Makefile
```diff
@@ -5,6 +5,8 @@ BUILD_DIR = $(CURDIR)/build
 LIB_DIR = $(BUILD_DIR)/debug
 INCLUDE_DIR = ../include
 
+# added for uthash
```
ghedo:
```suggestion
```

examples/README.md
```diff
@@ -0,0 +1,19 @@
+How to build C examples
+-----------------------
+
+### Requirements
+
+You will need the following softwares to build examples in this directory.
+Use your OS package manager(brew, apt, pkg) or install from the source.
```
ghedo:
```suggestion
Use your OS package manager (brew, apt, pkg, ...) or install from the source.
```

ghedo:
Merged as fe41660096d1747c0290c882d70231183ff2be60, thanks! I squashed the commits together and tweaked the README wording slightly.

PR #40, opened 2019-04-26
Support draft 20

LPardue:
Bare minimum to try out HTTP/3 draft 20. There is probably more work required in transport and recovery?


LPardue:
Updated all `hq` tokens, there were some that still referred to `-18` 😆

ghedo:
Looks good now (good catch on the *.c examples), though let's wait until a few more implementations support -20 before merging so we can do proper interop testing.

ghedo:
@LPardue while you are at it, could you also review the TODO comments in the h3 module? I think some of them might have already been addressed with the draft 19 support e.g. https://github.com/cloudflare/quiche/blob/8f02196ed3885954d70cd632adc4cb1f4ba89fc7/src/h3/mod.rs#L318 https://github.com/cloudflare/quiche/blob/8f02196ed3885954d70cd632adc4cb1f4ba89fc7/src/h3/mod.rs#L361

ghedo:
Also looks like this needs a rebase now.

LPardue:
> @LPardue while you are at it, could you also review the TODO comments in the h3 module?

Good idea, will do.

> I think some of them might have already been addressed with the draft 19 support e.g.
> 
> [quiche/src/h3/mod.rs](https://github.com/cloudflare/quiche/blob/8f02196ed3885954d70cd632adc4cb1f4ba89fc7/src/h3/mod.rs#L318)
> 
> Line 318 in [8f02196](/cloudflare/quiche/commit/8f02196ed3885954d70cd632adc4cb1f4ba89fc7)

This particular comment is a bit misleading. What it is trying to say is, when an Malformed frame error occurs (which is mostly going to happen when the frame contains less bytes than the stated length), we need to generate an on-wire error code that encodes the frame type (up to the limit). From draft 20:

```
HTTP_MALFORMED_FRAME (0x01XX):  An error in a specific frame type.
      If the frame type is "0xfe" or less, the type is included as the
      last byte of the error code.  For example, an error in a
      MAX_PUSH_ID frame would be indicated with the code (0x10D).  The
      last byte "0xff" is used to indicate any frame type greater than
      "0xfe".
```
The error code seems overly complex and supporting supporting it seems like a lot of code/effort, so I opened an issue on the H3 spec https://github.com/quicwg/base-drafts/issues/2551. I think we should postpone an work on this until that issue is resolved one way or the other.

src/h3/mod.rs
```diff
@@ -358,9 +358,9 @@ impl Error {
             Error::GeneralProtocolError => 0xFF,
             Error::MalformedFrame => 0x10,
 
-            Error::QpackDecompressionFailed => 0x20, // TODO: value is TBD
-            Error::QpackEncoderStreamError => 0x21,  // TODO: value is TBD
-            Error::QpackDecoderStreamError => 0x22,  // TODO: value is TBD
+            Error::QpackDecompressionFailed => 0x201,
```
```suggestion
            Error::QpackDecompressionFailed => 0x200,
```

PR #42, opened 2019-04-28
Some error handling improvements

ghedo:
Commit messages should be fairly self-explanatory. FYI @LPardue there's a few h3-related changes here on how errors from the transport layer and `octets` are handled/reported.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -46,7 +46,7 @@ Options:
   -h --help                Show this screen.
 ";
 
-fn main() -> Result<(), Box<std::error::Error>> {
+fn main() {
     let mut buf = [0; 65535];
     let mut out = [0; MAX_DATAGRAM_SIZE];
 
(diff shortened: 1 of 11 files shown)


LPardue:
This looks overall good. Now that the H3 library generates the connection close we should probably update the `poll()` description like you did for `recv()` to state that the library will close the connection e.g

``` ...On error the connection will be closed by
    /// calling [`close()`] with the appropriate error code.
```

I can see the argument for  making to_wire() non-public but now there is a mild inconsistency: that `Error` and `close()` remain public but we don't expose anything for the wire-format error code value. We could add some `pub const` definitions, or just go with your changes and see if any real problems come up.

ghedo:
Added a commit to update `poll()` docs as well.

I like the idea of exposing constants for the wire errors, so we can expose only the errors that the application would use (e.g. for transport it would be `NO_ERROR`, `INTERNAL_ERROR` and `SERVER_BUSY`, while for h3 we could expose `HTTP_NO_ERROR`, `HTTP_INTERNAL_ERROR`, `HTTP_EXCESSIVE_LOAD`, `HTTP_VERSION_FALLBACK` and possibly others) instead of just exposing all the ones that should only be used by the library. Though that should go in separate PRs I think.

LPardue:
SGTM

PR #44, opened 2019-05-01
h3: generate event for stream close

LPardue:


src/h3/mod.rs
```diff
@@ -456,6 +464,9 @@ pub enum Event {
 
     /// Data was received.
     Data(Vec<u8>),
+
+    // Stream was closed,
+    StreamClosed(),
```
```suggestion
    Finished,
```

So a couple of things: 1. I don't think you need `()` since you are not actually wrapping another type (this also means that matching this event would be somewhat more readable) 2. `Finished` seems like a better name (and shorter), but YMMV.

src/h3/ffi.rs
```diff
@@ -121,6 +123,8 @@ pub extern fn quiche_h3_event_for_each_header(
             },
 
         h3::Event::Data { .. } => unreachable!(),
+
+        h3::Event::StreamClosed { .. } => unreachable!(),
```
ghedo: You could probably just replace this and the previous case with `_ => unreachable!()`.

src/h3/ffi.rs
```diff
@@ -138,6 +142,19 @@ pub extern fn quiche_h3_event_data(
 
             data.len()
         },
+
+        h3::Event::StreamClosed { .. } => unreachable!(),
```
ghedo: Same, re: using `_ => unreachable!()`.

src/h3/ffi.rs
```diff
@@ -93,6 +93,8 @@ pub extern fn quiche_h3_event_type(ev: &h3::Event) -> u32 {
         h3::Event::Headers { .. } => 0,
 
         h3::Event::Data { .. } => 1,
+
+        h3::Event::StreamClosed { .. } => 2,
```
ghedo: You also need to add this to the `quiche_h3_event_type` enum in `quiche.h`.

src/h3/ffi.rs
```diff
@@ -138,6 +142,19 @@ pub extern fn quiche_h3_event_data(
 
             data.len()
         },
+
+        h3::Event::StreamClosed { .. } => unreachable!(),
+    }
+}
+
+#[no_mangle]
+pub extern fn quiche_h3_event_stream_closed(ev: &h3::Event) {
```
ghedo: I don't think this is needed? It also doesn't seem to do anything.

ghedo:
Oh also need to update https://docs.quic.tech/quiche/h3/index.html#detecting-end-of-stream

LPardue:
Thanks. I'll address your comments tomorrow

LPardue:
The latest changes are almost there but have an edge case that needs handling.

I also plan to rework unit tests slightly.

src/h3/mod.rs
```diff
@@ -1371,8 +1532,513 @@ mod tests {
...
+                &mut s.pipe.client,
+                s.client.peer_qpack_streams.decoder_stream_id.unwrap(),
+                b"Uh oh",
+                true
+            ),
+            Err(Error::WrongStream)
+        );
+
+        // server-side streams
+        assert_eq!(
```
ghedo: Do we really need to duplicate the same tests for both sides? I'd expect that the same code handles both cases anyway.

src/h3/mod.rs
```diff
@@ -952,6 +1009,15 @@ impl Connection {
 
     /// Opens QPACK encoder and decoder streams, if not already opened.
     fn open_qpack_streams(&mut self, conn: &mut super::Connection) -> Result<()> {
+        self.open_qpack_encoder_stream(conn)?;
```
ghedo: I guess it's kind of unrelated, but couldn't we just have a single "open_uni_stream()" that takes the stream type and returns the stream ID, so then we could do e.g. `self.local_qpack_streams.encoder_stream_id = open_uni_stream(QPACK_ENCODER_STREAM_TYPE_ID);` or something like that?

src/h3/mod.rs
```diff
@@ -1309,10 +1411,69 @@ mod tests {
 
             self.pipe.advance(buf).ok();
 
+            // server streams
```
```suggestion
            // Server streams.
```

src/h3/mod.rs
```diff
@@ -1294,13 +1389,20 @@ mod tests {
             })
         }
 
+        /// Do the HTTP/3 handshake so both ends are in sane initial state
         pub fn handshake(&mut self, buf: &mut [u8]) -> Result<()> {
             self.pipe.handshake(buf)?;
 
+            // client streams
```
```suggestion
            // Client streams.
```

src/h3/mod.rs
```diff
@@ -1309,10 +1411,69 @@ mod tests {
...
+            self.pipe.advance(buf).ok();
+
+            if self.pipe.server.grease {
+                self.server.open_grease_stream(&mut self.pipe.server)?;
+            }
+
+            self.pipe.advance(buf).ok();
+
+            loop {
+                match self.client.poll(&mut self.pipe.client) {
```
ghedo: I think you could just turn this into something like:
```rust
while self.client.poll(&mut self.pipe.client).is_ok() {
  // Do nothing.
}
```

Same for the one below.

src/h3/mod.rs
```diff
@@ -1309,10 +1411,69 @@ mod tests {
...
+                    Err(_) => (),
+                }
+            }
+
             Ok(())
         }
+
+        /// Calls poll on identified role and asserts the returned event matches
+        /// the expected event.
+        pub fn poll_assert(
```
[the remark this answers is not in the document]
LPardue: This function has some pros and cons:

pro - it avoids long assert lines that get wrapped by cargo fmt, so reduces line count
pro - it avoids code goofs where you poll a H3 connection with the wrong socket

con - any assert failures no longer happen in the test, so its very difficult to quickly diagnose why a test failed
con - Role enum

Suggest we keep poll_assert for now and refactor the way it works to be a macro that keeps pros but removes cons.

src/h3/mod.rs
```diff
@@ -657,6 +682,20 @@ impl Connection {
         // Return how many bytes were written, excluding the frame header.
         let written = conn.stream_send(stream_id, body, fin)?;
 
+        // tidy up streams
+        if !fin {
+            return Ok(written);
+        }
+
+        if self
```
RReverser: Try instead something like:

```rust
match self.streams.entry(stream_id) {
  Entry::Occupied(entry) if entry.get().peer_fin() => {
    entry.remove();
  }
  _ => {}
}
```

src/h3/mod.rs
LPardue: lol yes that is a much better suggestion

src/h3/mod.rs
LPardue: Change the test to focus on server.

If client-side tests are different/needed, we can add a different test later.

src/h3/mod.rs
```diff
@@ -657,6 +682,20 @@ impl Connection {
         // Return how many bytes were written, excluding the frame header.
         let written = conn.stream_send(stream_id, body, fin)?;
 
+        // tidy up streams
+        if !fin {
```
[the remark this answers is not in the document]

src/h3/mod.rs
RReverser: Oh? What failed?

LPardue: Compiler complained that Entry was unknown. A tried a few things but gave up.

RReverser: Unknown as in, you just need to `use ...` it to get in scope first?

PR #47, opened 2019-05-10
Strict frame checking

LPardue:
The main thrust of this change is to add more checks to frame parsing. This includes detecting if it is sent to an invalid peer, on an invalid stream, at the wrong time on a stream and so on.

Most of this happens in `poll()` when it parses each frame. When first writing this block, I created some helper functions but when adding more checks in this change I found that those functions reduced the readability of the code. So now we just have a simple pattern of

* For a given frame type
  * Test a validity condition
    * If fail
      * send a CONNECTION_CLOSE with the most suitable error code
      * Return error
   * otherwise do whatever needs doing with the frame (either generate an event, set some internal state, or nothing)

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -212,7 +212,7 @@
 //! ```
 //!
 //! ## Detecting end of request or response
-
+//!
 //! A single HTTP/3 request or response may consist of several HEADERS and DATA
 //! frames; it is finished when the QUIC stream is closed. Calling [`poll()`]
 //! repeatedly will generate an [`Event`] for each of these. The application may
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1338,16 +1338,16 @@ mod tests {
     /// Session is an HTTP/3 test helper structure. It holds a client, server
     /// and pipe that allows them to communicate.
     ///
-    /// default() creates a session with some sensible default
-    /// configuration. [`with_configs()`] allows for providing a specific
+    /// `default()` creates a session with some sensible default
+    /// configuration. `with_configs()` allows for providing a specific
     /// configuration.
     ///
-    /// handshake() performs all the steps needed to create a valid HTTP/3
+    /// `handshake()` performs all the steps needed to establish an HTTP/3
     /// connection.
     ///
     /// Some utility functions are provided that make it less verbose to send
-    /// request, responses and individual headers. The full quiche API remains a
-    /// avaialable for any tests that need to do unconventional things (such as
+    /// request, responses and individual headers. The full quiche API remains
+    /// avaialable for any test that need to do unconventional things (such as
     /// bad behaviour that triggers errors).
     struct Session {
         pipe: testing::Pipe,
@@ -1437,22 +1437,22 @@ mod tests {
             Ok(())
         }
 
-        /// Advance the session pipe over the buffer.
+        /// Advances the session pipe over the buffer.
         fn advance(&mut self) -> crate::Result<()> {
             self.pipe.advance(&mut self.buf)
         }
 
-        /// Poll the client for events.
+        /// Polls the client for events.
         fn poll_client(&mut self) -> Result<(u64, Event)> {
             self.client.poll(&mut self.pipe.client)
         }
 
-        /// Poll the server for events.
+        /// Polls the server for events.
         fn poll_server(&mut self) -> Result<(u64, Event)> {
             self.server.poll(&mut self.pipe.server)
         }
 
-        /// Send a request from client with default headers.
+        /// Sends a request from client with default headers.
         ///
         /// On success it returns the newly allocated stream and the headers.
         fn send_request(&mut self, fin: bool) -> Result<(u64, Vec<Header>)> {
@@ -1474,7 +1474,7 @@ mod tests {
             Ok((stream, req))
         }
 
-        /// Send a response from server with default headers.
+        /// Sends a response from server with default headers.
         ///
         /// On success it returns the headers.
         fn send_response(
@@ -1496,7 +1496,7 @@ mod tests {
             Ok(resp)
         }
 
-        /// Send some default payload from client.
+        /// Sends some default payload from client.
         ///
         /// On success it returns the payload.
         fn send_body_client(
@@ -1512,7 +1512,7 @@ mod tests {
             Ok(bytes)
         }
 
-        /// Send some default payload from server.
+        /// Sends some default payload from server.
         ///
         /// On success it returns the payload.
         fn send_body_server(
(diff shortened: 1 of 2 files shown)


LPardue:
I've hoisted the additional checks from an unrelated WIP branch into this PR.

src/h3/mod.rs
```diff
@@ -1295,19 +1329,30 @@ mod tests {
...
 
-    enum Role {
-        Client,
-        Server,
-    }
-
+    /// Session is an HTTP/3 test helper structure. It holds a client, server
+    /// and pipe that allows them to communicate.
+    ///
+    /// default() creates a session with some sensible default
```
```suggestion
    /// `default()` creates a session with some sensible default
```

src/h3/mod.rs
```diff
@@ -1295,19 +1329,30 @@ mod tests {
...
+    /// configuration. [`with_configs()`] allows for providing a specific
```
```suggestion
    /// configuration. `with_configs()` allows for providing a specific
```

src/h3/mod.rs
```diff
@@ -1295,19 +1329,30 @@ mod tests {
...
+    /// configuration.
+    ///
+    /// handshake() performs all the steps needed to create a valid HTTP/3
```
```suggestion
    /// `handshake()` performs all the steps needed to establish an HTTP/3
```

src/h3/mod.rs
```diff
@@ -1295,19 +1329,30 @@ mod tests {
...
+    /// connection.
+    ///
+    /// Some utility functions are provided that make it less verbose to send
+    /// request, responses and individual headers. The full quiche API remains a
```
```suggestion
    /// request, responses and individual headers. The full quiche API remains
```

src/h3/mod.rs
```diff
@@ -1295,19 +1329,30 @@ mod tests {
...
+    /// avaialable for any tests that need to do unconventional things (such as
```
```suggestion
    /// avaialable for any test that need to do unconventional things (such as
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
-        /// Calls poll on identified role and asserts the returned event matches
-        /// the expected event.
-        pub fn poll_assert(
-            &mut self, role: Role, expected: Result<(u64, Event)>,
-        ) {
-            match role {
-                Role::Client => {
-                    assert_eq!(self.client.poll(&mut self.pipe.client), expected);
-                },
+        /// Advance the session pipe over the buffer.
```
```suggestion
        /// Advances the session pipe over the buffer.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+        fn advance(&mut self) -> crate::Result<()> {
+            self.pipe.advance(&mut self.buf)
+        }
 
-                Role::Server => {
-                    assert_eq!(self.server.poll(&mut self.pipe.server), expected);
-                },
-            }
+        /// Poll the client for events.
```
```suggestion
        /// Polls the client for events.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
         }
-    }
 
-    /// Send a single HTTP/3 on a stream.
-    fn send_frame(
-        conn: &mut crate::Connection, frame: frame::Frame, stream_id: u64,
-        fin: bool,
-    ) -> Result<()> {
-        let mut d = [42; 65535];
+        /// Poll the server for events.
```
```suggestion
        /// Polls the server for events.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+        fn poll_server(&mut self) -> Result<(u64, Event)> {
+            self.server.poll(&mut self.pipe.server)
+        }
 
-        let mut b = octets::Octets::with_slice(&mut d);
+        /// Send a request from client with default headers.
```
```suggestion
        /// Sends a request from client with default headers.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+            self.advance().ok();
 
-        let off = b.off();
-        conn.stream_send(stream_id, &d[..off], fin)?;
+            Ok((stream, req))
+        }
 
-        Ok(())
-    }
+        /// Send a response from server with default headers.
```
```suggestion
        /// Sends a response from server with default headers.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
 
-    #[test]
-    fn request_no_body_response_no_body() {
-        let mut buf = [0; 65535];
+            Ok(resp)
+        }
 
-        let mut s = Session::default().unwrap();
-        s.handshake(&mut buf).unwrap();
+        /// Send some default payload from client.
```
```suggestion
        /// Sends some default payload from client.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
-        assert_eq!(stream, 0);
+            self.advance().ok();
 
-        s.pipe.advance(&mut buf).ok();
+            Ok(bytes)
+        }
 
-        s.poll_assert(Role::Server, Ok((stream, Event::Headers(req.to_vec()))));
-        s.poll_assert(Role::Server, Ok((stream, Event::Finished)));
+        /// Send some default payload from server.
```
```suggestion
        /// Sends some default payload from server.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+            self.advance().ok();
 
-        s.pipe.advance(&mut buf).ok();
+            Ok(bytes)
+        }
 
-        s.poll_assert(Role::Client, Ok((stream, Event::Headers(resp.to_vec()))));
-        s.poll_assert(Role::Client, Ok((stream, Event::Finished)));
-        s.poll_assert(Role::Client, Err(Error::Done));
+        /// Send a single HTTP/3 frame from the client.
```
```suggestion
        /// Sends a single HTTP/3 frame from the client.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+
+            let off = b.off();
+            self.pipe.client.stream_send(stream_id, &d[..off], fin)?;
+
+            self.advance().ok();
+
+            Ok(())
+        }
+
+        /// Send a single HTTP/3 frame from the server.
```
```suggestion
        /// Sends a single HTTP/3 frame from the server.
```

src/h3/mod.rs
```diff
@@ -1385,525 +1431,599 @@ mod tests {
...
+            self.advance().ok();
+
+            Ok(())
+        }
     }
 
     #[test]
-    fn request_no_body_response_one_chunk() {
-        let mut buf = [0; 65535];
+    /// Make sue that random GREASE values sit within the specified limit.
```
```suggestion
    /// Make sure that random GREASE values is within the specified limit.
```

src/h3/mod.rs
```diff
@@ -1925,11 +2045,10 @@ mod tests {
     }
 
     #[test]
+    // Client opens multiple control streams, which is forbidden.
```
ghedo: Other tests use `///` instead of `//`, same below. Don't really care which one is used, but we should be consistent.

LPardue: woops, I'll fix this up tomorrow

LPardue: fixed in the latest code review fixup commit

PR #48, opened 2019-05-13
h3: add support for processing PRIORITY frames

LPardue:
This change is based on the `strict-frame-checking` branch.

It adds PRIORITY frame parsing and many checks for ensuring spec-compliant requests can be handled robustly. There are a bunch of tests because the PRIORITY frame MAY be received on a request stream but only as the first frame, never after HEADERS, but its ok to omit it all together. sigh...

Server side, no events are exposed to indicate priority changes, and no extra logic beyond validation is present.
  
Client side, no API is exposed to for the client to set priorities. No PRIORITY frames are sent at any time.


LPardue:
There may be an easier way to achieve the goal of handling PRIORITY. I'll come up with an alternative PR that also rebases on most recent branches.

LPardue:
Prefer #50 to this PR

PR #53, opened 2019-05-22
examples: respond to non-GET requests with 405

LPardue:
The http3-server example doesn't respond at all to non-GET requests. This is not helpful to clients as it could be confused with some other (transport) error.

Since the example is simple, we just default to responding to these kinds of requests with a 405 Method Not Allowed.

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -507,6 +507,7 @@ fn build_response(
 ) -> Result<(std::vec::Vec<quiche::h3::Header>, std::vec::Vec<u8>), ()> {
     let mut file_path = std::path::PathBuf::from(root);
     let mut path = std::path::Path::new("");
+    let mut method = "";
 
     // Look for the request's path and method.
     for hdr in request {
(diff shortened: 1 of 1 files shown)


examples/http3-server.rs
```diff
@@ -440,32 +440,42 @@ fn build_response(
...
+            }
         }
-    }
 
-    let (status, body) = match std::fs::read(file_path.as_path()) {
-        Ok(data) => (200, data),
-        Err(_) => (404, b"Not Found!".to_vec()),
-    };
+        match std::fs::read(file_path.as_path()) {
+            Ok(data) => {
```
ghedo: FTR, from in person chat, this could probably be simplified by using tuples to return (status, data), rather than mutable variables.

ghedo:
Please rebase on latest master, this is pretty old.

LPardue:
d'oh. So when I rebase, this happens

```
cargo build --examples
error: failed to select a version for `winapi`.
    ... required by package `quiche v0.1.0-alpha4 (/home/lucas/workspace/quiche2)`
versions that meet the requirements `= 0.2.8` are: 0.2.8

the package `quiche` depends on `winapi`, with features: `wincrypt` but `winapi` does not have these features.


failed to select a version for `winapi` which could resolve this conflict
```

ghedo:
Yeah, it's always the same problem, removing `Cargo.lock` should "fix" it.

examples/http3-server.rs
```diff
@@ -501,39 +501,43 @@ fn handle_request(
     }
 }
 
-/// Builds an HTTP/3 response given a request.
```
ghedo: Don't hink this was on purpose, was it?

examples/http3-server.rs
```diff
@@ -501,39 +501,43 @@ fn handle_request(
...
 fn build_response(
     root: &str, request: &[quiche::h3::Header],
 ) -> Result<(std::vec::Vec<quiche::h3::Header>, std::vec::Vec<u8>), ()> {
     let mut file_path = std::path::PathBuf::from(root);
     let mut path = std::path::Path::new("");
+    let mut method = "";
 
-    // Look for the request's path and method.
```
ghedo: Same.

ghedo:
Oh, I just realized that for requests with a body we also need to shutdown the stream, otherwise poll() would still keep reporting Data events. Can be done in a separate PR though.

PR #54, opened 2019-05-26
h3: refactor stream buffering and state machine

ghedo:
This refactors how data is read from transport streams and buffered
at the HTTP/3 layer, as well as simplifies the HTTP/3 stream state
machine.

Instead of reading and buffering a fixed amount of bytes from readable
streams, only the amount of bytes required to complete the current state
is read from the transport. Data is still buffered in the HTTP/3
stream's state in order to deal with partial reads, but is now read
directly into the stream buffer to avoid one intermediate copy.

The stream's buffer is resized accordingly to the number of bytes
required by the next state. However the allocated memory for the buffer
is reused across states to avoid unnecessary allocations (an allocation
is only required if the new state requires more data than the current
state).

As for the state machine, all the states responsible for reading a
varint's length have been merged with the states actually reading the
varint itself, to simplify the state machine.

---

Note that all tests in the `h3::stream` module had to be updated, so need to review that in particular to make sure they are equivalent to what we had before.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -734,19 +734,11 @@ impl Connection {
         for s in streams {
             trace!("{} stream id {} is readable", conn.trace_id(), s);
 
-            loop {
-                match self.handle_stream(conn, s) {
-                    Ok(_) => break,
+            match self.handle_stream(conn, s) {
+                Ok(_) | Err(Error::Done) => continue,
 
-                    Err(Error::Done) => break,
-
-                    Err(Error::BufferTooShort) => {
-                        // Keep processing transport stream.
-                    },
-
-                    Err(e) => return Err(e),
-                };
-            }
+                Err(e) => return Err(e),
+            };
         }
 
         for (stream_id, stream) in
@@ -1139,37 +1131,21 @@ impl Connection {
     fn handle_stream(
         &mut self, conn: &mut super::Connection, stream_id: u64,
     ) -> Result<()> {
-        let mut d = [0; 32768];
-
         let stream = self
             .streams
             .entry(stream_id)
             .or_insert_with(|| stream::Stream::new(stream_id, false));
 
-        let (read, _fin) = conn.stream_recv(stream_id, &mut d)?;
-        stream.push(&d[..read])?;
-
-        trace!(
-            "{} read {} bytes on stream {}",
-            conn.trace_id(),
-            read,
-            stream_id
-        );
-
-        while stream.more() {
+        while stream.try_fill_buffer(conn).is_ok() {
             match stream.state() {
-                stream::State::StreamTypeLen => {
-                    let varint_byte = stream.buf_bytes(1)?[0];
-                    stream.set_next_varint_len(octets::varint_parse_len(
-                        varint_byte,
-                    ))?;
-                },
-
                 stream::State::StreamType => {
-                    let varint = stream.get_varint()?;
+                    let varint = match stream.try_consume_varint() {
+                        Ok(v) => v,
+
+                        Err(_) => continue,
+                    };
 
                     let ty = stream::Type::deserialize(varint)?;
-
                     stream.set_ty(ty)?;
 
                     match &ty {
@@ -1250,27 +1226,23 @@ impl Connection {
                     }
                 },
 
-                stream::State::FramePayloadLenLen => {
-                    let varint_byte = stream.buf_bytes(1)?[0];
-                    stream.set_next_varint_len(octets::varint_parse_len(
-                        varint_byte,
-                    ))?
-                },
+                stream::State::PushId => {
+                    let varint = match stream.try_consume_varint() {
+                        Ok(v) => v,
 
-                stream::State::FramePayloadLen => {
-                    let varint = stream.get_varint()?;
-                    stream.set_frame_payload_len(varint)?;
-                },
+                        Err(_) => continue,
+                    };
 
-                stream::State::FrameTypeLen => {
-                    let varint_byte = stream.buf_bytes(1)?[0];
-                    stream.set_next_varint_len(octets::varint_parse_len(
-                        varint_byte,
-                    ))?
+                    stream.set_push_id(varint)?;
                 },
 
                 stream::State::FrameType => {
-                    let varint = stream.get_varint()?;
+                    let varint = match stream.try_consume_varint() {
+                        Ok(v) => v,
+
+                        Err(_) => continue,
+                    };
+
                     match stream.set_frame_type(varint) {
                         Err(Error::UnexpectedFrame) => {
                             let msg = format!("Unexpected frame type {}", varint);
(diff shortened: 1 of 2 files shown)


src/h3/mod.rs
```diff
@@ -1311,11 +1293,29 @@ impl Connection {
...
+                    };
+
+                    conn.close(true, e.to_wire(), b"")?;
+
+                    return Err(e);
+                },
 
-                stream::State::Done => return Err(Error::Done),
+                stream::State::Drain => {
+                    let mut d = [0; 4096];
```
LPardue: Now that the Drain states seems to handle things a little differently, one thing we could do is send a STOP_SENDING. This could prevent us from wasting cycles on handling streams that are useless.

src/h3/stream.rs
```diff
@@ -370,51 +332,130 @@ impl Stream {
...
+            return Ok(());
         }
 
+        Err(Error::InternalError)
+    }
+
+    /// Tries to fill the state buffer by reading data from the corresponding
+    /// transport stream.
+    ///
+    /// When not enugh data can be read to complete the state, this returns
```
LPardue:
```suggestion
    /// When not enough data can be read to complete the state, this returns
```

src/h3/stream.rs
```diff
@@ -370,51 +332,130 @@ impl Stream {
...
+    /// `Error::Done`.
+    pub fn try_fill_buffer(
+        &mut self, conn: &mut crate::Connection,
+    ) -> Result<()> {
+        let buf = &mut self.state_buf[self.state_off..self.state_len];
+
+        let (read, _) = conn.stream_recv(self.id, buf)?;
```
LPardue: passing `conn` lets us avoid one of the copies of the read data right? nice.

src/h3/stream.rs
```diff
@@ -370,51 +332,130 @@ impl Stream {
...
-        let rem_bytes = self.buf_end_pos - self.buf_read_off;
-        rem_bytes > 0
+    /// Returns true if the state buffer has enough data to complete the state.
+    fn state_buffer_complete(&self) -> bool {
+        self.state_off == self.state_len
+    }
+
+    /// Transitions the stream to a new state, and resets the state buffer.
+    fn state_transition(&mut self, new_state: State, expected_len: usize) {
+        self.state = new_state;
```
LPardue: At some point in the distant past we had a function like. It was a convenient place to put transition sanity checking. We removed it though for a range of reasons.
We could add that back, however, given that all state transitions are handled in internally to stream.rs, the overhead of checks might be unnecessary.

src/h3/stream.rs
```diff
@@ -574,34 +608,33 @@ mod tests {
...
-        stream.set_next_varint_len(frame_ty_len).unwrap();
+        stream.try_fill_buffer_for_tests(&mut cursor).unwrap();
 
-        let frame_ty = stream.get_varint().unwrap();
+        let frame_ty = stream.try_consume_varint().unwrap();
         assert_eq!(stream.set_frame_type(frame_ty), Err(Error::MissingSettings));
     }
 
     #[test]
+    /// Process not allowed frame on control stream.
```
LPardue:
```suggestion
    /// Process not-allowed frame on control stream.
```

src/h3/stream.rs
```diff
@@ -370,51 +332,130 @@ impl Stream {
...
+
+        Ok(())
+    }
+
+    /// Tries to fill the state buffer by reading data from the given cursor.
+    ///
+    /// This is intended to replace `try_fill_buffer()` in tests, in order to
+    /// avoid having to setup a transport connection.
+    #[cfg(test)]
+    fn try_fill_buffer_for_tests(
```
LPardue: In some other files, I've seen these test utility functions placed closer to the tests. I don't mind either way but pointing it out in case you meant to move it.

src/h3/stream.rs
```diff
@@ -140,132 +195,43 @@ impl Stream {
...
-            let varint = octets::Octets::with_slice(
-                &mut self.buf[self.buf_read_off as usize..n],
-            )
-            .get_varint()?;
-
-            self.stream_offset += self.next_varint_len as u64;
-            self.buf_read_off += self.next_varint_len as u64;
+    /// Sets the push ID and transitions to the next state.
+    pub fn set_push_id(&mut self, _id: u64) -> Result<()> {
+        // TODO: implement push ID.
```
LPardue: missing an `assert!` for consistency?

src/h3/mod.rs
LPardue: Or, am I missing how Drain works in practice...

ghedo: This is pretty much equivalent to the previous behavior, with the difference that instead of filling the stream buffer, we just discard the data immediately.

But yes, we could probably send STOP_SENDING, if we supported it.

src/h3/stream.rs
ghedo: Right. Basically we can just read the bytes directly into the stream buffer, instead of using an intermediate buffer like before.

src/h3/stream.rs
ghedo: Yep, added now.

src/h3/mod.rs
```diff
@@ -2183,7 +2183,7 @@ mod tests {
     }
 
     #[test]
-    /// Client closes a qpack stream, which is forbidden.
+    /// Client sends data on a QPACK stream, which we don't allow.
     fn close_qpack_stream() {
```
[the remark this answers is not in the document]
ghedo: Yes, I added a TODO comment now so it will be clear what needs to be done when this starts failing.

src/h3/stream.rs
ghedo: The difference is that this is a `Connection` method, so it needs to be inside the `impl` block. It might be possible to make this a normal function and move inside test `tests` module though.

src/h3/stream.rs
ghedo: Well, this function is really just to avoid having to copy the same 3 lines all over the place. We already have asserts (which are already of dubious utility) to validate state transitions, so more runtime checks are unnecessary IMO.

What we might want to do is some form of static checking like the one proposed in https://hoverbear.org/2016/10/12/rust-state-machine-pattern/ but I'm not convinced it would improve the situation much TBH.

LPardue: The changes that would be needed for static checking seem quite onerous.

We have a bunch of tests that validate transitions. So on reflection I think what we have is ok.

src/h3/mod.rs
LPardue: Something else we could do, now that reading is more dynamic, is not drain unknown stream types. That gives us some way to put back pressure on a stream.

ghedo: So, that is how I implemented this originally (basically just return `Error::Done` like it was done before), but for whatever reason it broke fetching big files from cloudflare-quic.com. That is, the http3-client wouldn't receive all of the data anymore, and then the connection would time out without receiving the `Finished` event.

I just tried again (just reaplaced this block with `reuturn Err(Error::Done);` and I can still reproduce the problem, though it might just be that some changes in some other parts of the code are also needed to make this work.

We might want to do some more investigation in the future when we have more time.

LPardue: Agreed. This sounds like a separate optimisation task to investigate STOP_SENDING and lazy reads. No need to block this PR on that.

ghedo:
@LPardue anything else? If not please approve :)

LPardue:
🚢

PR #60, opened 2019-06-12
h3::stream: don't resize state buffer for DATA frames

ghedo:
Since we don't buffer DATA frames anymore, there's no point in resizing
the state buffer.

---

This depends on https://github.com/cloudflare/quiche/pull/59.

I can now run the `http3/5000000` benchmark without running our of memory :)

### DIFF
--- a/src/h3/stream.rs
+++ b/src/h3/stream.rs
@@ -222,7 +222,7 @@ impl Stream {
             Type::Unknown => State::Drain,
         };
 
-        self.state_transition(state, 1);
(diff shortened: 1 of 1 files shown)


src/h3/stream.rs
```diff
@@ -491,11 +491,18 @@ impl Stream {
     }
 
     /// Transitions the stream to a new state, and resets the state buffer.
```
LPardue:
```suggestion
    /// Transitions the stream to a new state, and optionally resets the state buffer.
```

PR #63, opened 2019-06-18
Fix http3-client on MacOS

junhochoi:
Problem is when we bind() udp socket, Linux understand '::' as v6 and v4 socket both, so connect() with v4 address will succeed. However in MacOS (or BSD I believe) we need to specify it's v4 or v6.

This patch try to read 1st resolved address from URL and set bind address accordingly. When IP address of 1st resolved hostname is v4 then it bind() to '0.0.0.0', otherwise bind() to '::'.

### DIFF
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -79,25 +79,19 @@ fn main() {
     // doesn't allow v4 and v6 can be bind() in one socket).
     // Note that linux doesn't need this because it can handle v4 and v6 socket
     // when bind() with "::".
-    let mut addrs_iter = url.to_socket_addrs().unwrap();
-    let addr = addrs_iter.next(); // get next IP address resolved from URL hostname
-    let bind_addr;
-    match addr {
-        Some(ip) => {
-            debug!("addr = {:?}", ip);
-            bind_addr = if ip.is_ipv4() {
-                "0.0.0.0:0" // v4
-            } else {
-                "[::]:0" // v6
-           }
-        },
-        None => panic!("IP address not found"),
-    }
+
+    // resolve server address
+    let peer_addr = url.to_socket_addrs().unwrap().next().unwrap();
+    info!("connecting to {:?}...", peer_addr);
+    let bind_addr = match peer_addr {
+        std::net::SocketAddr::V4(_) => "0.0.0.0:0",
+        std::net::SocketAddr::V6(_) => "[::]:0",
+    };
 
     // Create the UDP socket backing the QUIC connection, and register it with
     // the event loop.
     let socket = std::net::UdpSocket::bind(bind_addr).unwrap();
-    socket.connect(&url).unwrap();
+    socket.connect(peer_addr).unwrap();
 
     let socket = mio::net::UdpSocket::from_socket(socket).unwrap();
     poll.register(
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -82,7 +82,7 @@ fn main() {
     let mut bind_addr = "[::]:0"; // ipv6 by default
     match addr {
         Some(ip) => {
-            println!("addr = {:?}", ip);
+            debug!("addr = {:?}", ip);
             if ip.is_ipv4() {
                 bind_addr = "0.0.0.0:0";
             }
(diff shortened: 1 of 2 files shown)


examples/http3-client.rs
```diff
@@ -71,9 +72,31 @@ fn main() {
...
 
+    // Take a look at server address resolved to check if it's ipv4 or ipv6.
+    // Depending on the IP family, bind_addr will be default address of
+    // v4 or v6 for calling bind() later.
+    // This workaround is to work with MacOS (or BSD variants which
+    // doesn't allow v4 and v6 can be bind() in one socket).
+    // Note that linux doesn't need this because it can handle v4 and v6 socket
+    // when bind() with "::".
+    let mut addrs_iter = url.to_socket_addrs().unwrap();
+    let addr = addrs_iter.next(); // get next IP address resolved from URL hostname
```
ghedo: There's no need to keep the iterator as separate mutable variable, you can merge the two lines above as:
```rust
    // Resolve server address.
    let peer_addr = url.to_socket_addrs().unwrap().next().unwrap();

    info!("connecting to {:?}...", peer_addr);
```

examples/http3-client.rs
```diff
@@ -71,9 +72,31 @@ fn main() {
...
+    let bind_addr;
+    match addr {
```
ghedo: This match can be simplified by making better use of matching:
```rust
    let bind_addr = match peer_addr {
        std::net::SocketAddr::V4(_) => "0.0.0.0:0",

        std::net::SocketAddr::V6(_) => "[::]:0",
    };
```

examples/http3-client.rs
```diff
@@ -71,9 +72,31 @@ fn main() {
...
+           }
+        },
+        None => panic!("IP address not found"),
+    }
+
     // Create the UDP socket backing the QUIC connection, and register it with
     // the event loop.
-    let socket = std::net::UdpSocket::bind("[::]:0").unwrap();
+    let socket = std::net::UdpSocket::bind(bind_addr).unwrap();
     socket.connect(&url).unwrap();
```
ghedo: We should reuse the `peer_addr` variable from above here, to avoid doing another DNS resolution (best case is that we resolve the domain twice for no reason, worst case is that the second resolution returns a different result).

examples/http3-client.rs
```diff
@@ -71,9 +72,31 @@ fn main() {
...
+    // This workaround is to work with MacOS (or BSD variants which
+    // doesn't allow v4 and v6 can be bind() in one socket).
+    // Note that linux doesn't need this because it can handle v4 and v6 socket
+    // when bind() with "::".
+    let mut addrs_iter = url.to_socket_addrs().unwrap();
+    let addr = addrs_iter.next(); // get next IP address resolved from URL hostname
+    let bind_addr;
+    match addr {
+        Some(ip) => {
+            debug!("addr = {:?}", ip);
```
LPardue: Something that is a little annoying with http3-client is that it doesn't give much feedback on what is happening when no RUST_LOG var is set. If we are going to add some logging anyway, maybe we should consider to print!() a string that states "Making request to <URL>, resolved to IP".

examples/http3-client.rs
junhochoi: SGTM

examples/http3-client.rs
junhochoi: SGTM

examples/http3-client.rs
junhochoi: We can use `peer_addr` instead of `&url`.

examples/http3-client.rs
junhochoi: Yes we need to add a command line option somthing like `-v` and add more curl-like options for storing body into file, which can help easier troubleshooting. Let me work on another PR.

junhochoi:
> Added some comments. Also please run `cargo +nightly fmt` to fix format errors.
> 
> Once this is ready we should apply the same changes to `examples/client.rs` as well.

Updated with recommend changes and ran fmt. Also applied same change to `examples/client.rs` too.

ghedo:
Merged with some minor tweaks as cc8a23f152a385e2defce338aae05f5bbd4e0743, thanks!

PR #64, opened 2019-06-21
Fix ALPN failure detection in the server

LPardue:
NB: I think this is a bit of a workaround fix. There's probably a better way to do it. Opening this for discussion as much as anything.

First, detect the case when the server cannot pick any of the ALPN IDs that the client has stated support.
Second, when generating the CONNECTION_CLOSE frame, SSL_quic_write_level() will tell us the epoch is Application. This is confusing because as far as the client is concerned, the epoch is Initial; sending a CONNECTION_CLOSE using Application will cause the client to drop the packet on the floor and all hell breaks lose. So for this case choose to use Initial epoch always.


src/lib.rs
```diff
@@ -2362,6 +2370,11 @@ impl Connection {
 
     /// Selects the packet number space for outgoing packets.
     fn write_epoch(&self) -> Result<packet::Epoch> {
+        // On error during handshake, latest space is always Initial
+        if self.error.is_some() && !self.handshake_completed {
+            return Ok(packet::EPOCH_INITIAL);
```
junhochoi: Why we want to go back to Initial? I think we simply need to follow current crypto level. If we are in Application or Handshake level already, going back to Initial doesn't guarantee the packet is decrypted properly. https://tools.ietf.org/html/draft-ietf-quic-tls-20#section-4.10

ghedo:
Note that while sending empty ALPN is not optimal, it should be fine as long as the handshake doesn't complete without a protocol being negotiated, which we already enforce https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L2310

In any case I don't think the change here is correct as it assumes that ALPN is processed on the first call to `SSL_do_handshake` which is not guaranteed. Also, always sending handshake errors in Initial packets is probably wrong as the peer might have already dropped the Initial keys at that point and it also violates a SHOULD in https://tools.ietf.org/html/draft-ietf-quic-transport-20#section-10.3

>Prior to handshake confirmation, the peer might not have 1-RTT keys, so the endpoint SHOULD send CONNECTION_CLOSE frames in a Handshake packet. If the endpoint does not have Handshake keys, it SHOULD send CONNECTION_CLOSE frames in an Initial packet.

To implement this properly, I think not sending an empty ALPN needs to be enforced inside BoringSSL itself.

LPardue:
> To implement this properly, I think not sending an empty ALPN needs to be enforced inside BoringSSL itself.

Yeah I'm coming around to this now. Some of the server behavior I saw in testing depends on the client, which is why this proposed hack is not robust. Interestingly, this topic seems to have been discussed a while back - https://bugs.chromium.org/p/chromium/issues/detail?id=497770

LPardue:
Let's take a different approach, closing this PR.

Issue #65, opened 2019-06-21
Fix server timeout caused by a packet loss at Handshake


junhochoi:
When some of Handshake packet (especially ack of client) is missing, server still received h3 request and prepare to send response, it's not sent to the client. Also in this situation client sent ack to ack from server of Handshake packet which is not allowed.

junhochoi:
pr #66 created.

LPardue:
> When some of Handshake packet is missing

The client or the server Handshake packet?

I'm not too familiar with this area of the code. It seems odd the server would receive request before the handshake completed properly. Am I missing something obvious?

junhochoi:
> > When some of Handshake packet is missing
> 
> The client or the server Handshake packet?
> 
> I'm not too familiar with this area of the code. It seems odd the server would receive request before the handshake completed properly. Am I missing something obvious?

Commit (log and source) has more detailed scenario.

LPardue:
Thanks, I have a better understanding of the problem and the fix. Ileft some code comments on your PR.

PR #66, opened 2019-06-21
Fix server timeout caused by a packet loss at Handshake

junhochoi:
Update stream.send.max_data when handshake is complete.
- add new get_max_data() for accessing send.max_data
- change update_max_data()

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2354,10 +2354,7 @@ impl Connection {
                     // (handshake is complete before getting
                     // request) it will not do anything because
                     // send.get_max_data() != 0
-                    for (id, stream) in self
-                        .streams
-                        .iter_mut()
-                    {
+                    for (id, stream) in self.streams.iter_mut() {
                         // we use same value of Stream::new()
                         stream.send.update_max_data(
                             self.peer_transport_params
@@ -2954,7 +2951,20 @@ pub mod testing {
         }
     }
 
+    // receive and send packet
+    // read until consuming all "len" bytes of packets (can be multiple)
+    // and send 0 or more packets and return sent size
     pub fn recv_send(
+        conn: &mut Connection, buf: &mut [u8], len: usize,
+    ) -> Result<usize> {
+        let left = testing::recv(conn, buf, len)?;
+        assert_eq!(left, 0);
+
+        testing::send(conn, buf, 0)
+    }
+
+    // read packets up to "len" size
+    pub fn recv(
         conn: &mut Connection, buf: &mut [u8], len: usize,
     ) -> Result<usize> {
         let mut left = len;
@@ -2969,9 +2979,17 @@ pub mod testing {
             }
         }
 
-        assert_eq!(left, 0);
+        Ok(left)
+    }
 
+    // write packets and return total size sent
+    // count is the number of packets to be sent.
+    // if count = 0, no limit
+    pub fn send(
+        conn: &mut Connection, buf: &mut [u8], count: i32,
+    ) -> Result<usize> {
         let mut off = 0;
+        let mut count_left = count;
 
         while off < buf.len() {
             match conn.send(&mut buf[off..]) {
@@ -2980,6 +2998,11 @@ pub mod testing {
                 Err(Error::Done) => break,
 
                 Err(e) => return Err(e),
+            }
+
+            count_left -= 1;
+            if count_left == 0 {
+                return Ok(off);
             }
         }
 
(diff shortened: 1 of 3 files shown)


junhochoi:
#65

LPardue:
BTW this build failure is due to a change in rustfmt. Not that it matter too much, but we can squelch it with a a 1-liner https://github.com/cloudflare/quiche/pull/64/commits/591feb4233eb2d3f2529a93704f9b6c74e311527

src/lib.rs
```diff
@@ -2341,6 +2341,26 @@ impl Connection {
...
+                    // Handshake packet loss but received actual request already,
+                    // so the stream (usually id 0) is already opened and have some
+                    // data to write, but actual packetization is prevented because of
+                    // max_tx_data = 0.
+                    // What we do here is update max_data if max_data=0
+                    // In normal situation (handshake is complete before getting
+                    // request) it will not do anything.
+                    for (id, stream) in self.streams.iter_mut() {
+                        if stream.send.get_max_data() == 0 {
+                            stream.send.update_max_data(self.max_tx_data);
```
LPardue: `self.max_tx_data` is the connection-wide `peer_params.initial_max_data`, so I think this inadvertently sets the stream credit window to a size that is generally larger. In the case of quiche examples, there is an order of magnitude difference between the values. We should probably use `peer_params.initial_max_stream_data_bidi_remote` (which is what stream.rs does internally when it creates the stream object).

src/lib.rs
```diff
@@ -2341,6 +2341,26 @@ impl Connection {
...
+                    // already written. This can happen when there is a
+                    // Handshake packet loss but received actual request already,
+                    // so the stream (usually id 0) is already opened and have some
+                    // data to write, but actual packetization is prevented because of
+                    // max_tx_data = 0.
+                    // What we do here is update max_data if max_data=0
+                    // In normal situation (handshake is complete before getting
+                    // request) it will not do anything.
+                    for (id, stream) in self.streams.iter_mut() {
+                        if stream.send.get_max_data() == 0 {
```
LPardue: It is possible to condense this to:

```
for (id, stream) in self.streams.iter_mut().filter(|(_, s)| s.send.get_max_data() == 0) {
 ...
}
```

although you'll need to make `get_max_data` non-mutable

src/stream.rs
```diff
@@ -564,6 +570,11 @@ impl SendBuf {
         Ok(out)
     }
 
+    /// Return current max_data
+    pub fn get_max_data(&mut self) -> usize {
```
```suggestion
    pub fn get_max_data(&self) -> usize {
```
Can remove the mutable if just reading.

src/lib.rs
junhochoi: `self.max_tx_data` is initialized when handshake completes and will be used when a stream is created by Stream::new(). When stream is created (using `get_or_create()`) it will use `peer_params.initial_max_stream_data_bidi_local`, not `remote`. So I don't need to use peer_params value explicitly here.

src/lib.rs
junhochoi: Thanks. I tried it before and wondering why build failed...

src/lib.rs
LPardue: I don't follow. On the server side, the stream is created on reception of a Stream frame (https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L1362). It seems the problem is triggered when the server receives STREAM frame before do_handshake() returns Ok?

If thats the case then the stream is constructed from default transport paramers of 0 stream window size. In the linked code:

```
let max_tx_data = self
                        .peer_transport_params
                        .initial_max_stream_data_bidi_local
                        as usize;
```
that value propagates down to into the stream constructor

```
Stream {
            recv: RecvBuf::new(max_rx_data),
            send: SendBuf::new(max_tx_data),
        }
```

But with this PR, we read the peer's initial max data into the Connection (self) and then use that value to update the stream's send buff tx size. That can't be right unless I'm missing something.

```
self.max_tx_data = peer_params.initial_max_data as usize;
...
stream.send.update_max_data(self.max_tx_data);
```

junhochoi: Got it. So I was somehow confused with Connection.max_data and Stream.send.max_data. We need to use initial_max_stream_data_bidi_local. I will update.

junhochoi: Fixed in 6fe07a4

src/lib.rs
```diff
@@ -2341,6 +2341,37 @@ impl Connection {
...
+                        .filter(|(_, s)| s.send.get_max_data() == 0)
+                    {
+                        // we use same value of Stream::new()
+                        stream.send.update_max_data(
+                            self.peer_transport_params
+                                .initial_max_stream_data_bidi_local
+                                as usize,
+                        );
+                        trace!(
+                            "stream_id={} update stream.send.max_data={}",
```
LPardue: we probably want to log `&self.trace_id` too, so that server logs can disambiguate which connection is seeing this behaviour

ghedo:
Also, I fixed the rustfmt config in master, so please rebase to resolve the build failures.

LPardue:
> The fundamental problem here is that we process STREAM frames as well as allow stream writes (including creating stream state) before the handshake is complete, which is probably wrong, so maybe we should just prevent that instead?

That was one of my initial thoughts but I figured it was by done on purpose to avoid blocking request processing while waiting for Handshake retransmits.

As I get more familiar with this part of the library, it seems weirdly inconsistent that we would start doing  things like request processing, while simultaneously delaying the read of the peer's transport params until handshake_complete.

src/lib.rs
junhochoi: Updated

src/lib.rs
```diff
@@ -2341,6 +2341,37 @@ impl Connection {
...
+                    // What we do here is update stream.send.max_data if
+                    // max_data=0 to its initial value used in
+                    // Stream::new(). In normal situation
+                    // (handshake is complete before getting
+                    // request) it will not do anything because
+                    // send.get_max_data() != 0
+                    for (id, stream) in self
+                        .streams
+                        .iter_mut()
+                        .filter(|(_, s)| s.send.get_max_data() == 0)
```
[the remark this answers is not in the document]
junhochoi: LGTM

src/stream.rs
```diff
@@ -110,6 +110,12 @@ impl StreamMap {
                 };
 
                 let s = Stream::new(max_rx_data, max_tx_data);
+                trace!(
```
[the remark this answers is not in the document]
junhochoi: LGTM

src/lib.rs
```diff
@@ -2341,6 +2341,37 @@ impl Connection {
...
+                    {
+                        // we use same value of Stream::new()
+                        stream.send.update_max_data(
+                            self.peer_transport_params
+                                .initial_max_stream_data_bidi_local
```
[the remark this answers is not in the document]
junhochoi: In this case this stream (id 0) is initiated by the client but write is blocked from server side, so I think `initial_max_stream_data_bidi_local` is right and if handshake was done beforehand this was initialized with `initial_max_stream_data_bidi_local`.

> I'll look into doing some refactoring to make this work in the other places
So I think you will create another patch, so I will leave it as is here.

junhochoi:
> > The fundamental problem here is that we process STREAM frames as well as allow stream writes (including creating stream state) before the handshake is complete, which is probably wrong, so maybe we should just prevent that instead?
> 
> That was one of my initial thoughts but I figured it was by done on purpose to avoid blocking request processing while waiting for Handshake retransmits.
> 
> As I get more familiar with this part of the library, it seems weirdly inconsistent that we would start doing things like request processing, while simultaneously delaying the read of the peer's transport params until handshake_complete.

While looking at it I thought the similar thing but for now I think it is ok/right to delay stream write until handshake is done. On the other hand we may reject stream creation until handshake is complete but it may mean a packet loss from peer side, causing retransmit.

junhochoi:
> In any case we need a test for this case, can you look into adding one @junhochoi ?

Test case added at 7fe2f59

ghedo:
@LPardue yes, this was originally intended to allow support for receiving 0-RTT data though it hasn't been implemented yet. But note that applications can already enforce not reading data (e.g. process HTTP requests) before connection is established (and indeed we do in our applications), but the problem is that STREAM frames are still processed by the library, so stream state is initialized anyway.

@junhochoi I'm thinking maybe we could still process (decrypt and parse) a limited number of short header ("Application") packets before handshake is complete, but instead of processing the frames immediately they could be buffered and processed when the handshake is completed. This way we wouldn't need to drop packets, unless we receive more than a certain amount (5? 10? To avoid potential DoS).

ghedo:
I made 188588cdad2b6b65a31ed1cc98e3040444c8490b which should make it easier to implement the frame buffering I suggested in my previous comment, if someone wants to try.

src/lib.rs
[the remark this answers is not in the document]
junhochoi: Test case added at 7fe2f59

junhochoi:
> @junhochoi I'm thinking maybe we could still process (decrypt and parse) a limited number of short header ("Application") packets before handshake is complete, but instead of processing the frames immediately they could be buffered and processed when the handshake is completed. This way we wouldn't need to drop packets, unless we receive more than a certain amount (5? 10? To avoid potential DoS).

Sounds like a good idea. e.g. client is doing POST with a big file. up to initcwnd (10?) will be acceptable. I think QUIC transport draft has something similar.

ghedo:
I looked into implementing the frame buffering thing and made https://github.com/cloudflare/quiche/pull/68. Note that this still allows applications to send data before the handshake is completed, but that can be implemnted separately (and shouldn't be a problem right now for the HTTP/3 use-case).

junhochoi:
> I looked into implementing the frame buffering thing and made #68. Note that this still allows applications to send data before the handshake is completed, but that can be implemnted separately (and shouldn't be a problem right now for the HTTP/3 use-case).

Thanks. I think the major difference here is whether to allow Application read the packets before Handshake completes -- #66 branch allows and #68 branch doesn't allow (instead it buffers packets). Good thing of #66 is during recovering lost Handshake packet server can read request stream, while #68 waits until Handshake completes and let Application read packets afterwards. (so #66 can be a little faster when origin latency is higher). Another aspect is whether to allow Application read packets (stream data) before Handshake fully completes can be a security concern. I think this can be acceptable because until Handshake fully completes server cannot send response back (write is blocked) and if nothing recovered this connection will be closed by timed out eventually.

To limit reading packets while Handshake, I added b008b0f from your #68 branch, basically without buffering, so now server accepts no more than initcwnd packets before Handshake complete.

PTAL @ghedo. Probably we need to choose between one of them.

junhochoi:
#68 provides more compliant solution. Closing this one.

PR #67, opened 2019-06-27
http3-client (rust) improvement

junhochoi:
More curl-like options.
- store response header
- store response body
- add request headers
- timings
- verbose option
- packet loss simulation (simple - only one loss supported)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ url = "1"
 docopt = "1"
 criterion = "0"
 env_logger = "0"
+clap = "~2"
 
 [profile.bench]
 debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ url = "1"
 docopt = "1"
 criterion = "0"
 env_logger = "0"
-clap = "~2"
 
 [profile.bench]
 debug = true
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -27,9 +27,6 @@
 #[macro_use]
 extern crate log;
 
-#[macro_use]
-extern crate clap;
-
 use env_logger::Builder;
 use ring::rand::*;
 use std::fs::File;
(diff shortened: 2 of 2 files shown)


ghedo:
> docopt doesn't accept multiple options with same name.

It does actually, you can do it by specifying `...` in the option value:
```rust
const USAGE: &str = "Usage:
  example [options]

Options:
  --value VALUE ...   Test.
";
```

then you can run the command as `example --value A --value B --value C` and use `get_vec()` https://docs.rs/docopt/1.1.0/docopt/struct.ArgvMap.html#method.get_vec to get a list of values.

In any case I don't think we should implement the packet loss thing in the http3 example client as there are much better ways to achieve the same thing, like using qvalve https://github.com/NTAP/qvalve which has many more features than this, or quicly's udpfw tool https://github.com/h2o/quicly/blob/master/t/udpfw.c  

Some of the other options also seem to have better alternatives that don't require adding more code to the examples, e.g. the functionality of the `--output` option seems to be much more easily achieved by just piping the output of the command to a file (or another command), and the timing option also seems unnecessary to me as h2load built with QUIC and HTTP/3 support is much better at that anyway. Do we really need to implement options for that?

The examples are intended to show how to use the quiche API, so adding more complexity to them makes it harder for people to understand how they work so if we need more curl (or other tools)-like features, IMO we should just finish the HTTP/3 support for curl (or use tools that were specifically intended for that functionality) rather than try to build clones.

(Side comment: maybe we should just use h2load for HTTP/1.x, HTTP/2 and HTTP/3 testing in the lab? That would give more consistency and we could also test doing multiple requests in the same connection, and multiple concurrent connections and many other things more easily, not to mention that we wouldn't test our HTTP/3 server by using a client that uses the same code).

Anyways, looking at the changes here, you seem to have reverted some things from master in this PR, I don't think that was on purpose, was it?

And last thing, if you are implementing several new features I think it would be best if they at least were in separate commits, if not in separate PRs altogether, as it would make it easier to review.

junhochoi:
Revert changes with clap and back to Docopt at f01cb00

junhochoi:
Closing now. In the future I may add a subset of features in this patch.

PR #68, opened 2019-06-30
delay processing of frames in Application packets until after the han…

ghedo:
…dshake is complete

If an Application packet is received before the handshake is completed
it would get processed normally, which is potentially dangerous from a
security perspective (e.g. applications could potentially read STREAM
data even if the handshake wasn't verified), and causes stream state to
be initialized with the wrong transport parameters in some cases
(potentially causing a deadlock).

This change delays the processing of "early Application frames" (i.e.
frames in Application packets that have been received before the
handshake has completed) to after the connection is established, by
buffering the frames.

An arbitrary limit is put on the maximum number of early Application
packets that can be processed in such a way. If the limit is exceeded,
the additional packets are dropped instead, forcing the peer to resend
them.

Note that this is an optimization, as simply dropping all early Application
packets would have also fixed the problems mentioned above, but would
have introduced additional latency as packets would have needed to be
resent.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -623,6 +623,14 @@ pub struct Connection {
     /// Received address verification token.
     token: Option<Vec<u8>>,
 
+    /// List of frames from Application packets received before the handshake
+    /// was completed.
+    early_app_frames: Vec<frame::Frame>,
+
+    /// Number of Application packets received before the handshake was
+    /// completed.
+    early_app_pkts: usize,
+
     /// Error code to be sent to the peer in CONNECTION_CLOSE.
     error: Option<u16>,
 
@@ -877,6 +885,10 @@ impl Connection {
             odcid: None,
 
             token: None,
+
+            early_app_frames: Vec::new(),
+
+            early_app_pkts: 0,
 
             error: None,
 
@@ -1238,6 +1250,22 @@ impl Connection {
             return Err(Error::Done);
         }
 
+        // Keep track of the number of Application packets received before the
+        // handshake is completed, and drop any that exceed the initial
+        // congestion window packet count.
+        if hdr.ty == packet::Type::Application && !self.is_established() {
+            self.early_app_pkts += 1;
+
+            if self.early_app_pkts > recovery::INITIAL_WINDOW_PACKETS {
+                error!(
+                    "{} dropped early application packet len={} pn={}",
+                    self.trace_id, payload_len, pn,
+                );
+
+                return Ok(header_len + payload_len);
+            }
+        }
+
         // Keep track of how many bytes we received from the client, so we
         // can limit bytes sent back before address validation, to a multiple
         // of this. The limit needs to be increased early on, so that if there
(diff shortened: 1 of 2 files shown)


ghedo:
This can be tested independently by using [qvalve](https://github.com/NTAP/qvalve) with the following rule:
```
> h1 drop
```
save that into a file (`pr68.qv`) and then run qvalve, the http3-server and http3-client as follows:
```
 % qvalve -ra 127.0.0.1 -rp 4433 -la 127.0.0.1 -lp 4444 -r pr68.qv
```
```
 % RUST_LOG=info target/release/examples/http3-server --no-retry
```
```
 % RUST_LOG=info target/release/examples/http3-client https://127.0.0.1:4444/ --no-verify --wire-version ff000014
```

src/lib.rs
```diff
@@ -1260,6 +1288,13 @@ impl Connection {
                 ack_elicited = true;
             }
 
+            // If the packet this frame belongs to is an early Application one,
+            // buffer the frame for later processing.
+            if !self.is_established() && hdr.ty == packet::Type::Application {
```
LPardue: nit: you have a reverse order of logic on https://github.com/cloudflare/quiche/pull/68/files#diff-b4aea3e418ccdb71239b96952d9cddb6R1256, is there some reason for that?

src/lib.rs
```diff
@@ -2074,7 +2109,7 @@ impl Connection {
     /// Continues the handshake.
     ///
     /// If the connection is already established, it does nothing.
```
LPardue: Maybe we should describe the parameter `now`?

junhochoi:
Thanks. I left my comment at https://github.com/cloudflare/quiche/pull/66#issuecomment-507461284
BTW qvalue looks useful. 👍

junhochoi:
I looked QUIC TLS spec and it says the following - then this branch is more comformant to the spec.

https://tools.ietf.org/html/draft-ietf-quic-tls-20#section-5.7

> 5.7.  Receiving Out-of-Order Protected Frames
> 
>    Due to reordering and loss, protected packets might be received by an
>    endpoint before the final TLS handshake messages are received.  A
>    client will be unable to decrypt 1-RTT packets from the server,
>    whereas a server will be able to decrypt 1-RTT packets from the
>    client.
> 
>    However, a server MUST NOT process data from incoming 1-RTT protected
>    packets before verifying either the client Finished message or - in
>    the case that the server has chosen to use a pre-shared key - the
>    pre-shared key binder (see Section 4.2.11 of [TLS13]).  Verifying
>    these values provides the server with an assurance that the
>    ClientHello has not been modified.  Packets protected with 1-RTT keys
>    MAY be stored and later decrypted and used once the handshake is
>    complete.
> 
>    A server could receive packets protected with 0-RTT keys prior to
>    receiving a TLS ClientHello.  The server MAY retain these packets for
>    later decryption in anticipation of receiving a ClientHello.

src/lib.rs
ghedo: No reason. Changed this to match the other.

src/lib.rs
ghedo: IDK, to me it seems fairly self-explanatory what a `std::time::Instant` variable called `now` represents TBH.

PR #72, opened 2019-07-04
examples: add Dockerfile

LPardue:
Adds a Dockerfile that works in 2 stages:

1) creates a build image from official Rust sources, installs build prereqs and does a release build of examples with cargo

2) creates a distribution image, ensures system root store is up to date and copies just the binaries from stage 1.

Updated readme.md to show how to use this thing.

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,0 +1,25 @@
+FROM rust:1.35 as build
+
+WORKDIR /usr/src/
+
+RUN apt-get update && apt-get install -y \
+    git \
+    cmake \
+    golang \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN git clone --recurse-submodules --depth 1 https://github.com/cloudflare/quiche
+
+RUN cd quiche && cargo build --release --examples
+
+FROM debian:latest
+RUN apt-get update && apt-get install -y \
+    ca-certificates \
+ && rm -rf /var/lib/apt/lists/*
+RUN update-ca-certificates
+WORKDIR /usr/src/
+COPY --from=build /usr/src/quiche/target/release/examples/http3-client ./
+COPY --from=build /usr/src/quiche/target/release/examples/http3-server ./
+COPY --from=build /usr/src/quiche/target/release/examples/client ./
+COPY --from=build /usr/src/quiche/target/release/examples/server ./
+ENV RUST_LOG=trace
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -17,9 +17,9 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
  && rm -rf /var/lib/apt/lists/*
 RUN update-ca-certificates
-WORKDIR /usr/src/
-COPY --from=build /usr/src/quiche/target/release/examples/http3-client ./
-COPY --from=build /usr/src/quiche/target/release/examples/http3-server ./
-COPY --from=build /usr/src/quiche/target/release/examples/client ./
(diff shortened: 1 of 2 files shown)


examples/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN git clone --recurse-submodules --depth 1 https://github.com/cloudflare/quiche
+
+RUN cd quiche && cargo build --release --examples
+
+FROM debian:latest
+RUN apt-get update && apt-get install -y \
+    ca-certificates \
+ && rm -rf /var/lib/apt/lists/*
+RUN update-ca-certificates
+WORKDIR /usr/src/
```
junhochoi: Installing binaries into /usr/src doesn't looks a good practice. `/usr/local/quiche/bin` or `/usr/local/bin` looks good to me. Also we can use `cargo install ...` instead of copy.

examples/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY --from=build /usr/src/quiche/target/release/examples/http3-client ./
+COPY --from=build /usr/src/quiche/target/release/examples/http3-server ./
+COPY --from=build /usr/src/quiche/target/release/examples/client ./
+COPY --from=build /usr/src/quiche/target/release/examples/server ./
+ENV RUST_LOG=trace
```
junhochoi: Do we need this by default?

examples/Dockerfile
LPardue: Updated to COPY to `/usr/local/quiche/bin` and add that to path.

AFAICT the `cargo install` command won't help much here because we are copying from one container to the other. So even if we installed on the build machine, we'd still need the COPY. Separate COPYs create individual layers that make speed up docker build times if changes modify examples in isolation.

examples/Dockerfile
LPardue: I'm open to change this but I'd argue that most people willing to try this today are interested in seeing what is going on with QUIC and HTTP/3. info level hides a lot of detail and obscures some classes of error. I'd like to avoid fielding questions where it seems like the examples are not doing anything because of transient errors.

LPardue: FYI I changed this to `info` and it can be overridden by inquisitive souls.

examples/Dockerfile
junhochoi: > Updated to COPY to `/usr/local/quiche/bin` and add that to path.
In a new commit it's `/usr/local/bin/quiche`, not `/usr/local/quiche/bin`.

examples/Dockerfile
junhochoi: SGTM. Thanks

examples/README.md
```diff
@@ -18,3 +18,21 @@ Simply run `make` in this directory.
...
+To build:
+
+```
+docker build -t cloudflare-quiche .
+```
+
+To run http3-client once the image is built:
+
+```
+docker run --entrypoint '/bin/bash' cloudflare-quiche -c "http3-client https://cloudflare-quic.com"
```
ghedo: Do we need the `--entrypoint` parameter? Ideally we'd just do:
```
docker run -it cloudflare-quiche http3-client https://cloudflare-quic.com
```
or something like that.

examples/Dockerfile
LPardue: oops, I fixed it now

examples/README.md
LPardue: Default entrypoint is now http3-client, so all user needs to provide is a URL like your example. H@ck3r$ can override this if they really want to.

PR #73, opened 2019-07-05
Add support for sending POST (and other methods) requests with body to http3-client

ghedo:
This adds support for setting a custom HTTP method in http3-client and send request body.

It's not a whole lot of additional complexity IMO so adding this here shouldn't be too bad, though lsquic-client also supports this (and maybe h2load?), so I'm fine to leave this out as well.

### DIFF
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -38,6 +38,7 @@ const USAGE: &str = "Usage:
   http3-client -h | --help
 
 Options:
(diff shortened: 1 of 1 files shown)


junhochoi:
LGTM.

I have one question - what if POST request body is big? Do we have a size limit of send_body() or interface for streaming upload?

ghedo:
@junhochoi you can call `send_body()` multiple times to do streaming, if that's what you are asking. Though in the examples we just do a single call for simplicity.

PR #74, opened 2019-07-05
add stream_shutdown() API

ghedo:
This allows applications to discard incoming or outgoing stream data. For
example an application might want to ignore incoming HTTP request data
in case of errors. Without this, the application would have to consume
all the incoming data in the stream manually, otherwise the HTTP/3 layer
would keep reporting "Data" events forever.

Currently that's all this API does currently, but in the future it can
be used to trigger sending STOP_SENDING and RESET_STREAM frames for
specific streams.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -366,6 +366,20 @@ impl std::convert::From<octets::BufferTooShortError> for Error {
     }
 }
 
+/// The stream's side to shutdown.
+///
+/// This should be used when calling [`stream_shutdown()`].
+///
+/// [`stream_shutdown()`]: struct.Connection.html#method.stream_shutdown
+#[repr(C)]
+pub enum Shutdown {
+    /// Stop receiving stream data.
+    Read  = 0,
+
+    /// Stop sending stream data.
+    Write = 1,
+}
+
 /// Stores configuration shared between multiple connections.
 pub struct Config {
     local_transport_params: TransportParams,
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1918,6 +1932,40 @@ impl Connection {
         Ok(buf.len())
     }
 
+    /// Shuts down reading or writing from/to the specified stream.
+    ///
+    /// When the `direction` argument is set to [`Shutdown::Read`], outstanding
+    /// data in the stream's receive buffer is dropped, and no additional data
+    /// is added to it. Data received after calling this method is still
+    /// validated and ACKed but not stored, and [`stream_recv()`] will not
+    /// return it to the application.
```
LPardue: So IIUC until we add RESET_STREAM or STOP_SENDING, the library will continue to receive data, drop it and not insrease stream flow control window?

src/lib.rs
```diff
@@ -1918,6 +1932,40 @@ impl Connection {
...
+    ///
+    /// When the `direction` argument is set to [`Shutdown::Write`], outstanding
+    /// data in the stream's send buffer is dropped, and no additional data
+    /// is added to it. Data passed to [`stream_send()`] after calling this
+    /// method will be ignored.
```
LPardue: How does this work with recovery? Is unacked data still owned by the transport?

ghedo: Unacked data is part of the sent packets queue, so yes. When the packet gets acked we drop it along with the other frames in the packet, and if it doesn't get acked, when the packet is marked as lost we just drop the data anyway (i.e. we don't resend it).

src/lib.rs
ghedo: Yes, though I just realized this also consumes connection-level flow control credits, so maybe we should give those back.

src/lib.rs
junhochoi: I think dropping send buffer data before getting ack is a little concerning unless we want to close the stream here. e.g. let's say we send a stream data, packet lost, we call shutdown(). Then client still try to send with lost packet in its ACK range. Since sender already dropped the stream data, so we don't retransmit as you say (do we send ack??), then client may keep requesting it until timed out.

ghedo: @junhochoi `shutdown()` should only be called if the application is not interested in the stream anymore. In the future we will need to add `STOP_SENDING` and `RESET_STREAM` so that this becomes more robust, but for now we will only use this when terminating requests (e.g. due to errors).

PR #75, opened 2019-07-08
tls: load certs from Windows system store

### LINKED ISSUE #61: Certificate verification fails on Windows native build
The http3-client fails certificate verification when built natively, e.g.

```
$ http3-client https://cloudflare-quic.com
snip...
[2019-06-15T11:43:24.931343300Z TRACE quiche::tls] error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
[2019-06-15T11:43:24.937788700Z ERROR http3_client] 88dab5f9b993280072028d5b4d34d30a4e3c recv failed: TlsFail
```

The cert for cloudflare-quic.com is trusted by my machine. So I suspect this is something related to building quiche in a way that properly imports my system trust store.

LPardue:
Rough first stab at this, PTAL. Fixes #61 .

Since this problem and its solution are windows specific, there is liberal peppering of `#[cfg(windows)]`.  I'm keen to see if this actually works as it appears on paper.

Happy to hear more elegant ways to do this. Tidy up code is not great, windows API example code is a bit rubbish.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -81,6 +81,16 @@ struct SSL_CIPHER(c_void);
 #[repr(transparent)]
 struct X509_VERIFY_PARAM(c_void);
 
+#[allow(non_camel_case_types)]
+#[repr(transparent)]
+#[cfg(windows)]
+struct X509_STORE(c_void);
+
+#[allow(non_camel_case_types)]
+#[repr(transparent)]
+#[cfg(windows)]
+struct X509(c_void);
+
 #[repr(C)]
 #[allow(non_camel_case_types)]
 struct SSL_QUIC_METHOD {
@@ -123,11 +133,13 @@ pub struct Context(*mut SSL_CTX);
 impl Context {
     pub fn new() -> Result<Context> {
         unsafe {
-            let ctx = SSL_CTX_new(TLS_method());
+            let ctx_raw = SSL_CTX_new(TLS_method());
 
-            map_result(SSL_CTX_set_default_verify_paths(ctx))?;
+            let mut ctx = Context(ctx_raw);
 
-            Ok(Context(ctx))
+            ctx.load_ca_certs()?;
+
+            Ok(ctx)
         }
     }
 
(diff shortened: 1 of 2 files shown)


LPardue:
Ha! Clippy build failures because of unused functions...

LPardue:
> it really shouldn't be exposed as public API, and certainly not one that is only available on Windows. The loading should happen in the same place as other trust stores are configured:
https://github.com/cloudflare/quiche/blob/master/src/tls.rs#L128

That sounds way more logical, thnx.

junhochoi:
I have almost zero knowledge on Windows so it's hard to comment but I agree that user don't need to call anything additional API only for Windows.

This PR looks loading system certs, meaning in unix-like platforms BoringSSL does it automatically but not in Windows?

LPardue:
> This PR looks loading system certs, meaning in unix-like platforms BoringSSL does it automatically but not in Windows?

AFAICT, `SSL_CTX_set_default_verify_paths()` work on Linux and MacOS but not Windows (google of the function and windows finds loads of results). I missed that we use this function, so it is the natural place to add platform-specific CA loading.

LPardue:
Now the change only really affect tls.rs; this is much clearer.

src/tls.rs
```diff
@@ -152,6 +179,47 @@ impl Context {
         })
     }
 
+    #[cfg(windows)]
+    fn load_ca_certs(&mut self) -> Result<()> {
```
ghedo: This method name is still pretty generic for being Windows-specific.

How about you define a separate implementation of the method for ``#[cfg(not(windows))]` that only does the `SSL_CTX_set_default_verify_paths()` call, so then in `new()` you'd just call `load_ca_certs()` unconditionally.

src/tls.rs
```diff
@@ -679,6 +747,15 @@ extern {
 
     fn SSL_CTX_set_default_verify_paths(ctx: *mut SSL_CTX) -> c_int;
 
+    #[cfg(windows)]
+    fn SSL_CTX_get_cert_store(ctx: *mut SSL_CTX) -> *mut X509_STORE;
+    #[cfg(windows)]
+    fn X509_STORE_add_cert(ctx: *mut X509_STORE, x: *mut X509) -> c_int;
```
ghedo: Nitpick, but the functions in this block are organized by type, so you should move this and the `X509` functions to the end of the block and split them with `// X509_STORE` and `// X509` comments.

src/tls.rs
```diff
@@ -25,6 +25,18 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#[cfg(windows)]
+extern crate winapi;
+#[cfg(windows)]
+use winapi::um::wincrypt::{
+    CertCloseStore,
```
ghedo: Nit: generally we don't group imports like this, we simply expand all the `use`s.

src/tls.rs
```diff
@@ -25,6 +25,18 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#[cfg(windows)]
+extern crate winapi;
```
ghedo: Do you really need this?

ghedo: As in, you generally don't need to explicitly import crates that are defined in `Cargo.toml`, though I don't know if this is special in some way.

src/tls.rs
LPardue: bloody d2i_, I folded that into ``// X509`` but leem know if you want a separate block

src/tls.rs
LPardue: removed all imports/usage

src/tls.rs
LPardue: I've done this but don't have a non-windows machine handy, so may find it breaks

src/tls.rs
```diff
@@ -152,6 +164,60 @@ impl Context {
         })
     }
 
+    #[cfg(not(windows))]
+    fn load_ca_certs(&mut self) -> Result<()> {
+        unsafe { map_result(SSL_CTX_set_default_verify_paths(ctx_raw)) }
```
LPardue: self.as_ptr()
```suggestion
        unsafe { map_result(SSL_CTX_set_default_verify_paths(self.as_ptr())) }
```

PR #76, opened 2019-07-09
schedule writable streams fairly

ghedo:
Instead of always picking the first writable stream from the streams map
(depending on the HashMap iteration order, so not necessarily the stream
with the lower ID), keep track of writable streams in a separate FIFO
queue, and round-robin them.

This preserves the order of stream writes to what the application
decides (i.e. the order of stream_send() calls), and increases stream
concurrency and significantly reduces deviation in response times
between concurrent requests.

For example, running 250 identical HTTP/3 requests concurrently using
h2load shows the following before the change:

```
  finished in 78.66s, 3.18 req/s, 3.12MB/s
  requests: 250 total, 250 started, 250 done, 250 succeeded, 0 failed, 0 errored, 0 timeout
  status codes: 250 2xx, 0 3xx, 0 4xx, 0 5xx
  traffic: 245.25MB (257165011) total, 4.39KB (4500) headers (space savings 58.14%), 238.42MB (250000000) data
                       min         max         mean         sd        +/- sd
  time for request:   133.99ms      78.66s      26.53s      23.46s    60.80%
```

and after:

```
  finished in 76.32s, 3.28 req/s, 3.21MB/s
  requests: 250 total, 250 started, 250 done, 250 succeeded, 0 failed, 0 errored, 0 timeout
  status codes: 250 2xx, 0 3xx, 0 4xx, 0 5xx
  traffic: 245.25MB (257166916) total, 4.39KB (4500) headers (space savings 58.14%), 238.42MB (250000000) data
                       min         max         mean         sd        +/- sd
  time for request:     75.61s      76.31s      76.10s    170.77ms    59.20%
```

In the future this can be adapted to implement a more sophisticated
stream prioritization scheme.

Incidentally this also shows significant improvement in the http3
benchmarks due to the fact that we don't need to loop over all streams
(including the ones that have no data to send) when deciding which
stream to schedule, but only over the ones that are actually writable:

```
  http3/128000            time:   [4.1817 ms 4.8528 ms 5.4428 ms]
                          change: [-52.216% -45.968% -38.798%] (p = 0.00 < 0.05)
                          Performance has improved.
  http3/256000            time:   [10.424 ms 10.667 ms 10.949 ms]
                          change: [-49.763% -48.050% -46.369%] (p = 0.00 < 0.05)
                          Performance has improved.
  http3/512000            time:   [50.051 ms 50.258 ms 50.446 ms]
                          change: [-49.695% -44.301% -38.402%] (p = 0.00 < 0.05)
                          Performance has improved.
  http3/1000000           time:   [82.858 ms 83.813 ms 84.938 ms]
                          change: [-50.842% -49.848% -48.879%] (p = 0.00 < 0.05)
                          Performance has improved.
```

Other benchmarks are unchanged.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1472,7 +1472,15 @@ impl Connection {
 
                     self.tx_data -= data.len();
 
+                    let was_writable = stream.writable();
+
                     stream.send.push(data)?;
+
+                    // If the stream is now writable push it to the writable
+                    // queue, but only if it wasn't already queued.
+                    if stream.writable() && !was_writable {
+                        self.streams.push_writable(stream_id);
+                    }
                 },
 
                 frame::Frame::ACK { .. } => {
@@ -1699,10 +1707,9 @@ impl Connection {
             self.max_tx_data > self.tx_data &&
             left > frame::MAX_STREAM_OVERHEAD
         {
-            // TODO: round-robin selected stream instead of picking the first
-            for (id, stream) in
-                self.streams.iter_mut().filter(|(_, s)| s.writable())
-            {
+            while let Some(stream_id) = self.streams.pop_writable() {
+                let stream = self.streams.get_mut(stream_id).unwrap();
+
                 // Make sure we can fit the data in the packet.
                 let stream_len = cmp::min(
                     left - frame::MAX_STREAM_OVERHEAD,
@@ -1718,7 +1725,7 @@ impl Connection {
                 self.tx_data += stream_buf.len();
 
                 let frame = frame::Frame::Stream {
-                    stream_id: *id,
+                    stream_id,
                     data: stream_buf,
                 };
 
@@ -1729,6 +1736,13 @@ impl Connection {
 
                 ack_eliciting = true;
                 in_flight = true;
+
+                // If the stream is still writable, push it to the back of the
+                // queue again.
+                if stream.writable() {
+                    self.streams.push_writable(stream_id);
+                }
+
                 break;
             }
         }
@@ -1927,7 +1941,15 @@ impl Connection {
 
         // TODO: implement backpressure based on peer's flow control
 
+        let was_writable = stream.writable();
+
         stream.send.push_slice(buf, fin)?;
+
+        // If the stream is now writable push it to the writable queue, but
+        // only if it wasn't already queued.
+        if stream.writable() && !was_writable {
+            self.streams.push_writable(stream_id);
+        }
 
         Ok(buf.len())
     }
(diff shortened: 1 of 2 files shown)


ghedo:
BTW, something that the h2load test doesn't show is that the average and mean TTFB per-request goes down, even though total per-request time goes up, so this doesn't just make all requests go slower :)

junhochoi:
>finished in 76.32s, 3.28 req/s, 3.21MB/s
>                       min         max         mean         sd        +/- sd
>  time for request:     75.61s      76.31s      76.10s    170.77ms    59.20%

This means _every_ request took 75~76 sec? Can we get individual timing per request?
And are those requests are quic streams in a single quic connection or distributed over multiple quic connections?

ghedo:
@junhochoi it's 250 simultaneous requests on the same connection, each with a 1MB response body. So before this change, each response would be sent sequentially (so the first one is the fastest and then they get progressively slower until the last one which is the slowest), but with this change they get sent concurrently so the total time is higher on average, but time to first byte lower.

I don't have timing for each request right now, but I can run the h2load test to get them.

ghedo:
Also note that the test was run against the quiche example HTTP/3 server, I haven't done it against NGINX yet.

src/lib.rs
```diff
@@ -1927,8 +1941,16 @@ impl Connection {
 
         // TODO: implement backpressure based on peer's flow control
 
+        let was_writable = stream.writable();
+
         stream.send.push_slice(buf, fin)?;
 
+        // Push stream to the writable queue, but only if it wasn't already
+        // queued..
+        if !was_writable {
```
junhochoi: No need to check `stream.writable()`?

src/lib.rs
```diff
@@ -3853,6 +3922,58 @@ mod tests {
         let mut r = pipe.server.readable();
         assert_eq!(r.next(), None);
     }
+
+    #[test]
+    fn stream_round_robin() {
```
junhochoi: I think the goal of this test is to check writing order is preserved - write to stream 8, 0, 4 and client read stream 8, 0, 4. Correct? If so it's good to have some comment what this test try to validate.

src/lib.rs
ghedo: The original thinking was that since we are adding new data the stream would be always writable, but now that I think more about it, the stream might not have flow control credits so we should indeed check for `stream.writable()` here as well.

src/lib.rs
ghedo: Fair enough. I added descriptions to both new tests, let me know if they need clarification.

ghedo:
@LPardue I just figured out how to dismiss reviews :D

PR #77, opened 2019-07-10
http3-client: add request headers

junhochoi:
Adding -H/--header "field: value" option to http3-client
to add request headers in the command line. Can be multiple.

Useful when you need more http-related tests.

Currently it can't override default headers (e.g. user-agent).
If you add, there will be 2 headers with same name.

e.g.

```
http3-client -H 'cache-control: no-cache' https://cloudflare-quic.com
http3-client -H 'content-type: application/x-www-form-urlencoded' --method PUT --body test.dat https://httpbin-like...
```

### DIFF
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -248,10 +248,10 @@ fn main() {
             // Add custom headers to the request.
             for header in &req_headers {
                 let header_split: Vec<&str> = header.splitn(2, ": ").collect();
-                // Convert header field to lowercase
-                let (field, value) =
-                    (&header_split[0].to_lowercase().to_string(), header_split[1]);
-                req.push(quiche::h3::Header::new(field, value));
+                req.push(quiche::h3::Header::new(
+                    header_split[0],
+                    header_split[1],
+                ));
             }
 
             info!("sending HTTP request {:?}", req);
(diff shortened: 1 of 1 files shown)


examples/http3-client.rs
```diff
@@ -241,6 +245,15 @@ fn main() {
                 quiche::h3::Header::new("user-agent", "quiche"),
             ];
 
+            // Additional headers
+            for header in &req_headers {
+                let header_split: Vec<&str> = header.splitn(2, ": ").collect();
+                // Convert header field to lowercase
+                let (field, value) =
+                    (&header_split[0].to_lowercase().to_string(), header_split[1]);
```
ghedo: This is already done when calling `Header::new()` so you don't need to do it explciitly here:
https://github.com/cloudflare/quiche/blob/master/src/h3/mod.rs#L464

examples/http3-client.rs
```diff
@@ -71,6 +72,9 @@ fn main() {
 
     let url = url::Url::parse(args.get_str("URL")).unwrap();
 
+    // Request headers (can be multiple)
```
```suggestion
    // Request headers (can be multiple).
```

examples/http3-client.rs
```diff
@@ -241,6 +245,15 @@ fn main() {
                 quiche::h3::Header::new("user-agent", "quiche"),
             ];
 
+            // Additional headers
```
```suggestion
            // Add custom headers to the request.
```

examples/http3-client.rs
junhochoi: Updated. Thanks

ghedo:
Ugh, I forgot to squash before merging, oh well...

PR #82, opened 2019-07-12
http3-client: panic on custom header error

LPardue:
Due to fat fingers I stumbled across an unhelpful error message when I forgot to include a `:` in a header name. So this PR adds some very basic error checking and panics with the value of the bad header.


examples/http3-client.rs
```diff
@@ -248,6 +248,10 @@ fn main() {
             // Add custom headers to the request.
             for header in &req_headers {
                 let header_split: Vec<&str> = header.splitn(2, ": ").collect();
+                if header_split.len() == 1 {
```
ghedo: Seems like `!= 2` would be clearer here?

PR #84, opened 2019-07-20
draft 22

LPardue:
* update salt and test vectors
* update invariants
* update some error-related transport frames
* update H3 error codes
* update priority frame handling - only appears on control stream now
* update alpn IDs


src/packet.rs
```diff
@@ -211,31 +211,11 @@ impl Header {
             }
         };
 
-        let (dcil, scil) = match b.get_u8() {
-            Ok(v) => {
-                let mut dcil = v >> 4;
-                let mut scil = v & 0xf;
+        let dcid_len = b.get_u8()?;
```
ghedo: We should validate the limit for this value.

src/packet.rs
```diff
@@ -211,31 +211,11 @@ impl Header {
...
-
-                (dcil, scil)
-            },
-
-            Err(_) => return Err(Error::BufferTooShort),
-        };
-
-        let dcid = b.get_bytes(dcil as usize)?.to_vec();
-        let scid = b.get_bytes(scil as usize)?.to_vec();
+        let scid_len = b.get_u8()?;
```
ghedo: Same.

src/packet.rs
```diff
@@ -610,18 +573,9 @@ pub fn negotiate_version(
...
-    if !scid.is_empty() {
-        cil |= ((scid.len() - 3) as u8) << 4;
-    }
-
-    if !dcid.is_empty() {
-        cil |= ((dcid.len() - 3) as u8) & 0xf;
-    }
-
-    b.put_u8(cil)?;
+    b.put_u8(scid.len() as u8)?;
```
ghedo: We weren't doing this before, but it would probably be a good idea that the application doesn't pass too long connection IDs.

src/tls.rs
```diff
@@ -590,7 +590,7 @@ extern fn send_alert(ssl: *mut SSL, level: crypto::Level, alert: u8) -> c_int {
     );
 
     let error: u16 = TLS_ALERT_ERROR + u16::from(alert);
-    conn.error = Some(error);
+    conn.error = Some(u64::from(error));
```
ghedo: Probably better to just change `error` on the line above to be u64, so we don't go u8 -> u16 -> u64 (probably needs changing `TLS_ALERT_ERROR` type as well).

LPardue:
Updated to add DCID and SCID length checking for "QUIC v1", along with some unit tests.

aside: "v1" is assumed to be equivalent to `crate::PROTOCOL_VERSION`. At some future point it might help to label the constant so that it is clearly v1 and not something else.

src/packet.rs
```diff
@@ -211,31 +211,17 @@ impl Header {
...
-                    return Err(Error::InvalidPacket);
-                }
-
-                (dcil, scil)
-            },
-
-            Err(_) => return Err(Error::BufferTooShort),
-        };
+        let dcid_len = b.get_u8()?;
+        if version == crate::PROTOCOL_VERSION && dcid_len > 20 {
```
ghedo: Better use `MAX_CID_LEN` here instead of hardcoding the value.
```suggestion
        if version == crate::PROTOCOL_VERSION && dcid_len > MAX_CID_LEN {
```

src/packet.rs
```diff
@@ -211,31 +211,17 @@ impl Header {
...
+            return Err(Error::InvalidPacket);
+        }
+        let dcid = b.get_bytes(dcid_len as usize)?.to_vec();
 
-        let dcid = b.get_bytes(dcil as usize)?.to_vec();
-        let scid = b.get_bytes(scil as usize)?.to_vec();
+        let scid_len = b.get_u8()?;
+        if version == crate::PROTOCOL_VERSION && scid_len > 20 {
```
ghedo: Same.
```suggestion
        if version == crate::PROTOCOL_VERSION && scid_len > MAX_CID_LEN {
```

src/packet.rs
ghedo: This is still valid, but we can do this in a seaprate PR.

src/packet.rs
LPardue: derp, thnx.

src/packet.rs
LPardue: oops I missed this one, agree can do in a separate PR.

LPardue:
all nits addressed AFAICT

ghedo:
Ended up squashing everything together because I'm lazy.

PR #88, opened 2019-07-24
improve per-connection accounting of sent bytes

ghedo:
Instead of increasing the count when data is sent, and decreasing it
when data is marked as lost, only increase the count based on how
streams' maximum offsets change, so we don't need to decrease the count
on data loss.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1469,8 +1469,6 @@ impl Connection {
                         None => continue,
                     };
 
-                    self.tx_data -= data.len();
-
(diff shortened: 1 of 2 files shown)


ghedo:
This probably needs more tests (e.g. for the data lost case), but I haven't been able to write one so far.

ghedo:
After some more testing, it turns out that this doesn't work properly, so for the time being I made #96 to at least workaround the crash found during interop testing, but we'll need to debug this further.

PR #90, opened 2019-07-31
h3: don't error on QPACK instruction

LPardue:
Before this change, *any* data received on a QPACK encoder/decoder stream
would trigger a conn_close with encoder/decoder error type, which could
be the wrong thing to do in some cases. This change allows data on the
streams but since we don't support QPACK dynamic table yet, we simply
put the stream into a draining state and ignore instructions.

With this change in place we can properly detect stream closure, so I
took the chance to update the relevant test.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -2233,7 +2233,7 @@ mod tests {
     }
 
     #[test]
-    /// Client closes QPACK stream, which is forbidden
+    /// Client closes QPACK stream, which is forbidden.
     fn close_qpack_stream() {
         let mut s = Session::default().unwrap();
         s.handshake().unwrap();
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1353,20 +1353,12 @@ impl Connection {
                 },
 
                 stream::State::QpackInstruction => {
-                    stream.try_fill_buffer(conn)?;
+                    let mut d = [0; 4096];
 
-                    let e = match stream.ty() {
-                        Some(stream::Type::QpackEncoder) =>
-                            Error::QpackEncoderStreamError,
-                        Some(stream::Type::QpackDecoder) =>
-                            Error::QpackDecoderStreamError,
-
-                        _ => unreachable!(),
-                    };
-
-                    conn.close(true, e.to_wire(), b"")?;
-
-                    return Err(e);
+                    // Read data from the stream and discard immediately.
+                    loop {
+                        conn.stream_recv(stream_id, &mut d)?;
+                    }
                 },
 
                 stream::State::Drain => {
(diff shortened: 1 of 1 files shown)


src/h3/mod.rs
```diff
@@ -2288,21 +2289,23 @@ mod tests {
...
-        s.send_frame_client(
-            frame::Frame::MaxPushId { push_id: 1 },
-            s.client.local_qpack_streams.encoder_stream_id.unwrap(),
-            true,
-        )
-        .unwrap();
+        s.pipe
+            .client
+            .stream_send(
+                s.client.local_qpack_streams.encoder_stream_id.unwrap(),
```
ghedo: Might be a good idea to save this in a separate variable, so we would avoid the odd formatting, e.g.
```
let stream_id = s.client.local_qpack_streams.encoder_stream_id.unwrap();
s.pipe.client.stream_send(stream_id, ...).unwrap();
```

src/h3/mod.rs
```diff
@@ -2323,6 +2326,53 @@ mod tests {
...
+        // added so we just test that the data is safely ignored.
+        let mut s = Session::default().unwrap();
+        s.handshake().unwrap();
+
+        let mut abrupt_termination = false;
+
+        s.pipe
+            .client
+            .stream_send(
+                s.client.local_qpack_streams.encoder_stream_id.unwrap(),
```
ghedo: Same as above (also below).

src/h3/mod.rs
```diff
@@ -2323,6 +2326,53 @@ mod tests {
...
+        loop {
+            match s.server.poll(&mut s.pipe.server) {
+                Ok(_) => (),
+
+                Err(Error::Done) => {
+                    break;
+                },
+
+                Err(_) => {
+                    abrupt_termination = true;
```
ghedo: Could just `panic!()` here directly, so we wouldn't need this variable.

LPardue:
I agree with your suggestion but interestingly it causes the `close_qpack_stream` test to fail. I think because the stream closure detection fails - https://github.com/cloudflare/quiche/blob/master/src/h3/mod.rs#L997 :/

LPardue:
nits addressed in 041a6fb. I think detecting closure of drain-state streams needs some more thinking.

ghedo:
Thinking about this some more: `Drain` state basically causes `stream_shutdown()` to be called (and really, we should probably get rid of the state and just call shutdown directly instead of doing the whole state transition dance, but that's a separate issue) which sometimes soon will cause `STOP_SENDING` to be sent, which in turn causes `RESET_STREAM` to be sent by the peer, so we would end up closing the connection anyway because the peer tries to close the stream :man_facepalming:.

An alternative option could be to change the `QpackInstruction` state to do the stream draining like we did in the old `Drain` state, basically just looping and discarding all stream data https://github.com/cloudflare/quiche/commit/6e8f6243803613e0e63ea39ea43cf431a42a288f#diff-0679c3c7d9c74055118562f1b8d85df9

```rust
let mut d = [0; 4096];

// Read data from the stream and discard immediately.
loop {
    conn.stream_recv(stream_id, &mut d)?;
}
```
This is not great, but would only be necessary until we implement proper QPACK parsing.

LPardue:
Let me try that

src/h3/mod.rs
```diff
@@ -2241,21 +2233,20 @@ mod tests {
     }
 
     #[test]
-    /// Client sends data on a QPACK stream, which we don't allow.
+    /// Client closes QPACK stream, which is forbidden
```
```suggestion
    /// Client closes QPACK stream, which is forbidden.
```

PR #93, opened 2019-08-04
Update makefile to fix build error

VMatrix1900:
Use `-lquiche` instead of `-l:libquiche.a` to build.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -17,7 +17,7 @@ LIBSSL_DIR = $(dir $(shell find ${BUILD_DIR} -name libssl.a))
 
 LDFLAGS = -L$(LIBCRYPTO_DIR) -L$(LIBSSL_DIR) -L$(LIB_DIR)
 
-LIBS = $(LIB_DIR)/libquiche.a -lev -ldl -pthread
+LIBS = -lquiche -lev -ldl -pthread
 
 all: client server http3-client http3-server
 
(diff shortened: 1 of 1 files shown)


LPardue:
traceback to this change https://github.com/cloudflare/quiche/commit/57e2936bdf49c3b9239f468470109d5689066d77

ghedo:
Yeah, so, this forces building against the static library, because otherwise the examples would be built with the dynamic library which is not in a standard path, so the examples would fail to run.

@VMatrix1900 what build error are you trying to fix with this?

VMatrix1900:
In my Macbook, using `-l:libquiche.a` can not find the library. Changing it to `-l:quiche` fixes it.

examples/Makefile
```diff
@@ -17,7 +17,7 @@ LIBSSL_DIR = $(dir $(shell find ${BUILD_DIR} -name libssl.a))
 
 LDFLAGS = -L$(LIBCRYPTO_DIR) -L$(LIBSSL_DIR) -L$(LIB_DIR)
 
-LIBS = -l:libquiche.a -lev -ldl -pthread
+LIBS = -lquiche -lev -ldl -pthread
```
junhochoi: `-l:libquiche.a` is not understood in MacOS linker but it works with Linux.

If the goal is to build with libquiche.a always, I prefer to use

`
-LIBS = -l:libquiche.a -lev -ldl -pthread
+LIBS = $(LIB_DIR)/libquiche.a -lev -ldl -pthread
`

To work with both platform.

junhochoi: I mistakenly pushed changes (and reverted). my bad

ghedo:
Merged, thanks @VMatrix1900!

PR #96, opened 2019-08-04
avoid per-connection count of sent bytes going negative

ghedo:


ghedo:
Also see https://github.com/cloudflare/quiche/pull/88#issuecomment-518022848

PR #101, opened 2019-08-07
ffi: Fix odcid is present even when empty

nqv:


nqv:
With this handshake test, client got `InvalidTransportParam` if I replace NULL odcid with an empty slice:
https://github.com/goburrow/quiche/blob/master/deps/quiche_test.c

ghedo:
0-length connection ID is technically a valid connection ID, so I _think_ the existing behavior is correct, but the QUIC spec doesn't address this case specifically so I may be wrong.

ghedo:
The problem you are seeing with `InvalidTransportParam` is that the server has to include an ODCID transport parameter when it does stateless retry, but your code changes 0-length connection ID to not be included, so the client fails to validate the server's transport aprameters.

nqv:
Thank you

PR #102, opened 2019-08-07
h3: limit size of stream state buffer

ghedo:
A peer can influence the size of the state buffer (e.g. with the
payload size of a GREASE frame), so we need to limit the maximum
size to avoid DoS.

In the future we might want to have a dynamic limit set by the
application, but for now a static one will do.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1183,7 +1183,11 @@ impl Connection {
                     };
 
                     let ty = stream::Type::deserialize(varint)?;
-                    stream.set_ty(ty)?;
+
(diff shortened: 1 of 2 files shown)


ghedo:
Also added another related commit to close the HTTP/3 connection in some more error cases.

LPardue:
Cool. I think this also applies to insanely large SETTINGS frames which we had planned to fix up.

ghedo:
As far as this change is concerned, currently any non-DATA frame is handled the same way, so the test should apply for SETTINGS too.

PR #103, opened 2019-08-07
examples: shutdown client request streams after reading headers

LPardue:
this seems to fix the problem, alignment/formatting sucks.

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -499,6 +499,13 @@ fn handle_request(
             error!("{} failed to build response {:?}", conn.trace_id(), e);
         },
     }
+
+    // We decide the response based on headers alone.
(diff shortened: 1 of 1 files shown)


examples/http3-server.rs
```diff
@@ -331,6 +331,19 @@ fn main() {
                                 &headers,
                                 args.get_str("--root"),
                             );
+
+                            // We decide the response based on headers alone.
+                            // Stop reading the stream so that a request
+                            // containing a body does not generate pointless
+                            // Data events.
+                            client
```
ghedo: Might be better to put this directly in `handle_request()` to avoid the weird indentation.

LPardue: Yeah

PR #106, opened 2019-08-08
ffi: fix quiche_conn_application_proto function signature

nqv:
It should be similar to SSL_get0_alpn_selected


ghedo:
Merged, thanks @nqv!

PR #109, opened 2019-08-08
drop handshake keys when the handshake is confirmed

ghedo:


src/lib.rs
```diff
@@ -2463,21 +2479,18 @@ impl Connection {
...
     }
 
-    /// Drops the initial keys and recovery state.
-    fn drop_initial_state(&mut self) {
-        if self.pkt_num_spaces[packet::EPOCH_INITIAL]
-            .crypto_open
-            .is_none()
-        {
+    /// Drops the keys and recovery state for the given epoch.
+    fn drop_epoch_state(&mut self, epoch: usize) {
```
```suggestion
    fn drop_epoch_state(&mut self, epoch: packet::Epoch) {
```

We have a typedef for Epoch, so lets use it

ghedo: Oh, yeah, good point, I forgot about that.

ghedo:
Not sure why GitHub is being a jerk, but the PR build succeeded https://travis-ci.com/cloudflare/quiche/builds/122448234

PR #110, opened 2019-08-09
32bit architecture support

junhochoi:
Current quiche master is not working at 32bit OS.

Here is a simple test result in x86 Debian:

```
% uname
Linux buster32 4.19.0-5-686-pae #1 SMP Debian 4.19.37-5+deb10u1 (2019-07-19) i686 GNU/Linux
```

`cargo build` runs successfully, however `cargo test` fails:

```
test result: FAILED. 129 passed; 46 failed; 0 ignored; 0 measured; 0 filtered out
```

Also `http3-client` fails at runtime

```
$ target/debug/examples/http3-client https://cloudflare-quic.com
thread 'main' panicked at 'attempt to multiply with overflow', /usr/src/rustc-1.34.2/src/libcore/num/mod.rs:3516:24
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

In above case, main issue is in `src/frame.rs:863`, where `2usize.pow(62)` overflows in 32bit platform (`usize` is 32bit in 32bit OS). However, since `usize` is used for stream offset, which makes offset calculation is limited to 32bit where QUIC spec requires bigger integer.

This patch mainly changes stream offset and other related parameter (e.g. `max_data` related which specifies stream size) to `u64` and fix for related codes. Type conversion is done for `usize` to `u64` not to lose bits.

Result:
Tested in x86 (Debian Buster) and x86-64 (Debian Stretch) has no test failures and http3-client successfully download the URL.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,7 +252,7 @@ pub const MIN_CLIENT_INITIAL_LEN: usize = 1200;
 
 const PAYLOAD_MIN_LEN: usize = 4;
 
-const MAX_AMPLIFICATION_FACTOR: u64 = 3;
+const MAX_AMPLIFICATION_FACTOR: usize = 3;
 
 /// A specialized [`Result`] type for quiche operations.
 ///
@@ -625,7 +625,7 @@ pub struct Connection {
 
     /// Total number of bytes the server can send before the peer's address
     /// is verified.
-    max_send_bytes: u64,
+    max_send_bytes: usize,
 
     /// Streams map, indexed by stream ID.
     streams: stream::StreamMap,
@@ -1290,7 +1290,7 @@ impl Connection {
         // is an error there is enough credit to send a CONNECTION_CLOSE.
         if !self.verified_peer_address {
             self.max_send_bytes +=
-                ((header_len + payload_len) as u64) * MAX_AMPLIFICATION_FACTOR;
+                (header_len + payload_len) * MAX_AMPLIFICATION_FACTOR;
         }
 
         // To avoid sending an ACK in response to an ACK-only packet, we need
@@ -1499,7 +1499,7 @@ impl Connection {
         }
 
         // Calculate available space in the packet based on congestion window.
-        let mut left: u64 = cmp::min(self.recovery.cwnd(), b.cap()) as u64;
+        let mut left = cmp::min(self.recovery.cwnd(), b.cap());
 
         // Limit data sent by the server based on the amount of data received
         // from the client before its address is validated.
@@ -1532,7 +1532,7 @@ impl Connection {
         // length, the packet number and the AEAD overhead. We assume that
         // the payload length can always be encoded with a 2-byte varint.
         left = left
-            .checked_sub((b.off() + 2 + pn_len + overhead) as u64)
+            .checked_sub(b.off() + 2 + pn_len + overhead)
             .ok_or(Error::Done)?;
 
         let mut frames: Vec<frame::Frame> = Vec::new();
@@ -1557,11 +1557,11 @@ impl Connection {
                 ranges: self.pkt_num_spaces[epoch].recv_pkt_need_ack.clone(),
             };
 
-            if frame.wire_len() as u64 <= left {
+            if frame.wire_len() <= left {
                 self.pkt_num_spaces[epoch].ack_elicited = false;
 
                 payload_len += frame.wire_len();
-                left -= frame.wire_len() as u64;
+                left -= frame.wire_len();
 
                 frames.push(frame);
             }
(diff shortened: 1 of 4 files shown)


LPardue:
It might be useful to have multi-architecture CI builds, so we can spot regressions. Travis makes that a bit difficult because it only offers 64-bit VMs but there is an example of cross compilation here: https://medium.com/@tyrf/multi-target-rust-builds-with-auto-deployment-in-travis-ci-ab5214fb73ba

junhochoi:
Right. It would be useful to have multi architecture build since we supports multiple platforms already, including Android and iOS. However in this issue build runs cleanly on 32bit. Need to run cargo test or http3-client (or other command line tools) to make sure it works.

src/lib.rs
```diff
@@ -1499,7 +1499,7 @@ impl Connection {
         }
 
         // Calculate available space in the packet based on congestion window.
-        let mut left = cmp::min(self.recovery.cwnd(), b.cap());
+        let mut left: u64 = cmp::min(self.recovery.cwnd(), b.cap()) as u64;
```
ghedo: IMO `left` should be kept as `usize` since it tracks the size of a memory buffer.

src/stream.rs
```diff
@@ -300,27 +300,27 @@ pub struct RecvBuf {
     data: BinaryHeap<RangeBuf>,
 
     /// The lowest data offset that has yet to be read by the application.
-    off: usize,
+    off: u64,
 
     /// The total length of data received on this stream.
     len: usize,
```
ghedo: If `off` is `u64`, then this should be as well, as in general it is larger than `off`.

src/stream.rs
```diff
@@ -544,24 +544,24 @@ pub struct SendBuf {
     data: BinaryHeap<RangeBuf>,
 
     /// The maximum offset of data buffered in the stream.
-    off: usize,
+    off: u64,
 
     /// The amount of data that was ever written to this stream.
     len: usize,
```
ghedo: Same.

src/stream.rs
```diff
@@ -617,9 +617,10 @@ impl SendBuf {
     }
 
     /// Returns contiguous data from the send buffer as a single `RangeBuf`.
-    pub fn pop(&mut self, max_data: usize) -> Result<RangeBuf> {
+    pub fn pop(&mut self, max_data: u64) -> Result<RangeBuf> {
```
ghedo: I think `max_data` should stay as `usize` as its value depends on `left` and the size of the output buffer.

LPardue:
CI currently runs cargo test and executes all our unit tests. 

Having full coverage of build targets is tricky but if there is a way to edge towards that I think it helps, otherwise we risk just continually breaking some targets without ever knowing.

src/stream.rs
junhochoi: `max_data` is from transport parameter or MAX_DATA frame which is variable int, so need to be `u64` I think.

junhochoi:
> CI currently runs cargo test and executes all our unit tests.
> 
> Having full coverage of build targets is tricky but if there is a way to edge towards that I think it helps, otherwise we risk just continually breaking some targets without ever knowing.

I think something like https://github.com/jdub/travis-qemu-example looks useful.

src/stream.rs
junhochoi: Updated at 1116ab8

src/stream.rs
junhochoi: Updated at 1116ab8

src/lib.rs
junhochoi: Updated at 1116ab8

junhochoi:
Updated based on the review. No test failures on 32bit/64bit.

src/stream.rs
junhochoi: Oh I understand now. It's different from MAX_DATA... confused by name. I will make changes.

junhochoi: Fixed at 5fffab9

src/lib.rs
```diff
@@ -888,8 +888,8 @@ impl Connection {
             sent_count: 0,
 
             rx_data: 0,
-            max_rx_data: max_rx_data as usize,
-            max_rx_data_next: max_rx_data as usize,
+            max_rx_data: max_rx_data as u64,
+            max_rx_data_next: max_rx_data as u64,
```
```suggestion
            max_rx_data_next: max_rx_data,
```

src/lib.rs
```diff
@@ -888,8 +888,8 @@ impl Connection {
             sent_count: 0,
 
             rx_data: 0,
-            max_rx_data: max_rx_data as usize,
-            max_rx_data_next: max_rx_data as usize,
+            max_rx_data: max_rx_data as u64,
```
ghedo: The two values above should already be `u64` so you don't need to cast them I think.
```suggestion
            max_rx_data,
```

PR #120, opened 2019-08-15
tools: initial version of httpbin test tool

LPardue:
This adds a new binary application to the quiche project under /tools. 

In a nutshell, you can run a test against an httpbin server. Each test can consist of 1 or more requests, along with code to check the actual response against the expected one. The public API is intended to be simple, you create a test based on a string identifer, you call `send_requests()`, store responses with `append_response_{header|data}()`, and finally call `assert()`. The binary returns 0 in a good case, -1 in a test fail case, or panics in many other cases :D (we probably want to change that but it works for what we need today)

main.rs - is basically a modified version of http3-client.rs; some options removed and a new `--test` one added. IT would be nice for it to list the valid test values but didn't figure that out a way to do that without manually writing it.

lib.rs is where the action is at. There's code that helps to build and maintain a struct for test requests and assertions.

There are some scripts that show how this could be linked into pytest or other testing framework.


LPardue:
need to rebase to fix build failure.

ghedo:
I had an idea :smiling_imp: Instead of building a whole new tool for this, maybe we could just use Rust's test system... hear me out :smile: 

We would still have a separate crate in `tools/`, but instead it implementing the command line tool (with all the argument parsing), it would just be `src/lib.rs` with all the `test_*` functions turned into `#[test]` functions, as well as providing all the helpers you implemented here (`HttpBinTest` and the like), plus some kind of wrapper for setting up the connection (see e.g. [this](https://medium.com/@ericdreichert/test-setup-and-teardown-in-rust-without-a-framework-ba32d97aa5ab) or even [this](https://doc.rust-lang.org/unstable-book/language-features/custom-test-frameworks.html)).

Why the fuck would we want to do that, you ask. Well, Rust tests are already built into a single executable, which provides filtering by name from the command-line, as well as pretty output. This would also mean that we can write the tests as if they were normal Rust tests and use all the APIs already provided.

The downside is that tests don't take arbitrary arguments, but for passing the target server and path we can just use environment variables I think.

LPardue:
I like it! Why?

Http3bin-client.rs has next to no value - it is a ton of duplicate code from the client. That means we have to maintain yet another client, and experience to date has shown most people forget to update them.

Figuring out a nice way to bootstrap a real connection in rust test framework would have a ton of value because it's something we can reuse. For example, build a test runner for interop, other forms of online tests etc.

This changes the scope of original intent slightly but I think the spirit is the same.

LPardue:
So I took an attempt at this but because of the large changes I didn't nuke the previous python-oriented commandline version (also I somehow added some of your commits). I'll tidy all of this up before the PR is merge ready.

The design is as proposed, `http3client-test/src/lib.rs` is just concerned with building a HttpBinTest, which it feeds to `runner.rs` (a quiche client in disguise). This has the nice side effect of meaning that all concrete test code is arranged together in a named function. You can just run `cargo test` and it goes through the 29 test cases.

Because these are integration tests (not unit tests of the HttpBinTest class and helpers) I figured it made sense to create a separate `tests` folder, move the test cases to a new file and move runner.rs there too. However, I got into some Rust module mess when I tried that and I couldn't be bothered trying to fix it.

LPardue:
Also for giggles, I thought it might be funny to get Travis to run these tests too - I'm sure that will come back to bite us so we might not want it.

LPardue:
> Also for giggles, I thought it might be funny to get Travis to run these tests too - I'm sure that will come back to bite us so we might not want it.

I'm surprised but it actually worked. We could probably add an envvar to help manage migration between versions (e.g. if client version != endpoint version don't run the tests because they will just fail)

```
running 29 tests

test tests::basic_auth ... ok

test tests::bytes ... ok

test tests::cache ... ok

test tests::cookies ... ok

test tests::deflate ... ok

test tests::delete ... ok

test tests::delay ... ok

test tests::drip ... test tests::drip has been running for over 60 seconds

test tests::drip ... ok

test tests::encode_utf8 ... ok

test tests::form ... ok

test tests::get ... ok

test tests::gzip ... ok

test tests::headers ... ok

test tests::html ... ok

test tests::image ... ok

test tests::ip ... ok

test tests::links ... ok

test tests::patch ... ok

test tests::post ... ok

test tests::put ... ok

test tests::redirect ... ok

test tests::range ... ok

test tests::response_headers ... ok

test tests::robots ... ok

test tests::stream ... ok

test tests::stream_bytes ... ok

test tests::useragent ... ok

test tests::xml ... ok

test tests::status ... ok

test result: ok. 29 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

.travis.yml
```diff
@@ -20,12 +20,18 @@ matrix:
       - cargo clippy --examples -- -D warnings
       - cargo doc --no-deps
       - make -C examples
+      - cd tools/httpbintest
```
ghedo: Pro tip: you can pass `--manifest-path` to `cargo build` and `cargo test` directly, so you don't need to cd into the directory.

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//! 🗑️ HTTP/3 httpbin test client helper.
```
ghedo: I'm thinking maybe we shouldn't reference httpbin directly? This crate is useful for tests that involve httpbin too.

Maybe we should just call the tool something like `http3-integration-tests`?

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+
+use quiche::h3::Header;
+
+const USER_AGENT: &str = "quiche-http3bin";
+
+// Stores the request, the expected response headers, and the actual response.
+// The assert_hdrs() method is provided for convenience to validate the received
+// headers match the expected headers.
+#[derive(Clone)]
+pub struct HttpBinReq {
```
ghedo: Same regarding mentioning httpbin, maybe just call `HttpTestReq`? This also applies to the other types below.

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+        test.set_concurrency(concurrent);
+        assert!(crate::runner::run(&mut test));
+    }
+
+    // Build a single request and expected response with status code
+    fn request_check_status(testpoint: &str, status: usize) -> Vec<HttpBinReq> {
+        let expect_hdrs = Some(vec![Header::new(":status", &status.to_string())]);
+
+        let mut url = endpoint();
+        let path = format!("{}/{}", url.path(), &testpoint);
```
ghedo: This seems a pretty common operation, maybe `endpoint()` should just take the path as well and do this? Or have some other function.

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+            reqs,
+            assert,
+            issued_reqs: HashMap::new(),
+            concurrent: true,
+            current_idx: 0,
+        }
+    }
+
+    /// Set test request concurrency
+    pub fn set_concurrency(&mut self, concurrent: bool) {
```
ghedo: It doesn't look like the concurrency needs to be changed at runtime, so maybe it should just be an argument of `new()`?

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+    #[serde(rename = "content-type")]
+    content_type: Option<Vec<String>>,
+    origin: Option<String>,
+    url: Option<String>,
+}
+
+// A helper function pointer type for assertions.
+// Each test takes a vector of HttpBinReq for validation
+// and must return a pass or fail.
+type HttpBinTestAssert = (fn(&[HttpBinReq]) -> bool);
```
ghedo:
```suggestion
type HttpBinTestAssert = fn(&[HttpBinReq]) -> bool;
```

I don't think you need parentheses here.

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+        req.hdrs
+            .push(Header::new("content-type", "application/json"));
+
+        vec![req]
+    }
+
+    fn assert_request_body(reqs: &[HttpBinReq]) -> bool {
+        let mut ok = reqs[0].assert_hdrs();
+
+        let json = jsonify(&reqs[0].resp_body);
```
ghedo: Why is this needed?

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+            expect_hdrs,
+        );
+
+        req.hdrs
+            .push(Header::new("content-type", "application/json"));
+
+        vec![req]
+    }
+
+    fn assert_request_body(reqs: &[HttpBinReq]) -> bool {
```
ghedo: Why does this take a slice if it only checks a single request?

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+
+    pub fn assert_hdrs(&self) -> bool {
+        let mut ok = true;
+
+        if let Some(expect_hdrs) = &self.expect_resp_hdrs {
+            for hdr in expect_hdrs {
+                if let Some(h) =
+                    self.resp_hdrs.iter().find(|&x| x.name() == hdr.name())
+                {
+                    ok &= hdr.value() == h.value();
```
ghedo: I think we should just use `assert_eq!()` in these functions (not just `assert_hdrs()`), rather than returning a boolean. Also, it probably makes sense to turn them into macros, so it would be easier to debug problems in the tests (the assert means that you get the expected and actual values in a panic message, and the macro means that you get the correct line numbers without needing a stack trace).

tools/httpbintest/src/lib.rs
```diff
@@ -0,0 +1,1304 @@
...
+    where
+        T: FnOnce() -> () + std::panic::UnwindSafe,
+    {
+        INIT.call_once(|| {
+            env_logger::builder()
+                .default_format_timestamp_nanos(true)
+                .init()
+        });
+
+        let result = std::panic::catch_unwind(|| test());
```
ghedo: Instead of catching panics, it would probably be better to just let them crash the program, so we would get better debug output when a test fails (see also the other comments about using real asserts).

tools/httpbintest/src/lib.rs
ghedo: Actually nevermind, I figured it out...

tools/httpbintest/src/lib.rs
LPardue: This function needs to match the signature of `HttpBinTestAssert` which accepts a slice. Perhaps there is a way to rewrite code such that assert can signature can accept a slice or individual item...

tools/httpbintest/src/lib.rs
LPardue: Yeah probably, I'll see what I can do in a refactor.

tools/httpbintest/src/lib.rs
LPardue: This is hangover from previous design, so yeah probably can be nuked.

tools/httpbintest/src/lib.rs
LPardue: I was copying the link you suggested - https://medium.com/@ericdreichert/test-setup-and-teardown-in-rust-without-a-framework-ba32d97aa5ab

I guess it depends if there is any expensive setup that needs to be tidied up in the case of failure but I don't see a strong need for that in the case of this client-oriented testing code.

.travis.yml
LPardue: done, but I backed-out the change to run tests in Travis to avoid build failure noise

tools/httpbintest/src/lib.rs
LPardue: In bcab71a622 I renamed the folder and split it out to be Http3Test module with structs & methods plus a test runner, then a separate `tests` folder that shown a concrete example of using Http3Test to a httpbin server.

We can bike shed some more but I think this split makes things clearer

.travis.yml
```diff
@@ -20,12 +20,16 @@ matrix:
       - cargo clippy --examples -- -D warnings
       - cargo doc --no-deps
       - make -C examples
+      - RUSTFLAGS="-D warnings" cargo build --release --verbose
```
ghedo: I think we need to use "cargo test --dry-run" here, otherwise the tests don't actually get built.

```suggestion
      - RUSTFLAGS="-D warnings" cargo test --dry-run --verbose
```

ghedo: I think we need to use "cargo test --dry-run" here, otherwise the tests don't actually get built.

```suggestion
      - RUSTFLAGS="-D warnings" cargo test --dry-run --verbose
```

.travis.yml
```diff
@@ -20,12 +20,16 @@ matrix:
...
+        --manifest-path tools/http3-test/Cargo.toml
    - rust: beta
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
       - cargo doc --no-deps
       - make -C examples
+      - RUSTFLAGS="-D warnings" cargo build --release --verbose
```
ghedo:
```suggestion
      - RUSTFLAGS="-D warnings" cargo test --dry-run --verbose
```

.travis.yml
```diff
@@ -35,6 +39,8 @@ matrix:
       - cargo fmt -- --check
       - cargo doc --no-deps
       - make -C examples
+      - RUSTFLAGS="-D warnings" cargo build --release --verbose
```
ghedo:
```suggestion
      - RUSTFLAGS="-D warnings" cargo test --dry-run --verbose
```

tools/http3-test/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+//! let assert = |reqs: &[http3test::Http3Req]| reqs[0].assert_hdrs();
+//!
+//! let mut test = http3test::Http3Test::new(url, reqs, assert, true);
+//! ```
+//!
+//! ## Sending test requests
+//!
+//! Testing a server requires a quiche connection and a HTTP/3 connection.
+//!
+//! Request are issued with the [`send_request()`] method. By default, all
```
ghedo:
```suggestion
//! Request are issued with the [`send_requests()`] method. By default, all
```

tools/http3-test/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+#[macro_use]
+extern crate log;
+
+use std::collections::HashMap;
+
+use quiche::h3::Header;
+
+pub const USER_AGENT: &str = "quiche-http3-integration-client";
+
+// Stores the request, the expected response headers, and the actual response.
```
ghedo: Should use `///` instead of `//` for comments to be picked up by rustdoc.

Also, there should be an empty (still commented out though) line between the short description and the long one.

tools/http3-test/src/runner.rs
```diff
@@ -0,0 +1,314 @@
...
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::net::ToSocketAddrs;
+
+use ring::rand::*;
+
+#[allow(dead_code)]
```
ghedo: I guess this is because of the fact that it's only used in tests? Using `cargo test --dry-run` should make this unnecessary.

tools/http3-test/test_httpbin.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
```
ghedo: We can probably remove this file?

tools/http3-test/tests/httpbin_tests.rs
```diff
@@ -0,0 +1,692 @@
+#[cfg(test)]
+mod httpbin_tests {
```
ghedo: I think you can remove the two lines above, as files under `tests/` should only be built when testing (i.e. with  `cargo test`), and then you can remove the indentation for the rest of the file.

tools/http3-test/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "http3test"
```
ghedo:
```suggestion
name = "http3-test"
```

tools/http3-test/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+authors = ["Lucas Pardue <lucaspardue.24.7@gmail.com>"]
+edition = "2018"
+
```
ghedo:
```suggestion
publish = false

```
so we don't accidentally upload this to crates.io (I'll need to do this for the other tool as well).

.travis.yml
ghedo: Also, we should probably run clippy and rustfmt on this as well...

tools/http3-test/src/lib.rs
ghedo: Same below.

tools/http3-test/tests/httpbin_tests.rs
LPardue: the nice property of them being in `mod httpbin_tests` was that we could make different kinds of integration test and run them separately e.g. `cargo test httpbin_tests::drip` and `cargo test some_other_test::break_it`.

Is there some other way to achieve that ?

.travis.yml
```diff
@@ -20,12 +20,20 @@ matrix:
...
+        -- -D warnings
    - rust: beta
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
       - cargo doc --no-deps
       - make -C examples
+      - RUSTFLAGS="-D warnings" cargo test --no-run --verbose
+        --manifest-path tools/http3-test/Cargo.toml
+      - cargo clippy --manifest-path tools/http3-test/Cargo.toml
```
ghedo: We only have clippy in stable.

LPardue: do'h!

tools/http3-test/tests/httpbin_tests.rs
ghedo: I guess not, let's just leave as-is then.

tools/http3-test/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+//!
+//! However, because checking response headers is so common, for convenience
+//! the expected headers can be provided during [`Http3Assert`] construction:
+//!
+//! ```no_run
+//! let mut url = url::Url::parse("https://cloudflare-quic.com/b/get").unwrap();
+//! let mut reqs = Vec::new();
+//!
+//! let expect_hdrs = Some(vec![quiche::h3::Header::new(":status", "200")]);
+//! reqs.push(http3-test::Http3Req::new("GET", &url, None, expect_hdrs));
```
LPardue: errr, @ghedo so when I accepted your suggestion to rename the module from `http3test` to `http3-test` I had to update all the namespace usages, but now rustfmt wants to turn this line into 

```
reqs.push(http3 - test::Http3Req::new("GET", &url, None, expect_hdrs));
```

ghedo: Lol... I think yo need to replace `-` with `_` (to be consistent I guess we should always use `http3_test` instead of `http3-test` in code).

PR #123, opened 2019-08-17
limit stream send buffer size

ghedo:
In order to avoid buffering an infinite amount of data in the stream's
send buffer, we now only allow streams to buffer outgoing data up to the
amount that the peer allows us to send.

This also adds a new Connection method, stream_capacity() which can be
used by applications to know how much data they can write in advance.

This is required by the HTTP/3 code that needs to know the size of body
it can write in advance in order to set a proper size in the DATA frame
header. Without stream_capacity(), send_body() could potentially write
an incomplete DATA frame in the stream send buffer which would be
complicated to recover from.

---

TODO:
- [x] Merge #124.
- [x] Update server examples to resume sending data after stream is blocked (requires #124).

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -50,9 +50,18 @@ Options:
   -h --help         Show this screen.
 ";
 
+struct PartialResponse {
+    body: Vec<u8>,
+
+    written: usize,
+}
+
 struct Client {
     conn: Box<quiche::Connection>,
+
     http3_conn: Option<quiche::h3::Connection>,
+
+    partial_responses: HashMap<u64, PartialResponse>,
 }
 
 type ClientMap = HashMap<Vec<u8>, (net::SocketAddr, Client)>;
@@ -269,6 +278,7 @@ fn main() {
                 let client = Client {
                     conn,
                     http3_conn: None,
+                    partial_responses: HashMap::new(),
                 };
 
                 clients.insert(scid.to_vec(), (src, client));
@@ -319,14 +329,20 @@ fn main() {
                 client.http3_conn = Some(h3_conn);
             }
 
-            if let Some(http3_conn) = &mut client.http3_conn {
+            if client.http3_conn.is_some() {
+                // Handle writable streams.
+                for stream_id in client.conn.writable() {
+                    handle_writable(client, stream_id);
+                }
+
                 // Process HTTP/3 events.
                 loop {
+                    let http3_conn = client.http3_conn.as_mut().unwrap();
+
                     match http3_conn.poll(client.conn.as_mut()) {
                         Ok((stream_id, quiche::h3::Event::Headers(headers))) => {
                             handle_request(
-                                &mut client.conn,
-                                http3_conn,
+                                client,
                                 stream_id,
                                 &headers,
                                 args.get_str("--root"),
@@ -472,9 +488,12 @@ fn validate_token<'a>(
 
 /// Handles incoming HTTP/3 requests.
 fn handle_request(
-    conn: &mut quiche::Connection, http3_conn: &mut quiche::h3::Connection,
-    stream_id: u64, headers: &[quiche::h3::Header], root: &str,
+    client: &mut Client, stream_id: u64, headers: &[quiche::h3::Header],
+    root: &str,
 ) {
+    let conn = &mut client.conn;
+    let http3_conn = &mut client.http3_conn.as_mut().unwrap();
+
     info!(
         "{} got request {:?} on stream id {}",
         conn.trace_id(),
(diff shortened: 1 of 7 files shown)


src/h3/mod.rs
```diff
@@ -689,11 +689,24 @@ impl Connection {
...
+
+        let stream_cap = conn.stream_capacity(stream_id)?;
+
+        // Check if there is enough capacity to send the frame header
+        if stream_cap < overhead {
+            return Err(Error::Done);
+        }
+
+        // Cap the frame payload length to the stream's capacity.
+        let body_len = std::cmp::min(body.len(), stream_cap - overhead);
```
LPardue: If stream_cap == overhead, it could result in sending 0-payload-length DATA frames. They are technically fine, but I think it adds work for library and the calling app to deal with the fact that 0-non-header bytes were written.

ghedo: Yeah, good point. I can't imagine a use-case for this either, so we can just forbid it by changing the check above to `if stream_cap <= overhead`.

ghedo: Fixed.

ghedo:
Rebased on top of #124.

src/h3/mod.rs
```diff
@@ -2346,6 +2361,59 @@ mod tests {
...
+            );
+        }
+
+        assert_eq!(s.poll_server(), Ok((stream, Event::Data)));
+        assert_eq!(
+            s.recv_body_server(stream, &mut recv_buf),
+            Ok(bytes.len() - 2)
+        );
+
+        // Fin flag from last send_bidy() call was not sent as the buffer was
```
LPardue:
```suggestion
        // Fin flag from last send_body() call was not sent as the buffer was
```

ghedo: Fixed, thanks.

ghedo:
I have now updated the server examples to deal with partial writes as well (also found a bug in the HTTP/3 send_body() method which caused fin to be sent on partial writes). I'll leave the clients as an exercise for the reader :smile: (or for a future PR).

@LPardue Since these are pretty big changes, can you have another look please?

src/stream.rs
```diff
@@ -593,16 +593,30 @@ impl SendBuf {
     }
 
     /// Inserts the given slice of data at the end of the buffer.
```
LPardue: Might want to update the description to say that it returns the actual number of elements that were pushed.

src/lib.rs
```diff
@@ -2002,6 +2010,15 @@ impl Connection {
         Ok(())
     }
 
+    /// Returns the stream's outgoing flow control capacity in bytes.
+    pub fn stream_capacity(&self, stream_id: u64) -> Result<usize> {
```
LPardue: < speculation>
It might be useful in the future to know the capacity of read/inbound streams. So perhaps we want to name this method `stream_send_capacity`.
< /speculation>

examples/http3-server.rs
```diff
@@ -549,6 +580,39 @@ fn build_response(
...
 }
 
+/// Handles newly writable streams.
+fn handle_writable(client: &mut Client, stream_id: u64) {
+    let conn = &mut client.conn;
+    let http3_conn = &mut client.http3_conn.as_mut().unwrap();
+
+    debug!("{} stream {} is writable", conn.trace_id(), stream_id);
+
+    if !client.partial_responses.contains_key(&stream_id) {
```
LPardue: should this check be moved to the top of the function? That would prevent some unneeded prints, unless you specifically want to see when the transport can write to a stream but the app has nothing to write...

examples/server.rs
```diff
@@ -432,25 +465,68 @@ fn handle_stream(
...
     }
 }
 
+/// Handles newly writable streams.
+fn handle_writable(client: &mut Client, stream_id: u64) {
+    let conn = &mut client.conn;
+
+    debug!("{} stream {} is writable", conn.trace_id(), stream_id);
+
+    if !client.partial_responses.contains_key(&stream_id) {
```
LPardue: same as above

src/lib.rs
ghedo: I'd expect that to be called something other than "capacity" though, as that to me, mostly means what's the available space inside the buffer rather than what's the actual buffer size is (i.e. how much of the capacity is actually used). Though we can also always change this name later on as we are quite far from stabilizing the API :)

examples/http3-server.rs
ghedo: I left the log specifically outside the check as I thought it would be useful for debugging (e.g. stream is writable but we don't have any data to send).

In practice I never used it though, so I don't really care that much.

src/lib.rs
LPardue: yeah fair point, resolving this

src/stream.rs
ghedo: Yeah, good point, updated now.

PR #124, opened 2019-08-17
stream: implement Writable iterator

ghedo:
This implements an iterator for writable streams, that is, streams that
have enough flow control capacity to be written to.

This can be used by applications to resume sending on streams that were
previously blocked by flow control.

---

This also renames the previous uses of "writable" to "flushable" to avoid confusion.

TODO:
- [x] Merge #119.
- [x] Add some tests.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -293,6 +293,12 @@ impl Iterator for Readable {
     }
 }
 
+impl ExactSizeIterator for Readable {
+    fn len(&self) -> usize {
+        self.streams.len()
+    }
+}
+
 /// Receive-side stream buffer.
 ///
 /// Stream data received by the peer is buffered in a list of data chunks
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -200,6 +200,11 @@ impl StreamMap {
         Readable::new(&self.streams)
     }
 
+    /// Creates an iterator over streams that can be written to.
+    pub fn writable(&self) -> Writable {
+        Writable::new(&self.streams)
+    }
+
     /// Creates an iterator over all streams.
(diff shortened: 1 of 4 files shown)


ghedo:
This should also check that the stream is not finished (see #127).

ghedo:
Added tests now (and while at it, I also added tests for `Readable`), so this should be ready to go now.

nqv:
I see `strut Readable` and `struct Writable` are similar, wondering if they can be consolidated into a `StreamIterator` or something?

ghedo:
Yeah, that's a good idea, thanks @nqv. Though I'll leave that refactoring for a later PR as this is already getting big.

LPardue:
I'm late to the party for the review but I had a basic question. If I call `conn.writable()` and consume connection-level flow control window by writing to one stream, then the `conn.stream_send()` call will fail? At that point should the Writable iterator be treated as invalidated?

src/stream.rs
```diff
@@ -687,6 +776,18 @@ impl SendBuf {
         self.data.clear();
     }
 
+    /// Returns true if the send-side of the stream is complete.
+    ///
+    /// This happens when the stream's send final size is knwon, and the
```
LPardue: late review but typo `knwon`

ghedo: :+1: will fix in later PR.

ghedo:
@LPardue we might look at that as an optimization, though it would be pretty hard to implement as the iterator is completely independent from the connection itself.

src/stream.rs
ghedo: Fixed in https://github.com/cloudflare/quiche/commit/75d419fcf1446b9dda0b3a070e14900953841e87.

PR #126, opened 2019-08-17
tls: add feature to build against OpenSSL

ghedo:
This adds a new feature "openssl" to build against OpenSSL instead of
BoringSSL. We need this because the OpenSSL and BoringSSL APIs are not
fully compatible (e.g. some "functions" are actually macros wrapping
SSL_ctrl()).

This depends on https://github.com/openssl/openssl/pull/8797

To build OpenSSL do:

```
 % git clone --branch=master-quic-support https://github.com/akamai/openssl/
 % cd openssl
 % export OPENSSL_INSTALL_PATH=$PWD/build
 % ./config enable-tls1_3 --prefix=$OPENSSL_INSTALL_PATH
 % make -j$(nproc)
 % make install_sw
```

To build quiche using the above OpenSSL build do:

```
 % export PKG_CONFIG_PATH=$OPENSSL_INSTALL_PATH/lib/pkgconfig
 % export LD_LIBRARY_PATH=$OPENSSL_INSTALL_PATH/lib
 % cargo test --features openssl
```


src/build.rs
```diff
@@ -173,6 +173,14 @@ fn main() {
         println!("cargo:rustc-link-lib=static=ssl");
     }
 
+    if cfg!(feature = "openssl") {
+        #[cfg(feature = "openssl")]
```
jblazquez: The order of `libcrypto` and `libssl` below should be swapped. I failed to build quiche with unresolved symbols because the linker command line had `-lcrypto -lssl` in that order. Swapping the two here fixed it.

VMatrix1900:
Can we add the rustls and ring option? #129

splitice:
Any update on openssl support?

ghedo:
@splitice not really. The upstream OpenSSL PR is still open with no end in sight.

jsgf:
@ghedo 
> The upstream OpenSSL PR is still open with no end in sight.

What is it blocked on?

splitice:
perhaps target https://github.com/akamai/openssl/tree/OpenSSL_1_1_1d-quic ?

PR #127, opened 2019-08-17
stream: track stream's send final size

ghedo:
This enforces setting the stream's final size (i.e. calling
stream_send() with the "fin" argument set to true) only once, as the
last stream write (so no setting fin multiple times, or doing additional
writes after fin is set).


ghedo:
Will merge into #124.

PR #85, opened 2019-07-21
h3: use HashMap instead of BTreeMap to track streams

ghedo:
Some (but not all) microbenchamrks show this is faster, so it must be true:

```
http3/128000            time:   [3.0736 ms 3.2788 ms 3.4550 ms]
                        change: [-20.154% -11.160% -1.2155%] (p = 0.03 < 0.05)
                        Performance has improved.
http3/256000            time:   [10.542 ms 10.748 ms 10.924 ms]
                        change: [-12.792% -10.599% -8.2133%] (p = 0.00 < 0.05)
                        Performance has improved.
http3/512000            time:   [26.139 ms 26.438 ms 26.695 ms]
                        change: [-3.8094% -2.1481% -0.4609%] (p = 0.02 < 0.05)
                        Change within noise threshold.
```

PR #133, opened 2019-08-23
h3: streamline poll()

ghedo:
This changes the poll() function such that frames received on a stream
are handled immediately, rather than being queued for later processing.

This simplies the general architecture of the HTTP/3 implementation, and
removes a slight potential for abuse, since the stream's frame queue
didn't have a limit. In addition this removes the need to loop through
all streams when poll() is called, to handle previously queued frames.

Benchmarks show some improvement, though this also enables further
optimizations in the future:

```
  http3/128000            time:   [3.1744 ms 3.3412 ms 3.4806 ms]
                          change: [-19.469% -10.427% -0.5146%] (p = 0.04 < 0.05)
                          Change within noise threshold.
  http3/256000            time:   [11.336 ms 11.527 ms 11.689 ms]
                          change: [-10.013% -7.8772% -5.7020%] (p = 0.00 < 0.05)
                          Performance has improved.
  http3/512000            time:   [25.405 ms 25.766 ms 26.072 ms]
                          change: [-17.732% -16.439% -14.968%] (p = 0.00 < 0.05)
                          Performance has improved.
  http3/1000000           time:   [62.185 ms 63.543 ms 64.738 ms]
                          change: [-8.1289% -6.4240% -4.4967%] (p = 0.00 < 0.05)
                          Performance has improved.
```

As can be observed by the tests, there is a slight change in behavior,
as events might be returned even if there is a later error (though we
couldn't really depend on the previous behavior, as it depended on when
streams were received).

In addition, while before we would return events in order of stream ID
(as streams were processed in order due to the use of a BTreeMap), the
events order now depends on the order of streams returned by the
readable stream iterator (which is kinda random).

This required some refactoring of HTTP/3 internals:

 * Before this change, when trying to parse unknown frames,
   Frame::from_bytes would return Error::Done, but it now returns a
   special Unknown frame type. Incidentally, this fixes a problem we
   had before, where try_consume_frame() would ignore malformed frames.

 * Finished streams are queued in a separate queue, so that we can
   properly report Finished events.

 * Control streams are now processed separately, before any other
   readable stream.

 * Frame processing is now performed in a separate method.

This also incorporates https://github.com/cloudflare/quiche/pull/85.

---

This change kind of got out of hand, so I'm creating a PR now before I add more stuff to it.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -246,7 +246,8 @@
 //! [`send_response()`]: struct.Connection.html#method.send_response
 //! [`send_body()`]: struct.Connection.html#method.send_body
 
-use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::collections::VecDeque;
 
 use crate::octets;
 
@@ -505,7 +506,7 @@ pub struct Connection {
     highest_request_stream_id: u64,
     highest_uni_stream_id: u64,
 
-    streams: BTreeMap<u64, stream::Stream>,
+    streams: HashMap<u64, stream::Stream>,
 
     local_settings: ConnectionSettings,
     peer_settings: ConnectionSettings,
@@ -520,6 +521,8 @@ pub struct Connection {
     peer_qpack_streams: QpackStreams,
 
     max_push_id: u64,
+
+    finished_streams: VecDeque<u64>,
 }
 
 impl Connection {
@@ -532,7 +535,7 @@ impl Connection {
             highest_request_stream_id: 0,
             highest_uni_stream_id: initial_uni_stream_id,
 
-            streams: BTreeMap::new(),
+            streams: HashMap::new(),
 
             local_settings: ConnectionSettings {
                 num_placeholders: Some(config.num_placeholders),
@@ -565,6 +568,8 @@ impl Connection {
             },
 
             max_push_id: 0,
+
+            finished_streams: VecDeque::new(),
         })
     }
 
@@ -727,16 +732,6 @@ impl Connection {
         // Return how many bytes were written, excluding the frame header.
         let written = conn.stream_send(stream_id, &body[..body_len], fin)?;
 
-        // Tidy up streams.
-        if fin &&
-            self.streams
-                .get(&stream_id)
-                .filter(|s| s.peer_fin())
-                .is_some()
-        {
-            self.streams.remove(&stream_id);
-        };
-
         Ok(written)
     }
 
@@ -760,7 +755,15 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        Ok(stream.try_consume_data(conn, out)?)
+        let read = stream.try_consume_data(conn, out)?;
+
+        // While body is being received, the stream is marked as finished only
+        // when all data is read by the application.
+        if conn.stream_finished(stream_id) {
+            self.finished_streams.push_back(stream_id);
+        }
+
+        Ok(read)
     }
 
     /// Processes HTTP/3 data received from the peer.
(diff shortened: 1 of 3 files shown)


src/h3/mod.rs
```diff
@@ -1174,15 +960,40 @@ impl Connection {
...
+
+    fn process_readable_stream(
+        &mut self, conn: &mut super::Connection, stream_id: u64,
+    ) -> Result<(u64, Event)> {
+        self.streams
             .entry(stream_id)
             .or_insert_with(|| stream::Stream::new(stream_id, false));
 
         loop {
+            let stream = self.streams.get_mut(&stream_id).unwrap();
```
LPardue: `unwrap()` in library code. But also I don't understand why the assignment of `stream` was moved inside the loop. Since stream_id doesn't change inside the function, it seems like a wasted `get_mut()` call each iteration, and it makes the previous `.entry().or_insert_with()` weird.

src/h3/mod.rs
```diff
@@ -1838,40 +1866,55 @@ mod tests {
...
-        assert_eq!(s.recv_body_server(stream2, &mut recv_buf), Ok(body.len()));
-        assert_eq!(s.poll_server(), Ok((stream2, Event::Finished)));
-
-        assert_eq!(s.poll_server(), Ok((stream3, Event::Headers(req3))));
-        assert_eq!(s.poll_server(), Ok((stream3, Event::Data)));
-        assert_eq!(s.recv_body_server(stream3, &mut recv_buf), Ok(body.len()));
-        assert_eq!(s.poll_server(), Ok((stream3, Event::Data)));
-        assert_eq!(s.recv_body_server(stream3, &mut recv_buf), Ok(body.len()));
-        assert_eq!(s.poll_server(), Ok((stream3, Event::Finished)));
+        let (stream, ev) = s.poll_server().unwrap();
```
LPardue: The test code is nicely adapted to the new functionality but it is quite duplicative. We might want to simplify this in a future PR; for example a loop or add a testing util function that can do the checks.

src/h3/mod.rs
```diff
@@ -760,7 +755,13 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        Ok(stream.try_consume_data(conn, out)?)
+        let read = stream.try_consume_data(conn, out)?;
+
+        if conn.stream_finished(stream_id) {
+            self.finished_streams.push_back(stream_id);
```
LPardue: My spidersense was tingling at this code being here. I had to trace back through the code to remember that DATA frame case wouldn't be handled by other functions and that it is required here. Maybe worth a comment?

src/h3/mod.rs
ghedo: Removed the unwrap, thanks.

As for the `get_mut()` thing, this is necessary to avoid holding up `self` in a borrow for the entire duration of the loop, otherwise we wouldn't be able to borrow it again when calling `process_frame()` below (this is due to a borrow checker limitation which prevents it from working across function calls).

```
error[E0499]: cannot borrow `*self` as mutable more than once at a time
    --> src/h3/mod.rs:1186:27
     |
999  |             let stream = match self.streams.get_mut(&stream_id) {
     |                                ------------ first mutable borrow occurs here
...
1006 |             match stream.state() {
     |                   ------ first borrow later used here
...
1186 |                     match self.process_frame(conn, stream_id, frame) {
     |                           ^^^^ second mutable borrow occurs here

error: aborting due to previous error
```

I added a comment to make this somewhat more clear.

src/h3/mod.rs
ghedo: Yeah, I turned this and the response checkes below into a loop.

src/h3/mod.rs
ghedo: I'm not sure I understand what clarification is needed here, any suggestion for what the comment should say?

LPardue: I was thinking of something like 

```
While the in stream::State::Data, stream closure needs to be detected and handled when body bytes are read
```

but I don't know how much that helps anyone coming to this code.

ghedo: Ok, added a comment now.

PR #134, opened 2019-08-23
Optimize stream iterators

ghedo:
We currently iterate over all streams in many cases (e.g. to create readable/writable streams iterators, checking for streams that need to send MAX_STREAM_DATA, ...), but in most cases a more efficient option would be to keep track of streams that need some action separately from the main stream map, and simply iterate over that list.

The overall improvement is pretty fantastic:
```
http3/128000            time:   [1.1004 ms 1.1867 ms 1.2631 ms]                         
                        change: [-72.075% -69.050% -65.596%] (p = 0.00 < 0.05)
                        Performance has improved.
http3/256000            time:   [2.7900 ms 2.8375 ms 2.8781 ms]                          
                        change: [-77.717% -77.201% -76.668%] (p = 0.00 < 0.05)
                        Performance has improved.
http3/512000            time:   [4.4683 ms 4.5074 ms 4.5411 ms]                          
                        change: [-84.194% -83.943% -83.692%] (p = 0.00 < 0.05)
                        Performance has improved.
```

On top of that, the changes done in https://github.com/cloudflare/quiche/pull/133 enable a much bigger improvement, due to the fact that after that PR's changes, the only remaining bottleneck in `quiche::h3::poll()` was the readable stream iteration (keep in mind that the following is improvement over the already improved code):

```
http3/128000            time:   [284.18 us 285.05 us 286.12 us]                         
                        change: [-70.637% -68.590% -66.356%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  8 (8.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
http3/256000            time:   [571.44 us 573.43 us 575.47 us]                         
                        change: [-78.839% -78.508% -78.175%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 27 outliers among 100 measurements (27.00%)
  16 (16.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  7 (7.00%) high severe
http3/512000            time:   [1.1695 ms 1.1734 ms 1.1774 ms]                          
                        change: [-73.266% -73.031% -72.789%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 25 outliers among 100 measurements (25.00%)
  13 (13.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  9 (9.00%) high severe
```

TODO:
* [ ] Convert writable iterator as well (can probably be done as a separate PR though).
* [ ] Almost full iterator probably needs more tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1932,8 +1932,14 @@ impl Connection {
 
         self.max_rx_data_next = self.max_rx_data_next.saturating_add(read as u64);
 
+        let readable = stream.readable();
+
         if stream.recv.almost_full() {
             self.streams.mark_almost_full(stream_id, true);
+        }
+
+        if !readable {
+            self.streams.mark_readable(stream_id, false);
         }
 
         Ok((read, fin))
@@ -2016,7 +2022,13 @@ impl Connection {
 
         match direction {
             // TODO: send STOP_SENDING
-            Shutdown::Read => stream.recv.shutdown(),
+            Shutdown::Read => {
+                stream.recv.shutdown();
+
+                if !stream.readable() {
+                    self.streams.mark_readable(stream_id, false);
+                }
+            },
 
             // TODO: send RESET_STREAM
             Shutdown::Write => stream.send.shutdown(),
@@ -2503,6 +2515,10 @@ impl Connection {
                 let stream = self.get_or_create_stream(stream_id, false)?;
 
                 stream.recv.push(data)?;
+
+                if stream.readable() {
+                    self.streams.mark_readable(stream_id, true);
+                }
 
                 self.rx_data += data_len;
             },
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1616,13 +1616,20 @@ impl Connection {
                 let stream = match self.streams.get_mut(stream_id) {
                     Some(v) => v,
 
-                    None => continue,
+                    None => {
+                        // The stream doesn't exist anymore, so remove it from
+                        // the almost full set.
+                        self.streams.mark_almost_full(stream_id, false);
+                        continue;
+                    },
                 };
 
                 let frame = frame::Frame::MaxStreamData {
                     stream_id,
                     max: stream.recv.update_max_data() as u64,
                 };
+
+                self.streams.mark_almost_full(stream_id, false);
 
                 if frame.wire_len() > left {
                     break;
@@ -1924,6 +1931,10 @@ impl Connection {
         let (read, fin) = stream.recv.pop(out)?;
 
         self.max_rx_data_next = self.max_rx_data_next.saturating_add(read as u64);
+
+        if stream.recv.almost_full() {
+            self.streams.mark_almost_full(stream_id, true);
+        }
 
         Ok((read, fin))
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1612,13 +1612,15 @@ impl Connection {
 
         // Create MAX_STREAM_DATA frames as needed.
         if pkt_type == packet::Type::Application && !is_closing {
-            for (id, stream) in self
-                .streams
-                .iter_mut()
-                .filter(|(_, s)| s.recv.more_credit())
-            {
+            for stream_id in self.streams.almost_full() {
+                let stream = match self.streams.get_mut(stream_id) {
+                    Some(v) => v,
+
+                    None => continue,
+                };
+
                 let frame = frame::Frame::MaxStreamData {
-                    stream_id: *id,
+                    stream_id,
                     max: stream.recv.update_max_data() as u64,
                 };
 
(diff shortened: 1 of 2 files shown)


ghedo:
Combining the changes from this PR with the changes from #133, this is the overall improvement over master:
```
http3/128000            time:   [284.94 us 285.97 us 287.19 us]                         
                        change: [-91.286% -90.497% -89.585%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 25 outliers among 100 measurements (25.00%)
  10 (10.00%) low severe
  2 (2.00%) low mild
  3 (3.00%) high mild
  10 (10.00%) high severe
http3/256000            time:   [577.51 us 578.63 us 579.50 us]                         
                        change: [-95.532% -95.454% -95.375%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) low severe
  2 (2.00%) high mild
  2 (2.00%) high severe
http3/512000            time:   [1.1905 ms 1.1970 ms 1.2052 ms]                          
                        change: [-95.966% -95.915% -95.861%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  10 (10.00%) low severe
  2 (2.00%) high mild
  6 (6.00%) high severe
```

90-96% :rofl:

LPardue:
🤯

src/stream.rs
```diff
@@ -61,6 +62,17 @@ pub struct StreamMap {
...
     /// Streams are added to the back of the list, and removed from the front.
     flushable: VecDeque<u64>,
+
+    /// Set of stream IDs corresponding to streams that have outstanding data
+    /// to read. This is used to generate a `StreamIter` of streams without
+    /// having to iterate over the full list of streams.
+    readable: HashSet<u64>,
+
+    /// Set of stream IDs corresponding to streams that are almost out of flow
+    /// control credit and need to send MAX_STREAM_DATA.. This is used to
```
```suggestion
    /// control credit and need to send MAX_STREAM_DATA. This is used to
```

src/lib.rs
```diff
@@ -1610,16 +1610,16 @@ impl Connection {
 
         // Create MAX_STREAM_DATA frames as needed.
         if pkt_type == packet::Type::Application && !is_closing {
-            for (id, stream) in self
-                .streams
-                .iter_mut()
-                .filter(|(_, s)| s.recv.more_credit())
-            {
+            for stream_id in self.streams.almost_full() {
+                let stream = self.streams.get_mut(stream_id).unwrap();
```
LPardue: I think we don't want the `unwrap()` in library code.

But also, looking at this a bit more deeply, it kinda sucks that `almost_full()` returns a `stream_id` iterator and not a `stream` iterator. I don't know if it's possible to do that inside stream.rs while also maintaining the optimized lookup that this PR introduces.

src/lib.rs
```diff
@@ -1921,6 +1921,16 @@ impl Connection {
 
         self.max_rx_data_next = self.max_rx_data_next.saturating_add(read as u64);
 
+        let readable = stream.readable();
```
LPardue: It seems redundant to assign`readable` here; the code suggestion below just calls the function directly.

However, after looking through the whole PR, I wonder if we really need to expose these `push_readable()` and `clear_readable()` methods at all. We could just expose a single function like `set_readable(id: u64, readable: bool)`, which would reduce code size and avoid leaking stream.rs implementation detail.

src/lib.rs
```diff
@@ -1921,6 +1921,16 @@ impl Connection {
...
+
+        if stream.recv.almost_full() {
+            self.streams.push_almost_full(stream_id);
+        }
+
+        if !readable {
```
```suggestion
        if !stream.readable() {
```

This makes it more consistent with similar code around line 2000+

src/lib.rs
ghedo: Removed the unwrap().

As for the iterator thing, this is in part for consistency (so we use `StreamIter` everywhere instead of having different iterator in different cases) as well as because I think that returning stream directly might make the borrow checker sad.

src/lib.rs
ghedo: The assignment is yet again because of our friend the borrow checker. Without the assignment the `stream` variable would keep the borrow on `self` "alive", so we wouldn't be able to call `self.streams.mark_readable()` and the like.

Anyway, I merged the various `push_*()` and `clear_*()` into single `mark_*()` methods.

src/lib.rs
ghedo: See https://github.com/cloudflare/quiche/pull/134#discussion_r318059640 for why this can't be done.

src/stream.rs
ghedo: Fixed, thanks.

ghedo:
@LPardue also, FWIW I also think that `almost_full` is not a very good name, but it seemed better than `out_of_credit` and I couldn't come up with anything else. Feel free to make a suggestion if you have any.

ghedo:
I'll update the writable iterator in a separate PR, and turns out we already have some tests for almost full (might add some for MAX_STREAM_DATA specifically in the future though), so this is good to go.

PR #138, opened 2019-08-28
stream: optimize writable iterator

ghedo:
Instead of iterating over all streams when deciding which streams are
writable, keep track of writable streams separately.


ghedo:
Added a commit to prevent the `From<&HashSet<u64>>` implementation for `StreamIter` from showing up in the public API.

PR #142, opened 2019-09-03
tools: allow test runner to target different test server

LPardue:
Small change to `runner.rs` to allow it agility in what server it actually targets for tests. It decouples the server SNI (used by QUIC) from the IP and port used by UDP socket. 

The httpbin_tests.rs are updated with a new environment variable `HTTPBIN_HOST` to make this easy to use from a testing environment.

### DIFF
--- a/tools/http3_test/src/runner.rs
+++ b/tools/http3_test/src/runner.rs
@@ -24,11 +24,12 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::net::ToSocketAddrs;
-
 use ring::rand::*;
 
-pub fn run(test: &mut crate::Http3Test, verify_peer: bool) {
+pub fn run(
+    test: &mut crate::Http3Test, peer_addr: std::net::SocketAddr,
+    verify_peer: bool,
+) {
     const MAX_DATAGRAM_SIZE: usize = 1350;
 
     let mut buf = [0; 65535];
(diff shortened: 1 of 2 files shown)


tools/http3_test/src/runner.rs
```diff
@@ -28,7 +28,7 @@ use std::net::ToSocketAddrs;
 
 use ring::rand::*;
 
-pub fn run(test: &mut crate::Http3Test, verify_peer: bool) {
+pub fn run(test: &mut crate::Http3Test, host: &str, verify_peer: bool) {
```
ghedo: Instead of a string, maybe pass a `SocketAddr` directly? https://doc.rust-lang.org/std/net/enum.SocketAddr.html

tools/http3_test/tests/httpbin_tests.rs
```diff
@@ -25,6 +25,16 @@ mod httpbin_tests {
         url
     }
 
+    fn host() -> String {
+        let host = match std::env::var_os("HTTPBIN_HOST") {
+            Some(val) => val.into_string().unwrap(),
+
+            None => String::from("cloudflare-quic.com:443"),
```
ghedo: Do we always want users to set this when running the tests? An alternative would be to use `HTTPBIN_ENDPOINT` host by default, unless `HTTPBIN_HOST` is set. This way we also wouldn't need to hard-code cloudflare-quic.com in two places.

For example, `host()` could take the test URL as parameter, and return the value from that if `HTTPBIN_HOST` is not set.

tools/http3_test/src/runner.rs
LPardue: I did that first but then realised it's a bit annoying if you just want to pass a string directly into runner.rs. 

It's a pretty close call on which is better.

tools/http3_test/tests/httpbin_tests.rs
LPardue: Makes sense, ta.

tools/http3_test/src/runner.rs
LPardue: It was actually easier to handle some edge cases (related to your other comment) by only accepting a SocketAddr

tools/http3_test/tests/httpbin_tests.rs
LPardue: done

ghedo:
I tweaked the commit message slightly when merging.

PR #143, opened 2019-09-05
stream application data API

### LINKED ISSUE #112: API: custom pointer per (h3) stream?
When doing multiplexed h3 streams over the same QUIC connection.

It would be helpful for my code (and presumably others') if we could assign an opaque pointer to the h3 stream when such is created, and then quiche could deliver that pointer when it returns events for that stream. It will save me from having to implement the same stream-to-pointer hash lookup.

ghedo:
This allows applications to store stream-specific data inside quiche, so
as to avoid having to track streams by themselves.

One potential use-case was storing internal HTTP/3 stream state, so we
wouldn't need to keep separate streams maps for both transport and
HTTP/3 however the current structure of the HTTP/3 implementation does
not allow for this (due to cross-references that make borrow checker
sad). Though this can be useful for other applications.

Fixes #112.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -270,6 +270,19 @@ bool quiche_conn_is_in_early_data(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
+// Initializes the stream's application data.
+//
+// Stream data can only be initialized once. Additional calls to this method
+// will fail.
+//
+// Note that the application is responsible for freeing the data.
+int quiche_conn_stream_init_application_data(quiche_conn *conn,
+                                             uint64_t stream_id,
+                                             void *data);
+
+// Returns the stream's application data, if any was initialized.
+void *quiche_conn_stream_application_data(quiche_conn *conn, uint64_t stream_id);
+
 // Fetches the next stream from the given iterator. Returns false if there are
 // no more elements in the iterator.
 bool quiche_stream_iter_next(quiche_stream_iter *iter, uint64_t *stream_id);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2228,6 +2228,53 @@ impl Connection {
         stream.recv.is_fin()
     }
 
+    /// Initializes the stream's application data.
+    ///
+    /// This can be used by applications to store per-stream information without
+    /// having to maintain their own stream map.
+    ///
+    /// Stream data can only be initialized once. Additional calls to this
+    /// method will return [`Done`].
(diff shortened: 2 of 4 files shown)


LPardue:
From a client perspective the interaction with quiche to create streams is `quiche_h3_send_request`. So my mental model was that the user data would be supplied at request time, and then later on `quiche_h3_conn_poll` would return a ref to the application data. There are other cases (like yet-to-be-implemented-push-promise) where that model probably breaks down.

The code in this PR is more symmetric (for client and server) in probably all cases but requires a bit more work for the library-using-application

ghedo:
@LPardue that would require applications that don't use this to actively ignore whatever poll() returns. I don't expect that many applications will use this API, so it's better to optimize for the common case.

LPardue:
Yeah, and with your separate `poll2` proposal users that choose to use it can write that code into their callback handlers, so it is a least nicely logically partitioned.

I think maintaining symmetry of polling quiche is a desirable property so thats why I like this PR.

include/quiche.h
```diff
@@ -270,6 +270,16 @@ bool quiche_conn_is_in_early_data(quiche_conn *conn);
...
 
+// Initializes the stream's application data.
+//
+// Note that the application is responsible for freeing the data.
+int quiche_conn_stream_init_application_data(quiche_conn *conn,
+                                             uint64_t stream_id,
+                                             void *data);
+
+// Returns the stream's application data, if any was initialized.
+void *quiche_conn_stream_application_data(quiche_conn *conn, uint64_t stream_id);
```
bagder: This looks exactly what I want. If I'd bikeshed anything on this, it would perhaps be that the function names could be called *set* and *get* stream application-data, but I don't care strongly.

src/lib.rs
```diff
@@ -2209,6 +2209,53 @@ impl Connection {
         stream.recv.is_fin()
     }
 
+    /// Initializes the stream's application data.
+    ///
+    /// This can be used by application to store per-stream information without
```
```suggestion
    /// This can be used by applications to store per-stream information without
```

include/quiche.h
```diff
@@ -270,6 +270,16 @@ bool quiche_conn_is_in_early_data(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
+// Initializes the stream's application data.
```
LPardue: The internal docs state this can only be done once, is that worth repeating in the `.h`?

ghedo: Added.

PR #145, opened 2019-09-05
(re)start the idle timer when sending an ACK-eliciting packet

ghedo:
Spec says that we need to restart idle timer when sending an
ACK-eliciting packet, but only if we haven't sent an ACK-eliciting
packet since last receiving a packet.

---

TODO:
* [ ] Add test? Not sure yet how this can be tested though...


ghedo:
Build is broken because latest Rust toolchain nightly build of rustfmt failed. I verified locally that the formatting is ok with yesterday's build of rustfmt.

PR #147, opened 2019-09-06
Use find instead of glob pattern

VMatrix1900:
The glob pattern is not cross platform. At least not working in my Macbook. See [Stackoverflow question](https://stackoverflow.com/questions/4036191/sources-from-subdirectories-in-makefile).
Change it to `find` fix the problem.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -33,7 +33,7 @@ http3-client: http3-client.c $(INCLUDE_DIR)/quiche.h $(LIB_DIR)/libquiche.a
 http3-server: http3-server.c $(INCLUDE_DIR)/quiche.h $(LIB_DIR)/libquiche.a
 	$(CC) $(CFLAGS) $(LDFLAGS) $< -o $@ $(INCS) $(LIBS)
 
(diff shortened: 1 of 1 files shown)


examples/Makefile
```diff
@@ -33,7 +33,7 @@ http3-client: http3-client.c $(INCLUDE_DIR)/quiche.h $(LIB_DIR)/libquiche.a
 http3-server: http3-server.c $(INCLUDE_DIR)/quiche.h $(LIB_DIR)/libquiche.a
 	$(CC) $(CFLAGS) $(LDFLAGS) $< -o $@ $(INCS) $(LIBS)
 
-$(LIB_DIR)/libquiche.a: $(SOURCE_DIR)/**/*.rs
+$(LIB_DIR)/libquiche.a: $(shell find $(SOURCE_DIR) -name ".rs")
```
junhochoi:
```suggestion
$(LIB_DIR)/libquiche.a: $(shell find $(SOURCE_DIR) -type f -name ".rs")
```

ghedo:
Merged, thanks @VMatrix1900!

PR #148, opened 2019-09-06
Remove `set_max_level` of logger

VMatrix1900:
User is supposed to use `RUST_LOG` environment variable to config the log level. This function prevents that.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -70,7 +70,6 @@ pub extern fn quiche_enable_debug_logging(
     let logger = Box::new(Logger { cb, argp });
 
     log::set_boxed_logger(logger).unwrap();
-    log::set_max_level(log::LevelFilter::Trace);
 }
 
 #[no_mangle]


ghedo:
I don't think this is correct. `RUST_LOG` only works if you setup `env_logger` which C applications would definitely not have. The quiche library only ever logs at "trace" level, so if you do this the application will not get any logs.

junhochoi:
@VMatrix1900 In case of C examples, `RUST_LOG` environment variable is not used (I can confirm) because in `quiche_enable_debug_logging()` C FFI we don't use `env_logger` as @ghedo explained. In other words this patch only disables `trace` level logging.

I believe you want to have some flexibility on C example logging. If so, I think it's good to have a command line flag (e.g. -D) for enabling/disabling debug logging.

ghedo:
Closing this per the above comments. Feel free to open a new PR if you want to implement the `-D` option @junhochoi suggested.

PR #151, opened 2019-09-09
stream: properly propagate fin from zero-length writes

ghedo:
Before, we would propagate the fin flag from zero-length writes by
creating an empty RangeBuf and queueing it into the SendBuf.

However due to the `out_len > 0` loop check, SendBuf::pop() would fail
to process the empty buffer if the `max_data` parameter matched the
amount of data stored in SendBuf.

As an example, assume that SendBuf has two buffers:

 1. "hello", off=0, len=5, fin=false
 2. "",      off=5, len=0, fin=true

(the second buffer being zero-length).

If SendBuf::pop() is called with `max_data = 5`, the second buffer would
not be removed from the queue, and the fin flag wouldn't be returned.

Instead make SendBuf::pop() match the output buffer's maximum offset
against the known final offset, this way we don't need to actually queue
the fin=true empty range buffer in the first place.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -675,6 +675,8 @@ impl SendBuf {
         }
 
         if data.is_empty() {
+            // Create a dummy range buffer, in order to propagate the `fin` flag
+            // into `RangeBuf::push()`. This will be discarded later on.
             let buf = RangeBuf::from(&[], self.off, fin);
 
             return self.push(buf).map(|_| 0);
@@ -735,6 +737,12 @@ impl SendBuf {
             self.fin_off = Some(buf.max_off());
         }
 
+        // We already recorded the final offset, so we can just discard the
+        // empty buffer now.
+        if buf.is_empty() {
+            return Ok(());
+        }
+
         self.data.push(buf);
 
         Ok(())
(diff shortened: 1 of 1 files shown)


src/stream.rs
```diff
@@ -771,11 +779,17 @@ impl SendBuf {
...
             out_off = buf.max_off();
 
-            out.fin = out.fin || buf.fin();
-
             out.data.extend_from_slice(&buf.data);
         }
 
+        // Ignore the `fin` flag set for the range buffer itself, and instead
+        // match the buffer's maximum offset against the known final size (if
+        // any). This way we can avoid queueuing empty buffers just so we can
```
junhochoi: queueuing -> queueing

src/stream.rs
```diff
@@ -771,11 +779,17 @@ impl SendBuf {
...
+        // track the final size.
+        if self.fin_off == Some(out.max_off()) {
```
junhochoi: Do we need to have `>=` instead of `==`?

ghedo: No, the maximum offset of the outbut buffer needs to match the final offset exactly, otherwise we end up setting `fin=true` for multiple STREAM frames.

ghedo:
Rebased, and tweaked the logic to just do:
```rust
        out.fin = self.fin_off == Some(out.max_off());
```
instead of the `if ... { }`. Please review again.

PR #152, opened 2019-09-10
tools: http3_test detect client idle timeout

LPardue:
Previously, the runner would appear to complete successfully if the
idle timeout triggered without all requests completing. Test
assertions are only done once all requests are complete so, if the
idle timeout is hit before then, just panic.

The downside of this is that it may reveal problems that were previously masked, but in the long term this is a good thing.

Finally, the idle timeout is quite generous in order to support the example httpbin integration test suite. We should probably parameterize it but I figure we can do that in a separate PR.

### DIFF
--- a/tools/http3_test/src/runner.rs
+++ b/tools/http3_test/src/runner.rs
@@ -31,6 +31,7 @@ pub fn run(
     verify_peer: bool,
 ) {
     const MAX_DATAGRAM_SIZE: usize = 1350;
+    const IDLE_TIMEOUT: u64 = 60000;
 
     let mut buf = [0; 65535];
     let mut out = [0; MAX_DATAGRAM_SIZE];
@@ -83,7 +84,7 @@ pub fn run(
         .set_application_protos(quiche::h3::APPLICATION_PROTOCOL)
         .unwrap();
 
-    config.set_idle_timeout(60000);
+    config.set_idle_timeout(IDLE_TIMEOUT);
     config.set_max_packet_size(MAX_DATAGRAM_SIZE as u64);
     config.set_initial_max_data(max_data);
     config.set_initial_max_stream_data_bidi_local(max_stream_data);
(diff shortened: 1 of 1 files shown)


tools/http3_test/src/runner.rs
```diff
@@ -140,6 +141,11 @@ pub fn run(
 
                 conn.on_timeout();
 
+                if conn.is_closed() && reqs_complete != reqs_count {
```
ghedo: FWIW, we already check for `conn.is_closed()` in two places below, so maybe can just add the logic there?

LPardue: makes sense!

tools/http3_test/src/runner.rs
```diff
@@ -140,6 +141,11 @@ pub fn run(
...
+                    panic!("Client timed out after {}ms and only completed {}/{} requests",
+                    IDLE_TIMEOUT, reqs_complete, reqs_count);
```
ghedo: Also, we already track the start of the connection in `req_start`, so maybe we can just do `req_start.elapsed()` which would be more precise than just using the value of the idle timeout.

LPardue: example `Client timed out after 60.236662187s` :D

LPardue:
suggestions applied. PTAL.

tools/http3_test/src/runner.rs
```diff
@@ -180,6 +181,12 @@ pub fn run(
 
         if conn.is_closed() {
             info!("connection closed, {:?}", conn.stats());
+
+            if reqs_complete != reqs_count {
+                panic!("Client timed out after {:?} and only completed {}/{} requests",
+                req_start.elapsed() , reqs_complete, reqs_count);
```
ghedo:
```suggestion
                req_start.elapsed(), reqs_complete, reqs_count);
```

tools/http3_test/src/runner.rs
```diff
@@ -305,6 +312,12 @@ pub fn run(
...
```
ghedo:
```suggestion
                req_start.elapsed(), reqs_complete, reqs_count);
```

PR #153, opened 2019-09-10
http3-client C example: Pass hostname to h3 request

junhochoi:
In C http3-client example, when we create a h3 request hostname
is hard coded (quich.tech). Pass hostname to h3 authority field.

### DIFF
--- a/examples/http3-client.c
+++ b/examples/http3-client.c
@@ -49,6 +49,8 @@
 struct conn_io {
     ev_timer timer;
 
+    const char* host;
+
     int sock;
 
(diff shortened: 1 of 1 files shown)


lohith-bellad:
Isn't it quic.tech ?

examples/http3-client.c
```diff
@@ -49,6 +49,8 @@
 struct conn_io {
     ev_timer timer;
 
+    const char* host;
```
ghedo: Nit:
```suggestion
    const char *host;
```

PR #157, opened 2019-09-11
http3_tests: make idle timeout configurable in runner

LPardue:
Adds a timeout parameter to `runner.rs` and a env var to httpbin_tests that exercises it.


ghedo:
Need to fix formatting.

PR #158, opened 2019-09-11
stream: track completed streams

ghedo:
In order to track active streams, we need to be able to detect when a
stream is complete, that is, when all sent data has been acknowledged
and when all incoming data has been read by the application.

We already have RecvBuf::is_fin() to detect completion on the receive
side. For the send-side, we need to track all ranges of sent data that
have been acknowledged by the peer.

When all data has been sent and acknowledged up to the stream's
send-side final size, then the stream is complete.

In the future we can use this to update local stream limits by sending
MAX_STREAMS frames, as well as to free memory when streams are not in
use anymore.

However, the use of RangeSet to track stream data might have an impact
on performance (particularly when packet loss generates gaps in the
data), so need to do more testing.

---

* [ ] Handle streams reset by peer (and by ourselves). Can be done as a separate PR though.


src/ranges.rs
```diff
@@ -131,6 +131,25 @@ impl RangeSet {
     }
 }
 
+impl PartialEq<Range<u64>> for RangeSet {
+    fn eq(&self, other: &Range<u64>) -> bool {
```
LPardue: Would this function return false when comparing two empty ranges for equivalence? Or is it not possible to create such a RangeSet?

src/packet.rs
```diff
@@ -648,12 +648,18 @@ impl PktNumSpace {
             crypto_open: None,
             crypto_seal: None,
 
-            crypto_stream: stream::Stream::new(std::u64::MAX, std::u64::MAX),
+            crypto_stream: stream::Stream::new(
+                0,
```
LPardue: Presumably, the Stream ID property is meaningless for this crypto stream. So that raises the question, what purpose does the ID have other than to be used for determine if the stream is bidi or uni? Maybe you have some future use in mind.

src/ranges.rs
ghedo: Hmmm, I should add a test for that.

ghedo: Oh but actually `RangeSet` can be empty, but `Range` can't as it's literally just a struct with two fields (start and end).

I should probably add a comment on what this whole thing does, but to clarify, this is comparison between `RangeSet` (which is a custom struct implemented in this file) and `Range` (which is provided by Rust standard library).

src/packet.rs
ghedo: I don't remember why I did this, though I changed it now to track bidi directly, as you say it makes more sense.

src/ranges.rs
LPardue: well now you said it, it is clear from the method signature the equivalence check does that. I'm just less familiar with Rust. The comment helps though!

src/stream.rs
```diff
@@ -254,6 +256,14 @@ impl StreamMap {
         self.peer_max_streams_uni = cmp::max(self.peer_max_streams_uni, v);
     }
 
+    /// Drops completed stream.
+    ///
+    /// This should only be called when Stream::is_complete() returns true for
+    /// the given stream.
+    pub fn collect(&mut self, _stream_id: u64, _local: bool) {
+        // TODO: do something here
```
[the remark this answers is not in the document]
ghedo: In the short term we'll just need to track completed streams count against the current limit, but in the future we can do additional clean-up (e.g. maybe free some memory or something like that).

ghedo:
Rebased on master and fixed conflict. Also, added a somewhat related test for the pre-existing `ack_off` logic (though I'm thinking of removing that).

src/stream.rs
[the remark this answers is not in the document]
ghedo: I made https://github.com/cloudflare/quiche/pull/164 to show how this is used (I decided to make two separate PRs to avoid a huge PR that is hard to review).

ghedo:
I ended up completely refactoring the `ack_off` logic by using the new `acked` range set (see additional commit), this way we only track acked offsets in one place.

`ack_off` was also somewhat buggy, as it didn't take into considerations gaps in the acked data (added a test to show this behavior), though since this is basically just an optimization (to avoid resending already ACKed data), it doesn't matter that much. We might even decide to completely remove this to reduce complexity.

So, anyway, please review again (or at least just the last commit I added) :)

ghedo:
@junhochoi btw, it might be useful to look at what the overall performance impact of this is, since it's probably non-zero (particularly in case of packet loss).

src/stream.rs
```diff
@@ -885,6 +875,17 @@ impl SendBuf {
...
     }
 
+    /// Returns the highest contiguously ACKed offset.
+    pub fn ack_off(&self) -> u64 {
+        match self.acked.iter().next() {
+            // Only consider the initial range if it contiguously covers the
+            // start of the stream (i.e. from offset 0).
+            Some(std::ops::Range { start: 0, end }) => end,
+
+            Some(_) | None => 0,
```
junhochoi: `||`?

ghedo: Nope, `|` is the actual operator to match multiple patterns in Rust https://doc.rust-lang.org/1.2.0/book/patterns.html#multiple-patterns

junhochoi:
> @junhochoi btw, it might be useful to look at what the overall performance impact of this is, since it's probably non-zero (particularly in case of packet loss).

Did a quick test with 2% loss/2Mbytes. Slightly slower but I think it's marginal.

src/stream.rs
```diff
@@ -885,6 +875,17 @@ impl SendBuf {
         Ok(())
     }
 
+    /// Returns the highest contiguously ACKed offset.
+    pub fn ack_off(&self) -> u64 {
```
LPardue: does this need to be `pub`?

ghedo: Nope, fixed.

ghedo:
Will do the stream reset thing in a separate PR.

PR #164, opened 2019-09-13
increase local max streams limits with MAX_STREAMS 

ghedo:
This implements increasing the local max streams limits by sending
MAX_STREAMS frames, once a certain threshold is passed.

Stream credits are given back to the peer only once a non-locally
generated stream is complete, that is, once all incoming data has been
read by the application and all outgoing data has been ACKed.

---

TODO
* [x] Merge https://github.com/cloudflare/quiche/pull/158 first.

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -47,6 +47,7 @@ Options:
   --name <str>             Name of the server [default: quic.tech]
   --max-data BYTES         Connection-wide flow control limit [default: 10000000].
   --max-stream-data BYTES  Per-stream flow control limit [default: 1000000].
+  --max-streams STREAMS    Number of allowed concurrent streams [default: 100].
   --no-retry               Disable stateless retry.
   --no-grease              Don't send GREASE.
   -h --help                Show this screen.
@@ -86,6 +87,9 @@ fn main() {
     let max_stream_data = args.get_str("--max-stream-data");
     let max_stream_data = u64::from_str_radix(max_stream_data, 10).unwrap();
 
+    let max_streams = args.get_str("--max-streams");
+    let max_streams = u64::from_str_radix(max_streams, 10).unwrap();
+
     // Setup the event loop.
     let poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
@@ -122,8 +126,8 @@ fn main() {
     config.set_initial_max_stream_data_bidi_local(max_stream_data);
     config.set_initial_max_stream_data_bidi_remote(max_stream_data);
     config.set_initial_max_stream_data_uni(max_stream_data);
-    config.set_initial_max_streams_bidi(100);
-    config.set_initial_max_streams_uni(5);
+    config.set_initial_max_streams_bidi(max_streams);
+    config.set_initial_max_streams_uni(max_streams);
     config.set_disable_active_migration(true);
 
     if std::env::var_os("SSLKEYLOGFILE").is_some() {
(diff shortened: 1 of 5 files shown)


ghedo:
Rebased on top of latest master (there were a few conflicts).

src/lib.rs
```diff
@@ -1595,19 +1590,25 @@ impl Connection {
...
-            !is_closing
-        {
-            let frame = frame::Frame::MaxData {
-                max: self.max_rx_data_next,
-            };
+        if pkt_type == packet::Type::Short && !is_closing {
+            // Create MAX_STREAMS_BIDI frame.
+            if self.streams.should_update_max_streams_bidi() {
+                let max_streams = self.streams.update_max_streams_bidi();
+                let frame = frame::Frame::MaxStreamsBidi { max: max_streams };
```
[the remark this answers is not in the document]

ghedo:
@LPardue regarding the stream resets, it looks like we are not considering streams complete if they get reset by the peer and we don't receive all the data up to the "final size" from the reset. I'll look into fixing and testing that as part of #158 (or probably as a separate PR since #158 is already pretty big).

ghedo:
Rebased on master.

ghedo:
@LPardue @junhochoi please review again, as I had to change some things (e.g. added explicit count of peer opened streams instead of decreasing the max limit).

examples/http3-server.rs
```diff
@@ -47,6 +47,7 @@ Options:
   --name <str>             Name of the server [default: quic.tech]
   --max-data BYTES         Connection-wide flow control limit [default: 10000000].
   --max-stream-data BYTES  Per-stream flow control limit [default: 1000000].
+  --max-streams STREAMS    Number of allowed concurrent streams [default: 100].
```
LPardue: I think we might want to expose independent concurrency values for uni and bidi streams. The comment applies to server.rs too.

With the same concurrency, In the best case, a user is just going to create an application that opens themselves up for receiving more junk peer-initiated unidirectional streams. In the worst case, max-streams (<= 2) the application will never give the peer enough stream credit for a successful H3 connection. They could sleepwalk into this situation when trying to play around with different levels of request concurrency - it's reasonable to want to control how many requests are allowed.

src/lib.rs
```diff
@@ -4420,6 +4451,99 @@ mod tests {
...
     }
+
+    #[test]
+    /// Tests that the MAX_STREAMS frame is sent for bidirectional streams.
+    fn stream_limit_update_bidi() {
+        let mut buf = [0; 65535];
+
+        let mut pipe = testing::Pipe::default().unwrap();
+
+        assert_eq!(pipe.handshake(&mut buf), Ok(()));
```
LPardue: This test has an aspect that could be confusing and might break in the future. It relies on some implicit behaviour of `testing::Pipe::default()` to create the client and server with initial max_streams_* with a value of 3.

src/stream.rs
```diff
@@ -52,11 +52,19 @@ pub struct StreamMap {
     /// Peer's maximum unidirectional stream count limit.
     peer_max_streams_uni: u64,
 
+    /// The total number of bidirectional streams opened by the peer.
+    peer_opened_streams_bidi: u64,
+
+    /// The total number of unirectional streams opened by the peer.
```
```suggestion
    /// The total number of undirectional streams opened by the peer.
```

ghedo: Ugh, I wonder if I'll ever get this right.

examples/http3-server.rs
ghedo: Ok, done.

src/lib.rs
ghedo: So, I now changed the tests to use a custom config, so at least the case where we change the default values wouldn't affect this. But yeah, there are probably other ways to make these tests better, but for now this should be enough (might do that in a later PR).

PR #166, opened 2019-09-14
 update to draft 23

ghedo:
While at it also add tests for initial packet encryption.

Also note that I removed the `derive_initial_secrets2` test since the last update it became the same as `derive_initial_secrets` (originally it was using the vectors from https://github.com/quicwg/base-drafts/wiki/Test-Vector-for-the-Initial-AEAD-key-derivation but that hasn't been updated in a while) so there's no point anymore in having both.


ghedo:
@LPardue AFAICT from the changelog, HTTP/3 changes that would affect us are related to error codes only, is that correct or there are other things I missed?

LPardue:
And removal of PRORITY signal, which for us should just be removal of type and tests in h3/frame.rs

PR #168, opened 2019-09-16
h3: update to draft 23

LPardue:
Remove PRIORITY frame and tests. 

Update error codes, which leaves some holes in our internal enum. This also lets us address some existing TODOs, and I fixed a minor case of using the incorrect error code on the control stream.

### DIFF
--- a/src/h3/stream.rs
+++ b/src/h3/stream.rs
@@ -225,7 +225,7 @@ impl Stream {
                     (frame::SETTINGS_FRAME_TYPE_ID, false) =>
                         self.initialized = true,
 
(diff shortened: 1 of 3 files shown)


src/h3/stream.rs
```diff
@@ -225,25 +225,29 @@ impl Stream {
                     (frame::SETTINGS_FRAME_TYPE_ID, false) =>
                         self.initialized = true,
 
+                    // Non-SETTINGS frames no allowed on control stream
```
ghedo:
```suggestion
                    // Non-SETTINGS frames not allowed on control stream
```

PR #170, opened 2019-09-18
pad Initial packets from the server

ghedo:
We were previously only padding Initial packets from the client, but
it's useful (albeit not required) to also pad server Initial packets to
make sure the path supports packets of at least 1200 bytes as soon as
possible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1797,8 +1797,8 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        // Pad the client's initial packet.
-        if !self.is_server && pkt_type == packet::Type::Initial {
+        // Pad all initial packets.
+        if pkt_type == packet::Type::Initial {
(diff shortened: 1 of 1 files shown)


LPardue:
`tests::limit_handshake_data` is failing

ghedo:
Yeah, that test is a bit weird, need to figure out what's going on with it.

ghedo:
As far as I know PMUTD is done on the send path independently by both endpoints, so this is a simpler version of that.

But anyway the original idea was that not receiving an Initial packet in the first place is probably better than receiving one and then having to wait for the Handshake packets that might not arrive because they are too large. This way the client would be able to timeout sooner if the path doesn't allow the connection.

But thinking about this some more, this is probably more hassle than it's worth, so I'll just close it. We can think about this more later on.

PR #179, opened 2019-09-22
tools: initial fuzzers

ghedo:
This adds a couple of fuzzers based on honggfuzz that can be used to
fuzz incoming packet processing from both the server and client side.


ghedo:
This should be good to go now.

LPardue:
failing CI. Is this a Rust version problem or should cfg fuzzing be cfg feature fuzzing?

ghedo:
No, the build command was wrong (can't pass `--cfg` directly to cargo apparently). But in any case I changed this to use the actual command to build the fuzzers (using honggfuzz) rather than try to simulate it.

LPardue:
given the number of file additions, I wonder how this will affect the build times for Docker on Mac

ghedo:
:man_shrugging: though all those files can probably be left out in that case.

LPardue:
I understand this is focused on fuzzing the transport right now. But it seems not too hard to extend it to HTTP/3. For example, 

packet_server.rs
```
      fuzz!(|data: &[u8]| {
            let mut buf = data.to_vec();
            let mut conn =
                quiche::accept(&server_scid, None, &mut config).unwrap();

            let mut h3_conn = quiche::h3::Connection::with_transport(
                &mut conn,
                &h3_config,
            ).unwrap();

            conn.recv(&mut buf).ok();
            h3_conn.poll(conn.as_mut()).ok();
        });
```

inside http3::with_transport 
```
if cfg!(not(fuzzing)) {
       http3_conn.send_settings(conn)?;
}
```

ghedo:
So, I renamed the fuzzers and made the server fuzzer accept cert and key paths from environment. This is needed to run the fuzzers on Mayhem (also added configuration for that).

@LPardue yeah, we can add more fuzzers, though I'd like to concentrate on these two until we have the whole infrastructure properly setup.

LPardue:
Totally agree, after gaining a bit more context I realise there are complexities for HTTP/3. Let's avoid this this PR getting bogged down by that.

junhochoi:
Just curious where those corpus come from? Is it from actual packet dumps?

ghedo:
@junhochoi they were generated randomly by honggfuzz.

ghedo:
But also, I managed to generate a few test cases manually, which seem to increase the coverage, so I'll add those as well.

ghedo:
Too bad honggfuz doesn't support minimizing inputs...

ghedo:
In retrospect going with honggfuzz might have been a mistake, as the lack of corpus minimization is pretty annoying and it's generally not as well supported as libfuzzer in general. So i'll try to put something together based on libfuzzer/cargo-fuzz to see if that's better (and also see how it fares on Mayhem).

ghedo:
Closing in favor of #186.

PR #180, opened 2019-09-24
h3: refactor Config API and remove priority placeholder setting

LPardue:
HTTP/3 draft-23 removed prioritization. This change
removes the "num placeholders" settings which
means that we slightly modify the public API for HTTP/3
config: `h3::Config::new()` and `quiche_h3_config_new()`
by removing that parameter.

Placeholders were a server-advertised feature, so this
change has the nice effect of making the API more sane
for clients.

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -93,7 +93,7 @@ fn http3(c: &mut Criterion) {
 
     let mut config = make_bench_config();
 
-    let mut h3_config = quiche::h3::Config::new(0, 1024, 0, 0).unwrap();
+    let mut h3_config = quiche::h3::Config::new(1024, 0, 0).unwrap();
 
     let mut s =
         quiche::h3::testing::Session::with_configs(&mut config, &mut h3_config)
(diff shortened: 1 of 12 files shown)


junhochoi:
LGTM however only concern I have is API change. This will break existing applications which use libquiche (I don't know how many in the field). Of course QUIC/H3 draft is still evolving so expecting a few more change like this, I think it's good time to think about how we keep backward compatibility of APIs or making APIs more extensible with such changes.

LPardue:
@junhochoi I agree with your concerns.

Since we are making a breaking change, I decided to go the whole hog and make the creation of the h3::Config option more extensible by adding functions to change specific properties. This gives us better granularity over changes to the properties available today, and sets a framework for future additions.

ghedo:
This looks great, though we can probably avoid calling `set_max_header_list_size()` explicitly in most places (benchmarks, examples, tests, ...). We were setting an explicit size before because we had to, but we don't need to anymore.

LPardue:
Good point!

junhochoi:
LGTM. I think this API style is better.

bagder:
Thanks to @LPardue's PR, curl is ready for the API change.

PR #181, opened 2019-09-24
h3: enforce the max header list size setting

ghedo:


src/h3/qpack/mod.rs
```diff
@@ -53,6 +53,9 @@ pub enum Error {
 
     /// The decoded QPACK header name or value is not valid.
     InvalidHeaderValue,
+
+    /// The decoded QPACK header block exceeded the size limit.
+    HeaderBlockTooLarge,
```
LPardue: Bikeshedding warning, I would call this `HeaderListTooLarge`.

A "decoded header block" is a header list, and "encoded header list" is a header block. So to me, the only way a header block can be too large is if it exceeds the size of a stream (which should be caught as a frame size error in HEADERS).

In practice the more likely case is that header blocks are small but get decoded to something much bigger that blows out the limit of SETTINGS_MAX_HEADER_LIST_SIZE.

ghedo: Yeah, good point. Fixed.

src/h3/mod.rs
```diff
@@ -1257,10 +1257,18 @@ impl Connection {
...
                 }
 
+                // Local settings are always set, so this is safe.
+                let max_size = self.local_settings.max_header_list_size.unwrap();
+
                 let headers = self
                     .qpack_decoder
-                    .decode(&mut header_block[..])
-                    .map_err(|_| Error::QpackDecompressionFailed)?;
+                    .decode(&mut header_block[..], max_size as usize)
```
junhochoi: Is it safe to assume `max_size` as `usize`? `self.local_settings.max_header_list_size` is `u64` type.
I don't think practically it never exceed 32-bit, but just for clarity.

ghedo:
This needs to be updated for changes in #180 (local settings are not initialized by default).

src/h3/mod.rs
ghedo: My thinking was that the header list needs to be allocated in memory anyway, so it can't exceed `usize::MAX`, which in turn means that truncating the `max_size` value to that should be safe (the main annoyance with using `u64` is that we would need to cast `usize` values in the `decode()` function 3 times, rather than do a single `u64` -> `usize` cast here).

junhochoi: Sounds good. This `usize` is platform dependent keep annoying me :)

ghedo:
This is now updated and based on top of #180.

src/h3/mod.rs
ghedo: So, thinking about this some more, it's probably just safer to use `u64` everywhere even if it's slightly more annoying, so I changed this to use `u64` instead of `usize` now.

LPardue:
Looks good. Just to ask the question, since we changed the API to allow undefined == infinite but have an internal limit, whether we want to document that somewhere.  I think it doesn't really help because an application has bigger issues if the peer wants to send a u64::MAX sized header list.

ghedo:
I guess we can document default values in the various `Config::set_*` methods.

LPardue:
ack. Lets do that in a separate single-shot PR.

PR #185, opened 2019-09-27
octets: fix macro parameter expansion inside unsafe block

### LINKED ISSUE #184: Macros in src/octet.rs may have unexpected unsafe behavior
https://github.com/cloudflare/quiche/blob/d646a60b497be4a92046fbd705161164388b73d2/src/octets.rs#L69-L73

https://github.com/cloudflare/quiche/blob/d646a60b497be4a92046fbd705161164388b73d2/src/octets.rs#L99-L103

These both evaluate some of their parameters in an `unsafe` block. This allows the callers of these macros to use an unsafe expression there, without needing to specify an `unsafe` block.

This is almost always unintended (and indeed, there's no comments suggesting it was intended). I'd suggest doing `let len = $len;` outside of the `unsafe` block to avoid this.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=578b803a28085939cd5a4791a4fb4dd4 is an example of how this can fail.

LPardue:
Fixes #184

### DIFF
--- a/src/octets.rs
+++ b/src/octets.rs
@@ -55,18 +55,19 @@ impl std::error::Error for BufferTooShortError {
 
 macro_rules! peek_u {
     ($b:expr, $ty:ty, $len:expr) => {{
+        let len = $len;
         let src = &$b.buf[$b.off..];
 
-        if src.len() < $len {
+        if src.len() < len {
             return Err(BufferTooShortError);
         }
 
         let mut out: $ty = 0;
         unsafe {
             let dst = &mut out as *mut $ty as *mut u8;
-            let off = (mem::size_of::<$ty>() - $len) as isize;
+            let off = (mem::size_of::<$ty>() - len) as isize;
 
-            ptr::copy_nonoverlapping(src.as_ptr(), dst.offset(off), $len);
+            ptr::copy_nonoverlapping(src.as_ptr(), dst.offset(off), len);
         };
 
         Ok(<$ty>::from_be(out))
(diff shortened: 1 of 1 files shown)


src/octets.rs
```diff
@@ -93,14 +94,16 @@ macro_rules! put_u {
...
         }
 
+        let len = $len;
+
         #[allow(clippy::range_plus_one)]
-        let dst = &mut $b.buf[$b.off..($b.off + $len)];
+        let dst = &mut $b.buf[$b.off..($b.off + len)];
 
         unsafe {
             let src = &<$ty>::to_be($v) as *const $ty as *const u8;
```
alex: You've also got `$v` in this macro, sorry I missed that on first review.

LPardue: Fixed! thanks.

src/octets.rs
```diff
@@ -66,11 +66,12 @@ macro_rules! peek_u {
         }
 
         let mut out: $ty = 0;
+        let len = $len;
```
ghedo: It'd probably be nicer to put this above the `if src.len() < $len` (i.e. immediately before or after `let src = ...`) for consistency, so we would use `len` instead of `$len` everywhere.

src/octets.rs
```diff
@@ -93,14 +94,17 @@ macro_rules! put_u {
             return Err(BufferTooShortError);
         }
 
+        let v = $v;
+        let len = $len;
```
ghedo: Same.

src/octets.rs
LPardue: yeah I missed that, thanks.

ghedo: Updated.

PR #186, opened 2019-09-27
fuzz: add initial fuzzers based on libfuzzer

ghedo:
Supersedes #179. Even Mayhem seems to like this more so it seems like the way to go.

There are two downsides though:
1. cargo-fuzz only seems to work with Rust nightly.
2. cargo-fuzz doesn't support an "initialization" step (e.g. to create quiche `Config` object), so we need to hack one in using `lazy_static`.


ghedo:
Also, the `lazy_static` hack could be removed if https://github.com/rust-fuzz/libfuzzer-sys/pull/33 (or similar) was merged into `libfuzzer-sys`.

camshaft:
You should consider using [bolero](https://github.com/camshaft/bolero/#usage) 😄. I ran into similar downsides and wanted something better. You can see a feature comparison here: https://github.com/camshaft/bolero/#prior-work

PR #190, opened 2019-09-29
examples: log more connection information in clients

ghedo:


LPardue:
Look good. On my experimental qlog branch I chose a slightly different way to format hex strings. Either way, there are probably reuse cases so might want to consider if this is something more generally useful in `lib` or `utils`.

https://github.com/LPardue/quiche/blob/qlog/src/lib.rs#L3374

ghedo:
@LPardue yeah, that's nice, though in this case I just copied the function we are already using in `server.rs` and `http3-server.rs`, which is small although not particularly pretty.

LPardue:
Ah yes, there is already a fair amount of duplication across these examples anyway. In a way I like that the examples show how to use some of the data coming out of quiche.

 Guess we can come back an refactor later if we think its needed and/or useful.

examples/client.rs
```diff
@@ -134,6 +133,13 @@ fn main() {
     // Create a QUIC connection and initiate handshake.
     let mut conn = quiche::connect(url.domain(), &scid, &mut config).unwrap();
 
+    info!(
+        "connecting to {:} from {:} with scid {}",
```
LPardue: how about sticking in the QUIC version and ALPN in here - just so we can quickly spot obvious mistmatches?

ghedo: This is pre-handshake, so ALPN is always static and version is whatever the user sets with `--wire-version` (or the default value to trigger negotiation), so not sure this is particularly useful. Though it's easy enough to add.

LPardue: yeah. I guess that information might have been useful if the client examples would have needed to be more widely used but now we have curl + browser(s).

PR #198, opened 2019-10-03
ffi: avoid panic on multiple quiche_enable_debug_logging

LPardue:
Previously, multiple calls to `quiche_enable_debug_logging()` would
cause a Rust panic, requiring applications to implement their
own housekeeping to avoid the problem. With this change we
manage things more gracefully, returning 0 on success of or -1
if there was some problem.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -65,12 +65,17 @@ impl log::Log for Logger {
 #[no_mangle]
 pub extern fn quiche_enable_debug_logging(
     cb: extern fn(line: *const u8, argp: *mut c_void), argp: *mut c_void,
-) {
+) -> c_int {
     let argp = atomic::AtomicPtr::new(argp);
     let logger = Box::new(Logger { cb, argp });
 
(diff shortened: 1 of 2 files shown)


src/ffi.rs
```diff
@@ -65,12 +65,18 @@ impl log::Log for Logger {
...
 pub extern fn quiche_enable_debug_logging(
     cb: extern fn(line: *const u8, argp: *mut c_void), argp: *mut c_void,
-) {
+) -> c_int {
     let argp = atomic::AtomicPtr::new(argp);
     let logger = Box::new(Logger { cb, argp });
 
-    log::set_boxed_logger(logger).unwrap();
-    log::set_max_level(log::LevelFilter::Trace);
+    match log::set_boxed_logger(logger) {
```
ghedo: Instead of using match you could more simply do:
```rust
if log::set_boxed_logger(logger).is_err() {
  return -1;
}
```

PR #199, opened 2019-10-04
 stream: fix handling of remote max streams limit updates

ghedo:
Before this change we would decrease the maximum stream limit every time
we attempt creating a new local stream, without tracking how many
streams we actually opened.

This would work if the remote stream limits were static, however the
peer can send a MAX_STREAMS frame (which carries absolute values, rather
than incremental updates) which would cause us to reset the limits to
the maximum value. We would then attempt to create new streams without
keeping into consideration streams that were already opened.

This change adds counters for the local streams that were actually
opened (like we already did for remote ones in df75104), so that we can
simply compare that against the max limits.

---

While working on #164 I thought I'd implement support for retrying HTTP/3 requests when stream limits are hit in the example client to test the server-side support for sending MAX_STREAMS, but I noticed that the client behavior was wrong, hence this fix (though this problem is unrelated to the other PR and has been here since forever, we just never supported sending more requests once the initial max streams limits were hit).

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -66,6 +66,12 @@ pub struct StreamMap {
     local_max_streams_uni: u64,
     local_max_streams_uni_next: u64,
 
+    /// The total number of bidirectional streams opened by the local endpoint.
+    local_opened_streams_bidi: u64,
+
+    /// The total number of unidirectional streams opened by the local endpoint.
+    local_opened_streams_uni: u64,
+
     /// Queue of stream IDs corresponding to streams that have buffered data
     /// ready to be sent to the peer. This also implies that the stream has
     /// enough flow control credits to send at least some of that data.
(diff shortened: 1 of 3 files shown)


ghedo:
For some more context, in the updated tests in this PR, before the client wouldn't return the `Error::StreamLimit` error when creating the last stream (which exceeds the server limits) and instead tried to create the stream anyway (then the server would close the connection).

src/stream.rs
```diff
@@ -159,17 +165,25 @@ impl StreamMap {
...
-                            .checked_sub(1)
-                            .ok_or(Error::StreamLimit)?,
-
-                    (true, false) =>
-                        self.peer_max_streams_uni = self
-                            .peer_max_streams_uni
-                            .checked_sub(1)
-                            .ok_or(Error::StreamLimit)?,
+                    (true, true) => {
+                        self.local_opened_streams_bidi += 1;
```
LPardue: couldn't this screw up counts if the local peer keeps trying to open streams while the remote peer doesn't update the max? Would it be better to test the condition like `self.local_opened_streams_bidi +1 <= self.peer_max_streams_bidi` and then only increment `self.local_opened_streams_bidi` much later when we actually succeed in `Stream::new()`?

ghedo: Ah, nice catch. This actually explains a bug I was investigating where depending on the max streams limit, the client would still timeout sometimes. I moved the increments to after the checks (even for the remote streams case, even though in that case the connection would just fail anyway).

PR #200, opened 2019-10-07
pick lowest between idle and loss detection timers

ghedo:
Currently if idle timer is lower than loss timer, but loss timer is not
None, we will pick the loss timer over the idle one, so we might end up
never timing out the connection if there are outstanding packets that
have not been acknowledged.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2199,13 +2199,18 @@ impl Connection {
         }
 
         let timeout = if self.draining_timer.is_some() {
+            // Draining timer takes precedence over all other timers. If it is
+            // set it means the connection is closing so there's no point in
+            // processing the other timers.
             self.draining_timer
-        } else if self.recovery.loss_detection_timer().is_some() {
-            self.recovery.loss_detection_timer()
-        } else if self.idle_timer.is_some() {
-            self.idle_timer
         } else {
-            None
+            // Use the lowest timer value (i.e. "sooner") among idle and loss
+            // detection timers. If they are both unset (i.e. `None`) then the
+            // result is `None`, but if at least one of them is set then a
+            // `Some(...)` value is returned.
+            let timers = [self.idle_timer, self.recovery.loss_detection_timer()];
+
+            timers.iter().filter_map(|&x| x).min()
         };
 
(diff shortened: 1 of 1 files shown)


ghedo:
Oh wait, this doesn't work actually.

ghedo:
Correction, it does work, but causes a bunch of spurious timeouts when draining timer is set. Updated it with fix.

junhochoi:
> Currently if idle timer is lower than loss timer, but loss timer is not None, 

Can you describe what scenario it is? My understanding is that idle timeout is always bigger than PTO. ("10.2.  Idle Timeout" of -transport) and if there is unacked data we should continue with PTO  first.

ghedo:
@junhochoi the PTO timer has a backoff mechanism (`timeout * (2 ^ pto_count)`) in case of repeated loss, so the idea here is that if `pto_count` is big enough, then the PTO timer will exceed idle timer.

Indeed the spec says:

>  When a PTO timer expires, the PTO period MUST be set to twice its current value. This exponential reduction in the sender’s rate is important because the PTOs might be caused by loss of packets or acknowledgements due to severe congestion. **The life of a connection that is experiencing consecutive PTOs is limited by the endpoint’s idle timeout**.

That last sentence is what appears to not be working correctly here. Though we are actually still testing whether this works as intended so I added the `hold` label back.

junhochoi:
Also note that (-transport 10.2) it looks updated since draft 19:

> If the idle timeout is enabled, a connection is silently closed and the state is discarded when it remains idle for longer than both the advertised idle timeout (see Section 18.2) and three times the
   current Probe Timeout (PTO).
https://github.com/quicwg/base-drafts/commit/6f09658c18cfeca0df0c4b1fc435cceaf550b7e8

So idle timeout will be `max (<idle timeout advertised>, 3 * current_pto_timeout)`

And it seems like we restart idle timer only when sending packet, not when receiving packet. but this can be a separate issue.

ghedo:
We restart the idle timer when receiving packets as well https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L1363

PR #203, opened 2019-10-08
Idle timeout calculation update

junhochoi:
Follows QUIC transport Idle Timeout section:
```
  idle_timeout = max ( tp.idle_timeout, 3 * probe_timeout )
```

get_idle_timeout() function added.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1360,11 +1360,7 @@ impl Connection {
             cmp::max(self.pkt_num_spaces[epoch].largest_rx_pkt_num, pn);
 
         if self.local_transport_params.idle_timeout > 0 {
-            self.idle_timer = Some(
-                now + time::Duration::from_millis(
-                    self.local_transport_params.idle_timeout,
-                ),
-            );
+            self.idle_timer = Some(now + self.idle_timeout());
         }
 
         self.recv_count += 1;
@@ -1911,11 +1907,7 @@ impl Connection {
             !self.ack_eliciting_sent &&
             self.local_transport_params.idle_timeout > 0
         {
-            self.idle_timer = Some(
-                now + time::Duration::from_millis(
-                    self.local_transport_params.idle_timeout,
-                ),
-            );
+            self.idle_timer = Some(now + self.idle_timeout());
         }
 
         if ack_eliciting {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1049,6 +1049,17 @@ impl Connection {
         Ok(done)
     }
 
+    /// Return idle timeout value
```
```suggestion
    /// Returns the idle timeout value.
```

src/lib.rs
```diff
@@ -1049,6 +1049,17 @@ impl Connection {
...
+    fn get_idle_timeout(&mut self) -> time::Duration {
+        let idle_timeout = cmp::max(
+            time::Duration::from_millis(self.local_transport_params.idle_timeout),
+            3 * self.recovery.pto(),
+        );
+        trace!("{} idle timeout = {:?}", self.trace_id, idle_timeout);
```
ghedo: Do we really need this log? it seems like it would be quite spammy since it would log for almost every received and sent packets.

src/lib.rs
```diff
@@ -1049,6 +1049,17 @@ impl Connection {
         Ok(done)
     }
 
+    /// Return idle timeout value
+    fn get_idle_timeout(&mut self) -> time::Duration {
```
```suggestion
    fn idle_timeout(&mut self) -> time::Duration {
```

src/lib.rs
```diff
@@ -1360,11 +1371,7 @@ impl Connection {
             cmp::max(self.pkt_num_spaces[epoch].largest_rx_pkt_num, pn);
 
         if self.local_transport_params.idle_timeout > 0 {
-            self.idle_timer = Some(
-                now + time::Duration::from_millis(
-                    self.local_transport_params.idle_timeout,
-                ),
-            );
+            self.idle_timer = Some(now + self.get_idle_timeout());
```
```suggestion
            self.idle_timer = Some(now + self.idle_timeout());
```

src/lib.rs
```diff
@@ -1911,11 +1918,7 @@ impl Connection {
             !self.ack_eliciting_sent &&
             self.local_transport_params.idle_timeout > 0
         {
-            self.idle_timer = Some(
-                now + time::Duration::from_millis(
-                    self.local_transport_params.idle_timeout,
-                ),
-            );
+            self.idle_timer = Some(now + self.get_idle_timeout());
```
```suggestion
            self.idle_timer = Some(now + self.idle_timeout());
```

src/lib.rs
ghedo: Also, could you please move this method at the end of the `impl` block? That is, after `should_update_max_data()` (or above it).

Just so people reading from top to bottom get to the more meatier methods first, and the smaller helper methods don't get in the way.

src/lib.rs
junhochoi: It was for debugging (used trace!()) but I can remove it.

PR #204, opened 2019-10-08
Update idle timer after processing packets

junhochoi:
When receiving packets, idle timer is updated at every
`recv_single()` but we only need to update once when `recv()`
processed all received packets.
Also clears `ack_eliciting_sent` flag as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1046,6 +1046,13 @@ impl Connection {
             left -= read;
         }
 
+        // Update idle timer
+        if self.local_transport_params.idle_timeout > 0 {
+            self.idle_timer = Some(now + self.idle_timeout());
+        }
+
+        self.ack_eliciting_sent = false;
+
         Ok(done)
     }
 
@@ -1271,7 +1278,7 @@ impl Connection {
                     self.trace_id, payload_len, pn,
                 );
 
-                return Ok(header_len + payload_len);
+                return Err(Error::Done);
             }
         }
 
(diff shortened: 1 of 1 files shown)


ghedo:
I think the idle timeout should only be updated after _successfully_ processing incoming packets. However after this change packets that are ignored (e.g. see https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L1217, fixed typos in https://github.com/cloudflare/quiche/pull/205) would cause the timeout to be reset, so an attacker (or just random junk) could potentially keep the connection alive forever by injecting invalid packets.

junhochoi:
> I think the idle timeout should only be updated after _successfully_ processing incoming packets. However after this change packets that are ignored (e.g. see https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L1217, fixed in #205) would cause the timeout to be reset, so an attacker (or just random junk) could potentially keep the connection alive forever by injecting invalid packets.

Ok. Probably I missed that part. Then my question is, why `recv_single()` returns `Ok(size)` when packet is dropped, not returning `Error::Done`? L1229 and L1271 too.

ghedo:
The reason basically is to be able to process coalesced packets properly. For example, a client might send a coalesced packet with both Initial and Handshake packets in it, but the server might have dropped the initial keys already, so without returning `Ok(...)`, the server wouldn't be able to just ignore the Initial packet but process the Handshake one.

Now, this is not exactly a common case, so we might want to discuss whether we should even do this in the first place, though I don't think there's anything preventing the client from doing that either (i.e. I think it's valid behavior according to the spec) so we might end up forbidding valid behavior if we don't do this.

ghedo:
In any case we should probably update the `recv_single()` documentation to mention this, and maybe even add a test.

ghedo:
Also, in the case on L1271 we probably don't actually need to return `Ok(...)` though, we could just return `Error::Done`.

src/lib.rs
```diff
@@ -1046,6 +1046,13 @@ impl Connection {
             left -= read;
         }
 
+        // Update idle timer
+        if self.local_transport_params.idle_timeout > 0 {
+            self.idle_timer = Some(now + self.idle_timeout());
```
LPardue: `now` is not in scope here

junhochoi:
Closing now. This seems micro-optimization, we can take a look later.

PR #206, opened 2019-10-09
deps: update BoringSSL submodule

ghedo:
Requires a slight tweak to a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,11 @@ addons:
 matrix:
   include:
    - rust: stable
+     before_install:
+      # Install and use the current stable release of Go
+      - gimme --list
+      - eval "$(gimme stable)"
+      - gimme --list
      install:
       - rustup component add clippy
      script:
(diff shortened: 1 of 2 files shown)


LPardue:
Not a huge fan of all the asserts in the test but I guess its always been like that. Would it be better to have a single-line assert e.g. `assert(client.completed && !client.confirmed)`? I don't know. However, if this test started failing we'd have to look at the surrounding statements for context, whereas a single assert would express the failure condition.

ghedo:
@LPardue not sure I understand what you don't like here. But yeah, if you merge multiple conditions in the same assert then it becomes harder to figure out what is actually failing.

LPardue:
I guess what I was moaning about is that we end up asserting the same conditions multiple times e.g. `assert!(pipe.client.handshake_completed);` is done 5 times. If there is a risk that `pipe.client.handshake_completed` can return to false after having been set to true, then the current approach is fine. But my first inspection of the test implied that it cares more about detecting the progression of different states from false -> true.

But this is just a moan, happy to move on.

ghedo:
So, apparently some of the changes in the test aren't actually needed with newer BoringSSL (had originally written this with a slightly outdated BoringSSL), so updated that commit to only add the client-side testing and leave server-side alone.

@LPardue yeah, the duplicated asserts are mostly to make it clear what the full expected status is, they are not really necessary I think (i.e. the conditions should not be going back to false once true). It just seemed more clear to me :man_shrugging:

PR #209, opened 2019-10-09
Use include instead of exclude

junhochoi:
For internal testing, using `exclude` will cause a lot of
directory traversing when running `cargo test`, slow down
test run a much (adding a minutes) when there is a lot of
files to scan and where file system operation is slow, such
as MacOS + Docker environment.

Using `include` will prevent this from happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
+      - RUSTFLAGS="-D warnings" cargo package --verbose --allow-dirty
       - cargo clippy --examples -- -D warnings
       - cargo doc --no-deps
       - make -C examples
@@ -41,6 +42,7 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
+      - RUSTFLAGS="-D warnings" cargo package --verbose --allow-dirty
       - cargo fmt -- --check
       - cargo doc --no-deps
       - make -C examples
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/boringssl
 /target
 **/*.rs.bk
 Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,16 +10,25 @@ readme = "README.md"
 keywords = ["quic", "http3"]
 categories = ["network-programming"]
 license = "BSD-2-Clause"
-exclude = [
-    "deps/boringssl/crypto/**/*_test.txt",
-    "deps/boringssl/crypto/**/*_tests.txt",
-    "deps/boringssl/**/nist_cavp/*.txt",
-    "deps/boringssl/fuzz/*",
-    "deps/boringssl/go.mod",
-    "deps/boringssl/ssl/test/runner/*",
-    "deps/boringssl/third_party/wycheproof_testvectors/*",
-    "fuzz/*",
-    "tools/*",
+include = [
+     "/*.md",
+     "/*.toml",
+     "/CODEOWNERS",
+     "/COPYING",
+     "/benches",
+     "/deps/boringssl/**/*.[chS]",
+     "/deps/boringssl/**/*.cc",
+     "/deps/boringssl/**/*.errordata",
+     "/deps/boringssl/**/*.go",
+     "/deps/boringssl/**/*.pl",
+     "/deps/boringssl/**/CMakeLists.txt",
+     "/deps/boringssl/*.md",
+     "/deps/boringssl/LICENSE",
+     "/deps/boringssl/sources.cmake",
+     "/examples",
+     "/include",
+     "/quiche.svg",
+     "/src",
 ]
 
 [features]


ghedo:
It looks like this causes `cargo package` to fail (I should probably add it to Travis CI builds...).

In any case the point of the `exclude` is to leave unneeded files out of the quiche crate, as there's a 10MB limit for uploads to crates.io and if we included al BoringSSL files we wouldn't be able to publish new versions to crates.io.

junhochoi:
> It looks like this causes `cargo package` to fail (I should probably add it to Travis CI builds...).
> 
> In any case the point of the `exclude` is to leave unneeded files out of the quiche crate, as there's a 10MB limit for uploads to crates.io and if we included al BoringSSL files we wouldn't be able to publish new versions to crates.io.

Actually I don't mind updating the list. What files need to be included for crate packaging? Do we even need to include boringssl?

junhochoi:
Let me test `cargo package --dry-run` a little more.

Lekensteyn:
For future reference, the issue was reproduced as follows:
```
git clone --recursive https://github.com/cloudflare/quiche /tmp/quiche # tag 0.1.0
docker run --rm -it -v /tmp/quiche:/quiche rust:1.38
apt update && apt install -y golang cmake strace
# Ensure build results are preserved between source dirs
export CARGO_TARGET_DIR=/tmp/cargo-target-dir
```

With Docker desktop 2.1.0.3 on macOS 10.14.6, this took:
```
root@ab323d621b73:/quiche# time cargo test fjsf
    Updating crates.io index
   Compiling quiche v0.1.0 (/quiche)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 40s
     Running /tmp/cargo-target-dir/debug/deps/quiche-490e341714dee5b5

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 186 filtered out


real	1m40.519s
user	1m3.090s
sys	0m19.040s
```

With the Cargo.toml patch that adds `include`, it is a bit faster on macOS:
```
root@ffb488be23f0:/quiche# time cargo test sdfa
   Compiling quiche v0.1.0 (/quiche)
    Finished dev [unoptimized + debuginfo] target(s) in 37.80s
     Running /tmp/cargo-target-dir/debug/deps/quiche-490e341714dee5b5

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 186 filtered out

real	0m37.785s
user	1m4.570s
sys	0m15.710s
```

but is still slower than running the tests outside the osxfs mountpoint. Copying `/quiche` to `/quiche2` and then running tests from there completes in 1.3s. Using the `-v $PWD:/quiche:delegated` option made no significant difference (initial run 1m46s, second cached run 1m1s).

When quiche is a git submodule and part of a larger tree, Rust (via libgit2) starts scanning the whole parent repository. With slow lstat/stat on osxfs, this results in a very noticeable slowdown. <details><summary>perf trace excerpt</summary>

```
-   44.77%     0.00%  cargo            cargo                    [.] cargo::sources::path::PathSource::last_modified_file
   - 44.77% cargo::sources::path::PathSource::last_modified_file
      - 44.23% cargo::sources::path::PathSource::list_files
         - 44.22% cargo::sources::path::PathSource::list_files_git
            - 22.63% git2::repo::Repository::statuses
               - git_status_list_new
                  - 22.61% git_diff_index_to_workdir
                     - 22.60% git_diff__from_iterators
                        - 19.30% git_submodule__status
                           - 19.23% git_diff_index_to_workdir
                              - 19.21% git_diff__from_iterators
                                 - 15.78% git_diff__oid_for_entry
                                    - 11.96% git_filter_list_load
                                       - git_filter_list__load_ext
                                          - 11.93% git_attr_get_many_with_session
                                             - 7.98% collect_attr_files
                                                - 4.33% git_attr_cache__get
                                                   - 3.79% git_attr_file__load
                                                      + 3.44% __xstat64
                                                - 2.34% git_path_walk_up
                                                   - 2.33% push_one_attr
                                                      - 2.32% git_attr_cache__get
                                                         - 2.13% git_attr_file__load
                                                            + 1.76% __xstat64
                                                + 1.05% system_attr_file
                                             - 3.92% git_attr_path__init
                                                - 3.91% git_path_isdir
                                                   + 3.90% __xstat64
                                    + 1.96% __libc_close (inlined)
                                    + 1.80% git_futils_open_ro
                                 + 3.30% filesystem_iterator_advance_into
                        + 2.67% git_diff__oid_for_entry
                        + 0.60% filesystem_iterator_advance_into
            - 21.57% cargo::sources::path::PathSource::list_files_git
               - 21.50% git2::repo::Repository::statuses
                  - git_status_list_new
                     - 21.44% git_diff_index_to_workdir
                        - 21.41% git_diff__from_iterators
                           - 17.39% git_diff__oid_for_entry
                              - 13.54% git_filter_list_load
                                 - git_filter_list__load_ext
                                    - 13.52% git_attr_get_many_with_session
                                       - 8.89% collect_attr_files
                                          - 4.74% git_attr_cache__get
                                             - 4.20% git_attr_file__load
                                                + 3.83% __xstat64
                                          - 2.89% git_path_walk_up
                                             - 2.89% push_one_attr
                                                - 2.87% git_attr_cache__get
                                                   - 2.70% git_attr_file__load
                                                      + 2.22% __xstat64
                                          + 1.08% system_attr_file
                                       + 4.61% git_attr_path__init
                              + 2.10% __libc_close (inlined)
                              + 1.72% git_futils_open_ro
                           + 3.87% filesystem_iterator_advance_into
      + 0.53% std::sys::unix::fs::stat
```
</details>

junhochoi:
Ok updated include list. Now `cargo package` runs successfully and it's 3.4MB

```
-rw-r--r--   1 junho  staff  3580307 Oct  9 13:33 quiche-0.1.0.crate
```

Actually smaller than 0.1.0 crate (3.8M)

junhochoi:
Added `cargo package` to Travis CI. (looks working)

.travis.yml
```diff
@@ -60,6 +62,7 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
+      - RUSTFLAGS="-D warnings" cargo package --verbose --allow-dirty
```
ghedo: We can avoid doing this on macOS and Windows, since realistically we'll only run this on Linux.

junhochoi: Ok. I left `cargo package` in linux build only.

PR #210, opened 2019-10-10
log more TLS info when connection is established

ghedo:
Also did some cleaning while at it.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -355,7 +355,7 @@ impl Handshake {
         })
     }
 
-    pub fn get_quic_transport_params(&self) -> &[u8] {
+    pub fn quic_transport_params(&self) -> &[u8] {
         let mut ptr: *const u8 = ptr::null();
         let mut len: usize = 0;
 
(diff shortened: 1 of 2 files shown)


src/tls.rs
```diff
@@ -395,12 +395,48 @@ impl Handshake {
...
+
+    pub fn sigalg(&self) -> Option<String> {
+        let sigalg = unsafe {
+            let sigalg_id = SSL_get_peer_signature_algorithm(self.as_ptr());
+            if sigalg_id == 0 {
+                return None;
+            }
+
+            let sigalg_name = SSL_get_signature_algorithm_name(sigalg_id, 1);
+            match std::ffi::CStr::from_ptr(sigalg_name).to_str() {
```
LPardue: since we don't seem to care about the Ok case could we make this 
```
if std::ffi::CStr::from_ptr(sigalg_name).to_str().is_err() {
  return None;
}
```

ghedo: We do need the return value. The `match` "return" value in the `Ok(v)` case is then "returned" into `let curve = ...`.

LPardue: Oops!

LPardue: Oops!

PR #211, opened 2019-10-10
expand documentation for some transport parameters

ghedo:


LPardue:
Looks good. But I guess someone could misread the description where it says `increase the limit automatically`, especially for the case of max_streams where the concurrency stays the same and is managed by quiche.

Is the limit increased when the we receive a stream close, or when the application reads all data from the stream and obtains the stream close?

ghedo:
Yeah I don't like the wording either, but I couldn't come up with anything better. Any suggestion?

Also, yes, "when the stream is closed" is not correct, will fix.

LPardue:
Maybe it will pay to just be more explicit. For example, 

instead of 

```
    /// When set to a non-zero value quiche will only allow at most `v` bytes
    /// of incoming stream data to be buffered for the whole connection (that
    /// is, data that is not yet read by the application) and will increase the
    /// limit automatically as the buffer is emptied.
```

say

```
    /// When set to a non-zero value quiche will only allow at most `v` bytes
    /// of incoming stream data to be buffered for the whole connection (that
    /// is, data that is not yet read by the application). When an application reads
    /// data, quiche schedules a `MAX_STREAM_DATA` frame that increases the limit
    /// by the number of bytes read.
```

but I'm not totally sold on that because it is a bit cumbersome

ghedo:
I reworded it a bit, which I think makes it clearer. I also fixed the closed vs. completed thing and expanded on that a bit as well.

PR #213, opened 2019-10-11
Add iOS build to CI

junhochoi:
To distinguish from MacOS build, I added `xcode11.2`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,14 @@ matrix:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
    - rust: stable
+     os: osx
+     osx_image: xcode11.2
+     install:
+      - rustup target add aarch64-apple-ios x86_64-apple-ios
+      - cargo install cargo-lipo
+     script:
+      - RUSTFLAGS="-D warnings" cargo lipo --release --verbose
+   - rust: stable
      os: windows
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -60,6 +60,14 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
+   - rust: stable
+     os: osx
+     osx_image: xcode11.2
+     install:
+      - rustup target add aarch64-apple-ios armv7-apple-ios armv7s-apple-ios x86_64-apple-ios i386-apple-ios
```
ghedo: From the build log it seems only the `x86_64-apple-ios` target is used, do we need to install all the other ones as well?

junhochoi: lipo default is `aarch64-apple-ios,x86_64-apple-ios` and from the build log it built both platforms as follows:
```
...
"/Users/travis/build/cloudflare/quiche/target/universal/release/libquiche.a" "/Users/travis/build/cloudflare/quiche/target/aarch64-apple-ios/release/libquiche.a" "/Users/travis/build/cloudflare/quiche/target/x86_64-apple-ios/release/libquiche.a"
...
```
which means I can remove 32bit buid (armv7, armv7s, i386). I think apple already dropped 32bit support earlier, so those are only for supporting old version of iOS.

PR #214, opened 2019-10-12
Add Android build to CI

junhochoi:
Add android target, build 4 platforms. Since android
language support doesn't include rust and cmake version is too
old, install it separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-language: rust
-
 addons:
   apt:
     packages:
@@ -9,7 +7,8 @@ addons:
 
 matrix:
   include:
-   - rust: stable
+   - language: rust
+     rust: stable
      before_install:
       # Install and use the current stable release of Go
       - gimme --list
@@ -30,7 +29,8 @@ matrix:
       # quic-trace-log
       - RUSTFLAGS="-D warnings" cargo build --release --verbose --manifest-path tools/quic-trace-log/Cargo.toml
       - cargo clippy --manifest-path tools/quic-trace-log/Cargo.toml -- -D warnings
-   - rust: nightly
+   - language: rust
+     rust: nightly
      before_install:
       # Install and use the current stable release of Go
       - gimme --list
@@ -57,12 +57,14 @@ matrix:
       # quic-trace-log
       - RUSTFLAGS="-D warnings" cargo build --release --verbose --manifest-path tools/quic-trace-log/Cargo.toml
       - cargo fmt --manifest-path tools/quic-trace-log/Cargo.toml -- --check
-   - rust: stable
+   - language: rust
+     rust: stable
      os: osx
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
-   - rust: stable
+   - language: rust
+     rust: stable
      os: osx
      osx_image: xcode11.2
      install:
@@ -70,7 +72,8 @@ matrix:
       - cargo install cargo-lipo
      script:
       - RUSTFLAGS="-D warnings" cargo lipo --release --verbose
-   - rust: stable
+   - language: rust
+     rust: stable
      os: windows
      before_install:
       - choco install golang nasm
@@ -79,7 +82,8 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
-   - rust: stable
+   - language: rust
+     rust: stable
      env:
       NGINX_VER=1.16.1
      before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -69,6 +71,36 @@ matrix:
...
+      - export PATH=$HOME/.cargo/bin:$PATH
+      - rustup default stable
+      - rustup target add aarch64-linux-android arm-linux-androideabi armv7-linux-androideabi i686-linux-android
+      # Install android ndk
+      - NDK_VER=r13b
+      - curl -ondk.zip -q https://dl.google.com/android/repository/android-ndk-$NDK_VER-linux-x86_64.zip
+      - unzip -q ndk.zip -d $HOME
+      - export ANDROID_NDK_HOME=$HOME/android-ndk-$NDK_VER
+      # Install latest cmake. preinstalled version is too old (3.2.2)
+      - CMAKE_URL=https://cmake.org/files/v3.15/cmake-3.15.4-Linux-x86_64.tar.gz
```
ghedo: What version of cmake is required? The ubuntu bionic image has 3.12 https://docs.travis-ci.com/user/reference/bionic/#compilers-and-build-toolchain if we could just use that image we wouldn't need to install cmake manually.

.travis.yml
```diff
@@ -69,6 +71,36 @@ matrix:
...
+        - build-tools-26.0.1
+        # Minimum API level supported
+        - android-21
+     install:
+      # Install rust manually
+      - curl https://build.travis-ci.org/files/rustup-init.sh -sSf | sh -s -- -y --default-toolchain stable
+      - export PATH=$HOME/.cargo/bin:$PATH
+      - rustup default stable
+      - rustup target add aarch64-linux-android arm-linux-androideabi armv7-linux-androideabi i686-linux-android
+      # Install android ndk
```
ghedo: Can we use the `sdkmanager` command for this? https://github.com/android/ndk-samples/blob/master/.travis.yml#L18

.travis.yml
```diff
@@ -69,6 +71,36 @@ matrix:
...
+     dist: trusty
+     android:
+       components:
+        - build-tools-26.0.1
+        # Minimum API level supported
+        - android-21
+     install:
+      # Install rust manually
+      - curl https://build.travis-ci.org/files/rustup-init.sh -sSf | sh -s -- -y --default-toolchain stable
+      - export PATH=$HOME/.cargo/bin:$PATH
```
ghedo: Should probably just put this in an `env` section https://docs.travis-ci.com/user/environment-variables/

.travis.yml
```diff
@@ -69,6 +71,36 @@ matrix:
...
+      - rustup default stable
+      - rustup target add aarch64-linux-android arm-linux-androideabi armv7-linux-androideabi i686-linux-android
+      # Install android ndk
+      - NDK_VER=r13b
```
ghedo: Same about `env` https://docs.travis-ci.com/user/environment-variables/

.travis.yml
junhochoi: We need cmake 3.6.0 and later otherwise boringssl build will fail. For Android build we need to use `trusty` image. https://docs.travis-ci.com/user/languages/android/
And `trusty` image has cmake 3.2.2 (`/usr/bin/cmake`) so had to install newer version. This is outside of NDK build, so NDK cmake doesn't help.

.travis.yml
junhochoi: Here I am reusing how other rust project building with ndk (e.g. libc, ring).  I will take a look again.

.travis.yml
junhochoi: Ok. I was able to install `cmake` using `sdkmanager`, replacing this installation procedure.

.travis.yml
junhochoi: We still need to install `rust` this way unless we have a way to have an environment with 2 language (android and rust).

.travis.yml
junhochoi: Moved some constants for NDK and its url to env/global section.

.travis.yml
junhochoi: Now we use `sdkmanager` for `cmake`. I tested installing `ndk-bundle` with `sdkmanager` but it fails... I think it's because `ndk-bundle` coming with this `sdkmanager` is NDK r20 and we are not still compatible with this. There is toolchain script change somewhere between 13b and 20, breaking backward compatibility with old NDK. r13b was chosen based on current practice. If we want to be compatible with NDK 20 then we need a separate work.

ghedo:
Reabased on master, and also moved the `env` declaration to the actual build that uses those.

junhochoi:
> Reabased on master, and also moved the `env` declaration to the actual build that uses those.

Thanks. I think it's good to go.

junhochoi:
> > Reabased on master, and also moved the `env` declaration to the actual build that uses those.
> 
> Thanks. I think it's good to go.

Moved NDK_URL inside build script. It's too long so not looking good at CI matrix table.

PR #217, opened 2019-10-14
initial support for receiving 0-RTT packets

ghedo:
This adds initial support for enabling early data and processing 0-RTT
packets. In practice this only works on the server-side, as there is no
support for actually writing early data (or for the client to resume a
connection to beging with).

It *looks* like sending 0.5-RTT data from the server works as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -458,6 +458,11 @@ impl Config {
         self.tls_ctx.enable_keylog();
     }
 
+    /// Enables sending or receiving early data.
+    pub fn enable_early_data(&mut self) {
+        self.tls_ctx.set_early_data_enabled(true);
+    }
+
     /// Configures the list of supported application protocols.
     ///
     /// The list of protocols `protos` must be in wire-format (i.e. a series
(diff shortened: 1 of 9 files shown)


ghedo:
Since there's no support for actually sending 0-RTT data it's pretty hard to add tests, but I'm going to do some interop testing with other clients.

src/lib.rs
```diff
@@ -458,6 +458,11 @@ impl Config {
         self.tls_ctx.enable_keylog();
     }
 
+    /// Enables sending or receiving early data.
+    pub fn enable_early_data(&mut self) {
```
LPardue: do we want to expose this in FFI?

ghedo: Yeah, good point. Fixed.

ghedo:
We don't have to do anything on the server (except drop the packets). From the spec:

> After a client receives a Retry packet, 0-RTT packets are likely to have been lost or discarded by the server. A client SHOULD attempt to resend data in 0-RTT packets after it sends a new Initial packet.

PR #218, opened 2019-10-15
[wip] qlog: add a new crate for qlog and add example usage

LPardue:
This is not ready for merging but would benefit from more eyes.

The PR has 2 parts: the first is adding a QLOG data model, the second is some rough integration with http3-client to show the bare minimum of how we might integrate this. 

The QLOG datamodel that makes heavy use of serde to allow easy creation of QLOG compliant JSON. The structure is heavily based on the TypeScript schema - https://github.com/quiclog/qlog/blob/master/TypeScript/draft-01/QLog.ts. Sometimes this causes counter-intuitive code I'm hoping there are more Rust tricks to play.

The quiche integration needs more work - we shouldn't buffer all packets and frames and write at the end.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,10 @@ libc = "0.2"
 ring = "0.16"
 lazy_static = "1"
 
+serde_json = "1"
+
+qlog = {path = "./qlog"}
+
 [target."cfg(windows)".dependencies]
 winapi = { version = "0.3", features = ["wincrypt"] }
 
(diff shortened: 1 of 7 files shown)


LPardue:
Builds mostly failing in "method not used" type errors, we would benefit from unit tests and more integration (at worst, disable the error type).

LPardue:
It makes better sense to split this across multiple PRs, so closing this one.

PR #220, opened 2019-10-15
extras: add NGINX integration

ghedo:
This adds the integration of quiche with NGINX as a stand-alone patch.

Instructions for building and configuring NGINX with quiche support are
also provided, as well as documentation for contiguration directives.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,22 @@ matrix:
      script:
       - RUSTFLAGS="-D warnings" cargo build --release --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
+   - rust: stable
+     env:
+      NGINX_VER=1.16.1
+     before_install:
+      # Install and use the current stable release of Go
+      - gimme --list
+      - eval "$(gimme stable)"
+      - gimme --list
+     script:
+      - curl -O https://nginx.org/download/nginx-$NGINX_VER.tar.gz
+      - tar xvf nginx-$NGINX_VER.tar.gz
+      - |
+        cd nginx-$NGINX_VER &&
+        patch -p01 < ../extras/nginx/nginx-1.16.patch &&
+        ./configure --with-http_ssl_module --with-http_v2_module --with-http_v3_module --with-openssl="../deps/boringssl" --with-quiche=".." &&
+        make -j`nproc`
 
 deploy:
   provider: pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ matrix:
       - gimme --list
      script:
       - curl -O https://nginx.org/download/nginx-$NGINX_VER.tar.gz
-      - tar xvf nginx-$NGINX_VER.tar.gz
+      - tar xzf nginx-$NGINX_VER.tar.gz
       - |
         cd nginx-$NGINX_VER &&
         patch -p01 < ../extras/nginx/nginx-1.16.patch &&
--- a/extras/nginx/README.md
+++ b/extras/nginx/README.md
@@ -21,7 +21,7 @@ only works with the 1.16.x release branch (the latest stable release being
 
 ```
  % curl -O https://nginx.org/download/nginx-1.16.1.tar.gz
- % tar xvf nginx-1.16.1.tar.gz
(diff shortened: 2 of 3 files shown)


extras/nginx/README.md
```diff
@@ -0,0 +1,154 @@
...
+events {
+    worker_connections  1024;
+}
+
+http {
+    server {
+        listen 0.0.0.0:443 ssl http2;
+        listen [::]:443 ssl http2;
+
+        listen 0.0.0.0:443 quic reuseport;
```
LPardue: Would it help to describe the `quic` directive in appropriate section of the readme?

ghedo: This mostly follows the NGINX HTTP/2 module documentation https://nginx.org/en/docs/http/ngx_http_v2_module.html which also doesn't do this. I think this is because the `listen` directive actually comes from a different module.

.travis.yml
```diff
@@ -79,6 +79,22 @@ matrix:
...
+      - eval "$(gimme stable)"
+      - gimme --list
+     script:
+      - curl -O https://nginx.org/download/nginx-$NGINX_VER.tar.gz
+      - tar xvf nginx-$NGINX_VER.tar.gz
+      - |
+        cd nginx-$NGINX_VER &&
+        patch -p01 < ../extras/nginx/nginx-1.16.patch &&
+        ./configure --with-http_ssl_module --with-http_v2_module --with-http_v3_module --with-openssl="../deps/boringssl" --with-quiche=".." &&
+        make -j`nproc`
```
junhochoi: Do we add a sample `nginx.conf`? If so I think it's good to run `nginx -c nginx-quic.conf -t` at the end, just for sanity check.

extras/nginx/README.md
```diff
@@ -0,0 +1,154 @@
...
+--------
+
+The first step is to [download and unpack the NGINX source
+code](https://nginx.org/en/download.html). Note that the HTTP/3 and QUIC patch
+only works with the 1.16.x release branch (the latest stable release being
+1.16.1).
+
+```
+ % curl -O https://nginx.org/download/nginx-1.16.1.tar.gz
+ % tar xvf nginx-1.16.1.tar.gz
```
junhochoi:
```suggestion
 % tar xzvf nginx-1.16.1.tar.gz
```

.travis.yml
```diff
@@ -79,6 +79,22 @@ matrix:
...
+     env:
+      NGINX_VER=1.16.1
+     before_install:
+      # Install and use the current stable release of Go
+      - gimme --list
+      - eval "$(gimme stable)"
+      - gimme --list
+     script:
+      - curl -O https://nginx.org/download/nginx-$NGINX_VER.tar.gz
+      - tar xvf nginx-$NGINX_VER.tar.gz
```
junhochoi:
```suggestion
      - tar xzf nginx-$NGINX_VER.tar.gz
```

extras/nginx/README.md
LPardue: that's fine by me

ghedo: I added comments to the example configuration to make this a bit more clear.

.travis.yml
ghedo: Why? The current version seems to be working fine in CI.

extras/nginx/README.md
ghedo: Same as below.

.travis.yml
ghedo: Maybe in the future we'll probably want to run the HTTP/3 tests, but for now I just want to make sure this builds as a start.

junhochoi: `nginx -t` will check configuration is valid, not serving anything (exit immediately). If new patch breaks anything config related it will detect.

.travis.yml
junhochoi: `v` might be unnecessary (too verbose) and `z` is added because it's gzipped. I think GNU tar is auto detecting gzip, but still a good practice to add it. Not a big deal so you might ignore.

.travis.yml
ghedo: Yeah but we don't have the configuration as file, so once we set that up we might as well run the tests as well.

PR #221, opened 2019-10-15
Confirm the handshake by Handshake ack.

junhochoi:
Currently we drop the handshake key when the handshake
completed (but not in sync yet) and send some 1-RTT data
and received 1-RTT ack.

In addition to this, mark the handshake confirmed when
Handshake packet + ACK is received and this clears all
crypto data in sender buffer.

This can happen when 1-RTT ack is lost but Handshake ack
is received, currently resulting Handshake key is not dropped during
the request.

Found during the interop with ATS.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2534,15 +2534,22 @@ impl Connection {
                     &self.trace_id,
                 )?;
 
-                // When we receive an ACK for a 1-RTT packet after handshake
-                // completion, it means the handshake has been confirmed.
-                if epoch == packet::EPOCH_APPLICATION && self.handshake_completed
-                {
-                    self.handshake_confirmed = true;
+                if self.handshake_completed {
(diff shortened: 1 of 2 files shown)


ghedo:
TLS draft is pretty clear:

> In this document, the TLS handshake is considered confirmed at an endpoint when the following two conditions are met: the handshake is complete, and the endpoint has received an acknowledgment for a packet sent with 1-RTT keys

This PR does not respect those conditions. Alsom I have a feeling that this breaks the connection if the client sends additional Handshake packets after the ACK (client certificate maybe?). Need to do some tests.

ghedo:
> client certificate maybe?

I guess in this case the handshake wouldn't be complete actually, so that should still work.

junhochoi:
Ok. Also notified ATS engineer what is the issue, expecting some fix on their side in the future.

PR #224, opened 2019-10-16
update nginx HTTP/3 instructions

centminmod:
auto/configure seems to be for git cloned nginx repo builds ? for source tarball ./configure would be the correct one to use


ghedo:
Merged, thanks you!

PR #234, opened 2019-10-22
h3: add has_body flag to HTTP/3 Header Event

LPardue:
~~Previously, when `quiche::h3::Connection::poll()`
returned a `Headers` or `Data` event, there was
no way to tell if that event simultaneously closed
the indicated stream. Instead, `poll()` would need
to be called until the `Finished` event occurred.~~

~~With this change a boolean `fin` flag is returned
with the Event, which allows for some more
optimised work flows.~~

~~This changes the poll API and so has ripple effects. I chose to refactor `http3-client.rs` to act on the fin flag in every event handler. While for http3_test/runner.rs I chose to just ignore the flag and continue with the existing `Finished` event.~~

~~Implementing the feature the way I have reveals two things: 1) it's not possible to receive a a Data event with fin=true (because the stream has data on it that the app needs to read, 2) it's not possible to receive a Finished event with fin=false (because otherwise we seriously messed up).~~

The previous design was quite intrusive and as the final paragraph above highlighted had some edge cases, so I reworked it to only modify the Headers event.

Previously, when `quiche::h3::Connection::poll()`
returned a `Headers` event, there was no way to tell
if it simultaneously closed the indicated stream.
Instead, `poll()` would need to be called again
until the `Finished` event occurred.
   
With this change a boolean `fin` flag is returned
with the Event, which allows for some more
optimised work flows.
    
This is changes the Rust API, the Headers event
is changes from a `Vec<Header>` to a struct.
    
This changes the FFI, the `quiche_h3_event_for_each_header`
method takes an additional parameter.

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -93,7 +93,7 @@ fn http3(c: &mut Criterion) {
 
     let mut config = make_bench_config();
 
-    let h3_config = quiche::h3::Config::new().unwrap();
+    let mut h3_config = quiche::h3::Config::new().unwrap();
 
     let mut s =
         quiche::h3::testing::Session::with_configs(&mut config, &mut h3_config)
@@ -132,7 +132,7 @@ fn http3(c: &mut Criterion) {
                     s.pipe.flush_client(&mut buf).unwrap();
 
                     match s.server.poll(&mut s.pipe.server) {
-                        Ok((stream, quiche::h3::Event::Headers(_))) => {
+                        Ok((stream, quiche::h3::Event::Headers { .. })) => {
                             s.server
                                 .send_response(
                                     &mut s.pipe.server,
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -328,10 +328,10 @@ fn main() {
             // Process HTTP/3 events.
             loop {
                 match http3_conn.poll(&mut conn) {
-                    Ok((stream_id, quiche::h3::Event::Headers(headers))) => {
+                    Ok((stream_id, quiche::h3::Event::Headers { list, .. })) => {
                         info!(
                             "got response headers {:?} on stream id {}",
-                            headers, stream_id
+                            list, stream_id
                         );
                     },
 
--- a/extras/nginx/nginx-1.16.patch
+++ b/extras/nginx/nginx-1.16.patch
@@ -1546,7 +1546,7 @@ new file mode 100644
 index 000000000..972ecd223
 --- /dev/null
 +++ b/src/http/v3/ngx_http_v3.c
-@@ -0,0 +1,2093 @@
+@@ -0,0 +1,2092 @@
 +
 +/*
 + * Copyright (C) Cloudflare, Inc.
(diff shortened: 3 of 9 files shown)


src/h3/mod.rs
```diff
@@ -490,11 +490,23 @@ impl Header {
...
+/// Structure for `Event::Headers`
+///
+/// [`Event`]: enum.Event.html#Headers
+#[derive(Clone, Debug, PartialEq)]
+pub struct Headers {
+    /// Represents a Header list.
+    pub list: Vec<Header>,
+
+    /// Indicate if the Headers finish the stream
+    pub fin: bool,
```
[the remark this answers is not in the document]
LPardue: ah yeah now I remember and I prefer that, blame jet lag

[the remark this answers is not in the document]
ghedo: IIRC we don't actually return events for trailers right now, do we? But I'd still leave that as a separate PR to make review easier (don't think it matters too much if this goes in before that).

examples/http3-client.rs
```diff
@@ -347,36 +357,30 @@ fn main() {
...
-                            "{}/{} responses received",
-                            reqs_complete, reqs_count
-                        );
 
-                        if reqs_complete == reqs_count {
-                            info!(
-                                "{}/{} response(s) received in {:?}, closing...",
-                                reqs_complete,
+                            if process_stream_finished(
+                                false,
```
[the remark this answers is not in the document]
LPardue: I refactored things a little, the common code for handling a completed response is now called conditionally depending on if the Headers has a body or not.

LPardue:
This is ready for another review and should be GTG as long as the builds pass

extras/nginx/nginx-1.16.patch
```diff
@@ -1806,6 +1806,7 @@ index 000000000..972ecd223
 +    ngx_http_v3_stream_t      *stream;
 +    ngx_http_v3_srv_conf_t    *h3scf;
 +    ngx_http_v3_connection_t  *h3c;
++    bool                       has_body;
```
junhochoi: I think we need to initialize `has_body` to `false` (`http3-server.c` does that too)

LPardue: I think that doesn't suit nginx style (but I could be wrong). However after some offline discussion, we have a better API design that will avoid the problem. I'll push an update tomorrow.

benches/benches.rs
```diff
@@ -93,7 +93,7 @@ fn http3(c: &mut Criterion) {
 
     let mut config = make_bench_config();
 
-    let h3_config = quiche::h3::Config::new().unwrap();
+    let mut h3_config = quiche::h3::Config::new().unwrap();
```
junhochoi: Seems `mut` is unnecessary? (then you need to change L99 too)

extras/nginx/nginx-1.16.patch
```diff
@@ -1849,6 +1846,8 @@ index 000000000..972ecd223
 +        return;
 +    }
 +
++    stream->in_closed = !quiche_h3_event_header_has_body(ev);
```
junhochoi: In general I agree with API changes to report and of header separately, but why is have `has_body` suffix? It confuses me which body of header (because http header doesn't have a body...) so I suggest something like 'quiche_h3_event_header_finished()` to be clear with the intention (stop reading headers). I think it used to be `fin` but curious why the name is changed.

LPardue: The method name is a bit clunky but it basically allows for a C client to fetch the field from Rust Event::Headers structure that is documented in rustdoc.

Conceptually, we are trying to articulate whether the stream has more data on it, so applications can decide what to do (if anything).  The problem with the field name fin was that it is confusable with Event::Finished.

benches/benches.rs
ghedo: No, this file currently doesn't build, so this fixes that.

PR #239, opened 2019-10-27
update submodle boringssl url from G source to github mirror

HustCoderHu:
some countries can't access to G source

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,5 +1,4 @@
 [submodule "boringssl"]
 	path = deps/boringssl
 	url = https://github.com/google/boringssl.git
-    branch = master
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,4 +1,5 @@
 [submodule "boringssl"]
 	path = deps/boringssl
-	url = https://boringssl.googlesource.com/boringssl
+	url = https://github.com/google/boringssl.git
+    branch = master
```
ghedo: This is not needed (it also has odd indentation).

```suggestion
```

ghedo:
Merged, thanks @HustCoderHu!

PR #249, opened 2019-11-07
pin Connection

### LINKED ISSUE #45: Why is the Connection boxed by default?
Hey,

I was just skimming over the API and I was wondering why `connect` and `accept` return a boxed `Connection`? `Connection` is neither a trait object nor dynamically sized, so why the need for the box?

ghedo:
Due to the fact that we store a pointer to the Connection object inside
its own Handshake context, Connection objects cannot be moved (which is
why we return a boxed object from its constructor). Using Pin on it
makes the compiler enforce this.

Fixes #45.

---

Note this requires Rust 1.39.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ be linked directly into C/C++ applications.
 Building
 --------
 
-quiche requires Rust 1.38 or later to build. The latest stable Rust release can
+quiche requires Rust 1.39 or later to build. The latest stable Rust release can
 be installed using [rustup](https://rustup.rs/).
 
 Once the Rust build environment is setup, the quiche source code can be fetched
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -62,7 +62,7 @@ struct PartialResponse {
 }
 
 struct Client {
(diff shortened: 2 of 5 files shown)


LPardue:
Nightly builds are failing due to errors like

```
error: `extern` fn uses type `Config`, which is not FFI-safe

   --> src/ffi.rs:81:50

    |

81  | pub extern fn quiche_config_new(version: u32) -> *mut Config {

    |                                                  ^^^^^^^^^^^ not FFI-safe

    |

    = note: `-D improper-ctypes` implied by `-D warnings`

    = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct

    = note: this struct has unspecified layout

note: type defined here

   --> src/lib.rs:372:1
```

ghedo:
Looks like it's just the newer rust nightly, as master also fails :/

ghedo:
https://github.com/cloudflare/quiche/pull/251 should "fix" it.

ghedo:
Added some more context to the commit message.

PR #251, opened 2019-11-07
silence improper_ctypes warning

ghedo:
This is actually a false positive, as the structs are only exposed to
FFI as pointers, so they are effectively opaque from the non-Rust
perspective

The warning showed up in nightly after the following PR was merged:
https://github.com/rust-lang/rust/pull/65134


CryZe:
I've opened an issue discussing this here: https://github.com/rust-lang/rust/issues/66220

Issue #248, opened 2019-11-07
Unexpected http0.9 protocol when using cURL


SecurityInsanity:
:wave: Hey There (Thanks for the great library!),

I seem to be running into something unexpected while testing quiche (through the C FFI), and using a locally built cURL. When using cURL I must specify the: `--http0.9` option, but only when using my version (not the rust example).

In my C++ Code I run:

```
 // _config has been initialized with: quiche_config_new(QUICHE_PROTOCOL_VERSION);
quiche_config_set_application_protos(
      _config.get(),
      const_cast<uint8_t *>(
          reinterpret_cast<const uint8_t *>(QUICHE_H3_APPLICATION_PROTOCOL)),
      sizeof(QUICHE_H3_APPLICATION_PROTOCOL) - 1);
```

The definition of: `QUICHE_H3_APPLICATION_PROTOCOL` comes from `quiche.h`, and seems to be properly set to: `#define QUICHE_H3_APPLICATION_PROTOCOL "\x05h3-23"`, which is what I expect.

When I try to use cURL though I get:

```./src/curl -k --verbose --tlsv1.3 --http3 https://localhost:8080/ && echo ""
*   Trying ::1:8080...
* Sent QUIC client Initial, ALPN: h3-23
* quiche: recv() unexpectedly returned -1 (errno: 111, socket 3)
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1:8080...
* Sent QUIC client Initial, ALPN: h3-23
* h3 [:method: GET]
* h3 [:path: /]
* h3 [:scheme: https]
* h3 [:authority: localhost:8080]
* h3 [user-agent: curl/7.67.0-DEV]
* h3 [accept: */*]
* Using HTTP/3 Stream ID: 0 (easy handle 0x55e202f95280)
> GET / HTTP/3
> Host: localhost:8080
> user-agent: curl/7.67.0-DEV
> accept: */*
> 
* Received HTTP/0.9 when not allowed

* Connection #0 to host localhost left intact
curl: (1) quiche: recv() unexpectedly returned -1 (errno: 111, socket 3)
```

This error goes away if I provide the: `--http0.9` flag to cURL, and the request completes successfully.

When running the example http-server written in rust: `./target/debug/examples/http3-server --listen 0.0.0.0:8080` with cURL I am able to successfully query without providing the http0.9 flag:

```
./src/curl -k --tlsv1.3 --http3 https://localhost:8080/ && echo ""
[...]

PR #253, opened 2019-11-08
h3: document that app should validate headers

LPardue:
Make it clear that applications are responsible for validating HTTP/3 pseudo-headers and headers, to help avoid issues such as https://github.com/cloudflare/quiche/issues/248

The nginx patch included in this repo has an [example](https://github.com/cloudflare/quiche/blob/a27c2b898beae734c771bcc50803737540d288f3/extras/nginx/nginx-1.16.patch#L1728) of the sort of checks that should be done.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -350,9 +350,11 @@ enum quiche_h3_event_type quiche_h3_event_type(quiche_h3_event *ev);
 
 // Iterates over the headers in the event.
 //
-// The `cb` callback will be called for each header in `ev`. If `cb` returns
-// any value other than `0`, processing will be interrupted and the value is
-// returned to the caller.
+// The `cb` callback will be called for each header in `ev`. `cb` should check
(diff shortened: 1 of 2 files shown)


include/quiche.h
```diff
@@ -350,9 +350,11 @@ enum quiche_h3_event_type quiche_h3_event_type(quiche_h3_event *ev);
...
 // Iterates over the headers in the event.
 //
-// The `cb` callback will be called for each header in `ev`. If `cb` returns
-// any value other than `0`, processing will be interrupted and the value is
-// returned to the caller.
+// The `cb` callback will be called for each header in `ev`. `cb` should check
+// the validity of pseudo-headers and headers. If `cb` returns any value other
+// than `0`, processing will be interrupted and the value is returned to the
+// caller.
+
```
ghedo:
```suggestion
```

PR #261, opened 2019-11-16
support draft-24 alongside draft-23

ghedo:
This adds initial support for draft-24, while still retaining draft-23
support. In practice they are basically the same protocol, so the only
change is supporting multiple versions on the server and doing some kind
of version negotiation.

This also allows PING frames in non-1-RTT packets (technically it should
be draft-24-only, but it doesn't really matter).

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -38,7 +38,7 @@ extern "C" {
 //
 
 // The current QUIC wire version.
-#define QUICHE_PROTOCOL_VERSION 0xff000017
+#define QUICHE_PROTOCOL_VERSION 0xff000018
 
 // The maximum length of a connection ID.
 #define QUICHE_MAX_CONN_ID_LEN 20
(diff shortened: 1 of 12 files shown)


ghedo:
Note that, as mentioned in the PR description, draft-23 and draft-24 are basically the same, so there isn't really a risk of version downgrade attacks, since the security should be the same.

In the future if we want to support versions that e.g. fix security problems we need to consider whether we want to support the older version as well.

src/h3/mod.rs
```diff
@@ -252,7 +252,7 @@ use std::collections::VecDeque;
 use crate::octets;
 
 /// The current HTTP/3 ALPN token.
```
LPardue: The comment is a bit weird with this change.

include/quiche.h
```diff
@@ -38,7 +38,7 @@ extern "C" {
 //
 
 // The current QUIC wire version.
-#define QUICHE_PROTOCOL_VERSION 0xff000017
+#define QUICHE_PROTOCOL_VERSION 0xff000018
```
LPardue: I think we want to expose "old" in `quiche.h`

ghedo: The Rust value is also private. I think we shouldn't encourage people to use that version directly so leaving it as private seems a better option.

LPardue: fine by me

src/h3/mod.rs
ghedo: Fixed. I also expanded this comment with more details on how to use the value.

PR #263, opened 2019-11-16
QNS fixes

ghedo:


ghedo:
Also, @junhochoi do we really need to run the test suite on travis? It doesn't seem very useful since we are already running them as part of other builds.

extras/docker/Dockerfile
```diff
@@ -41,6 +41,7 @@ COPY --from=build /build/quiche/target/release/examples/client .
 COPY --from=build /build/quiche/target/release/examples/server .
 COPY --from=build /build/quiche/target/release/examples/http3-client .
 COPY --from=build /build/quiche/target/release/examples/http3-server .
+ENV RUST_LOG=trace
```
junhochoi: It may impact perf test (if added). If no plan to do in the near future it will be good.

junhochoi:
> Also, @junhochoi do we really need to run the test suite on travis? It doesn't seem very useful since we are already running them as part of other builds.

Which test suite do you mean?

ghedo:
@junhochoi the quiche one, i.e. https://github.com/cloudflare/quiche/blob/master/extras/docker/Makefile#L25

junhochoi:
> @junhochoi the quiche one, i.e. https://github.com/cloudflare/quiche/blob/master/extras/docker/Makefile#L25

If so, change here: https://github.com/cloudflare/quiche/blob/master/.travis.yml#L41
`all test` -> `all` will build docker images only.

ghedo:
Yeah, the question was more if you had a specific reason to run the tests that I missed.

extras/docker/Dockerfile
ghedo: I spoke with the QNS people, and they say that implementations shouldn't expect to be CPU-limited by the interop or QNS tests (that is, logging everything should be fine, e.g. quicly does that as well).

We can reconsider this later on if it becomes a problem though.

ghedo:
But anyway, I disabled running the tests now, so the CI builds should be a bit faster.

PR #267, opened 2019-11-22
Additional Fuzz Targets & Fuzzing improvements

alpire:
This pull request suggests adding two fuzz targets and improving an existing one:
- `qpack_decode`: The target now checks that `decode(encode(hdrs)) == hdrs` while not panicking instead of just checking `decode(input)` doesn't panic. 
- `qpack_huffman_rt` (NEW): Fuzzer for the QPACK huffman codec. Checks that `decode(encode(input)) == input` while not raising a panic.
- `h3_frame_rt` (NEW): Fuzzer for h3::frame (de)serialization. Checks that `deserialize(serialize(frame)) == frame` while not raising a panic.

As an aside, all of those fuzzers define a buffer to store the encoded (or serialized) data. Those buffers have to be bigger than the original `data` array given by libfuzzer. This is necessary because libfuzzer sometimes find a smaller representation than what the encoder would create. Reducing the size of those buffers would likely help find places where the encoder could be improved. For instance, it helped me find this potential optimization in the QPACK encoder (not included in this PR):

```diff
diff --git a/src/h3/qpack/encoder.rs b/src/h3/qpack/encoder.rs
index 1388323..8cf8b28 100644
--- a/src/h3/qpack/encoder.rs
+++ b/src/h3/qpack/encoder.rs
@@ -83,9 +83,13 @@ impl Encoder {
                     let name_len =
                         super::huffman::encode_output_length(h.0.as_bytes())?;

-                    encode_int(name_len as u64, LITERAL | 0x08, 3, &mut b)?;
-
-                    super::huffman::encode(h.0.as_bytes(), &mut b)?;
+                    if name_len > h.0.len() {
+                        encode_int(h.0.len() as u64, LITERAL, 3, &mut b)?;
+                        b.put_bytes(h.0.as_bytes())?;
+                    } else {
+                        encode_int(name_len as u64, LITERAL | 0x08, 3, &mut b)?;
+                        super::huffman::encode(h.0.as_bytes(), &mut b)?;
+                    }

                     encode_str(&h.1, 7, &mut b)?;
                 },
@@ -278,9 +282,13 @@ fn encode_int(
 fn encode_str(v: &str, prefix: usize, b: &mut octets::Octets) -> Result<()> {
     let len = super::huffman::encode_output_length(v.as_bytes())?;

-    encode_int(len as u64, 0x80, prefix, b)?;
-
-    super::huffman::encode(v.as_bytes(), b)?;
+    if len > v.len() {
+        encode_int(v.len() as u64, 0x00, prefix, b)?;
+        b.put_bytes(v.as_bytes())?;
+    } else {
+        encode_int(len as u64, 0x80, prefix, b)?;
+        super::huffman::encode(v.as_bytes(), b)?;
+    }

     Ok(())
 }
```

This PR also adds a couple of additional fuzzing improvements:
- Install llvm-symbolizer into the mayhem docker image in order for stack traces to have symbols
- -O3 for the fuzzing harnesses to speed up fuzzing (requires cargo-fuzz 0.5.5+ to have one of our fix: https://github.com/rust-fuzz/cargo-fuzz/pull/193. See https://github.com/rust-fuzz/cargo-fuzz/issues/63 for extra info)

We are working in the background to improve fuzzing support in rust as well. We pushed a fix to cargo-fuzz to handle optimized harnesses, and we have an open PR on libfuzzer-sys (https://github.com/rust-fuzz/libfuzzer-sys/pull/44), which will improve the information printed during a crash.

From your friends at ForAllSecure :)

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
+[profile.dev]
+opt-level = 3
+
 [dependencies]
 quiche = { path = "..", features = ["fuzzing"] }
 lazy_static = "1"
--- a/fuzz/Dockerfile
+++ b/fuzz/Dockerfile
@@ -3,6 +3,9 @@ LABEL maintainer="alessandro@cloudflare.com"
 
 WORKDIR /home/mayhem/
 
+# Install llvm-symbolizer to have source code information in stack traces
+RUN apt-get update && apt-get install llvm -y
+
 COPY ./cert.crt ./
 COPY ./cert.key ./
 
--- a/fuzz/src/qpack_decode.rs
+++ b/fuzz/src/qpack_decode.rs
@@ -3,9 +3,26 @@
 #[macro_use]
 extern crate libfuzzer_sys;
 
+// Fuzzer for qpack codec. Checks that decode(encode(hdrs)) == hdrs. To get the
+// initial hdrs, the fuzzer deserializes the input, and skips inputs where
+// deserialization fails.
+//
+// The fuzzer could have been written to instead check encode(decode(input)) ==
+// input. However, that transformation is not guaranteed to be the identify
+// function, as there are multiple ways the same hdr list could be encoded.
 fuzz_target!(|data: &[u8]| {
-    let mut buf = data.to_vec();
     let mut decoder = quiche::h3::qpack::Decoder::new();
+    let mut encoder = quiche::h3::qpack::Encoder::new();
+    let hdrs = match decoder.decode(&mut data.to_vec(), std::u64::MAX) {
+        Err(_) => return,
+        Ok(hdrs) => hdrs,
+    };
+    let mut encoded_hdrs = vec![0; data.len() * 10 + 1000];
+    let encoded_size = encoder.encode(&hdrs, &mut encoded_hdrs).unwrap();
 
-    decoder.decode(&mut buf, std::u64::MAX).ok();
+    let decoded_hdrs = decoder
+        .decode(&mut encoded_hdrs[..encoded_size], std::u64::MAX)
+        .unwrap();
+
+    assert_eq!(hdrs, decoded_hdrs)
 });


ghedo:
Thanks for the PR. This looks good generally, though I have some reservations about the new fuzzers as I'm not sure thay provide much value:
* `qpack_huffman_rt`: this should already be covered by `qpack_decode`, isn't it?
* `h3_frame_rt`: we are trying to move away from parsing HTTP/3 frames from buffers and we'll instead implement streaming parsers for them (this is already the case for `DATA`, though `HEADERS` and `SETTINGS` will follow as well, the other frames are pretty trivial), so this fuzzer will likely stop working. More generally, fuzzing frame parsing alone is not all that interesting as HTTP/3 frames are pretty simple, but what we would like to do is fuzz how the frames are handled within a connection. This is more complicated of course, as it requires completing a connection handshake as well as exchanging requests and resposes within a single test case, but it's likely to provide the most value (I've already started doing something along these lines for nginx-quic, though there's probably room for improvement particularly regarding HTTP/3 fuzzing itself).

So maybe split those commits into a separate PR so we can discuss them separately without blocking merging of the other fixes?

Also, just curious, did you intend to also make changes to disable LibFuzzer's `reduce_inputs` option, or is that not required anymore?

alpire:
Thanks for the feedback! I just updated the PR to remove the two additional fuzzers. After thinking about it, I agree with both of your comments: 1) huffman is already tested by qpack_decode, and 2) fuzzing frames within a connection would be much more interesting. Let me know if you still would like me to create another PR with those 2 fuzzers, as I'm not planning on creating one right now.

> Also, just curious, did you intend to also make changes to disable LibFuzzer's reduce_inputs option, or is that not required anymore?

I didn't intend to set that option right now. We are thinking about better ways to handle this on our side that would not require setting that option. I understand the corpus are getting quite large right now. Is it starting to be a blocking issue on your side?

ghedo:
> as I'm not planning on creating one right now

That's ok, we can reconsider later on if we want this once we've looked into HTTP/3 fuzzing more.

> Is it starting to be a blocking issue on your side?

It's not ideal as IIRC the regression tests phase for some of the targets takes a few hours, but it's not really critical or urgent either, so I think we are ok for now.

ghedo:
Merged now, thanks for the help @alpire!

PR #279, opened 2019-11-28
collect memory from completed streams

ghedo:
Drop state of streams that are completed to save memory. We still need
to track completed streams though, to prevent a peer from reusing closed
ones.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -715,6 +715,10 @@ impl Connection {
         )?;
         conn.stream_send(stream_id, &header_block, fin)?;
 
+        if fin && conn.stream_finished(stream_id) {
+            self.streams.remove(&stream_id);
+        }
+
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


src/h3/mod.rs
```diff
@@ -715,6 +715,10 @@ impl Connection {
         )?;
         conn.stream_send(stream_id, &header_block, fin)?;
 
+        if fin && conn.stream_finished(stream_id) {
```
LPardue: I think this is fine for normal operation, but makes me wonder how we handle stream collection when we do an early response. In that case we would trigger STOP_SENDING with H3_EARLY_RESPONSE, so its probably handled in code located elsewhere.

ghedo: That's what the TODO later is supposed to handle, among other things :)

LPardue:
:shipit:

PR #283, opened 2019-12-02
recovery: slightly optimize detection of duplicate acks

ghedo:
This is a slight optimization to avoid BTreeMap lookups when a packet is
clearly not newly acked.

### DIFF
--- a/src/recovery.rs
+++ b/src/recovery.rs
@@ -454,6 +454,14 @@ impl Recovery {
     }
 
     fn on_packet_acked(&mut self, pkt_num: u64, epoch: packet::Epoch) -> bool {
+        // If the acked packet number is lower than the lowest unacked packet
+        // number it means that the packet is not newly acked, so return early.
+        if let Some(lowest) = self.sent[epoch].values().nth(0) {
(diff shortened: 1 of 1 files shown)


src/recovery.rs
```diff
@@ -454,6 +454,14 @@ impl Recovery {
     }
 
     fn on_packet_acked(&mut self, pkt_num: u64, epoch: packet::Epoch) -> bool {
+        // If the acked packet number is lower than the lowest unacked packet
+        // number it means that the packet is not newly acked, so return early.
+        if let Some(oldest) = self.sent[epoch].values().nth(0) {
+            if oldest.pkt_num > pkt_num {
+                return false;
+            }
```
LPardue: From the comment, I'm not sure if the intent is to exit early if `oldest.pkt_num == pkt_num`.

Perhaps renaming `oldest` to `lowest` and inverting the test logic would let the code line up with comment a bit more e.g.

```
if pkt_num < lowest.pkt_num {
  return false;
}
```

ghedo: Ok, done.

PR #288, opened 2019-12-03
extend httpbin tests

schien:
introduce two environment variables for customizing "headers" test case
`EXTRA_HEADERS` for adding additional request headers in JSON format.
`EXPECT_REQ_HEADERS` for request header validation in JSON format.
Headers with null value represents non-existence in the forwarded request headers.

### DIFF
--- a/tools/http3_test/tests/httpbin_tests.rs
+++ b/tools/http3_test/tests/httpbin_tests.rs
@@ -94,6 +94,29 @@ mod httpbin_tests {
         }
     }
 
+    fn extra_headers() -> Option<serde_json::Map<String, serde_json::Value>> {
+        if let Some(val) = std::env::var_os("EXTRA_HEADERS") {
+            let json_string = val.into_string().unwrap();
+            let parsed: serde_json::Value =
+                serde_json::from_str(&json_string).unwrap();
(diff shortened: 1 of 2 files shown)


tools/http3_test/tests/httpbin_tests.rs
```diff
@@ -68,6 +68,20 @@ mod httpbin_tests {
         }
     }
 
+    fn extra_headers() -> Option<String> {
+        match std::env::var_os("EXTRA_HEADERS") {
+            Some(val) => Some(val.into_string().unwrap()),
+            None => None,
+        }
+    }
```
LPardue: I guess most tests would want the the parsed key-value map from the environment variable, so perhaps this function could do all of that (i.e. pull in the `serde_json` calls in headers() test below) and then just return a `Some(serde_json::Map<String, serde_json::Value>)`

schien: sure, I'll do it.

ghedo:
Build is failing. Need to run `cargo +nightly fmt` inside the crate.

ghedo:
Merged now after tweaking the commit message a bit, thanks @schien!

PR #285, opened 2019-12-03
Fix 32bit x86 build

junhochoi:
Pass 32bit toolchain to cmake for cross-building x86 linux from
x86_64 linux. Also CI target is added.

Ref: https://github.com/google/boringssl/blob/master/BUILDING.md

Fixes #277

PR #289, opened 2019-12-03
Fix CI build

junhochoi:
#285 triggers to perform a `deploy` when building `x86 cross-build` target.

Add a condition not to trigger it. Also removed `true` which is unnecessary.


ghedo:
Ah, I already pushed https://github.com/cloudflare/quiche/commit/08ca097f131dab1b6df3e37672c252187f3f977b to fix this.

junhochoi:
Ok. I understand this is a small fix but better to do make a PR for visibility...

PR #290, opened 2019-12-05
WIP: Find and retransmit lost packets in all pn_space

junhochoi:
Currently there is some case lost packet never retransmitted.

e.g.

Client send Initial#0

Server receive Initial#0
Server send Initial#0+ACK#0

Server send Handshake#0
Server send Handshake#1 -> Lost

Client receive Handshake#0
Client send Initial#1+ACK#0
Client send Handshake#0+ACK#0

Now Server's Handshake#1 packet is lost, server need to retransmit Handshake#1
as a TLP (tail loss probe), because Client doesn't know about Handshake#1 is
coming (no way to send ACK range).
However, when this happens, we never retransmit Handshake#1, making client timeout.

There is 2 issues:
1) Recovery::on_loss_detection_timeout() doesn't handle the case
   when earliest_loss_time() is None which can happen if we need to send TLP.
   Also, in the scenario above, we need to retransmit Handshake
   packet, but because earliest_loss_time() doesn't return
   Handshake level, so we don't know which packet to retransmit.
   To solve this, if loss_time is not available, scan all
   packet space to find if there are lost packets.
   This will add Handshake#1 to lost packet list.

2) Connection::write_epoch() is returning which packet space
   to send a packet. However, in this scenario, write_epoch()
   may not return a correct packet space for lost packet to be retransmitted.
   To solve this, lost packet is always scanned in all packet
   space in the beginning, so any lost packet in any packet space
   can be retransmitted regardless of current write level.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2565,6 +2565,14 @@ impl Connection {
 
     /// Selects the packet number space for outgoing packets.
     fn write_epoch(&self) -> Result<packet::Epoch> {
+        // If there is lost packet in any epoch, handle it first.
+        for epoch in packet::EPOCH_INITIAL..packet::EPOCH_COUNT {
+            // There are lost frames in this packet number space.
+            if !self.recovery.lost[epoch].is_empty() {
+                return Ok(epoch);
+            }
+        }
+
         // On error or PTO send packets in the latest epoch available, but only
         // send 1-RTT ones when the handshake is completed.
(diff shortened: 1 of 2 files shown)


LPardue:
overtaken by #291 ?

junhochoi:
> overtaken by #291 ?

Seems related. Let me test it today.

junhochoi:
Closing this. #291 is much more complete.

PR #291, opened 2019-12-05
recovery: PTO per packet number space

ghedo:
This makes PTO consistent with loss detection, which is already per
packet number space.

Besides aligning the implementation to the -recovery draft, this also
fixes some recovery bugs that were introduced in 5c99278, which caused
some packets to not be retransmitted when lost during the handshake.

---

I tested this manually using the http3-client and http3-server examples, and qvalve with the following scripts:
```
> h1 drop
```

```
< i1 drop
```

```
< h2 drop
```

Though it's pretty obvious the recovery code needs tests. I'll make a separate PR for that once I figure out how to do it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1756,7 +1756,7 @@ impl Connection {
         }
 
         // Create PING for PTO probe.
-        if self.recovery.probes > 0 && left >= 1 && !is_closing {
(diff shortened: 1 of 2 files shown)


ghedo:
Silenced a clippy warning that is not helpful...

PR #295, opened 2019-12-12
Travis: update build config

junhochoi:
- Group similar platform builds into one.
- Total number of build is reduced from 10 to 6
- When possible build everything as debug to utilize
  build cache as much as possible

Result: total CI build time is reduced from 28 min (master) to 21 min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -175,8 +175,7 @@ deploy:
     skip-cleanup: true
     script: >-
       echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin &&
-      make -C extras/docker &&
-      make -C extras/docker publish
+      make -C extras/docker all publish
     on:
       branch: master
       condition: $TRAVIS_RUST_VERSION = stable && $TRAVIS_OS_NAME = linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,12 +101,13 @@ matrix:
      rust: stable
      os: osx
      osx_image: xcode11.2
+     install:
+      - rustup target add aarch64-apple-ios x86_64-apple-ios
      script:
       # MacOS X
       - RUSTFLAGS="-D warnings" cargo build --verbose
       - RUSTFLAGS="-D warnings" cargo test --verbose
       # iOS
-      - rustup target add aarch64-apple-ios x86_64-apple-ios
       - cargo install --force cargo-lipo
       - RUSTFLAGS="-D warnings" cargo lipo --verbose
    - name: "rust/stable Windows x86_64/x86"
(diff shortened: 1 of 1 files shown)


junhochoi:
Ideally, want to 5 build only, but looks like it's hard. Can merge Docker build into rust/linux x86_64 but it only increase a total build time because previous build cache can't be reused for Docker build. But 6th build usually runs right after which build completed early.

ghedo:
I just realized that the docker image build always pulls the code from GitHub rather than using the local copy, so there isn't much point in running it for every PR/branch update, and we can just do the whole build as part of deploy (which is only run on master builds).

ghedo:
Also, I wonder how much of the build time decrease is due to not building in release mode anymore....

Anyway, I moved the docker image build to deploy, and also applied the same fix as https://github.com/cloudflare/quiche/commit/36589d6e3e88180d5cc4de5cc6516841e9aefc87 for the Windows build, let's see how that goes....

junhochoi:
> Also, I wonder how much of the build time decrease is due to not building in release mode anymore....

Build time release vs debug:

```
% cargo clean; time cargo build; cargo clean; time cargo build --release
    Finished dev [unoptimized + debuginfo] target(s) in 39.83s
    Finished release [optimized + debuginfo] target(s) in 1m 16s
```

Also inconsistent use of release/debug build will make build cache less efficient.
 
> Anyway, I moved the docker image build to deploy, and also applied the same fix as [36589d6](https://github.com/cloudflare/quiche/commit/36589d6e3e88180d5cc4de5cc6516841e9aefc87) for the Windows build, let's see how that goes....

Thanks. Moving Docker build to release is a good idea.

ghedo:
Looks like we can't use Go oldstable, as BoringSSL wants stable... :man_facepalming:

ghedo:
Not sure what changed, but the Go installation doesn't seem to work anymore, will try to debug.

.travis.yml
```diff
@@ -33,7 +31,6 @@ matrix:
       - rustup component add clippy
       - rustup target add $TARGET_32
      script:
-      - RUSTFLAGS="-D warnings" cargo build --verbose
```
junhochoi: Is it related to cache?

ghedo: Ooops, no, this was an accident :/

Though now I wonder how useful this is, since `cargo package` basically already covers this. Maybe we should just remove this from the nightly build as well...

ghedo: Made https://github.com/cloudflare/quiche/pull/323 to do that. We can go the other way and restore this if you disagree though.

junhochoi: Yes I think we can do this way too. Thanks.

Issue #298, opened 2019-12-12
max_packet_size is not used in recovery


junhochoi:
max_packet_size need to be is configurable, not using a fixed number.

PR #299, opened 2019-12-13
recovery: configurable congestion control

junhochoi:
Configurable Congestion Control

The purpose of this mega patch is to make congestion control
configurable. Also it's extensive so that we can add more congestion
module later and make it easier to switch between different CC
algorithms using config API.

New Modules
-----------

`cc/` directory is created.

- `cc/mod.rs` contains basic CC constants such as
  INITIAL_WINDOW. Also it defines CongestonControlTrait which has
  method of CC related functions and hooks.

- new_congestion_control(name) is a factory method to return
  a new CC module based on its name.

- `cc/reno.rs` implements `Reno` CC algotirhm, following
  QUIC recovery draft 24 based.

Currently besides of basic method, you need to implement
the following CC-specific hooks called from the stack:

- on_packet_sent_cc()
- on_packet_acked_cc()
- congestion_event()

Note that there is still some CC hooks missing or unimplmented.

Notably
- `IsAppLimited()``: moved to cc/mod.rs but still dummy
- `InPersistentCongestion()``: still in recovery.rs
- `ProcessECN()``: not implemented
- `OnPacketLost()``: still in recovery.rs. It has no CC specific
  other than calling `CongestionEvent()`` already moved to CC module.

lib.rs
------

- CC is now initialized using `cc::new_congestion_control()`.

recovery.rs
-----------

- CC constants moved to `cc/mod.rs`
- CC fields such as `bytes_in_flight`, `cwnd`, `recovery_start_time`, `ssthresh`
  are moved into cc modules. Also added `cc` field which is CC algorithm object.
  Due to the change above, code accessing fields above now use the following
  methods for reading/writing those values in CC module:
  `self.bytes_in_flight` -> `self.cc.bytes_in_flight()`
  reading `self.cwnd` -> use `self.cc.cwnd()`
  updating `self.cwnd` -> use `self.cc.set_cwnd()`
  reading `self.ssthresh` -> use `self.cc.ssthresh()`
- `OnPacketSentCC`, `OnPacketAckedCC`, `CongestionEvent` moved to
  CC module, so only calling cc methods here.
- `in_recovery()` moved to CC module.
- recovery.cwnd() is renamed to recovery.cwnd_available().

Config API
----------

`Config` struct now has `cc_algorithm`.

There is 2 APIs added for configuration those new parameters:

* config.set_cc_algorithm_name(<name>): set which CC algorithm to use by
  name.
* config.set_cc_algorithm(<enum>): set which CC algorithm to use. values
  are from quiche::CongestionControlAlgorithm.

ffi.rs
------

Corresponing C Config API is also added.

`quiche_config_set_cc_algorithm_name()`
`quiche_config_set_cc_algorithm()`

Those config should be called before new connection is made or accepted.

Examples
--------

Rust example of http3-server., server, http3-client, client has
a new options name `--cc-algorithm` to specify which CC algorithms can be used.

C example of http3-server and server calls quiche_config_set_cc_*() APIs.

Others
------
- Documentation in the source
- Test cases added

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -88,6 +88,9 @@ enum quiche_error {
 
     // The received data exceeds the stream's final size.
     QUICHE_ERR_FINAL_SIZE = -13,
+
+    // Error in congestion control.
+    QUICHE_ERR_CONGESTION_CONTROL = -14,
 };
 
 // Returns a human readable string with the quiche version number.
@@ -160,6 +163,13 @@ void quiche_config_set_max_ack_delay(quiche_config *config, uint64_t v);
 
 // Sets the `disable_active_migration` transport parameter.
 void quiche_config_set_disable_active_migration(quiche_config *config, bool v);
+
+enum quiche_cc_algorithm {
+    QUICHE_CC_RENO = 0,
+};
+
+// Sets the congestion control algorithm used.
+void quiche_config_set_cc_algorithm(quiche_config *config, enum quiche_cc_algorithm algo);
 
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
--- a/src/cc/mod.rs
+++ b/src/cc/mod.rs
@@ -1,0 +1,154 @@
+// Copyright (C) 2019, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::str::FromStr;
+
+use std::time::Duration;
+use std::time::Instant;
+
+use crate::cc;
+use crate::recovery::Sent;
+
+pub const INITIAL_WINDOW_PACKETS: usize = 10;
+
+pub const INITIAL_WINDOW: usize = INITIAL_WINDOW_PACKETS * MAX_DATAGRAM_SIZE;
+
+pub const MINIMUM_WINDOW: usize = 2 * MAX_DATAGRAM_SIZE;
+
+pub const MAX_DATAGRAM_SIZE: usize = 1452;
+
+pub const LOSS_REDUCTION_FACTOR: f64 = 0.5;
+
+/// Available congestion control algorithms.
+///
+/// This enum provides currently available list of congestion control
+/// algorithms.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum Algorithm {
+    /// Reno congestion control algorithm (default). `reno` in a string form.
+    Reno = 0,
+}
+
+impl FromStr for Algorithm {
+    type Err = crate::Error;
+
+    /// Converts a string to `CongestionControlAlgorithm`.
+    ///
+    /// If `name` is not valid, `Error::CongestionControl` is returned.
+    fn from_str(name: &str) -> Result<Self, Self::Err> {
+        match name {
+            "reno" => Ok(Algorithm::Reno),
+            _ => Err(crate::Error::CongestionControl),
+        }
+    }
+}
+
+/// Congestion control algorithm.
+pub trait CongestionControl
+where
+    Self: std::fmt::Debug,
+{
+    fn new() -> Self
+    where
+        Self: Sized;
+
+    fn cwnd(&self) -> usize;
+
+    fn bytes_in_flight(&self) -> usize;
+
+    fn decrease_bytes_in_flight(&mut self, bytes_in_flight: usize);
+
+    fn congestion_recovery_start_time(&self) -> Option<Instant>;
+
+    fn is_app_limited(&self) -> bool {
+        false
+    }
+
+    /// Resets the congestion window to the minimum size.
+    fn collapse_cwnd(&mut self);
+
+    /// OnPacketSentCC(bytes_sent)
+    fn on_packet_sent_cc(&mut self, bytes_sent: usize, trace_id: &str);
+
+    /// InCongestionRecovery(sent_time)
+    fn in_congestion_recovery(&self, sent_time: Instant) -> bool {
+        match self.congestion_recovery_start_time() {
+            Some(congestion_recovery_start_time) =>
+                sent_time <= congestion_recovery_start_time,
+
+            None => false,
+        }
+    }
+
+    /// OnPacketAckedCC(packet)
+    fn on_packet_acked_cc(
+        &mut self, packet: &Sent, srtt: Duration, min_rtt: Duration,
+        trace_id: &str,
+    );
+
+    /// CongestionEvent(time_sent)
+    fn congestion_event(
+        &mut self, time_sent: Instant, now: Instant, trace_id: &str,
+    );
+}
+
+/// Instances a congestion control implementation based on the CC algorithm ID.
+pub fn new_congestion_control(algo: Algorithm) -> Box<dyn CongestionControl> {
+    trace!("Initializing congestion control: {:?}", algo);
+    match algo {
+        Algorithm::Reno => Box::new(cc::reno::Reno::new()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn new_cc() {
+        let cc = new_congestion_control(Algorithm::Reno);
+
+        assert!(cc.cwnd() > 0);
+        assert_eq!(cc.bytes_in_flight(), 0);
+    }
+
+    #[test]
+    fn lookup_cc_algo_ok() {
+        let algo = Algorithm::from_str("reno").unwrap();
+
+        assert_eq!(algo, Algorithm::Reno);
+    }
+
+    #[test]
+    fn lookup_cc_algo_bad() {
+        assert_eq!(
+            Algorithm::from_str("???"),
+            Err(crate::Error::CongestionControl)
+        );
+    }
+}
+
+mod reno;
--- a/src/cc/reno.rs
+++ b/src/cc/reno.rs
@@ -1,0 +1,264 @@
+// Copyright (C) 2019, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::time::Duration;
+use std::time::Instant;
+
+use crate::cc;
+use crate::recovery::Sent;
+
+/// Reno congestion control implementation.
+pub struct Reno {
+    congestion_window: usize,
+
+    bytes_in_flight: usize,
+
+    congestion_recovery_start_time: Option<Instant>,
+
+    ssthresh: usize,
+    /* TODO: ECN is not implemented.
+     * ecn_ce_counters: [usize; packet::EPOCH_COUNT] */
+}
+
+impl cc::CongestionControl for Reno {
+    fn new() -> Self
+    where
+        Self: Sized,
+    {
+        Reno {
+            congestion_window: cc::INITIAL_WINDOW,
+
+            bytes_in_flight: 0,
+
+            congestion_recovery_start_time: None,
+
+            ssthresh: std::usize::MAX,
+            /* TODO: ECN is not implemented.
+             * ecn_ce_counters: [0; packet::EPOCH_COUNT], */
+        }
+    }
+
+    fn cwnd(&self) -> usize {
+        self.congestion_window
+    }
+
+    fn collapse_cwnd(&mut self) {
+        self.congestion_window = cc::INITIAL_WINDOW;
+    }
+
+    fn bytes_in_flight(&self) -> usize {
+        self.bytes_in_flight
+    }
+
+    fn decrease_bytes_in_flight(&mut self, bytes_in_flight: usize) {
+        self.bytes_in_flight =
+            self.bytes_in_flight.saturating_sub(bytes_in_flight);
+    }
+
+    fn congestion_recovery_start_time(&self) -> Option<Instant> {
+        self.congestion_recovery_start_time
+    }
+
+    fn on_packet_sent_cc(&mut self, bytes_sent: usize, _trace_id: &str) {
+        self.bytes_in_flight += bytes_sent;
+    }
+
+    fn on_packet_acked_cc(
+        &mut self, packet: &Sent, _srtt: Duration, _min_rtt: Duration,
+        _trace_id: &str,
+    ) {
+        self.bytes_in_flight -= packet.size;
+
+        if self.in_congestion_recovery(packet.time) {
+            return;
+        }
+
+        if self.is_app_limited() {
+            return;
+        }
+
+        if self.congestion_window < self.ssthresh {
+            // Slow start.
+            self.congestion_window += packet.size;
+        } else {
+            // Congestion avoidance.
+            self.congestion_window +=
+                (cc::MAX_DATAGRAM_SIZE * packet.size) / self.congestion_window;
+        }
+    }
+
+    fn congestion_event(
+        &mut self, time_sent: Instant, now: Instant, _trace_id: &str,
+    ) {
+        // Start a new congestion event if packet was sent after the
+        // start of the previous congestion recovery period.
+        if !self.in_congestion_recovery(time_sent) {
+            self.congestion_recovery_start_time = Some(now);
+
+            self.congestion_window = (self.congestion_window as f64 *
+                cc::LOSS_REDUCTION_FACTOR)
+                as usize;
+            self.congestion_window =
+                std::cmp::max(self.congestion_window, cc::MINIMUM_WINDOW);
+            self.ssthresh = self.congestion_window;
+        }
+    }
+}
+
+impl std::fmt::Debug for Reno {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "cwnd={} ssthresh={} bytes_in_flight={}",
+            self.congestion_window, self.ssthresh, self.bytes_in_flight,
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const TRACE_ID: &str = "test_id";
+
+    fn init() {
+        let _ = env_logger::builder().is_test(true).try_init();
+    }
+
+    #[test]
+    fn reno_init() {
+        let cc = cc::new_congestion_control(cc::Algorithm::Reno);
+
+        assert!(cc.cwnd() > 0);
+        assert_eq!(cc.bytes_in_flight(), 0);
+    }
+
+    #[test]
+    fn reno_send() {
+        init();
+
+        let mut cc = cc::new_congestion_control(cc::Algorithm::Reno);
+
+        cc.on_packet_sent_cc(1000, TRACE_ID);
+
+        assert_eq!(cc.bytes_in_flight(), 1000);
+    }
+
+    #[test]
+    fn reno_slow_start() {
+        init();
+
+        let mut cc = cc::new_congestion_control(cc::Algorithm::Reno);
+
+        let p = Sent {
+            pkt_num: 0,
+            frames: vec![],
+            time: std::time::Instant::now(),
+            size: 5000,
+            ack_eliciting: true,
+            in_flight: true,
+        };
+
+        // Send 5k x 4 = 20k, higher than default cwnd(~15k)
+        // to become no longer app limited.
+        cc.on_packet_sent_cc(p.size, TRACE_ID);
+        cc.on_packet_sent_cc(p.size, TRACE_ID);
+        cc.on_packet_sent_cc(p.size, TRACE_ID);
+        cc.on_packet_sent_cc(p.size, TRACE_ID);
+
+        let cwnd_prev = cc.cwnd();
+
+        cc.on_packet_acked_cc(
+            &p,
+            Duration::new(0, 1),
+            Duration::new(0, 1),
+            TRACE_ID,
+        );
+
+        // Check if cwnd increased by packet size (slow start).
+        assert_eq!(cc.cwnd(), cwnd_prev + p.size);
+    }
+
+    #[test]
+    fn reno_congestion_event() {
+        init();
+
+        let mut cc = cc::new_congestion_control(cc::Algorithm::Reno);
+        let prev_cwnd = cc.cwnd();
+
+        cc.congestion_event(
+            std::time::Instant::now(),
+            std::time::Instant::now(),
+            TRACE_ID,
+        );
+
+        // In Reno, after congestion event, cwnd will be cut in half.
+        assert_eq!(prev_cwnd / 2, cc.cwnd());
+    }
+
+    #[test]
+    fn reno_congestion_avoidance() {
+        init();
+
+        let mut cc = cc::new_congestion_control(cc::Algorithm::Reno);
+        let prev_cwnd = cc.cwnd();
+
+        // Send 20K bytes.
+        cc.on_packet_sent_cc(20000, TRACE_ID);
+
+        cc.congestion_event(
+            std::time::Instant::now(),
+            std::time::Instant::now(),
+            TRACE_ID,
+        );
+
+        // In Reno, after congestion event, cwnd will be cut in half.
+        assert_eq!(prev_cwnd / 2, cc.cwnd());
+
+        let p = Sent {
+            pkt_num: 0,
+            frames: vec![],
+            time: std::time::Instant::now(),
+            size: 5000,
+            ack_eliciting: true,
+            in_flight: true,
+        };
+
+        let prev_cwnd = cc.cwnd();
+
+        // Ack 5000 bytes.
+        cc.on_packet_acked_cc(
+            &p,
+            Duration::new(0, 1),
+            Duration::new(0, 1),
+            TRACE_ID,
+        );
+
+        // Check if cwnd increase is smaller than a packet size (congestion
+        // avoidance).
+        assert!(cc.cwnd() < prev_cwnd + 1111);
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -229,6 +229,35 @@
 //! [`readable()`]: struct.Connection.html#method.readable
 //! [`stream_recv()`]: struct.Connection.html#method.stream_recv
 //! [HTTP/3 module]: h3/index.html
+//!
+//! ## Congestion Control
+//!
+//! The quiche library provides a high-level API for configuring which
+//! congestion control algorithm to use throughout the QUIC connection.
+//!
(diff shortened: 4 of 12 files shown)


src/lib.rs
```diff
@@ -622,6 +630,20 @@ impl Config {
     pub fn set_disable_active_migration(&mut self, v: bool) {
         self.local_transport_params.disable_active_migration = v;
     }
+
+    /// Sets the congestion control algorithm used (string).
+    ///
+    /// The default value is 'reno'.
+    pub fn set_cc_algorithm(&mut self, name: &str) {
```
ghedo: We should just make an enum with all the possible CC algorithms, rather than bypass the type system by using strings. This will make it much easier to validate that the user passed a valid value (as the rust compiler will do that), as well as make it easier to document the allowed values.

src/lib.rs
```diff
@@ -622,6 +630,20 @@ impl Config {
...
+        self.cc_algorithm = name.to_string();
+    }
+
+    /// Sets the initial congestion control window size (bytes).
+    ///
+    /// The default value is `cc:INITIAL_WINDOW` (10 x `max_datagram_size`).
+    pub fn set_cc_initcwnd(&mut self, v: usize) {
```
ghedo: Why is this needed? I don't think we should let the application pick this, it should just give us what the `max_datagram_size` is. Or if we really need to this should be in number of packets, rather than bytes, so it's lees likely for the application to mess this up.

Maybe we should remove this and move it to a separate PR?

src/recovery.rs
```diff
@@ -34,10 +34,32 @@ use std::collections::BTreeMap;
...
 
+use crate::cc;
 use crate::frame;
 use crate::packet;
 use crate::ranges;
 
+// Provides shortcuts for the following cases:
+//   cc.as_ref().unwrap() -> cc.unwrap_ref()
+//   cc.as_mut().unwrap() -> cc.unwrap_mut()
+trait OptionExt {
```
ghedo: This doesn't seem to be much of a shortcut, and IMO it makes reading the code harder by using custom methods on standard types that normally don't have them. It's also not used consistently AFAICT.

src/recovery.rs
```diff
@@ -103,15 +117,10 @@ pub struct Recovery {
...
-    cwnd: usize,
-
-    recovery_start_time: Option<Instant>,
-
-    ssthresh: usize,
-
     pub loss_probes: [usize; packet::EPOCH_COUNT],
+
+    // Congestion Control
+    pub cc: Option<Box<dyn cc::CongestionControlTrait>>,
```
ghedo: I don't think this needs to be an `Option`. Instead we can just turn the `Recovery::default()` constructor into `Recovery::new(cc: CongestionControlID)` (or whatever the enum used for selecting a CC in `Config` is), and create the object when the `Recovery` object is also created. This also means that this field wouldn't need to be public, and the whole `OptionExt` thing can go away too.

src/recovery.rs
```diff
@@ -307,13 +314,33 @@ impl Recovery {
...
         if self.loss_probes.iter().any(|&x| x > 0) {
             return std::usize::MAX;
         }
 
-        self.cwnd.saturating_sub(self.bytes_in_flight)
+        self.cwnd().saturating_sub(self.bytes_in_flight())
+    }
+
+    // Shortcut to CC module's cwnd()
+    pub fn cwnd(&self) -> usize {
```
ghedo: All of these "shortcuts" just make it harder to read the code IMO, as it's never clear where a given value comes from without having to follow the shortcuts every time.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    // Access to internal variables
+    fn cwnd(&self) -> usize;
+
+    fn set_cwnd(&mut self, cwnd: usize);
+
+    fn ssthresh(&self) -> usize;
+
+    fn bytes_in_flight(&self) -> usize;
+
+    fn set_bytes_in_flight(&mut self, bytes_in_flight: usize);
```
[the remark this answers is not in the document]
ghedo: Actually, the only thing that actually reads and writes to this seems to be the recovery code, so maybe just leave the `bytes_in_flight` field in `Recovery` itself? So we can also get rid of these two methods.

src/recovery.rs
```diff
@@ -514,13 +521,15 @@ impl Recovery {
...
 
+    // TODO: move to Congestion Control and implement draft 24
     fn in_persistent_congestion(&mut self, _largest_lost_pkt: &Sent) -> bool {
         let _congestion_period = self.pto() * PERSISTENT_CONGESTION_THRESHOLD;
 
         // TODO: properly detect persistent congestion
         false
     }
 
+    // TODO: move to Congestion Control
```
ghedo: Should we though? Each CC implementation is going to have its own `CongestionEvent()` method, but I think the rest would stay the same?

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    fn congestion_recovery_start_time(&self) -> Option<Instant>;
+
+    // Reset to minimum window.
+    fn set_min_cwnd(&mut self);
```
ghedo: This should be called `collapse_cwnd()` IMO.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // Congestion Control hooks in QUIC recovery draft.
+    //
+
+    // OnPacketSentCC(bytes_sent)
```
ghedo: This comments are not very useful, they just repeat the method's signature...

src/recovery.rs
ghedo: By removing the `Option` as per a previous comment, this wouldn't be needed as we could just do `self.cc.method()` directly btw.

src/cc/reno.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn set_min_cwnd(&mut self) {
+        self.congestion_window = cc::MINIMUM_WINDOW;
+    }
+
+    // OnPacketSentCC
+    fn on_packet_sent_cc(&mut self, bytes_sent: usize) {
+        self.bytes_in_flight += bytes_sent;
+
+        trace!(
+            "{} {} CC_DEBUG OnPacketSentCC() bytes_in_flight={}",
```
ghedo: All these logs shouldn't be specific to each CC implementation I think (again, implementing `Debug` would make this easier). It will also become harder to log this information when we add qlog support I think.

But also, do we really need them? Trace logs are already very expensive to the point where they become useless when doing performance measurements, so I'm not sure these would be useful.

src/cc/dummy.rs
```diff
@@ -0,0 +1,177 @@
+// Copyright (C) 2019, Cloudflare, Inc.
```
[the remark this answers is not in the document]
junhochoi: Already mentioned in the PR description. "This is exactly same as Reno for now and works as a template of congestion control module and also demonstrate that there can be multiple CC modules.". I need this to test if I can switch over between different CCs. In the future it can be removed/replaced.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub const MAX_DATAGRAM_SIZE: usize = 1452;
+
+pub const LOSS_REDUCTION_FACTOR: f64 = 0.5;
+
+// Default congestion control algorithm
+pub const DEFAULT_CC_ALGORITHM: &str = cc::reno::CC_NAME_RENO;
+
+// Congestion Control Trait
+pub trait CongestionControlTrait {
```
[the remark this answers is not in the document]
junhochoi: SGTM

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn new() -> Self
+    where
+        Self: Sized;
+
+    fn get_name(&self) -> &str;
+
+    fn to_string(&self) -> String {
```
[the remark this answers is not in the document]
junhochoi: Looks like some leftover from last try... I will remove it those.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+// Default congestion control algorithm
+pub const DEFAULT_CC_ALGORITHM: &str = cc::reno::CC_NAME_RENO;
+
+// Congestion Control Trait
+pub trait CongestionControlTrait {
+    fn new() -> Self
+    where
+        Self: Sized;
+
+    fn get_name(&self) -> &str;
```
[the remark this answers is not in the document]
junhochoi: See below.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn to_string(&self) -> String {
+        self.get_name().to_string()
+    }
+
+    fn set_trace_id(&mut self, id: String);
+
+    // Access to internal variables
+    fn cwnd(&self) -> usize;
+
+    fn set_cwnd(&mut self, cwnd: usize);
```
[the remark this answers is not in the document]
junhochoi: This is for setting initcwnd via config.set_cc_initcwnd() API. See initcwnd API comment below.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: 1. set_bytes_in_flight() is called in recovery.on_packets_lost() and recovery.drop_unacked_data(), so I exposed the API.
2. bytes_in_flight is originally mentioned in congestion control section in the draft, so I basically moved. Another issue is that if we have bytes_in_flight in Recovery it's hard to access from CC module, but Recovery can have access to CC module's variable in this way.

src/lib.rs
junhochoi: Currently this API is similar to setsockopt(...TCP_CONGESTION...) so string is used. I was also thinking about enum... let me update.

src/lib.rs
junhochoi: Currently there is 2 issues.
1. People/company is playing with initcwnd values, including us, for performance optimization. Also Linux/FreeBSD allow to change initcwnd when socket is created.
2. bytes vs packets - even if we set the number of packet size, it will be converted to actual bytes size -- I am more concerned when we don't know max_datagram_size so actual window size is not known, so choose to set the size as bytes.
Also currently there is a bug #298 also need be aligned with this.

However, separating to another PR is a good idea. I will move to a new PR.

src/recovery.rs
junhochoi: I was wondering too... let me try.

src/recovery.rs
junhochoi: OnPacketLost() is mentioned in Congestion Control pseudocode in the draft, so try to follow it first. In reality as you say it looks good enough to keep OnPacketsLost() in Recovery and CC need to implement CongestionEvent() and InPersistentCongestion(). However set_bytes_in_flight() is called here too, so probably a good idea to move into CC, but there are many other fields accessed from Recovery (such as lost_count, sent[], etc).

So, for me it's hard to decide to move it or not... to keep the changes minimum I only left the TODO comment. This is a typical real implementation vs pseudocode issue. I don't want to push but basically a reminder to think it again if there is a good way to move into CC.

src/cc/mod.rs
junhochoi: Will change.

junhochoi: Also looks like I can move it to the trait default method.

src/cc/mod.rs
junhochoi: This is basically to tell readers which function name in the draft pseudo code. e.g. OnPacketSentCC() in the draft -> on_packet_sent_cc() in our code.

src/cc/reno.rs
junhochoi: When we add a proper qlog support we can replace this with qlog functions but at this point this is particularly useful for debugging to check when the packet is sent bytes_in_flight is increased properly, so want to keep it.

Basically I need some log when each packets are sent/received. Yes trace!() usage is getting expensive now... If there is something can be used other than trace!() I'd love to try, or maybe I made another macro only for cc logging.

src/cc/dummy.rs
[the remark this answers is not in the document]
junhochoi: Well. I removed `dummy`.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    where
+        Self: Sized;
+
+    fn get_name(&self) -> &str;
+
+    fn to_string(&self) -> String {
+        self.get_name().to_string()
+    }
+
+    fn set_trace_id(&mut self, id: String);
```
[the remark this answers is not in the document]
junhochoi: Removed this part too, however I think there is a need to log trace_id when we debug concurrent connections. Currently `Connection` is passing `trace_id` to `Recovery` module in many functions but looks not consistent.

e.g.

detect_lost_packets() have trace_id parameter, but not passing to on_packet_lost() (which will call congestion_event())
on_packet_acked() doesn't have trace_id (which will call on_packet_acked_cc()).

I think it's good to pass trace_id to Recovery object when generated and make it available for all methods.

src/cc/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // Access to internal variables
+    fn cwnd(&self) -> usize;
+
+    fn set_cwnd(&mut self, cwnd: usize);
+
+    fn ssthresh(&self) -> usize;
```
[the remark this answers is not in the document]
junhochoi: All shortcut methods in Recovery are removed, so now simply use recovery.cc.cwnd() and others.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: Changed as follows. I think it's good enough for now.

recovery.set_bytes_in_flight() -> recovery.cc.set_bytes_in_flight()
recovery.bytes_in_flight() -> recovery.cc.bytes_in_flight()

src/lib.rs
junhochoi: Changed to enum cc::Algorithm.

However, for examples, I added config.set_cc_algorithm_name() API which accepts a string. Otherwise when command line is parsed there need to be something which translates a string to cc::Algorithm enum. e.g. "reno" -> cc::Algorithm.

I added cc::lookup_cc_algorithm() method for that, but `cc` module itself is not visible to the application, so added a new config API. Any better idea is welcome.

src/recovery.rs
junhochoi: Now `Option<Box<dyn cc::CongestionControlTrait>>` -> `Box<dyn cc::CongestionControlTrait>` and all helper method are removed.

src/recovery.rs
junhochoi: OptionExt and all its usage are removed.

src/recovery.rs
junhochoi: All shortcut added to access CC modules is removed. Now simply use
recovery.cwnd -> recovery.cc.cwnd() and so on.

src/cc/reno.rs
junhochoi: `Debug` trait is implemented for `CongestionControl` and `trace_id` is removed (see related comment above for issues). But I still leave some trace!() logging for debuging cwnd and other CC variable changes.

junhochoi:
> Looks like the general idea is good, though I think the internal and external APIs can be improved so I added a bunch of comments (I left minor nits like ending comments with a period out, to avoid making this harder to follow, we can figure that out once we are happy with the general implementation).

Thank you very much for helpful comments! I think I addressed most of them (and reduced a lot of codes) but feel free to open any resolved comment if you have further.

I think going forward when we add more CC modules here, there will be following issues.

- Note that current design of CC module is 1) keeping all variables defined in QUIC recovery (Congestion Control) draft in cc module. 2) Implement all hooks (even some are still empty) defined in the draft.
- There is an exception, such as OnPacketsLost() and InPersistentCongestion(). Preferably I want to move them into CongestionControl, but some issues:
  - In actual implementations such as linux/BSD cc modules, they also have read/modify some of connection level stats/variables such as RTT, SRTT etc. Soon we will need them too, but accessing those (in our case, rtt is in Recovery) from CC module is not easy. One idea is to have CongestionControl trait for Recovery (but when it comes to multiple different CC modules, not sure it can be done easily)
  - However, accessing some variable in CC (e.g. bytes_in_flight) from Recovery is possible via methods, that's how this PR is doing.
  - Same reason with `trace_id`: currently defined in Connection and Connection methods are passing trace_id values only for necessary Recovery methods, it looks.

junhochoi:
For trace_id, I added 8c4aa6b for passing trace_id to CC events for logging purpose. It's a separate commit, so let me know if this works @ghedo

src/lib.rs
ghedo: My point was that we should also make the enum public, so the config API can just take a value from that and then you don't need to convert the string.

You can export the `cc::Algorithm` enum in `lib.rs` like:
```rust
pub use crate::cc::Algorithm CongestionControlAlgorithm;
```

src/recovery.rs
ghedo: As per my previous comment, this also doesn't need to be public, by making the `Recovery` consructor take the CC algorithm and create the correct one, rather than create a default CC state, and then change it later.

src/cc/mod.rs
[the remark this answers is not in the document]
ghedo: So, it looks like this is now only needed because of `collapse_cwnd()`, can we just make that the method that implementations need to implement? That would simplify the API.

src/cc/mod.rs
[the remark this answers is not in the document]
ghedo: Still, I think we can expose a more specific API, rather than always having to do the `saturating_sub()` dance. Something like, a `drop_packet()` method, that does the subtraction internally seems better IMO.

src/lib.rs
```diff
@@ -2515,7 +2543,7 @@ impl Connection {
             recv: self.recv_count,
             sent: self.sent_count,
             lost: self.recovery.lost_count,
-            cwnd: self.recovery.cwnd(),
+            cwnd: self.recovery.cc.cwnd(),
```
ghedo: This changes the meaning of the field I think, shouldn't it be be `self.recovery.cwnd_available()`?

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: Ok. set_cwnd() removed.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: Ok. set_bytes_in_flight() is replaced with decrease_bytes_in_flight(usize). I think this is basically controlling bytes_in_flight variable, so better to have the name in its API name.

src/lib.rs
junhochoi: Good tip! Added 'as' in between.

src/recovery.rs
junhochoi: So currently we don't have a constructor of Recovery other than default() which doesn't accept a paramter, so I changed Default implementation into a new() constructor in Recovery which accept Config as a paramter. In that way we can use Config object other than CC initilization going forward, but let me know this approach is ok.

src/cc/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+pub fn new_congestion_control(algo: Algorithm) -> Box<dyn CongestionControl> {
+    trace!("CC_DEBUG get_congestion_control: {:?}", algo);
+    match algo {
+        Algorithm::Reno => Box::new(cc::reno::Reno::new()),
+    }
+}
+
+// Return Algorithm enum from the string. This is mainly for command line
+// options support.
+pub fn lookup_cc_algorithm(name: &str) -> Algorithm {
```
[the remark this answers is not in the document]
junhochoi: Done.

src/cc/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+// Available CC algorithms.
+// This is defined in include/quiche.h as well.
+// When you add a new CC, update here and include/quiche.h both.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum Algorithm {
+    Reno = 0, // "reno"
+}
+
+impl std::fmt::Debug for dyn CongestionControl {
```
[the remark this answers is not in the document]
junhochoi: Which is fine, but I got the following if I removed L51:

```
error[E0277]: `dyn cc::CongestionControl` doesn't implement `std::fmt::Debug`
   --> src/recovery.rs:552:28
    |
552 |         write!(f, "{:?} ", self.cc)?;
    |                            ^^^^^^^ `dyn cc::CongestionControl` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
    |
    = help: the trait `std::fmt::Debug` is not implemented for `dyn cc::CongestionControl`
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `std::boxed::Box<dyn cc::CongestionControl>`
    = note: required by `std::fmt::Debug::fmt`
```

Which seems, recovery.cc is a trait object so it looks for Debug trait on the trait object first.

Also, Debug on Reno is removed (used to use for test, but no more).

src/cc/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    // CongestionEvent(time_sent)
+    // now is passed as well not to look up current time again.
+    fn congestion_event(
+        &mut self, time_sent: Instant, now: Instant, trace_id: &str,
+    );
+}
+
+// Returns a congestion control module, specified as `name`.
+pub fn new_congestion_control(algo: Algorithm) -> Box<dyn CongestionControl> {
+    trace!("CC_DEBUG get_congestion_control: {:?}", algo);
```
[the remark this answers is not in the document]
junhochoi: I removed CC_DEBUG string itself (this was for easily checking the messages using grep) from trace!() calls but still left some logging inside Reno module. Do you really mean not to have any logging in the CC module? I understand your concern this may be too verbose but I need those for checking cwnd and other internal CC changes during transfer. Also it would be great to know what is incosistent part here.

Another approach is that I can create a new macro for cc modules not logged by RUST_LOG... such as logged only when RUST_LOG=cc or something like that, until we have a proper qlog support.

src/lib.rs
junhochoi: Assuming stats() is for users to get some stats of the Connection, so providing current cwnd() is better I think (sorry I didn't mention it in the changelog). Is there any reason you want to return cwnd_available()? If someone calls stats() in the middle of connection cwnd() may return a smaller value than real (it returns cwnd-bytes_in_flight, or even usize::MAX).

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: So I tried to remove L51 block, and I got the following error:
```
error[E0277]: `dyn cc::CongestionControl` doesn't implement `std::fmt::Debug`
   --> src/recovery.rs:552:28
    |
552 |         write!(f, "{:?} ", self.cc)?;
    |                            ^^^^^^^ `dyn cc::CongestionControl` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
    |
    = help: the trait `std::fmt::Debug` is not implemented for `dyn cc::CongestionControl`
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `std::boxed::Box<dyn cc::CongestionControl>`
    = note: required by `std::fmt::Debug::fmt`

error: aborting due to previous error
```

Which seems that, what we want is that when `{:?}` is for `recovery.cc` we want to call Debug trait of Reno (actual object), but since this is `dyn CongestonControl` type. so still looking for Debug of that trait object.

junhochoi:
> Getting closer :)
> Also, we need more documentation I think. Both for the public API (`cc::Algorithm`) as well as for the trait methods, which seem to be documented not consistently (some methods are, some aren't, also we need to use the proper syntx and style, with `///` and a first line summary, newline, and then a more detailed description, see `Connection` for some examples).

Thanks! Most of the comment is addressed now but please see unresolved converstions. I will update the documentation once we settle down with the code.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: This is solved. I moved Debug from mod.rs to reno.rs

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: Reimplemented using FromStr trait.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: https://github.com/cloudflare/quiche/tree/cclog contains some idea for separating CC logs.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: However I still keep ssthresh(), currently it's only required in the testing code. (e.g. testing ssthresh changed after congestion event). I tried to move out of the trait but it didn't work well yesterday, but try to figure it out.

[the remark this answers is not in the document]
junhochoi: ssthresh() removed at 962f379

src/cc/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    // CongestionEvent(time_sent)
+    // now is passed as well not to look up current time again.
+    fn congestion_event(
+        &mut self, time_sent: Instant, now: Instant, trace_id: &str,
+    );
+}
+
+// Returns a congestion control module, specified as `name`.
```
LPardue: what does specified as `name` mean?

src/cc/reno.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    // Congestion Control hooks
+    fn on_packet_sent_cc(&mut self, bytes_sent: usize, trace_id: &str) {
+        self.bytes_in_flight += bytes_sent;
+
+        trace!("{} OnPacketSentCC() {:?}", trace_id, &self);
+    }
+
+    fn on_packet_acked_cc(
+        &mut self, packet: &Sent, _srtt: Duration, _min_rtt: Duration,
```
LPardue: will `srtt` and `min_rtt` ever get used? If not maybe we want to make them Options

src/cc/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Return Algorithm enum from the string.
+impl FromStr for Algorithm {
+    type Err = std::convert::Infallible;
+
+    /// This is mainly for command line options support.
+    fn from_str(name: &str) -> Result<Self, Self::Err> {
+        match name {
+            "reno" => Ok(Algorithm::Reno),
+            _ => Ok(Algorithm::Reno),
```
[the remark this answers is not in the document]
junhochoi: I was thinking Reno as "default" or "fall-back".  However panic will make things more clear. I will make changes.

src/cc/mod.rs
junhochoi: I will fix it.

src/cc/reno.rs
junhochoi: Those are reserved to other CCs (e.g. cubic and others) which need it. Reno doesn't use it, so it has underscore names. This parameters are not Options because the value is always available from Recovery module (there is no None case)

junhochoi:
> > Getting closer :)
> > Also, we need more documentation I think. Both for the public API (`cc::Algorithm`) as well as for the trait methods, which seem to be documented not consistently (some methods are, some aren't, also we need to use the proper syntx and style, with `///` and a first line summary, newline, and then a more detailed description, see `Connection` for some examples).
> 
> Thanks! Most of the comment is addressed now but please see my comment on unresolved conversions. I will update the documentation once we settle down with the code.

I updated documentation too. Please take a look.

src/cc/mod.rs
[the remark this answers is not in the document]
junhochoi: I changed the API again to return `Error` instead of panic!() so application can decide, when the string is not a known one.

src/lib.rs
```diff
@@ -229,6 +229,36 @@
...
+//! available congestion control algorithms.
+//!
+//! For example:
+//!
+//! ```
+//! let mut config = quiche::Config::new(quiche::PROTOCOL_VERSION).unwrap();
+//! config.set_cc_algorithm(quiche::CongestionControlAlgorithm::Reno);
+//! ```
+//!
+//! Alternatively, you can configure a congestion control algorithm to use
```
```suggestion
//! Alternatively, you can configure the congestion control algorithm to use
```

src/lib.rs
```diff
@@ -229,6 +229,36 @@
...
+//! by its name.
+//!
+//! ```
+//! let mut config = quiche::Config::new(quiche::PROTOCOL_VERSION).unwrap();
+//! config.set_cc_algorithm_name("reno").unwrap();
+//! ```
+//!
+//! Note that those API should be called before you call [`connect()`]
```
```suggestion
//! Note that the CC algorithm should be configured before calling [`connect()`]
```

src/lib.rs
```diff
@@ -229,6 +229,36 @@
...
+//! or [`accept()`]. Otherwise the connection will use a default CC algorithm.
+//! By default `Reno` will be used as a default congestion control.
```
```suggestion
```
Don't think we should repeat what the default is.

src/lib.rs
```diff
@@ -229,6 +229,36 @@
...
+//!
+//! [CongestionControlAlgorithm]: cc/enum.Algorithm.html
```
```suggestion
//! [`CongestionControlAlgorithm`]: cc/enum.Algorithm.html
```

src/lib.rs
```diff
@@ -381,6 +417,9 @@ pub struct Config {
     application_protos: Vec<Vec<u8>>,
 
     grease: bool,
+
+    // Congestion control algorithm to use
```
```suggestion
```

src/lib.rs
```diff
@@ -622,6 +662,35 @@ impl Config {
     pub fn set_disable_active_migration(&mut self, v: bool) {
         self.local_transport_params.disable_active_migration = v;
     }
+
+    /// Sets the congestion control algorithm used by string.
+    ///
+    /// The default value is `reno`. If failed, `Error::CongestionControl`
```
```suggestion
    /// The default value is `reno`. On error `Error::CongestionControl`
```

src/recovery.rs
```diff
@@ -103,19 +97,14 @@ pub struct Recovery {
...
-
-    cwnd: usize,
-
-    recovery_start_time: Option<Instant>,
-
-    ssthresh: usize,
-
     pub loss_probes: [usize; packet::EPOCH_COUNT],
+
+    // Congestion Control
```
ghedo:
```suggestion
```

include/quiche.h
```diff
@@ -161,6 +164,16 @@ void quiche_config_set_max_ack_delay(quiche_config *config, uint64_t v);
...
 
+enum quiche_cc_algorithm {
+    QUICHE_CC_RENO = 0,
+};
+
+// Sets the congestion control algorithm used (string).
+void quiche_config_set_cc_algorithm(quiche_config *config, enum quiche_cc_algorithm algo);
+
+// Sets the initial congestion control window size (bytes).
+void quiche_config_set_cc_initcwnd(quiche_config *config, size_t v);
```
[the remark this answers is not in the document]
junhochoi: Will be removed.

src/cc/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::fmt::Debug;
```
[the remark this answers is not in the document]
junhochoi: I need this for per-cc module Debug impl. It doesn't build without this.

src/cc/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn lookup_cc_algo_ok() {
+        let algo = Algorithm::from_str("reno").unwrap();
+
+        assert_eq!(algo, Algorithm::Reno);
+    }
+
+    #[test]
+    #[should_panic]
+    fn lookup_cc_algo_bad() {
+        let _ = Algorithm::from_str("???").unwrap(); // should panic!()
```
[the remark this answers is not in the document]
junhochoi: Oh this is a leftover from previous implementation. Nice catch.

junhochoi:
> Also, I see we still have a bunch of `trace!()` lines. Given the other PR on CC logging shouldn't these be removed?

Ok. I only left one at `new_congestion_control()` and removed ones in `reno.rs`.

src/cc/mod.rs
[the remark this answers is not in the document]
ghedo: Oh, right, I thought it would be included automatically, but actually in the other places we just use it directly like:
```rust
impl std::fmt::Debug for X {
    ...
}
```
to avoid importing something that is only used once.

src/lib.rs
```diff
@@ -622,6 +660,35 @@ impl Config {
...
+    ///
+    /// ## Examples:
+    ///
+    /// ```
+    /// # let mut config = quiche::Config::new(0xbabababa)?;
+    /// config.set_cc_algorithm_name("reno");
+    /// # Ok::<(), quiche::Error>(())
+    /// ```
+    pub fn set_cc_algorithm_name(&mut self, name: &str) -> Result<()> {
+        match CongestionControlAlgorithm::from_str(name) {
```
ghedo: I only just noticed this, sorry, but this whole thing can just be written as:
```rust
self.cc_algorithm = CongestionControlAlgorithm::from_str(name)?;

Ok(())
```

junhochoi:
> I made the following tweaks myself that I missed from the previous review (they are not worth going through another round of review) and rebased, will merge once build is complete.

LGTM. Thanks!

PR #300, opened 2019-12-13
try to fix the issue that "GCC is no longer supported" duration building android library

bachue:


junhochoi:
@bachue Thanks. However I see android build failures from CI (see Travis CI result above).
Currently we are testing with NDK 13b which is gcc. If you want to add clang support, you need to split the case gcc vs clang, maybe based on NDK version to have a different build config.

bachue:
@junhochoi  My NDK version is 21. And actually the version you are testing with is too old because I even cannot install this version from Android Studio. The oldest version for Android Studio is 21.

junhochoi:
@bachue I opened #327 to support latest NDK including this issue. Please take a look and let us know this works for you.

ghedo:
This should be fixed in https://github.com/cloudflare/quiche/commit/9704e9a2cf3b903693f7ff8bc039d5698ffb7684

PR #304, opened 2019-12-17
WIP recovery: separate logging for congestion control events

### LINKED ISSUE #303: Logging CC
#299 use trace!() for logging CC variable changes but probably too expensive and impact actual performance. Better to separate CC logging function and later it can be replaced (or added) to qlog support when we have it.

junhochoi:
Related to #299 and fixes #303 

The idea is to have a `cclog!()` macro basically same to other logging macro but basically it's not enabled without QUICHE_CC_LOG=1, so only we need logging from CC we can set this environment variable.

I used CC_LOG static mut variable, but it only changed in the call_once() and not changed, so using it under unsafe is fine (or I can't find not to use unsafe in the scenario).

p.s. This PR assumes #299.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -47,6 +47,7 @@ Options:
   --wire-version VERSION   The version number to send to the server [default: babababa].
   --dump-packets PATH      Dump the incoming packets as files in the given directory.
   --no-verify              Don't verify server's certificate.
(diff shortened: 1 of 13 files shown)


junhochoi:
Closing this one since we have qlog support now.

PR #307, opened 2019-12-19
WIP recovery: implement IsAppLimited()

junhochoi:
IsAppLimited() can be defined by bytes_in_flght < cwnd.

Let's consider the following scenario.

* initcwnd=3000

Send pkt #0 size  500 -> bytes_in_flight =  500 cwnd=3000
Recv ack #0           -> bytes_in_flight =    0 cwnd=3000
Send pkt #1 size 1000 -> bytes_in_flight = 1000 cwnd=3000
Send pkt #2 size 1000 -> bytes_in_flight = 2000 cwnd=3000
Send pkt #3 size 1000 -> bytes_in_flight = 3000 cwnd=3000
Recv ack #1           -> bytes_in_flight = 2000 cwnd=4000 (slow start)
Send pkt #4 size 1000 -> bytes_in_flight = 3000 cwnd=4000
Send pkt #5 size 1000 -> bytes_in_flight = 4000 cwnd=4000

Currently is_app_limited() returns always false,
We are actually doing like this:

Send pkt #0 size  500 -> bytes_in_flight =  500 cwnd=3000
Recv ack #0           -> bytes_in_flight =    0 cwnd=3500 (slow start)
Send pkt #1 size 1000 -> bytes_in_flight = 1000 cwnd=3500
Send pkt #2 size 1000 -> bytes_in_flight = 2000 cwnd=3500
Send pkt #3 size 1000 -> bytes_in_flight = 3000 cwnd=3500
Recv ack #1           -> bytes_in_flight = 2000 cwnd=4500 (slow start)
Send pkt #4 size 1000 -> bytes_in_flight = 3000 cwnd=4500
Send pkt #5 size 1000 -> bytes_in_flight = 4000 cwnd=4500

It means cwnd may be bloated by acks delivered before
bytes_in_flight reached cwnd. This happens because in the
beginning of QUIC connection has Initial and Handshake which
exchanges data before Application starts.

To solve this, is_app_limited() need to be defined.

However we can't define is_app_limited() as
bytes_in_flight < cwnd. This is because in on_packet_acked_cc(),
we decrease bytes_in_flight first, so when we call
self.is_app_limited() later, it always return true because
at this point bytes_in_flight < cwnd. This has an adverse impact
because now is_app_limited() returns true all the time, cwnd
will not grow.

To prevent this, now app_limited is a field in Reno. We update
this value in on_packet_sent_cc() only. When bytes_in_flight
is same or higher than cwnd, app_limited will be false.

Then, when on_packet_acked_cc() is called and when bytes_in
_flight is same to cwnd or higher, we use app_limited updated
last time when on_packet_sent_cc() is called. So when
bytes_in_flight == cwnd, on_packet_sent_cc() will decrease
bytes_in_flight but still see app_limited == false, so
cwnd will be updated.

Note that same technique can be used for cubic and other similar CCs.

p.s. This PR assumes #299.

### DIFF
--- a/src/cc/reno.rs
+++ b/src/cc/reno.rs
@@ -39,6 +39,8 @@ pub struct Reno {
     congestion_recovery_start_time: Option<Instant>,
 
     ssthresh: usize,
+
+    app_limited: bool,
     /* TODO: ECN is not implemented.
      * ecn_ce_counters: [usize; packet::EPOCH_COUNT] */
 }
@@ -56,8 +58,8 @@ impl cc::CongestionControl for Reno {
             congestion_recovery_start_time: None,
 
             ssthresh: std::usize::MAX,
-            /* TODO: ECN is not implemented.
-             * ecn_ce_counters: [0; packet::EPOCH_COUNT], */
+
+            app_limited: false,
         }
     }
 
@@ -84,6 +86,8 @@ impl cc::CongestionControl for Reno {
 
     fn on_packet_sent_cc(&mut self, bytes_sent: usize, _trace_id: &str) {
         self.bytes_in_flight += bytes_sent;
+
+        self.app_limited = self.bytes_in_flight < self.congestion_window;
     }
 
     fn on_packet_acked_cc(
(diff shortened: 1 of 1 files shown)


junhochoi:
Need to generalize it. Marking it `hold` for now.

junhochoi:
I opened a new PR #335 for more general solution not limited to Reno. Closing this.

PR #308, opened 2019-12-19
WIP recovery: quick recovery from a simple random loss

junhochoi:
Simple lost detection: when the recovery episode ends
and the duration of current recovery episode is
short than a given threshold, currently defined by
2 x min_rtt. This means there was a small loss
recovered very quickly, indicating this has a
low probability of congestion but a simple packet lost.
In this case we restore previous cwnd/ssthresh
to resume faster after recovery.

This gives a good performance improvement in case of high
random loss, not by congestion. Test result shows that
it can improve more than 40% in case of 1% random loss.

Note that it can be used other than Reno (cubic etc) too. (not BBR)

p.s. This PR assumes #299.

### DIFF
--- a/src/cc/reno.rs
+++ b/src/cc/reno.rs
@@ -31,6 +31,8 @@ use crate::cc;
 use crate::recovery::Sent;
 
 /// Reno congestion control implementation.
+const SIMPLE_LOST_THRESHOLD: u32 = 2;
+
 pub struct Reno {
     congestion_window: usize,
 
@@ -39,6 +41,10 @@ pub struct Reno {
     congestion_recovery_start_time: Option<Instant>,
 
     ssthresh: usize,
+
+    prev_cwnd: usize,
+
+    prev_ssthresh: usize,
     /* TODO: ECN is not implemented.
      * ecn_ce_counters: [usize; packet::EPOCH_COUNT] */
 }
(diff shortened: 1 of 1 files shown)


src/cc/reno.rs
```diff
@@ -95,6 +103,29 @@ impl cc::CongestionControl for Reno {
...
+            // and the duration of current recovery episode is
+            // short than a given threshold, currently defined by
+            // 2 x min_rtt. This means there was a small loss
+            // recovered very quickly, indicating this has a
+            // low probability of congestion but a simple packet lost.
+            // In this case we restore previous cwnd/ssthresh
+            // to resume faster after recovery.
+            if packet.time > congestion_recovery_start_time &&
+                congestion_recovery_start_time.elapsed() <
+                    SIMPLE_LOST_THRESHOLD * _min_rtt
```
LPardue: Realted to my comment on the other PR https://github.com/cloudflare/quiche/pull/299#discussion_r359852031

if using `_min_rtt` you can change it to `min_rtt`,

junhochoi: Fixed. Thanks!

junhochoi:
Rebased to master - I am thinking to make it configurable. Mark as hold for now.

junhochoi:
Closing this PR. I have another PR for CUBIC (https://github.com/cloudflare/quiche/pull/890) and once this is merged, I will create another PR for Reno later.

PR #309, opened 2019-12-19
extras/nginx: sync patch

ghedo:
Re-added all the whitespace back, because removing it from the patch got annoying.


junhochoi:
> LGTM but please fix the build failure.

#310 will fix build failure.

PR #311, opened 2020-01-06
[wip] lib: declare some public enums as non-exhaustive

LPardue:
not sure if we want to do this but...

rust 1.40 adds `non_exhaustive`, which seems to align well to making the quiche crate more extensible because it forces downstream crates to code for the possibility of new fields in structs or new enums variants.

I've focused on just quiche's public enums here, the small pact can be seen in the rust example client and server. There's probably also some other things like the integration tests or C examples that CI will catch - once it's uplifted to 1.40+.

### DIFF
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -377,6 +377,10 @@ fn main() {
                         }
                     },
 
+                    Ok((stream_id, _)) => {
+                        warn!("unsupported event on stream {}", stream_id);
+                    },
+
                     Err(quiche::h3::Error::Done) => {
                         break;
                     },
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -390,6 +390,10 @@ fn main() {
 
                         Ok((_stream_id, quiche::h3::Event::Finished)) => (),
 
+                        Ok((stream_id, _)) => {
+                            warn!("unsupported event on stream {}", stream_id);
+                        },
+
                         Err(quiche::h3::Error::Done) => {
                             break;
                         },
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -56,6 +56,7 @@ impl Level {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum Algorithm {
     #[allow(non_camel_case_types)]
     AES128_GCM,
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -37,6 +37,7 @@ pub const MAX_STREAM_OVERHEAD: usize = 12;
 pub const MAX_STREAM_SIZE: u64 = 1 << 62;
 
 #[derive(Clone, PartialEq)]
+#[non_exhaustive]
 pub enum Frame {
     Padding {
         len: usize,
--- a/src/h3/frame.rs
+++ b/src/h3/frame.rs
@@ -42,6 +42,7 @@ const SETTINGS_MAX_HEADER_LIST_SIZE: u64 = 0x6;
 const SETTINGS_QPACK_BLOCKED_STREAMS: u64 = 0x7;
 
 #[derive(Clone, PartialEq)]
+#[non_exhaustive]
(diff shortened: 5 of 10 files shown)


ghedo:
Note that most of those enums are not actually exposed in the public API, just internally in the crate, so those should remain unchanged I think.

It seems to make sense to do this for the public errors and h3 events enums (like you already did), though I don't think anyone would try to exhaustively match the errors in a real application, so might not be super useful in that case.

Also, need to consider what the potential side effects are, that is, if we add an event type and it doesn't get handled by the application, would that cause problems potentially?

LPardue:
Good points, thanks. I really should have marked this as WIP to make it clear that the PR was mainly to elicit some discussion on **if** we should do this and to understand the impact.

> Also, need to consider what the potential side effects are, that is, if we add an event type and it doesn't get handled by the application, would that cause problems potentially?

I think this is the pertinent question, regardless of if we have `non-exhaustive` or not. Matching the event enum doesn't necessarily mean the event is "handled", it sits in the library waiting for `poll` to be called again. This is a little unintuitive and has manifested in unintentional behaviour were people want to skip an event. So maybe we should think about generating some error if the pending event queue grows too large, or if the `poll` method is called too many times without the queue changing.

The example I have in mind comes from the Datagram extension to HTTP/3 (not opened against quiche yet), which adds a new event type to `poll`. This should only be triggered if the applications negotiated the use of the Datagram frame (aka dependent on Config). The datagram events are buffered to some arbitrary threshold number and so an application that reads them too slowly (or not at all) will miss datagrams. One might implement the queue as LIFO, so old datagrams get dropped if the library is polled too slowly. Regardless, would it be kind of annoying to have to add exhaustive matching for a Datagram event that an application implicitly or explicitly opted out of in Config?

junhochoi:
> Regardless, would it be kind of annoying to have to add exhaustive matching for a Datagram event that an application implicitly or explicitly opted out of in Config?

Not sure if I understand this fully but as you have in this patch we can use `_` to match all other events which doesn't have a specific processing. Also we assume quiche is built with customer application every time, so we can simply recommend to use `_` for any non-standard events. I don't know what is a good scenario for making enums non-exhaustive.

LPardue:
The issue is a bit of a meta one. Imagine today third-party applications that use quiche's HTTP/3 API just like we write in the docs.

```
match h3_conn.poll(&mut conn) {
        Ok((stream_id, quiche::h3::Event::Headers{list, has_body})) => {
            let mut headers = list.into_iter();

            // Look for the request's method.
            let method = headers.find(|h| h.name() == ":method").unwrap();

            // Look for the request's path.
            let path = headers.find(|h| h.name() == ":path").unwrap();

            if method.value() == "GET" && path.value() == "/" {
                let resp = vec![
                    quiche::h3::Header::new(":status", &200.to_string()),
                    quiche::h3::Header::new("server", "quiche"),
                ];

                h3_conn.send_response(&mut conn, stream_id, &resp, false)?;
                h3_conn.send_body(&mut conn, stream_id, b"Hello World!", true)?;
            }
        },

        Ok((stream_id, quiche::h3::Event::Data)) => {
            // Request body data, handle it.
        },

        Ok((stream_id, quiche::h3::Event::Finished)) => {
            // Peer terminated stream, handle it.
        }

        Err(quiche::h3::Error::Done) => {
            // Done reading.
            break;
        },

        Err(e) => {
            // An error occurred, handle it.
            break;
        },
}
```

If we add any new Event to `poll()`, suddenly every one of those apps breaks with a compile error caused by `match` not handling that case. *If* the applications had added the `_` case, like you suggest, then they would at least still build. By adding `non-exhaustive`, it forces everyone to have a `_` match, which is a vaccination against the issue ever coming up.

junhochoi:
I see, it makes sense to me.

This was useful for understanding: https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md

LPardue:
On deeper reflection, this seems like an annoying change to make

PR #312, opened 2020-01-06
 h3: return early from poll() when the connection is closing

ghedo:
When "close()" is called (e.g. due to an internal error) the connection
(both transport and HTTP/3 one) might be in a broken state, so we should
avoid doing any work that might cause undefined behavior.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -826,6 +826,12 @@ impl Connection {
     /// [`send_body()`]: struct.Connection.html#method.send_body
     /// [`close()`]: ../struct.Connection.html#method.close
     pub fn poll(&mut self, conn: &mut super::Connection) -> Result<(u64, Event)> {
+        // When the connection is closing (e.g. due to a protocol error) it might
+        // be in a broken state, so return early.
+        if conn.is_closing() {
+            return Err(Error::Done);
+        }
+
         // Process control streams first.
         if let Some(stream_id) = self.peer_control_stream_id {
             self.process_control_stream(conn, stream_id)?;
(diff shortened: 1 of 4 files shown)


src/h3/mod.rs
```diff
@@ -826,6 +826,12 @@ impl Connection {
     /// [`send_body()`]: struct.Connection.html#method.send_body
     /// [`close()`]: ../struct.Connection.html#method.close
     pub fn poll(&mut self, conn: &mut super::Connection) -> Result<(u64, Event)> {
+        // When the connection is closing (e.g. due to a protocol error) it might
+        // be in a broken state, so return early.
+        if conn.is_closing() {
+            return Err(Error::Done);
+        }
```
LPardue: Just wondering how this would relate to TODO GOAWAY behaviour? 

Imagine the scenario where a client sends 10 requests that are received and buffered up ready to be received via `poll()`, it issues a GOAWAY with MAX INT, sends all the responses, then sends CONNECTION_CLOSE(NO_ERROR). Would this early close cause all the pending response data to be "stuck" inside the library and blown away before the client could read it?

ghedo: That's regardless of this change I think. If `close()` is called we won't send `STREAM` frames anymore https://github.com/cloudflare/quiche/blob/530b207197b33631d264dedaf69ab3009e3f5529/src/lib.rs#L1842

Not sure how `GOAWAY` is supposed to work though. If the server cares about the data being received by the client it really needs to send `CONNECTION_CLOSE` only after the data has been sent (e.g. application could be blocked because of not enough flow control) and acked.

In HTTP/2 I don't think we close the TCP socket immediately after sending `GOAWAY` in most cases, so I guess HTTP/3 should work like that as well.

LPardue: So in this case the data would be "received" by the client because it was read out of the QUIC layer, but at some point a CONNECTION_CLOSE could come in disrupt the flushing of our H3 layer queuing. We can probably handle that by modulating the check to consider graceful close when we get around to implementing it. 

For now I think the PR makes things better.

ghedo:
Discussed this in person. We don't really need to add the public `is_closing()` API, and the name is wrong anyway as it really only check if the _local_ application initiated the close.

Also, I realized that the existing `is_closing` checks are wrong as, again, they only check if the _local_ application initiated the close, so will fix that in a separate PR.

PR #316, opened 2020-01-08
examples: add client option for requests to different URLs

LPardue:
Previously we allows only a single URL value on the CLI meaning
that we restricted `client` to making a single HTTP/0.9
request per conection, and `http3-client` to making multiple requests
for the same resource in a single connection if `-n` was provided.

With this change, we extend the CLI to allow a list of URLs.
`client` and `http3-client` then create one connection to the first
host in that list and issue as many requests as required.

When `http3-client` is invoked with multiple URLs together with
`-n` each URL is requested that many times before moving to the next.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -35,10 +35,8 @@ use ring::rand::*;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
-const HTTP_REQ_STREAM_ID: u64 = 4;
-
 const USAGE: &str = "Usage:
-  client [options] URL
+  client [options] URL...
   client -h | --help
 
 Options:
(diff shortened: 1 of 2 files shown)


examples/client.rs
```diff
@@ -35,10 +35,8 @@ use ring::rand::*;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
-const HTTP_REQ_STREAM_ID: u64 = 4;
-
 const USAGE: &str = "Usage:
-  client [options] URL
+  client [options] URL...
```
ghedo: While at it, do we want to add the `-n` option here as well? To make this more consistent with the http3-client.

examples/client.rs
```diff
@@ -77,14 +75,21 @@ fn main() {
...
+        .get_vec("URL")
+        .into_iter()
+        .map(|x| url::Url::parse(x).unwrap())
+        .collect();
 
     // Setup the event loop.
     let poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
 
+    // We'll only connect to the first server provided in URL list
```
ghedo:
```suggestion
    // We'll only connect to the first server provided in URL list.
```

examples/client.rs
```diff
@@ -161,9 +167,25 @@ fn main() {
...
+    // Create a single list of all the requests to be issued.
+    let mut reqs = std::collections::HashMap::new();
+    let mut stream_id = 0;
+
+    for url in &urls {
+        let req = format!("GET {}\r\n", url.path());
+
+        reqs.insert(stream_id, req);
+
+        stream_id += 4;
```
ghedo: Rather than associate the stream ID with the request here, we can just make a `Vec` and then calculate the stream ID based on `i` when sending them. This seems more consistent with http3-client.

examples/http3-client.rs
```diff
@@ -85,8 +89,11 @@ fn main() {
     let poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
 
+    // We'll only connect to the first server provided in URL list
```
```suggestion
    // We'll only connect to the first server provided in URL list.
```

examples/client.rs
LPardue: that's more or less how I started but then realised that when it comes to stream closure things were quite naiive. With H3 a lot of the mapping is done by the library but with H0.9 there is nothing, and being too simplistic with stream closures might cause some weirdness like counting a unidirectional stream closures against requests complete count.

PR #321, opened 2020-01-14
travis: split NGINX into its own build

ghedo:
Currently the "stable Linux" build is the one that takes the most time
to complete, which means that the entire build will have to wait for
that to complete.

By splitting NGINX into its own build we lower the time of the "stable
Linux" build. While this goes over the 5 concurrent builds limit, the
"stable Android" is very short, which means we might be able to complete
both that one and the NGINX one before "stable Linux" (or around the
same time).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ addons:
 
 matrix:
   include:
-   - name: "stable Linux x86_64/x86 + NGINX"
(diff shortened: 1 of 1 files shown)


ghedo:
This looks good now. It seems we might get a few minutes back in some cases, but not always.

PR #322, opened 2020-01-14
Enable cargo cache again

junhochoi:
Fix go version to 1.13.5 seems to fix the issue with cargo cache.

Last build failure is because `gimme` installed go 1.13.6 but boringssl build tried to find go 1.13.5, possibly from cache. BoringSSL build need go 1.13 or later, so any version of go 1.13.x works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ matrix:
      before_install:
       # Install and use the current stable release of Go
       - gimme --list
-      - eval "$(gimme -f 1.13.5)"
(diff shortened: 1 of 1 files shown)


ghedo:
BoringSSL targets the latest Go stable, so whenever a new GO is released we would have to update this manually, which doesn't seem great... Does the build still fail if you just use `stable`? Maybe the cache got cleared when we disabled it.

junhochoi:
Yeah it seems ok with `gimme stable`. Probably a travis issue but I don't know why cargo cache is impacting go binary version :<

PR #329, opened 2020-01-17
extra: add multiconnect test for quic-interop-runner

junhochoi:
multiconnect test requires:
The client is expected to establish a separate connection for each of the files requested in the $REQUESTS parameter. The connections may be established serially or in parallel. 

This is exactly how we are doing right now, so adding test name will be good enough.

### DIFF
--- a/extras/docker/qns/run_endpoint.sh
+++ b/extras/docker/qns/run_endpoint.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 # Set up the routing needed for the simulation
 /setup.sh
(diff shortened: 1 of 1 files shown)


junhochoi:
New result (quiche vs quiche only):

```
+--------+--------+
|        | quiche |
+--------+--------+
| quiche |        |
|        |        |
|        |   L1   |
+--------+--------+
+--------+----------------------+
|        |        quiche        |
+--------+----------------------+
| quiche | G: 8643 (± 143) kbps |
|        | C: 3784 (± 469) kbps |
+--------+----------------------+
```

L1 is failing due to the following error (but out of scope of this PR):

```
Expected 50 handshakes. Got: 54
```

PR #331, opened 2020-01-20
tools: add quiche-client and quiche-server

LPardue:
This adds a client and server application that
supports both HTTP/0.9 an HTTP/3 by default,
with preference towards the latter.

Other major differences compared to the example
binaries are the ability to multiplex requests
to different URLS, and for the client to log
response bodies to an indicated folder.

Common code is extracted into a library crate
but there is suprisingly little in common
between these two apps.


tools/apps/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        let vec: Vec<String> = buf.iter().map(|b| format!("{:02x}", b)).collect();
+
+        vec.join("")
+    }
+
+    /// ALPN helpers.
+    ///
+    /// This module contains constants and functions for working with ALPN.
+    pub mod alpns {
+        pub const HTTP_09: [&str; 3] = ["hq-24", "hq-23", "http/0.9"];
```
junhochoi: Do we need to support hq-23 with this new app?

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        pub const HTTP_3: [&str; 2] = ["h3-24", "h3-23"];
```
junhochoi: Same here with above comment.

tools/apps/src/lib.rs
LPardue: The library supports it, so there doesn't seem harm in allowing it

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,790 @@
...
+// Represents an HTTP/3 formatted request.
+struct Http3Request {
+    url: url::Url,
+    cardinal: u64,
+    hdrs: Vec<quiche::h3::Header>,
+}
+
+// Used for application tracking of HTTP requests and responses.
+#[allow(clippy::large_enum_variant)]
+enum HttpConn {
```
ghedo: It seems like a better alternative would be to define an `HttpConn` trait, with two implementations for /0.9 and /3. They could all live in the library crate and have both client and server methods, so these files would simply call those methods without having to `match` every time.

Same for requests I think.

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,790 @@
...
+        path = format!("{}.{}", path, cardinal);
+    }
+
+    match std::fs::File::create(&path) {
+        Ok(f) => std::io::BufWriter::new(f),
+        Err(e) => panic!("Bad times: {}", e),
+    }
+}
+
+fn report_http_incomplete(
```
ghedo: This can just be an `HttpConn` method?

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,790 @@
...
+
+// Used for application tracking of HTTP requests and responses.
+#[allow(clippy::large_enum_variant)]
+enum HttpConn {
+    Http09Conn {
+        stream_id: u64,
+        reqs_sent: usize,
+        reqs_complete: usize,
+        reqs: Vec<Http09Request>,
+        fds: HashMap<u64, std::io::BufWriter<std::fs::File>>,
```
ghedo: Why isn't the "fd" just part of the request struct? Also, would propbably call it something else, like, "writer" or something.

tools/apps/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "quiche_apps"
+version = "0.1.0"
+authors = ["Lucas Pardue <lucaspardue.24.7@gmail.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+docopt = "1"
+env_logger = "0.6"
+http = "0.2"
```
ghedo: Is this used anywhere?

Also, `ring` and `serde*`

junhochoi:
@LPardue can you update qns run_endpoint.sh as well? Eventually we need to test with it.

LPardue:
@junhochoi please see https://github.com/cloudflare/quiche/commit/5b242cd2ef47ddc8631d7dd3ac7c8a4b595679d0 and let me know if that looks like what you have me mind. Initially I figured it could be done as a separate PR (so we can manage switchover independently) but I can pull it into here if we want to do it simultaenously

tools/apps/src/bin/quiche-client.rs
LPardue: good idea

tools/apps/src/bin/quiche-client.rs
LPardue: I think this is down to being an afterthought feature mixed with wanting to keep a sane order for issuing requests. When not doing anything with the responses we didn't need any look up based on their stream ID. Let me take another look at this and see if I can find something more logical.

junhochoi:
> @junhochoi please see [5b242cd](https://github.com/cloudflare/quiche/commit/5b242cd2ef47ddc8631d7dd3ac7c8a4b595679d0) and let me know if that looks like what you have me mind. Initially I figured it could be done as a separate PR (so we can manage switchover independently) but I can pull it into here if we want to do it simultaenously

Yes I think we need some change to run_endpoint.sh to support multiple download. And first use of this PR will be interop-runner, why not include it here for easier testing?
Also please rebase to master. There was some change on run_endpoint.sh recently.

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,790 @@
...
+                        debug!(
+                            "stream {} has {} bytes (fin? {})",
+                            s,
+                            stream_buf.len(),
+                            fin
+                        );
+
+                        print!("{}", unsafe {
+                            std::str::from_utf8_unchecked(&stream_buf)
+                        });
```
junhochoi: Since we have `--dump-responses` option, I think when we set this option, no need to print the body. to stdout

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,790 @@
...
+                                h3_conn.recv_body(conn, stream_id, buf)
+                            {
+                                debug!(
+                                    "got {} bytes of response data on stream {}",
+                                    read, stream_id
+                                );
+
+                                print!("{}", unsafe {
+                                    std::str::from_utf8_unchecked(&buf[..read])
+                                });
```
junhochoi: (same as L599) Since we have `--dump-responses` option, I think when we set this option, no need to print the body. to stdout

junhochoi:
@LPardue for run_endpoint.sh, the following patch is needed to pass transfer test (after rebase)

```diff --git a/extras/docker/qns/run_endpoint.sh b/extras/docker/qns/run_endpoint.sh
index 94d9eb1..2b02a1f 100644
--- a/extras/docker/qns/run_endpoint.sh
+++ b/extras/docker/qns/run_endpoint.sh
@@ -45,16 +45,13 @@ check_testcase () {
 }

 run_quiche_client_tests () {
-    for req in $REQUESTS
-    do
-        $QUICHE_DIR/$QUICHE_CLIENT $QUICHE_CLIENT_OPT \
-            $CLIENT_PARAMS $req 2> $LOG || exit 127
-    done
+    $QUICHE_DIR/$QUICHE_CLIENT $QUICHE_CLIENT_OPT \
+        $CLIENT_PARAMS $REQUESTS 2>&1 > $LOG
 }

 run_quiche_server_tests() {
     $QUICHE_DIR/$QUICHE_SERVER --listen 0.0.0.0:443 --root $WWW_DIR \
-        $SERVER_PARAMS $QUICHE_SERVER_OPT 2> $LOG || exit 127
+        $SERVER_PARAMS $QUICHE_SERVER_OPT 2>&1 > $LOG
 }
```

tools/apps/Cargo.toml
LPardue: ring is used by quiche-server, the others are bogus

tools/apps/src/bin/quiche-client.rs
LPardue: it matches the behaviour of the example apps and is useful to see stuff happen when running interactively. Happy to make this more conditional if we really want

LPardue:
Think I've addressed most review comments. Seems sensible to wait until the draft 25 changes land on master, then I can rebase FTW

tools/apps/src/bin/quiche-client.rs
junhochoi: When quiche-clients runs with multiple urls, stdout output can be mixed up -- so need to use --dump-responses option any way to verify. Also in interop runner, log file will contain all body dump, make it harder to look at -- found during I run interop script using this branch.

LPardue: Agreed. I'll make the change to not print if `--dump-response` is set.

extras/docker/qns/run_endpoint.sh
```diff
@@ -49,10 +47,8 @@ check_testcase () {
 run_quiche_client_tests () {
     for req in $REQUESTS
```
junhochoi: No need to use `for`. Can you apply my patch https://github.com/cloudflare/quiche/pull/331#issuecomment-579557222

LPardue: oops, will fix that tomorrow

LPardue:
@junhochoi PTAL

tools/apps/src/bin/quiche-server.rs
```diff
@@ -0,0 +1,504 @@
...
+            debug!("{} processed {} bytes", client.conn.trace_id(), read);
+
+            // Create a new HTTP connection as soon as the QUIC connection
+            // is established.
+            if client.http_conn.is_none() &&
+                (client.conn.is_in_early_data() ||
+                    client.conn.is_established())
+            {
+                let app_proto = client.conn.application_proto();
+                let app_proto = &std::str::from_utf8(&app_proto).unwrap();
```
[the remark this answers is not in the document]
LPardue: Since I need value, do you want me to pattern match rather than unwrap?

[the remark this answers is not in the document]
ghedo: Oh, wait, I just realized that this is always going to be valid, otherwise we close the connection because of failed ALPN negotiation, so `unwrap()` is probably fine here, though maybe add a comment.

extras/docker/Dockerfile
```diff
@@ -39,10 +43,8 @@ WORKDIR /quiche
 COPY --from=build /build/quiche/examples/cert.crt \
      /build/quiche/examples/cert.key \
      examples/
-COPY --from=build /build/quiche/target/release/examples/client \
-     /build/quiche/target/release/examples/server \
-     /build/quiche/target/release/examples/http3-client \
-     /build/quiche/target/release/examples/http3-server \
+COPY --from=build /build/quiche/tools/apps/target/release/quiche-client \
+    /build/quiche/tools/apps/target/release/quiche-client \
```
junhochoi: quiche-client -> quiche-server

junhochoi:
```suggestion
    /build/quiche/tools/apps/target/release/quiche-server \
```

extras/docker/qns/run_endpoint.sh
```diff
@@ -12,9 +12,9 @@ set -e
 QUICHE_DIR=/quiche
 WWW_DIR=/www
 DOWNLOAD_DIR=/downloads
-QUICHE_CLIENT=client
-QUICHE_SERVER=server
-QUICHE_CLIENT_OPT="--no-verify"
+QUICHE_CLIENT=quiche-client
+QUICHE_SERVER=quiche-server
+QUICHE_CLIENT_OPT="--no-verify --dump-responses ${DOWNLOAD_DIR}"
 QUICHE_SERVER_OPT="--no-retry"
```
junhochoi:
```suggestion
QUICHE_SERVER_OPT="--max-streams-bidi 3000 --max-streams-uni 3000 --no-retry --cert examples/cert.crt --key examples/cert.key"
```

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,917 @@
...
+                            },
+
+                            None => {
+                                print!("{}", unsafe {
+                                    std::str::from_utf8_unchecked(&buf[..read])
+                                });
+                            },
+                        }
+
+                        if let Some(f) = &mut req.response_writer {
```
junhochoi: L747-L749 is making stored file doubled.

ghedo:
Also need to import the changes from https://github.com/cloudflare/quiche/pull/348.

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,917 @@
...
+                            f.write_all(&buf[..read]).ok();
+                        }
```
junhochoi:
```suggestion
```

LPardue: ah yeah this is stupid, thanks for finding it

extras/docker/qns/run_endpoint.sh
ghedo: We shouldn't just increase the stream limit for the purpose of passing the test. If there is a bug in the implementation we should fix it.

junhochoi: Oh. Probably leftover from testing. I will remove.

extras/docker/qns/run_endpoint.sh
```diff
@@ -12,10 +12,10 @@ set -e
...
 WWW_DIR=/www
 DOWNLOAD_DIR=/downloads
-QUICHE_CLIENT=client
-QUICHE_SERVER=server
-QUICHE_CLIENT_OPT="--no-verify"
-QUICHE_SERVER_OPT="--no-retry"
+QUICHE_CLIENT=quiche-client
+QUICHE_SERVER=quiche-server
+QUICHE_CLIENT_OPT="--no-verify --dump-responses ${DOWNLOAD_DIR}"
+QUICHE_SERVER_OPT="--max-streams-bidi 3000 --max-streams-uni 3000 --no-retry --cert examples/cert.crt --key examples/cert.key"
```
junhochoi:
```suggestion
QUICHE_SERVER_OPT="--no-retry --cert examples/cert.crt --key examples/cert.key"
```

extras/docker/qns/run_endpoint.sh
```diff
@@ -47,18 +45,14 @@ check_testcase () {
...
 run_quiche_client_tests () {
-    for req in $REQUESTS
-    do
-        # get path only from the url
-        file=$(echo $req | perl -F'/' -an -e 'print $F[-1]')
-        $QUICHE_DIR/$QUICHE_CLIENT $QUICHE_CLIENT_OPT \
-            $CLIENT_PARAMS $req > $DOWNLOAD_DIR/$file 2> $LOG || exit 127
-    done
+    $QUICHE_DIR/$QUICHE_CLIENT $QUICHE_CLIENT_OPT \
+        $CLIENT_PARAMS $REQUESTS 2>&1 > $LOG
```
junhochoi:
```suggestion
        $CLIENT_PARAMS $REQUESTS >& $LOG
```

extras/docker/qns/run_endpoint.sh
```diff
@@ -47,18 +45,14 @@ check_testcase () {
...
+
 }
 
 run_quiche_server_tests() {
     $QUICHE_DIR/$QUICHE_SERVER --listen 0.0.0.0:443 --root $WWW_DIR \
-        $SERVER_PARAMS $QUICHE_SERVER_OPT 2> $LOG || exit 127
+        $SERVER_PARAMS $QUICHE_SERVER_OPT 2>&1 > $LOG
```
junhochoi:
```suggestion
        $SERVER_PARAMS $QUICHE_SERVER_OPT >& $LOG
```

junhochoi:
Now I can get mostly green:
```
+--------+--------+--------+
|        | quiche | quicgo |
+--------+--------+--------+
| quiche |   M    |        |
|        |        |        |
|        |        |   M    |
+--------+--------+--------+
| quicgo |   M    |   M    |
|        |        |        |
|        |        |        |
+--------+--------+--------+
+--------+----------------------+----------------------+
|        |        quiche        |        quicgo        |
+--------+----------------------+----------------------+
| quiche | G: 9029 (± 168) kbps | G: 9170 (± 297) kbps |
|        | C: 3039 (± 168) kbps | C: 6583 (± 221) kbps |
+--------+----------------------+----------------------+
| quicgo | G: 9385 (± 271) kbps | G: 9392 (± 129) kbps |
|        | C: 5083 (± 306) kbps | C: 5511 (± 344) kbps |
+--------+----------------------+----------------------+
```

Only left issue is

```
Server: quicgo. Client: quiche. Running test case: multiplexing
Downloaded the wrong number of files. Got 1336, expected 1999.
```

But it's better to have another PR to fix this.

LPardue:
I addressed all of @ghedo's comments (to my knowledge).

extras/docker/qns/run_endpoint.sh
```diff
@@ -47,18 +45,15 @@ check_testcase () {
...
 
 run_quiche_client_tests () {
-    for req in $REQUESTS
-    do
-        # get path only from the url
-        file=$(echo $req | perl -F'/' -an -e 'print $F[-1]')
-        $QUICHE_DIR/$QUICHE_CLIENT $QUICHE_CLIENT_OPT \
-            $CLIENT_PARAMS $req > $DOWNLOAD_DIR/$file 2> $LOG || exit 127
-    done
+    sleep 3
```
ghedo: Is this required? Maybe add a comment to say why?

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,381 @@
...
+                http_conn = Some(Http3Conn::with_urls(
+                    &mut conn,
+                    &args.urls,
+                    args.reqs_cardinal,
+                    &args.req_headers,
+                    &args.body,
+                    &args.method,
+                ));
+            } else {
+                panic!("Negotiated unhandled protocol {:?}", app_proto);
```
ghedo: Remove? Don't think we can get this far.

tools/apps/src/bin/quiche-client.rs
```diff
@@ -0,0 +1,381 @@
...
+        let version = args.get_str("--wire-version");
+        let version = u32::from_str_radix(version, 16).unwrap();
+
+        let dump_response_path = if args.get_str("--dump-responses") != "" {
+            Some(args.get_str("--dump-responses").to_string())
+        } else {
+            None
+        };
+
+        // URLs (can be multiple)
```
ghedo:
```suggestion
        // URLs (can be multiple).
```

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+        req_start: &std::time::Instant,
+    );
+
+    fn report_incomplete(&self, start: &std::time::Instant);
+
+    fn handle_requests(
+        &mut self, conn: &mut std::pin::Pin<Box<quiche::Connection>>,
+        partial_responses: &mut HashMap<u64, PartialResponse>, root: &str,
+        buf: &mut [u8],
+    ) -> i8;
```
ghedo: This is not C :) so it should return `Result`. It seems the only error condition is when quiche has an error, so you can probably just use `quiche::Result<()>` and simply propagate the quiche error on failure, and return `Ok(())` on success.

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+    // pub fn default() -> Box<dyn HttpConn> {
+    // let h_conn = Http09Conn {
+    // stream_id: 0,
+    // reqs_sent: 0,
+    // reqs_complete: 0,
+    // reqs: Vec::new(),
+    // };
+    //
+    // Box::new(h_conn)
+    // }
```
ghedo: Remove above commented out code?

tools/apps/src/bin/quiche-server.rs
```diff
@@ -0,0 +1,507 @@
...
+    if token.len() < addr.len() || &token[..addr.len()] != addr.as_slice() {
+        return None;
+    }
+
+    let token = &token[addr.len()..];
+
+    Some(&token[..])
+}
+
+// Application-specific arguments that compliment the `CommonArgs`.
```
ghedo:
```suggestion
/// Application-specific arguments that compliment the `CommonArgs`.
```

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+    pub conn: std::pin::Pin<Box<quiche::Connection>>,
+
+    pub http_conn: Option<Box<dyn crate::HttpConn>>,
+
+    pub partial_responses: std::collections::HashMap<u64, PartialResponse>,
+}
+
+pub type ClientMap = HashMap<Vec<u8>, (net::SocketAddr, Client)>;
+
+// Makes a buffered writer for a resource with a target URL. The
```
ghedo:
```suggestion
/// Makes a buffered writer for a resource with a target URL. The
```
Also, the rest of the lines. Also, should have first line until first period, then empty line, then rest of the comment.

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+        buf: &mut [u8],
+    ) -> i8;
+
+    fn handle_writable(
+        &mut self, conn: &mut std::pin::Pin<Box<quiche::Connection>>,
+        partial_responses: &mut HashMap<u64, PartialResponse>, stream_id: u64,
+    );
+}
+
+// Represents an HTTP/0.9 formatted request.
```
ghedo:
```suggestion
/// Represents an HTTP/0.9 formatted request.
```

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+pub struct Http09Request {
+    url: url::Url,
+    cardinal: u64,
+    request_line: String,
+    stream_id: Option<u64>,
+    response_writer: Option<std::io::BufWriter<std::fs::File>>,
+}
+
+// Represents an HTTP/3 formatted request.
```
ghedo:
```suggestion
/// Represents an HTTP/3 formatted request.
```

tools/apps/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+            reqs_sent: 0,
+            reqs_complete: 0,
+            reqs: Vec::new(),
+            body: None,
+        };
+
+        Box::new(h_conn)
+    }
+
+    // Builds an HTTP/3 response given a request.
```
ghedo:
```suggestion
    /// Builds an HTTP/3 response given a request.
```

extras/docker/qns/run_endpoint.sh
junhochoi: https://github.com/marten-seemann/quic-interop-runner/issues/61
I added a comment

tools/apps/src/lib.rs
LPardue: ha yeah this is my fault. I did try to use result to start with but it really screwed something up (the loop breaking maybe?). Now the code has been refactored then it might be easy to put this in.

tools/apps/src/lib.rs
LPardue: oops

tools/apps/src/lib.rs
ghedo: In practice you can basically do the same thing you are doing now, except that instead of checking ` == -1` you can just do `.is_err()`. e.g.
```rust
if blah().is_err() {
  // handle error
}
```

LPardue:
PTA(nother)L

PR #335, opened 2020-01-24
recovery: implement IsAppLimited()

junhochoi:
IsAppLimited() can be defined by bytes_in_flght < cwnd.

Let's consider the following scenario.

```
(initcwnd=3000)
Send pkt #0 size  500 -> bytes_in_flight =  500 cwnd=3000
Recv ack #0           -> bytes_in_flight =    0 cwnd=3000
Send pkt #1 size 1000 -> bytes_in_flight = 1000 cwnd=3000
Send pkt #2 size 1000 -> bytes_in_flight = 2000 cwnd=3000
Send pkt #3 size 1000 -> bytes_in_flight = 3000 cwnd=3000
Recv ack #1           -> bytes_in_flight = 2000 cwnd=4000 (slow start)
Send pkt #4 size 1000 -> bytes_in_flight = 3000 cwnd=4000
Send pkt #5 size 1000 -> bytes_in_flight = 4000 cwnd=4000
```

Currently is_app_limited() returns always false,
We are actually doing like this:

```
Send pkt #0 size  500 -> bytes_in_flight =  500 cwnd=3000
Recv ack #0           -> bytes_in_flight =    0 cwnd=3500 (slow start)
Send pkt #1 size 1000 -> bytes_in_flight = 1000 cwnd=3500
Send pkt #2 size 1000 -> bytes_in_flight = 2000 cwnd=3500
Send pkt #3 size 1000 -> bytes_in_flight = 3000 cwnd=3500
Recv ack #1           -> bytes_in_flight = 2000 cwnd=4500 (slow start)
Send pkt #4 size 1000 -> bytes_in_flight = 3000 cwnd=4500
Send pkt #5 size 1000 -> bytes_in_flight = 4000 cwnd=4500
```

It means cwnd may be bloated by acks delivered before
bytes_in_flight reached cwnd. This happens because in the
beginning of QUIC connection has Initial and Handshake which
exchanges data before Application starts.

To solve this, is_app_limited() need to be defined properly.

However we can't define is_app_limited() as
bytes_in_flight < cwnd simply. This is because in on_packet_acked_cc(),
we decrease bytes_in_flight first, so when we call
self.is_app_limited() later, it always return true because
at this point bytes_in_flight < cwnd. This has an adverse impact
because now is_app_limited() returns true all the time, cwnd
will not grow.

The fix is create a bool variable "app_limited" in Recovery.
This value will be updated when we send a new data.
(true when bytes_in_flight < cwnd)

Also cc.is_app_limited() is removed. Instead,
app_limited flag will be passed to on_packet_acked_cc(),
so that it will be available on congestion control module.

### DIFF
--- a/src/recovery.rs
+++ b/src/recovery.rs
@@ -100,6 +100,8 @@ pub struct Recovery {
     pub loss_probes: [usize; packet::EPOCH_COUNT],
 
     pub cc: Box<dyn cc::CongestionControl>,
+
+    app_limited: bool,
 }
 
 impl Recovery {
@@ -138,6 +140,8 @@ impl Recovery {
             loss_probes: [0; packet::EPOCH_COUNT],
 
             cc: cc::new_congestion_control(config.cc_algorithm),
+
+            app_limited: false,
         }
     }
 
@@ -158,6 +162,9 @@ impl Recovery {
             if ack_eliciting {
                 self.time_of_last_sent_ack_eliciting_pkt[epoch] = Some(now);
             }
+
+            self.app_limited =
+                (self.cc.bytes_in_flight() + sent_bytes) < self.cc.cwnd();
 
             // OnPacketSentCC
             self.cc.on_packet_sent_cc(sent_bytes, trace_id);
@@ -470,6 +477,7 @@ impl Recovery {
                     &p,
                     self.rtt(),
                     self.min_rtt,
+                    self.app_limited,
                     trace_id,
                 );
             }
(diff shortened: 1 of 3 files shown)


src/recovery.rs
```diff
@@ -159,6 +163,10 @@ impl Recovery {
                 self.time_of_last_sent_ack_eliciting_pkt[epoch] = Some(now);
             }
 
+            // Set app_limited state
```
ghedo: This comment doesn't seem to be particularly useful.
```suggestion
```

PR #336, opened 2020-01-25
qlog: Add remaining EventTypes and helper functions

LPardue:
qlog: Add remaining EventTypes and helper functions
    
    Previously, the qlog data model was missing some
    event types (e.g. HTTP/3 and QPACK) and there were
    a few incorrect fields peppered throughout the code.
    
    While fixing up the code, I also spotted some of the
    burden on calling applications. I've attempted to
    improve life by adding an Event module that couples
    EventCategory, EventType and EventData. It provides
    methods to concisely construct valid events, and if
    people want to manually construct such events they
    can use an is_valid() method to catch obvious errors
    such as trying to log a TransportEventType with
    HTTP/3 EventData.
    
    Applications that use some or all of the helpers
    should find it much more difficult to produce bogus
    data.

cc: @hawkinsw @agrover @rmarx

### DIFF
--- a/tools/qlog/src/event.rs
+++ b/tools/qlog/src/event.rs
@@ -1,0 +1,1008 @@
+// Copyright (C) 2020, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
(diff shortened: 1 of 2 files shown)


tools/qlog/src/helpers.rs
```diff
@@ -0,0 +1,841 @@
...
+            },
+        }
+    }
+
+    /// Returns:
+    /// * `EventType`=`QpackEventType::HeadersEncoded`
+    /// * `EventData`=`QpackHeadersEncoded`.
+    pub fn qpack_headers_encoded(
+        stream_id: Option<String>, headers: Option<HttpHeader>,
+        block_prefix: QpackHeaderBlockPrefix,
```
hawkinsw: I am not a sophisticated rust programmer, but I cannot determine how a user of this function can build a `QpackHeaderBlockPrefix`. While the struct itself is public, none of the fields are. Therefore, it seems impossible to create such an object outside the package and pass it as a parameter here. The same seems to apply to `QpackDynamicTableEntry`. Please forgive me if this is something that any reasonable rust programmer should know and I am saying something stupid. 

No matter what, thank you for creating this PR. I am looking forward to all these helper methods :-)

LPardue: Hehe, this is not sophistication but rather a bug caused by not testing out the code from external modules.

We should make the fields public too, which will fix the problem

hawkinsw:
Whew! I'm glad I didn't miss something silly! Thanks for doing this!

LPardue:
Iterating on this PR a bit, so any reviews might get invalidated...

LPardue:
I've reworked things a bit in order to simplify some of the implementation code and reduce split responsibility. I don't plan any other changes other than to address review feedback

This is a breaking change to the API, sowwy.

The `event::Event` structure handles category, type and data - so there is no need to have separate categorized push functions. People that want an easy life should create a suitable `event::Event` with the functions provided and then submit that using the singular `push_event()` function. People that want a hard life can construct their own Vec<EventFields> and push them straight to Trace::events.

agrover:
:+1:

LPardue:
@ghedo I think we should merge this as is and I'll refine some aspects of it in a different PR

PR #340, opened 2020-01-29
Draft 25

ghedo:
This adds support for draft-25, while more or less keeping support for previous drafts.

Note that Retry is only supported with the latest draft (this was actually the case before as well, so :man_shrugging:).

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -243,7 +243,7 @@ int main(int argc, char *argv[]) {
     }
 
     quiche_config_set_application_protos(config,
-        (uint8_t *) "\x05hq-24\x05hq-23\x08http/0.9", 15);
+        (uint8_t *) "\x05hq-25\x05hq-24\x05hq-23\x08http/0.9", 15);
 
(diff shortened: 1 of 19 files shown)


junhochoi:
Do we still need to support h3-23 as well?

ghedo:
Added the idle_timeout changes as well. Also, apparently the fuzzers are not happy, so need to look into that.

ghedo:
> Do we still need to support h3-23?

That's what Firefox stable supports, and if we keep draft-24 we don't really need to do anything to keep draft-23 as well.

LPardue:
FWIW the change for `idle_timeout` also affects tools/http3_test/src/runner.rs and the `nginx-1.16.patch`

ghedo:
I patched the NGINX patch, and also update the NGINX alt-svc example configration.

ghedo:
And updated http3_test.

ghedo:
Aaaand fixed the panic found by the fuzzers.

LPardue:
:shipit:

PR #342, opened 2020-01-30
WIP: recovery: handling max_datagram_size

### LINKED ISSUE #298: max_packet_size is not used in recovery

junhochoi:
recovery: handling max_datagram_size

Currently quiche use MAX_DATAGRAM_SIZE(1452) as a fixed value.
This value is used for congestion window update, so getting a real
value is important for stable congestion control.

According to the draft, this value need to be set as a result of
PMTU discovery. However right now quiche and its server doesn't handle
PMTU and quiche library itself is not handling network sockets.
To support updating max_datagram_size, a new API is added.

  Rust: config.set_max_datagram_size(v)
  C: quiche_config_set_max_datagram_size(config, v)

Note that until Handshake completes, we will use 1200 as a
max_datagram_size. Also when peer transport parameter max_packet_size
is available (after Handshake completes), make sure
max_datagram_size is same or smaller than peer's
max_packet size.

For example, when server has max_datagram_size=1400
but client advertised max_packet_size=1300, it means client
may drop the packet size > 1300. For this case, server's
max_datagram_size will be updated to 1300 when handshake
completes.

Other changes:
- examples: use MAX_UDP_SIZE for maximum udp packet size received.
- cc: CongestionControlParams contains a default parameter
  when congestion control is initialized, such as initcwnd,
  minimum_window and max_datagram_size. Those values are
  updated when params.max_datagram_size() is called again.
- cc: set_max_datagram_size() to update max_datagram_size.
- cc: max_datagram_size() to get max_datagram_size.
- lib: defined MIN_PACKET_LEN (1200)
- lib: use self.recovery.cc.max_datagram_size() when getting
  max_pkt_len.

Fixes #298

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -44,6 +44,7 @@
 
 #define LOCAL_CONN_ID_LEN 16
 
+#define MAX_UDP_SIZE 65527
 #define MAX_DATAGRAM_SIZE 1350
 
 struct conn_io {
@@ -93,7 +94,7 @@ static void recv_cb(EV_P_ ev_io *w, int revents) {
 
     struct conn_io *conn_io = w->data;
 
-    static uint8_t buf[65535];
+    static uint8_t buf[MAX_UDP_SIZE];
 
     while (1) {
         ssize_t read = recv(conn_io->sock, buf, sizeof(buf), 0);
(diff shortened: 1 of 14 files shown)


junhochoi:
Ready for review.

junhochoi:
nginx patch need no change - it use MAX_DATAGRAM_SIZE = 1200 which is library's default.

junhochoi:
Ready for review again. I had some misunderstanding.

junhochoi:
Won't do

PR #343, opened 2020-01-30
build: make clippy happy

junhochoi:
Fix the following recent CI warning:
```
error: unneeded return statement
   --> src/build.rs:78:5

error: useless use of `format!`
  --> src/build.rs:63:16

error: `if _ { .. } else { .. }` is an expression
  --> src/build.rs:80:13
```


junhochoi:
Incorrect go version problem happened again. After removing cache files, the build went well.

PR #345, opened 2020-02-03
WIP: Implementing ack-frequency draft

junhochoi:
Implementing https://github.com/janaiyengar/ack-frequency

- ACK_FREQUENCY frame support (done)
- min_ack_delay TP parameter support (done)
- ACK strategy changes based on above.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -138,6 +138,13 @@ pub enum Frame {
     },
 
     HandshakeDone,
+
+    AckFrequency {
+        sequence_number: u64,
+        packet_tolerance: u64,
+        update_max_ack_delay: u64,
+        ignore_order: bool,
+    },
 }
 
 impl Frame {
@@ -255,6 +262,13 @@ impl Frame {
             },
 
             0x1e => Frame::HandshakeDone,
+
+            0xaf => Frame::AckFrequency {
+                sequence_number: b.get_varint()?,
+                packet_tolerance: b.get_varint()?,
+                update_max_ack_delay: b.get_varint()?,
+                ignore_order: b.get_u8()? != 0,
+            },
 
             _ => return Err(Error::InvalidFrame),
         };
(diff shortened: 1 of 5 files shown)


src/frame.rs
```diff
@@ -676,6 +704,19 @@ impl Frame {
...
+                sequence_number,
+                packet_tolerance,
+                update_max_ack_delay,
+                ignore_order,
+            } => {
+                1 + // frame type
+                octets::varint_len(*sequence_number) + // sequence_number
+                octets::varint_len(*packet_tolerance) + // packet_tolerance
+                octets::varint_len(*update_max_ack_delay) + // update_max_ack_delay
+                octets::varint_len(*ignore_order as u64)
```
LPardue: since ignrore_order is just a bool, you can probably just add 1 here

src/lib.rs
```diff
@@ -2953,6 +2974,40 @@ impl Connection {
                 // Once the handshake is confirmed, we can drop Handshake keys.
                 self.drop_epoch_state(packet::EPOCH_HANDSHAKE);
             },
+
+            frame::Frame::AckFrequency {
+                sequence_number,
+                packet_tolerance,
+                update_max_ack_delay,
+                ignore_order,
+            } => {
```
LPardue: QUIC extension frames require opt in, so we should blow up if the frames is received in a connection that does not support the extension. 

This is actually going to be a common pattern, so we can probably write a common extension checker and have a single unit test for extensions.

src/frame.rs
junhochoi: Fixed.

src/lib.rs
junhochoi: We can detect is using min_ack_delay value. I added a check to return InvalidFrame is min_ack_delay is not set. Supporting an extension need to be controlled via config APIs. I added Config.ext_set_min_ack_delay() for this extension -- by default it's not enabled. But it will be good to have some common stuff for extensions.

LPardue:
Any plans to update this PR to the latest draft @junhochoi ;)

junhochoi:
> Any plans to update this PR to the latest draft @junhochoi ;)

This is pretty old... I forgot to close it. If you have anyone to take it over please go ahead.

junhochoi:
Closing now. If anyone interested please pick it up.

PR #346, opened 2020-02-03
WIP: bundle an ACK frame every 4 packets

ghedo:
When sending a packet, we generally don't bundle ACK frames unless an
ack was elicited. This means that if a client has a large flow control
window we will not ack the client's acks for a long while. The client
will then keep sending bigger and bigger ACK frames, which end-up being
very expensive to process.

Note however that this will not trigger sending an ACK-only packet, just
bundling an ACK frame when other frames were already going to be sent.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1734,7 +1734,11 @@ impl Connection {
         let mut payload_len = 0;
 
         // Create ACK frame.
-        if self.pkt_num_spaces[epoch].ack_elicited && !is_closing {
+        if (self.pkt_num_spaces[epoch].ack_elicited ||
+            (!self.pkt_num_spaces[epoch].recv_pkt_need_ack.is_empty() &&
+                pn % 4 == 0)) &&
+            !is_closing
+        {
             let ack_delay =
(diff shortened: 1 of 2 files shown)


ghedo:
Even if we merge https://github.com/cloudflare/quiche/pull/347, we might want to still do this, though need to decide what the actual number of packets is.

VMatrix1900:
Is this duplicate with #345 ?

junhochoi:
> Is this duplicate with #345 ?

#345 is implementing ack-frequency draft. Related, but not directly.

VMatrix1900:
> > Is this duplicate with #345 ?
> 
> #345 is implementing ack-frequency draft. Related, but not directly.

The ack_ok function basically do the same thing as this pull request.

junhochoi:
> > > Is this duplicate with #345 ?
> > 
> > 
> > #345 is implementing ack-frequency draft. Related, but not directly.
> 
> The ack_ok function basically do the same thing as this pull request.

RIght.. But it's not complete. I will remove WIP when it's ready.

ghedo:
This is not going to be merged in the current form and it's also not urgent, so closing.

PR #347, opened 2020-02-03
recovery: do lowest unacked packet number check earlier

ghedo:
We already check if an acked packet number has already been acked by
comparing it to the lowest unacked packet number we hold. However this
means that we will still need to loop through all acked packets.

By doing the check earlier, we can break out of the loop without having
to check all following packet numbers. Given that we process acked
packet numbers from largest to lowest, we can assume that as soon as we
find an already-acked packet number all following numbers are also
already acked.


ghedo:
This was confirmed to fix the problem, so is ready for review.

PR #348, opened 2020-02-04
examples: deterministically generate source conn ID on servers

ghedo:
When we receive a ClientHello split across multiple Initial packets we
end up creating multiple connections in the example servers due to the
fact that we only store the server-generated destination connection ID,
and not the original client-generated destination connection ID, in the
HashMap used to associate incoming packets to connections.

This changes how we generate the server-generated destination connection
ID by making it more deterministic and based on the client-generated
connection ID, and then do a double lookup for both the raw dcid and the
deterministically-generated one.

Unfortunately due to the fact that we use the normal Rust HashMap we
can't have multiple keys pointing to the same connection, so we can't
store the client-generated connection ID directly. In practice we could
have a double layer HashMap, but then we'd run into problems when
removing connections because we might not have the original destination
connection ID (e.g. when we don't do Retry), so we would leak memory.

---

This is kind of an hack, so we will need to figure out a way to store multiple connection IDs (and be able to clear them later) as we will need that for migration anyway.

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -154,6 +154,10 @@ fn main() {
 
     let h3_config = quiche::h3::Config::new().unwrap();
 
+    let rng = SystemRandom::new();
+    let conn_id_seed =
+        ring::hmac::Key::generate(ring::hmac::HMAC_SHA256, &rng).unwrap();
+
     let mut clients = ClientMap::new();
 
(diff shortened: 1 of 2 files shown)


ghedo:
@junhochoi NGINX doesn't use connection IDs right now, so it's not affected.

PR #349, opened 2020-02-04
tls: use SSL_in_init() to check for handshake completion

ghedo:
In some cases (e.g. 0-RTT) BoringSSL will report the handshake as
completed by returning "1" from SSL_do_handshake(), even if the
handshake is not in fact complete.

So use "SSL_in_init()" to check for handshake completion.


ghedo:
I added a check to avoid parsing the peer's transport params multiple times.

PR #351, opened 2020-02-06
ignore frames on closed and collected streams

### LINKED ISSUE #350: interop-runner: server quicgo client quiche multiplexing test failing
It's a little early report because it requires #331 to reproduce, but reporting here because it seems consistent:

output.txt:
```
Running with server quicgo (martenseemann/quic-go-interop:latest) and client quiche (cloudflare/quiche-qns:latest)
Server: quicgo. Client: quiche. Running test case: multiplexing
Downloaded the wrong number of files. Got 1336, expected 1999.
```

client log.txt:

```
[2020-02-05T13:44:34.743837400Z INFO  quiche_apps] sending HTTP request "GET /uciduixidq\r\n"
 [2020-02-05T13:44:34.744768000Z ERROR quiche_client] recv failed: InvalidStreamState
 [2020-02-05T13:44:34.744795200Z INFO  quiche_apps] sending HTTP request "GET /uciduixidq\r\n"
```

ghedo:
Due to the fact that we have already dropped the state of the stream, we
cannot know whether the frame is valid or not. Before we would assume
that this is an error, but in some cases (e.g. spurious retransmissions
due to loss of ACK frames) it's possible to receive a valid frame on a
dropped stream.

Fixes #350.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2801,8 +2801,23 @@ impl Connection {
                     return Err(Error::InvalidStreamState);
                 }
 
-                // Get existing stream or create a new one.
-                let stream = self.get_or_create_stream(stream_id, false)?;
+                // Get existing stream or create a new one, but if the stream
+                // has already been closed and collected, ignore the frame.
+                //
+                // This can happen if e.g. an ACK frame is lost, and the peer
+                // retransmits another frame before it realizes that the stream
+                // is gone.
+                //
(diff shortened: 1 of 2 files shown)


junhochoi:
Now I get a successful test.

```
+--------+--------+
|        | quicgo |
+--------+--------+
| quiche |   M    |
|        |        |
|        |        |
+--------+--------+```

PR #352, opened 2020-02-10
Cargo: only depend on criterion when building benchmarks

ghedo:
This is a bit of a hack as it requires setting `--features bench` when
invoking `cargo bench`, but it allows us to avoid pulling in criterion
when building tests and examples, which reduces the number of
dependencies for those build in half.


ghedo:
The alternative would be to just remove these benchmarks which TBF haven't been particularly useful for a while :man_shrugging:

PR #353, opened 2020-02-10
Simplify examples

ghedo:
Mostly this allows us to drop some "expensive" (as in, slow to build) dependencies.

Along with https://github.com/cloudflare/quiche/pull/352 this cuts the time required to run `cargo test` on my laptop in half.

Before:
```
 % time cargo test
cargo test  400.72s user 29.41s system 584% cpu 1:13.54 total
```
After:
```
 % time cargo test
cargo test  177.36s user 18.62s system 516% cpu 37.933 total
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,6 @@ winapi = { version = "0.3", features = ["wincrypt"] }
 [dev-dependencies]
 mio = "0.6"
 url = "1"
-env_logger = "0.6"
 
 [profile.bench]
 debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,6 @@ winapi = { version = "0.3", features = ["wincrypt"] }
 [dev-dependencies]
 mio = "0.6"
 url = "1"
-docopt = "1"
 env_logger = "0.6"
 
 [profile.bench]
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -29,27 +29,11 @@ extern crate log;
 
 use std::net::ToSocketAddrs;
 
-use std::io::prelude::*;
-
 use ring::rand::*;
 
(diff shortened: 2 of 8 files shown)


LPardue:
Seems like win:win of simplifying our examples back to the purpose of showing quiche API, and improving build times. 

People that relied on the examples more complex features can use the new `quiche-client` and `quiche-server` but there there might be some pain in the transition period. It would be good if we could signpost this somehow, we can document stuff somewhere but I wonder would it be possible to detect a CLI error and print a message saying "you want to use tools/apps now"?

junhochoi:
I think `cargo test` doesn't build examples/*, so changes on examples/* won't improve test time?

Also, simplifying too much will make those examples less helpful for people who want to test. For example we usually ask people to show the output of `RUST_LOG=trace ./http3-client ...` to get a debugging logs for troubleshooting. If we remove all options and debugging logs from examples and suggest to use `quiche-{client,server}` for such purposes, `examples/*` will be forgotten soon because there is no reason to build/use it.

I'm ok with changes on src/cc/reno.rs (I didn't realize it's still there)

ghedo:
> I wonder would it be possible to detect a CLI error and print a message saying "you want to use tools/apps now"?

I can add something based on the number of arguments maybe.

> I think cargo test doesn't build examples/, so changes on examples/ won't improve test time?

It does. That's why there is a single `dev-dependencies` list which gets built when `cargo test` or `cargo build --examples` is run.

> suggest to use quiche-{client,server} for such purposes, examples/* will be forgotten soon because there is no reason to build/use it.

Well, the same applies to the documentation, it's up to us to keep them up to date. But also, examples are built on CI as per the previous point, so that helps.

ghedo:
> I can add something based on the number of arguments maybe.

Also, I'm planning to add some examples on how to use the apps to the README itself.

PR #358, opened 2020-02-14
apps: process requests ending with '/'

ghedo:
A requests ending in "/" will instead be processed as "/<index>", where
index is a file pass via the --index argument (by default "index.html").

This way we won't serve 404 for "/" requests (e.g. from browsers).

### DIFF
--- a/tools/apps/src/bin/quiche-server.rs
+++ b/tools/apps/src/bin/quiche-server.rs
@@ -48,6 +48,7 @@ Options:
   --cert <file>               TLS certificate path [default: src/bin/cert.crt]
   --key <file>                TLS certificate key path [default: src/bin/cert.key]
   --root <dir>                Root directory [default: src/bin/root/]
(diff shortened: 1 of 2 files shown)


ghedo:
Fixed clippy warning.

@LPardue well, yeah, but I don't think that's a bug though, as we can't really know what the path is supposed to be. The end result is that the client panics and the error is fairly clear, which IMO is good enough in that situation.
```
thread 'main' panicked at 'Bad times: Is a directory (os error 21)', src/lib.rs:185:23
```

LPardue:
I can live with that

ghedo:
FWIW, curl also errors out:
```
 % curl -O https://quic.tech:8443/
curl: Remote file name has no length!
```
Not sure if that's a better error message though...

ghedo:
Fixed clippy warning on the fix of the previous clippy warning :man_facepalming:

Issue #237, opened 2019-10-25
nginx patch breaks content-type header


xfalcox:
I added support for HTTP3 in the Discourse project here: https://github.com/discourse/discourse_docker/compare/http3

After deploying it to production and testing using `google-chrome-unstable --enable-quic --quic-version=h3-23 --origin-to-force-quic-on=example.com`, the requests for my HTML routes are missing the `content-type` header. Is this a know bug?

Testing with curl for HTTP2 has content-type:

```
curl -I https://forum.muito.ninja/latest
HTTP/2 200
...
content-type: text/html; charset=utf-8
....
```

Testing with quiche client doesn't have the content-type:

```
RUST_LOG=info target/debug/examples/http3-client https://forum.muito.ninja/latest

**Comment by junhochoi** [comment]
$ curl --http3 -vso/dev/null 'https://forum.muito.ninja/images/wizard/bubbles.png'
*   Trying 159.203.98.162:443...
* Sent QUIC client Initial, ALPN: h3-23
* h3 [:method: GET]
* h3 [:path: /images/wizard/bubbles.png]
* h3 [:scheme: https]
* h3 [:authority: forum.muito.ninja]
* h3 [user-agent: curl/7.67.0-DEV]
* h3 [accept: */*]
* Using HTTP/3 Stream ID: 0 (easy handle 0x7fde6700c000)
> GET /images/wizard/bubbles.png HTTP/3
> Host: forum.muito.ninja
> user-agent: curl/7.67.0-DEV
> accept: */*
>
< HTTP/3 200
< server: nginx
< date: Fri, 25 Oct 2019 22:48:01 GMT
< content-type: image/png
< content-length: 29108
< last-modified: Fri, 25 Oct 2019 15:55:32 GMT
< accept-ranges: bytes
<
{ [1032 bytes data]
* Connection #0 to host forum.muito.ninja left intact

**Comment by xmapst** [comment]
@junhochoi How did you get curl to support http3?

**Comment by junhochoi** [comment]
@xmapst https://github.com/curl/curl/blob/master/docs/HTTP3.md

**Comment by xmapst** [comment]
Thank you

**Comment by xfalcox** [comment]
Hey @junhochoi I just compiled `curl` with quiche HTTP3 enabled and if I do two requests:


```bash
curl --http3 -vso/dev/null 'https://forum.muito.ninja/latest'

curl  -vso/dev/null 'https://forum.muito.ninja/latest'
```

[...]

PR #362, opened 2020-02-15
nginx: vary header fix

hguandl:
https://github.com/cloudflare/quiche/commit/63a9096e4a90459f547eae90bdabc16ab39b1936 fixes an error that forgets to allocate new space for the `vary` header. This caused the `content-type` header to be overwritten by `vary`, which was reported in https://github.com/cloudflare/quiche/issues/237.

### DIFF
--- a/extras/nginx/nginx-1.16.patch
+++ b/extras/nginx/nginx-1.16.patch
@@ -100,7 +100,7 @@ new file mode 100644
 index 000000000..2c8995dd7
 --- /dev/null
 +++ b/auto/lib/quiche/conf
-@@ -0,0 +1,19 @@
+@@ -0,0 +1,23 @@
 +
 +# Copyright (C) Cloudflare, Inc.
 +
@@ -118,6 +118,10 @@ index 000000000..2c8995dd7
 +    CORE_INCS="$CORE_INCS $QUICHE/include"
 +    CORE_DEPS="$CORE_DEPS $QUICHE/target/$QUICHE_BUILD_TARGET/libquiche.a"
 +    CORE_LIBS="$CORE_LIBS $QUICHE/target/$QUICHE_BUILD_TARGET/libquiche.a $NGX_LIBPTHREAD"
++
++    if [ "$NGX_SYSTEM" = "Darwin" ]; then
++        CORE_LIBS+=" -framework Security"
++    fi
 +
 +fi
 diff --git a/auto/lib/quiche/make b/auto/lib/quiche/make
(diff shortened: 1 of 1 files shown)


junhochoi:
Thanks! Both patch looks good to me (FYI, I have homebrew formula https://github.com/cloudflare/homebrew-cloudflare/blob/master/nginx.rb and it has a correct link flag already. but it will be a good idea to include it into nginx patch).

However in this PR 2 commits are doing very different things, so I'd propose to split into 2 PRs for easier review and merge.

extras/nginx/nginx-1.16.patch
```diff
@@ -3338,6 +3342,7 @@ index 000000000..0877fb9a2
 +#if (NGX_HTTP_GZIP)
 +    /* Generate Vary header. */
 +    if (r->gzip_vary) {
++        h = ngx_array_push(headers);
```
ghedo: This needs to check the return value:
```suggestion
+        h = ngx_array_push(headers);
+        if (h == NULL) {
+            return NGX_ERROR;
+        }
+
```

extras/nginx/nginx-1.16.patch
```diff
@@ -1560,7 +1564,7 @@ new file mode 100644
 index 000000000..0877fb9a2
 --- /dev/null
 +++ b/src/http/v3/ngx_http_v3.c
-@@ -0,0 +1,2094 @@
+@@ -0,0 +1,2095 @@
```
ghedo:
```suggestion
@@ -0,0 +1,2099 @@
```

hguandl:
@junhochoi So how would be split this one into two PRs? Should I open another PR which only include https://github.com/cloudflare/quiche/commit/e1f9d135cae6f8aedaaebb31457b1234f115bc60 or the maintainers do so?

junhochoi:
> @junhochoi So how would be split this one into two PRs? Should I open another PR which only include [e1f9d13](https://github.com/cloudflare/quiche/commit/e1f9d135cae6f8aedaaebb31457b1234f115bc60) or the maintainers do so?

Right. Please open 1 PR for MacOS build and another PR for nginx header fix.

hguandl:
> Right. Please open 1 PR for MacOS build and another PR for nginx header fix.

@junhochoi Just finished that. Now MacOS build is on https://github.com/cloudflare/quiche/pull/367 and this PR focuses on header fix. I think it should work by merging https://github.com/cloudflare/quiche/pull/367 first and then this one.

ghedo:
Merged as https://github.com/cloudflare/quiche/commit/35a2f77199e3a4947dfd35973f73ea9473f6e307, thanks @hguandl!

PR #364, opened 2020-02-17
Handle partial writes of HTTP/3 HEADERS

ghedo:
See commit messages.

The first commit is actually a follow-up to https://github.com/cloudflare/quiche/commit/56fc60bfa2d92e3abfd6c06a47c191dcb439eac8. It fixes the fact that the flow control limit didn't actually decrease until a STREAM frame was sent on the wire, so the application could keep writing data without having the connecion flow control matter. It's not strictly necessary for the HTTP/3 fix, but it makes writing a test for that easier so I put them in the same PR.

### DIFF
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -36,6 +36,8 @@ use ring::rand::*;
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
 struct PartialResponse {
+    headers: Option<Vec<quiche::h3::Header>>,
+
     body: Vec<u8>,
 
     written: usize,
@@ -503,14 +505,28 @@ fn handle_request(
 
     let (headers, body) = build_response(root, headers);
 
-    if let Err(e) = http3_conn.send_response(conn, stream_id, &headers, false) {
-        error!("{} stream send failed {:?}", conn.trace_id(), e);
+    match http3_conn.send_response(conn, stream_id, &headers, false) {
+        Ok(v) => v,
+
+        Err(quiche::h3::Error::StreamBlocked) => {
+            let response = PartialResponse {
+                headers: Some(headers),
+                body,
+                written: 0,
+            };
+
+            client.partial_responses.insert(stream_id, response);
+            return;
+        },
+
+        Err(e) => {
+            error!("{} stream send failed {:?}", conn.trace_id(), e);
+            return;
+        },
     }
 
     let written = match http3_conn.send_body(conn, stream_id, &body, true) {
         Ok(v) => v,
-
-        Err(quiche::h3::Error::Done) => 0,
 
         Err(e) => {
             error!("{} stream send failed {:?}", conn.trace_id(), e);
@@ -519,7 +535,12 @@ fn handle_request(
     };
 
     if written < body.len() {
-        let response = PartialResponse { body, written };
+        let response = PartialResponse {
+            headers: None,
+            body,
+            written,
+        };
+
         client.partial_responses.insert(stream_id, response);
     }
 }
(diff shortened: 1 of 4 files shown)


examples/http3-server.rs
```diff
@@ -586,6 +607,24 @@ fn handle_writable(client: &mut Client, stream_id: u64) {
...
+            },
+
+            Err(e) => {
+                error!("{} stream send failed {:?}", conn.trace_id(), e);
+                return;
+            },
+        }
+    }
+
+    resp.headers = None;
```
LPardue: Does this mean the handling of blocked headers writes only work for one blocked iteration, or does the `return` prevent that?

ghedo: Yeah, the return means that next time the stream is writable it will try headers again.

src/h3/mod.rs
```diff
@@ -268,6 +268,8 @@ pub const APPLICATION_PROTOCOL: &[u8] = b"\x05h3-25\x05h3-24\x05h3-23";
 /// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
 pub type Result<T> = std::result::Result<T, Error>;
 
+const MAX_STREAM_GREASE_OVERHEAD: usize = 36;
```
LPardue: AFAICT this is use only in the case of sending GREASE frames, which can contain arbitrarily long data. It's just that the the value 36 is driven only by our choice to send 2 grease frames with one containing some payload of a specific length. Since we only do that once per connection in `send_grease_frames()`, can't we calculate the value there?

ghedo: I realized that we don't actually have to allocate the whole buffer on the stack, as we only need to write one varint at a time, so using a const is not really necessary (otherwise we would have had to allocate the buffer on the heap).

I now changed it to calculate the grease length dynamically, and reduced the size of the buffer.

src/h3/mod.rs
```diff
@@ -936,20 +978,42 @@ impl Connection {
...
+
+        let overhead = octets::varint_len(grease_frame1) + // frame type
+            1 + // payload len
+            octets::varint_len(grease_frame2) + // frame type
+            1 + // payload len
+            grease_payload.len(); // payload
+
+        // Don't send GREASE if there is not enough capacity for it.
+        if stream_cap < overhead {
+            return Ok(());
```
LPardue: the downside with returning Ok(() here is that the function returned to then immediately calls `self.frames_greased = true;` and will never try to grease again, so we'll miss greasing certain classes of client connection.

ghedo: Yeah, that was kind of on purpose, as a slightly more complicated handling of this doesn't seem worth it. I expanded the comment to make it more clear.

src/h3/mod.rs
```diff
@@ -642,19 +649,39 @@ impl Connection {
     /// a newly allocated stream.
     ///
     /// On success the newly allocated stream ID is returned.
+    ///
+    /// The [`StreamBlocked`] error is returned when the underlying QUIC stream
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application should retry the operation once the stream is
+    /// reported as writable again.
+    ///
+    /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
```
LPardue: When a request attempt gets StreamBlocked, retrying the the request will allocate a new stream ID. This wastes a stream but also, depending on the conditions that trigger the blockage, a client might burn through a load of streams that have been allocated in memory but never reused.

ghedo: Oh, good point, though AFAICT this is already the case (e.g. if the server doesn't give enough stream ID credits).

I added an additional commit to only increase the stream IDs after the relevant operation complete (this also fixes the overflow checks, though I doubt anyone would ever be able to hit the limit anyway).

src/h3/mod.rs
```diff
@@ -642,19 +650,47 @@ impl Connection {
...
+        conn.stream_send(stream_id, b"", false)?;
+
         self.send_headers(conn, stream_id, headers, fin)?;
 
+        // To avoid skipping stream IDs, we only calculate the next available
+        // stream ID when a request has been successfully buffered.
+        self.next_request_stream_id = self
+            .next_request_stream_id
+            .checked_add(4)
+            .ok_or(Error::IdError)?;
```
LPardue: oh this reminded me of the latent bug where effective maximum stream ID is not u64::MAX but rather 2^62-1. But the chances of either being hit are very small :)

ghedo: I guess the transport will end up returning an error in that case, so the client will keep failing to send a request, which is probably as good as it can get.

PR #365, opened 2020-02-17
Clean-up HTTP/3 errors

ghedo:


LPardue:
The errors are only unused in so far as we have yet to implement features that depend on them. Is the intention to add the errors back as we need them?

ghedo:
@LPardue yeah, we can add more errors later when we actually need them, but it's kind of confusing to have errors that we never use.

LPardue:
I guess the original intention was to allow applications to issue an application_close with something like

```
conn.close(true, quiche::h3::Error::ConnectError.to_wire(), b"")
```

Some errors will only be issued internally, but with Connect error this is only something the application using quiche can determine.

ghedo:
Noe that `to_wire()` is not public, so that's not actually possible right now. It would also require some more work to be able to call it from FFI.

I'm not really opposed to it if someone wants to do that work and re-add the errors as part of that, but until then it seems kind of pointless to keep these around.

LPardue:
heh, good point. This reminds me of some discussion about 1 year ago that ran similar. I can live with this change.

ghedo:
Rebased on master and fixed conflicts.

PR #366, opened 2020-02-17
README: add brief examples on how to use the quiche apps

ghedo:


ghedo:
I killed the CI builds on purpose as this is only a README change.

PR #23, opened 2019-02-08
Adjust CFLAGS on macos for examples

jaysonsantos:
This should fix #17

Issue #17, opened 2019-02-04
CPP examples fail to compile


VMatrix1900:
When I try to build the CPP version of client and server, `make all` returns error:
```
Undefined symbols for architecture x86_64:
  "_SecRandomCopyBytes", referenced from:
      ring::rand::darwin::fill::hf5f04717e928fa1c in libquiche.a(ring-b556caff1d96e9fb.ring.99s7mboi-cgu.7.rcgu.o)
  "_kSecRandomDefault", referenced from:
      ring::rand::darwin::fill::hf5f04717e928fa1c in libquiche.a(ring-b556caff1d96e9fb.ring.99s7mboi-cgu.7.rcgu.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
System: Mac OS 10.14.3

ghedo:
I don't have a mac so I don't really know what's going on here. Will try to find a way to reproduce.

ghedo:
Apparently we need to pass `-framework Security` to the compiler when on macOS, will need to figure out how to make the Makefile do that.

jaysonsantos:
This change made the trick here:

```diff
diff --git a/examples/Makefile b/examples/Makefile
index 9d19794..cfdef9a 100644
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,9 +1,15 @@
+OS := $(shell uname)
+
 SOURCE_DIR = ../src
 BUILD_DIR = ../target/debug
 INCLUDE_DIR = ../include
 
 CFLAGS = -I. -Wall -pedantic -fsanitize=address
 
+ifeq ($(OS), Darwin)
+CFLAGS += -framework Security
+endif
+
 LIBCRYPTO_DIR = $(dir $(shell find ${BUILD_DIR} -name libcrypto.a))
 LIBSSL_DIR = $(dir $(shell find ${BUILD_DIR} -name libssl.a))
 

```

VMatrix1900:
It's working. Thanks @jaysonsantos

ghedo:
Cool, thanks @jaysonsantos! Would you mind opening a pull request? So we can attribute the change properly.

jaysonsantos:
Sure, I will commit what I have locally

PR #367, opened 2020-02-18
nginx: MacOS build

hguandl:
As #17 and #23 discussed, the patched nginx needs to be linked with Apple's Security framework when compiling on macOS. https://github.com/cloudflare/quiche/commit/e1f9d135cae6f8aedaaebb31457b1234f115bc60 should fix that.


ghedo:
Merged, thanks @hguandl!

PR #372, opened 2020-02-24
WIP: path MTU discovery

ghedo:
This adds initial support for PMTUD. The PMTU starts at 1200 bytes
(minimum QUIC packet size), then 1452 is tried (maximum QUIC packet
size). If the probe fails we keep using 1200 and not attempt any
additional probe.

Due to the fact that the probe might be lost due to a lower path MTU, we
also need to ignore PMTUD probe losses for congestion control purposes,
otherwise we might decrease the cwnd even though there is no actual
congestion on the path.

The probing happens immediately after the handshake is completed (on the
server it is sent after HANDSHAKE_DONE, to avoid delaying the handshake
confirmation), so it might be too aggressive as it means the connection
incurs the cost of sending the additional packet even when the total
number of bytes is very small (e.g. potentially if the whole data would
have taken a single flight to send, it will now take 2).

For these reasons, PMTUD is disabled by default, and can be enabled with
a config flag. In the future we can experiment with smarter strategies.

---

I need to some additional testing, but this should be ready for review now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -422,9 +422,11 @@ pub struct Config {
 
     application_protos: Vec<Vec<u8>>,
 
+    cc_algorithm: cc::Algorithm,
+
     grease: bool,
 
-    cc_algorithm: cc::Algorithm,
+    pmtud: bool,
 }
 
 impl Config {
@@ -444,8 +446,9 @@ impl Config {
             version,
             tls_ctx,
             application_protos: Vec::new(),
-            grease: true,
             cc_algorithm: cc::Algorithm::Reno, // default cc algorithm
+            grease: true,
+            pmtud: false,
         })
     }
 
@@ -493,6 +496,11 @@ impl Config {
     /// The default value is `true`.
     pub fn grease(&mut self, grease: bool) {
         self.grease = grease;
+    }
+
+    /// Configures whether to do path MTU discovery.
+    pub fn discover_pmtu(&mut self, discover: bool) {
+        self.pmtud = discover;
     }
 
     /// Enables logging of secrets.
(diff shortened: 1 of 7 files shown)


junhochoi:
A couple of thoughts:

- C FFI for discover_pmtu() is needed too?
- I recommend to try 1350 (not 1452) first and fall back to 1200: For example Google uses 1350 as a default value - See sectino 7.1 of https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46403.pdf
- When PMTU is updated by a new PMTU discovery, congestion control need to be notified for ths change of max_datagram_size to tune cwnd changes properly. e.g. let's say we start with initcwnd 10 x 1452, when we found that actual MTU is 1200 later, then it's better to decrease cwnd value to 10 x 1200 (or de-scale properly) to make sure we don't oversend the data -- most likely at this point we already have sent 10 packets, so this cwnd decrease will make sure we won't over-send the data from this point. Also max_datagram_size is used for other places (e.g. collapse_cwnd()).

src/lib.rs
```diff
@@ -1734,9 +1749,51 @@ impl Connection {
...
+        // In addition, the PMTUD probe is only generated when the handshake is
+        // confirmed, to avoid interfering with the handshake (e.g. due to the
+        // anti-amplification limits).
+        if pkt_type == packet::Type::Short &&
+            (self.handshake_confirmed || self.handshake_done_sent) &&
+            self.pmtu_next > self.pmtu &&
+            self.recovery.cwnd_available() > self.pmtu_next &&
+            out_len > self.pmtu_next
+        {
+            // Add PADDING up to the PMTUD probe size (accounting for the QUIC
```
LPardue: Would it be useful to add some trace-level logging to indicate that a probe is being sent out?

ghedo:
This seems to be making transfers slower, so marking as WIP until I figure out why.

ghedo:
Superseded by https://github.com/cloudflare/quiche/pull/1635

PR #373, opened 2020-02-25
Don't decrease tx_data when retransmit

junhochoi:
tx_data is "The sum of the largest received offsets on all streams"
so it need to be increased when we send a new data. If we decrease
tx_data while retransmitting it will result sending more data,
so the receiver will get more data than reported in MAX_DATA frame,
causing FLOW_CONTROL_ERROR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1660,11 +1660,6 @@ impl Connection {
                         None => continue,
                     };
 
-                    // TODO: due to a packet loss edge case the following could
-                    // go negative, though it's not clear why, so will need to
-                    // figure it out.
-                    self.tx_data = self.tx_data.saturating_sub(data.len() as u64);
-
                     let was_flushable = stream.is_flushable();
 
                     let empty_fin = data.is_empty() && data.fin();


junhochoi:
To reproduce, I use the following condition:

1) sender packets are lost and retransmitted (to trigger tx_data decrease)
2) use same config value as nginx patch uses. e.g.
```
diff --git a/tools/apps/src/bin/quiche-server.rs b/tools/apps/src/bin/quiche-server.rs
index 4ec3f31..eb32fbd 100644
--- a/tools/apps/src/bin/quiche-server.rs
+++ b/tools/apps/src/bin/quiche-server.rs
@@ -37,7 +37,7 @@ use ring::rand::*;

 use quiche_apps::*;

-const MAX_DATAGRAM_SIZE: usize = 1350;
+const MAX_DATAGRAM_SIZE: usize = 1200;

 const USAGE: &str = "Usage:
   quiche-server [options]
@@ -50,10 +50,10 @@ Options:
   --root <dir>                Root directory [default: src/bin/root/]
   --index <name>              The file that will be used as index [default: index.html].
   --name <str>                Name of the server [default: quic.tech]
-  --max-data BYTES            Connection-wide flow control limit [default: 10000000].
-  --max-stream-data BYTES     Per-stream flow control limit [default: 1000000].
-  --max-streams-bidi STREAMS  Number of allowed concurrent streams [default: 100].
-  --max-streams-uni STREAMS   Number of allowed concurrent streams [default: 100].
+  --max-data BYTES            Connection-wide flow control limit [default: 10485760].
+  --max-stream-data BYTES     Per-stream flow control limit [default: 1048576].
+  --max-streams-bidi STREAMS  Number of allowed concurrent streams [default: 128].
+  --max-streams-uni STREAMS   Number of allowed concurrent streams [default: 3].
   --dump-packets PATH         Dump the incoming packets as files in the given directory.
   --early-data                Enables receiving early data.
   --no-retry                  Disable stateless retry.
```
3) using curl is easier way to reproduce.
4) server and client(curl) on localhost

not 100% but failing more than 50% without this patch.

LPardue:
In the diff above, you're modifying several of the server's receiving flow control settings. I thought the problem was observed at the client? It seems strange that these values would cause a client issue, so maybe it can be simplified down to the MAX_DATAGRAM_SIZE change?

junhochoi:
> In the diff above, you're modifying several of the server's receiving flow control settings. I thought the problem was observed at the client? It seems strange that these values would cause a client issue, so maybe it can be simplified down to the MAX_DATAGRAM_SIZE change?

Yes changing MAX_DATAGRAM_SIZE was enough to reproduce -- I updated the diff.

ghedo:
Nice catch. Before https://github.com/cloudflare/quiche/commit/56fc60bfa2d92e3abfd6c06a47c191dcb439eac8 this was needed because `tx_data` was increased when sending `STREAM` frames, so we had to decrease it on loss to avoid increasing it twice when retransmitting. But now we only increase `tx_data` when data is received by the application so don't need to do this anymore.

ghedo:
I tweaked the commit message with the information above and merged, thanks @junhochoi.

ghedo:
Actually, the bug was probably introduced by https://github.com/cloudflare/quiche/commit/7ccf77879f93c6be660ba051c758deae4c9046ee which is a follow-up to https://github.com/cloudflare/quiche/commit/56fc60bfa2d92e3abfd6c06a47c191dcb439eac8.

PR #379, opened 2020-02-26
add optional qlogging to quiche

LPardue:
This is still WIP but ready for review. I need to add tests for the new capability but will await design feedback.

The PR contains 2 commits, the older one uses a buffered serialization approach, the newer one replaces this with a streaming serialized approach. I left these so that we could compare the two, but we'd definitely squash this down before any merge.

To test this out, you can run quiche-client or quiche-server with the `QLOGDIR` environment pointing at a folder. The resulting trace will be written to a file such as `client-{scid}.qlog` which can then be imported to the visualization tool https://qvis.edm.uhasselt.be.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ libc = "0.2"
 ring = "0.16"
 criterion = { version = "0.3", optional = true }
 lazy_static = "1"
+qlog = { version = "0.2", optional = true }
 
 [target."cfg(windows)".dependencies]
 winapi = { version = "0.3", features = ["wincrypt"] }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -29,6 +29,9 @@ use std::ptr;
 use std::slice;
 use std::sync::atomic;
 
+#[cfg(all(unix, feature = "qlog"))]
+use std::os::unix::io::FromRawFd;
+
 use libc::c_char;
 use libc::c_int;
 use libc::c_void;
@@ -229,6 +232,25 @@ pub extern fn quiche_config_set_cc_algorithm(
     config: &mut Config, algo: CongestionControlAlgorithm,
 ) {
     config.set_cc_algorithm(algo);
+}
+
+#[no_mangle]
+#[cfg(all(unix, feature = "qlog"))]
+pub extern fn quiche_conn_set_qlog_fd(
+    conn: &mut Connection, fd: c_int, log_title: *const c_char,
+    log_desc: *const c_char,
+) {
+    let f = unsafe { std::fs::File::from_raw_fd(fd) };
+    let writer = std::io::BufWriter::new(f);
+
+    let title = unsafe { ffi::CStr::from_ptr(log_title).to_str().unwrap() };
+    let description = unsafe { ffi::CStr::from_ptr(log_desc).to_str().unwrap() };
+
+    conn.set_qlog(
+        std::boxed::Box::new(writer),
+        title.to_string(),
+        format!("{} id={}", description, conn.trace_id),
+    );
 }
 
 #[no_mangle]
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -689,6 +689,154 @@ impl Frame {
             _ => true,
         }
     }
+
+    #[cfg(feature = "qlog")]
+    pub fn to_qlog(&self) -> qlog::QuicFrame {
+        match self {
+            Frame::Padding { .. } => qlog::QuicFrame::padding(),
+
+            Frame::Ping { .. } => qlog::QuicFrame::ping(),
+
+            Frame::ACK { ack_delay, ranges } => {
+                let ack_ranges =
+                    ranges.iter().map(|r| (r.start, r.end - 1)).collect();
+                qlog::QuicFrame::ack(
+                    Some(ack_delay.to_string()),
+                    Some(ack_ranges),
+                    None,
+                    None,
+                    None,
+                )
+            },
+
+            Frame::ResetStream {
+                stream_id,
+                error_code,
+                final_size,
+            } => qlog::QuicFrame::reset_stream(
+                stream_id.to_string(),
+                *error_code,
+                final_size.to_string(),
+            ),
+
+            Frame::StopSending {
+                stream_id,
+                error_code,
+            } =>
+                qlog::QuicFrame::stop_sending(stream_id.to_string(), *error_code),
+
+            Frame::Crypto { data } => qlog::QuicFrame::crypto(
+                data.off().to_string(),
+                data.len().to_string(),
+            ),
+
+            Frame::NewToken { token } => qlog::QuicFrame::new_token(
+                token.len().to_string(),
+                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                    .to_string(),
+            ),
+
+            Frame::Stream { stream_id, data } => qlog::QuicFrame::stream(
+                stream_id.to_string(),
+                data.off().to_string(),
+                data.len().to_string(),
+                data.fin(),
+                None,
+            ),
+
+            Frame::MaxData { max } => qlog::QuicFrame::max_data(max.to_string()),
+
+            Frame::MaxStreamData { stream_id, max } =>
+                qlog::QuicFrame::max_stream_data(
+                    stream_id.to_string(),
+                    max.to_string(),
+                ),
+
+            Frame::MaxStreamsBidi { max } => qlog::QuicFrame::max_streams(
+                qlog::StreamType::Bidirectional,
+                max.to_string(),
+            ),
+
+            Frame::MaxStreamsUni { max } => qlog::QuicFrame::max_streams(
+                qlog::StreamType::Unidirectional,
+                max.to_string(),
+            ),
+
+            Frame::DataBlocked { limit } =>
+                qlog::QuicFrame::data_blocked(limit.to_string()),
+
+            Frame::StreamDataBlocked { stream_id, limit } =>
+                qlog::QuicFrame::stream_data_blocked(
+                    stream_id.to_string(),
+                    limit.to_string(),
+                ),
+
+            Frame::StreamsBlockedBidi { limit } =>
+                qlog::QuicFrame::streams_blocked(
+                    qlog::StreamType::Bidirectional,
+                    limit.to_string(),
+                ),
+
+            Frame::StreamsBlockedUni { limit } =>
+                qlog::QuicFrame::streams_blocked(
+                    qlog::StreamType::Unidirectional,
+                    limit.to_string(),
+                ),
+
+            Frame::NewConnectionId {
+                seq_num,
+                retire_prior_to,
+                conn_id,
+                ..
+            } => qlog::QuicFrame::new_connection_id(
+                seq_num.to_string(),
+                retire_prior_to.to_string(),
+                conn_id.len() as u64,
+                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                    .to_string(),
+                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                    .to_string(),
+            ),
+
+            Frame::RetireConnectionId { seq_num } =>
+                qlog::QuicFrame::retire_connection_id(seq_num.to_string()),
+
+            Frame::PathChallenge { .. } => qlog::QuicFrame::path_challenge(Some(
+                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                    .to_string(),
+            )),
+
+            Frame::PathResponse { .. } => qlog::QuicFrame::path_response(Some(
+                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                    .to_string(),
+            )),
+
+            Frame::ConnectionClose {
+                error_code, reason, ..
+            } => qlog::QuicFrame::connection_close(
+                qlog::ErrorSpace::TransportError,
+                *error_code,
+                *error_code,
+                String::from_utf8(reason.clone()).unwrap(),
+                Some(
+                    "TODO: https://github.com/quiclog/internet-drafts/issues/36"
+                        .to_string(),
+                ),
+            ),
+
+            Frame::ApplicationClose { error_code, reason } =>
+                qlog::QuicFrame::connection_close(
+                    qlog::ErrorSpace::ApplicationError,
+                    *error_code,
+                    *error_code,
+                    String::from_utf8(reason.clone()).unwrap(),
+                    None, /* Application variant of the frame has no trigger
+                           * frame type */
+                ),
+
+            _ => qlog::QuicFrame::unknown(std::u64::MAX),
+        }
+    }
 }
 
 impl std::fmt::Debug for Frame {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,6 +271,9 @@ use std::time;
 use std::pin::Pin;
 use std::str::FromStr;
 
+#[cfg(feature = "qlog")]
+use qlog::event::Event;
+
 /// The current QUIC wire version.
 pub const PROTOCOL_VERSION: u32 = PROTOCOL_VERSION_DRAFT27;
 
@@ -850,6 +853,14 @@ pub struct Connection {
 
     /// Whether to send GREASE.
     grease: bool,
+
+    /// Qlog streaming output.
+    #[cfg(feature = "qlog")]
+    qlog_streamer: Option<qlog::QlogStreamer>,
+
+    /// Whether peer transport parameters were qlogged.
+    #[cfg(feature = "qlog")]
+    qlogged_peer_params: bool,
 }
 
 /// Creates a new server-side connection.
@@ -1029,6 +1040,21 @@ macro_rules! push_frame_to_pkt {
     }};
 }
 
+/// Conditional qlog action.
+///
+/// Executes the provided body if the qlog feature is enabled and quiche
+/// has been condifigured with a log writer.
+macro_rules! qlog_with {
+    ($qlog_streamer:expr, $qlog_streamer_ref:ident, $body:block) => {{
+        #[cfg(feature = "qlog")]
+        {
+            if let Some($qlog_streamer_ref) = &mut $qlog_streamer {
+                $body
+            }
+        }
+    }};
+}
+
 impl Connection {
     fn new(
         scid: &[u8], odcid: Option<&[u8]>, config: &mut Config, is_server: bool,
@@ -1126,6 +1152,12 @@ impl Connection {
             closed: false,
 
             grease: config.grease,
+
+            #[cfg(feature = "qlog")]
+            qlog_streamer: None,
+
+            #[cfg(feature = "qlog")]
+            qlogged_peer_params: false,
         });
 
         if let Some(odcid) = odcid {
@@ -1155,6 +1187,60 @@ impl Connection {
         }
 
         Ok(conn)
+    }
+
+    /// Sets qlog output to the designated [`Writer`].
+    ///
+    /// [`Writer`]: https://doc.rust-lang.org/std/io/trait.Write.html
+    #[cfg(feature = "qlog")]
+    pub fn set_qlog(
+        &mut self, writer: Box<dyn std::io::Write>, title: String,
+        description: String,
+    ) {
+        let vp = if self.is_server {
+            qlog::VantagePointType::Server
+        } else {
+            qlog::VantagePointType::Client
+        };
+
+        let trace = qlog::Trace::new(
+            qlog::VantagePoint {
+                name: None,
+                ty: vp,
+                flow: None,
+            },
+            Some(title.to_string()),
+            Some(description.to_string()),
+            Some(qlog::Configuration {
+                time_offset: Some("0".to_string()),
+                time_units: Some(qlog::TimeUnits::Ms),
+                original_uris: None,
+            }),
+            None,
+        );
+
+        let mut streamer = qlog::QlogStreamer::new(
+            qlog::QLOG_VERSION.to_string(),
+            Some(title),
+            Some(description),
+            None,
+            std::time::Instant::now(),
+            trace,
+            writer,
+        );
+
+        streamer.start_log().ok();
+
+        let ev = self.peer_transport_params.to_qlog(
+            qlog::TransportOwner::Local,
+            self.version,
+            self.handshake.alpn_protocol(),
+            self.handshake.cipher(),
+        );
+
+        streamer.add_event(ev).ok();
+
+        self.qlog_streamer = Some(streamer);
     }
 
     /// Processes QUIC packets received from the peer.
@@ -1501,6 +1587,30 @@ impl Connection {
             pn
         );
 
+        qlog_with!(self.qlog_streamer, q, {
+            let packet_size = b.len();
+
+            let qlog_pkt_hdr = qlog::PacketHeader::with_type(
+                hdr.ty.to_qlog(),
+                pn,
+                Some(packet_size as u64),
+                Some(payload_len as u64),
+                Some(hdr.version),
+                Some(&hdr.scid),
+                Some(&hdr.dcid),
+            );
+
+            q.add_event(Event::packet_received(
+                hdr.ty.to_qlog(),
+                qlog_pkt_hdr,
+                Some(Vec::new()),
+                None,
+                None,
+                None,
+            ))
+            .ok();
+        });
+
         let mut payload =
             packet::decrypt_pkt(&mut b, pn, pn_len, payload_len, &aead).map_err(
                 |e| drop_pkt_on_err(e, self.recv_count, &self.trace_id),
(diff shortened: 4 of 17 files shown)


src/lib.rs
```diff
@@ -2724,6 +2961,79 @@ impl Connection {
         Ok(())
     }
 
+    #[cfg(feature = "qlogging")]
+    fn qlog_recovery_metrics_updated(&mut self) {
```
ghedo: It seems in general these methods return the event, and then the caller writes them to the writer, so maybe this could be a `to_qlog()` or whatever method of `recovery::Recovery` that returns the event, and then we can use the `qlog_with!()` to avoid a lot of duplication.

(yes, this would require modifying the recovery module, but not the actual recovery code which was the main annoyance with the previous version, as this method would be self-contained and isolated from the rest of the code, also, this would allow accessing some additional internal values without exposing them outside of the module).

src/lib.rs
```diff
@@ -2724,6 +2961,79 @@ impl Connection {
...
+                None, // pacing_rate
+            );
+
+            qlog_streamer.write_event(ev).ok();
+        }
+    }
+
+    /// Creates a qlog event for connection transport parameters and TLS fields
+    #[cfg(feature = "qlogging")]
+    pub fn transport_params_to_qlog(&self, local: bool) -> qlog::event::Event {
```
ghedo: This shouldn't be public.

Cargo.toml
```diff
@@ -50,6 +50,11 @@ ndk-old-gcc = []
 # optional, so we don't have to build them for other targets.
 bench = ["criterion"]
 
+# Conditional compilation of qlog support into quiche in order to make
+# dependencies optional. Applications that enable the feature still need to
+# explicitly enable logging via quiche configuration.
+qlogging = ["qlog"]
```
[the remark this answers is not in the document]
LPardue: I started with that but Rust complained that we can't have a feature (`qlog`) that is the same name as the crate (`qlog`).

src/frame.rs
```diff
@@ -689,6 +689,164 @@ impl Frame {
...
+    #[cfg(feature = "qlogging")]
+    pub fn to_qlog(&self) -> qlog::QuicFrame {
+        match self {
+            Frame::Padding { .. } => qlog::QuicFrame::padding(),
+
+            Frame::Ping { .. } => qlog::QuicFrame::ping(),
+
+            Frame::ACK { ack_delay, ranges } => {
+                // TODO: improve formatting acked ranges
+                let mut ack_ranges = Vec::new();
```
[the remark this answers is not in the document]
LPardue: sounds good, i'll need to remind myself of the exact qlog format and test your suggestion satisfies that (which I think it will)

src/lib.rs
```diff
@@ -818,6 +835,49 @@ pub struct Connection {
...
+    qlog_streamer: Option<qlog::QlogStreamer>,
+
+    /// Whether peer transport parameters were qlogged.
+    #[cfg(feature = "qlogging")]
+    qlogged_peer_params: bool,
+}
+
+/// Quiche connection event tracing.
+#[cfg(feature = "qlogging")]
+pub struct ConnectionQlog {
```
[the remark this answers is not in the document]
LPardue: Hiding the structure makes sense, `config.set_x` methods tend to be lightweight but if you're happy with several params I'll do that.

Setting the time is a hangover from the old design. I think the API now is incoherent so lets go with your suggestion and think later if we want to allow apps to start logging before quiche does.

src/lib.rs
```diff
@@ -1122,6 +1188,63 @@ impl Connection {
             conn.derived_initial_secrets = true;
         }
 
+        #[cfg(feature = "qlogging")]
+        {
+            // Once the connection is created, create a qlog objects if
+            // instructed. This takes ownership from Config and puts
+            // it into Connection for subsequent usage.
+
+            if let Some(log) = config.qlog.take() {
```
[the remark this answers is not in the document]
LPardue: I think my motivation was to use config to cheat at adding an optional thing without affecting existing method signtures. But you're right, it works but is ugly and probably fragile.

Cargo.toml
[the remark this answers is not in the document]
ghedo: I think you can just not add an explicit feature to `Cargo.toml` and simply mark the qlog dependency as optional, and that should work automatically. The `qlog` feature should be created automatically.

src/lib.rs
```diff
@@ -1488,13 +1638,43 @@ impl Connection {
...
+                }
+
+                // only log the remote transport parameters once the connection is
+                // established (i.e. after frames have been fully parsed) and only
+                // once per connection
+                if self.is_established() && !self.qlogged_peer_params {
+                    let ev = self.transport_params_to_qlog(false);
+
+                    if let Some(qlog_streamer) = self.qlog_streamer.as_mut() {
+                        qlog_streamer.write_event(ev).unwrap();
```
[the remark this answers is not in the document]
LPardue: .ok()

src/lib.rs
```diff
@@ -2536,6 +2756,13 @@ impl Connection {
             if draining_timer <= now {
                 trace!("{} draining timeout expired", self.trace_id);
 
+                #[cfg(feature = "qlogging")]
+                {
+                    if let Some(qlog_streamer) = self.qlog_streamer.as_mut() {
```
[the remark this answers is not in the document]
LPardue: yeah this is a great idea

src/ranges.rs
```diff
@@ -244,7 +244,8 @@ mod tests {
     fn insert_non_overlapping() {
         let mut r = RangeSet::default();
         assert_eq!(r.inner.len(), 0);
-        assert_eq!(&r.flatten().collect::<Vec<u64>>(), &[]);
```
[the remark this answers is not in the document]
LPardue: I don't know!!! (but in seriousness I agree, so let's see if I can bisect the root cause)

src/lib.rs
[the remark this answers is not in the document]
ghedo: I just realized there was an error in the macro code, should be fixed now.

tools/apps/src/bin/quiche-client.rs
```diff
@@ -139,6 +139,21 @@ fn main() {
     let mut conn =
         quiche::connect(connect_url.domain(), &scid, &mut config).unwrap();
 
+    // Only bother with qlog if the user specified it.
+    #[cfg(feature = "qlog")]
```
ghedo: FTR, https://github.com/rust-lang/rust/pull/69201 will be in Rust 1.43, so we can probably simplify this a bit in the future.

tools/apps/src/bin/quiche-server.rs
```diff
@@ -287,7 +287,23 @@ fn main() {
                     hex_dump(&scid)
                 );
 
-                let conn = quiche::accept(&scid, odcid, &mut config).unwrap();
+                let conn = if cfg!(feature = "qlog") && std::env::var_os("QLOGDIR").is_some() {
```
ghedo: Why not just do the `if` and `set_qlog()` after the connection is created like the client does? So we wouldn't suplicate the `accept()` call.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+    pub fn log_finish(&mut self) -> Result<()> {
+        if !self.log_started || self.log_finished {
+            return Err(Error::InvalidState);
+        }
+
+        self.writer.as_mut().write_all(b"]}]}")?;
+
+        self.log_finished = true;
+
+        self.writer.as_mut().flush()?;
```
ghedo: Seems like this could impact performance, do we really need to do this?

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+    /// Writes a JSON-serialized `EventField`s.
+    ///
+    /// Some qlog events can contain `QuicFrames`. If this is detected `true` is
+    /// returned and the streamer enters a frame-serialization mode that is only
+    /// concluded by `finish_frames()`. In this mode, attempts to log additional
+    /// events are ignored.
+    ///
+    /// If the event contains no array of `QuicFrames` return `false`.
+    pub fn write_event(&mut self, event: event::Event) -> Result<bool> {
+        if !self.log_started ||
```
ghedo: Maybe just have a `State` enum, instead of defining a different flag for each state.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+            self.log_finished ||
+            self.writing_event ||
+            self.writing_frames
+        {
+            return Err(Error::InvalidState);
+        }
+
+        let event_time = self.start_time.elapsed();
+
+        if let Some(t) = self.qlog.traces.get(0) {
```
ghedo: AFAICT this is always true, as the traces Vec will always have at least one element, so I think you can just do:
```rust
let t = self.qlog.traces[0];
```
and then avoid the additional block and indentation level.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+                let out = format!(
+                    "{}[{},{},{},{}{}",
+                    maybe_comma, ev_time, ev_cat, ev_ty, ev_data, maybe_terminate
+                );
+
+                self.writer.as_mut().write_all(out.as_bytes())?;
+                self.writing_event = contains_frames;
+
+                return Ok(contains_frames);
+            } else {
```
ghedo: No need for the `else` here,, can jut do the `return` outside of the block.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+                };
+
+            let maybe_comma = if self.first_event {
+                self.first_event = false;
+                ""
+            } else {
+                ","
+            };
+
+            let maybe_terminate = if contains_frames { "" } else { "]" };
```
ghedo: Related to the other comment about `Serialize`, if we manage to use `serialize_seq()` for this we would remove quite a bit of code here I think.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+                    out.insert(0, ',');
+                } else {
+                    self.first_frame = false;
+                }
+
+                self.writing_frames = true;
+
+                self.writer.as_mut().write_all(out.as_bytes())?;
+
+                if last {
```
ghedo: Looking at the quiche code, this is never set to `true` so do we really need it?

ghedo: Though if needed, should maybe just call `finish_frames()` to avoid duplicating the code.

src/ffi.rs
```diff
@@ -231,6 +234,26 @@ pub extern fn quiche_config_set_cc_algorithm(
     config.set_cc_algorithm(algo);
 }
 
+#[allow(unused_variables)]
```
ghedo: Why is this needed?

src/ffi.rs
```diff
@@ -231,6 +234,26 @@ pub extern fn quiche_config_set_cc_algorithm(
...
+#[no_mangle]
+#[cfg(all(unix, feature = "qlog"))]
+pub extern fn quiche_conn_set_qlog_fd(conn: &mut Connection, fd: c_int) -> c_int {
+    let f = unsafe { std::fs::File::from_raw_fd(fd) };
+    let writer = Some(std::io::BufWriter::new(f));
```
ghedo: Why is this wrapped in an `Option`? The `if` condition below will always be true.

src/lib.rs
```diff
@@ -1125,6 +1157,60 @@ impl Connection {
         Ok(conn)
     }
 
+    #[cfg(feature = "qlog")]
```
ghedo: Nit, this should be after the comment.

src/lib.rs
```diff
@@ -1488,13 +1598,47 @@ impl Connection {
...
 
             self.process_frame(frame, epoch, now)?;
         }
 
+        qlog_with!(self.qlog_streamer, q, {
+            // always conclude frame writing
+            q.finish_frames().ok();
+        });
+
+        // only log the remote transport parameters once the connection is
```
```suggestion
        // Only log the remote transport parameters once the connection is
```

src/lib.rs
```diff
@@ -1488,13 +1598,47 @@ impl Connection {
...
+        // established (i.e. after frames have been fully parsed) and only
+        // once per connection
```
```suggestion
        // once per connection.
```

src/lib.rs
```diff
@@ -1488,13 +1598,47 @@ impl Connection {
...
+        if self.is_established() {
+            qlog_with!(self.qlog_streamer, q, {
+                // always conclude frame writing
```
```suggestion
                // Always conclude frame writing.
```

src/recovery.rs
```diff
@@ -672,6 +675,25 @@ impl Recovery {
             self.app_limited_at_pkt = if limited > 0 { limited } else { 1 };
         }
     }
+
+    #[cfg(feature = "qlog")]
+    pub fn to_qlog(&self) -> qlog::event::Event {
+        // qvis can't use all these fields and they can be large
```
ghedo:
```suggestion
        // QVis can't use all these fields and they can be large.
```

tools/apps/Cargo.toml
```diff
@@ -5,18 +5,15 @@ authors = ["Lucas Pardue <lucaspardue.24.7@gmail.com>"]
 edition = "2018"
 publish = false
 
-[features]
```
ghedo: This should be kept.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+        }
+
+        self.writer.as_mut().write_all(b"]}]")?;
+        self.writing_frames = false;
+        self.writing_event = false;
+
+        Ok(())
+    }
+
+    /// Return the writer.
```
ghedo:
```suggestion
    /// Returns the writer.
```

tools/qlog/src/lib.rs
```diff
@@ -1027,6 +1325,24 @@ pub enum EventData {
     },
 }
 
+impl EventData {
+    /// Detect if `EventData` contains an array of `QuicFrame`s.
```
ghedo:
```suggestion
    /// Detects if `EventData` contains an array of `QuicFrame`s.
```

tools/apps/Cargo.toml
LPardue: yeah I fat fingered it soz

tools/apps/src/bin/quiche-server.rs
LPardue: so I tried that first but Clippy gets itself tied in knots, if qlog is not enabled, `conn` doesn't need to be `mut`. If it's not ` mut` and qlog is enabled then `set_qlog` will fail. We could disable the cliipy warning in this case?

ghedo: Yeah, silencing should be fine in this case.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+
+        // A qlog contains a trace holding a vector of events that we want to
+        // serialize in a streaming manner. So at the start of serialization,
+        // take off all closing delimiters, and leave us in a state to accept
+        // new events.
+        // TODO: this unfortunately removes any events already existing in the
+        // trace, which might be undesirable.
+        match serde_json::to_string(&self.qlog) {
+            Ok(mut out) => {
+                if let Some(index) = out.find("\"events\":[") {
```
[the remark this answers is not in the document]
LPardue: So far with these changes, they allow both a buffered and a streaming mode of using the qlog crate. Modifying Serialize risks breaking that. Furthermore, the generated serialization code for `Trace` is 150 lines of code, maintaining a manual equivalent code is a bit brittle.

Reverse-scanning is going to be an improvement. Additionally, if it scans for end of the array, it can avoid blatting over any existing events, which is a problem highlighted in the TODO.

tools/qlog/src/lib.rs
```diff
@@ -258,6 +298,264 @@ impl Default for Qlog {
...
+        }
+
+        Ok(())
+    }
+
+    /// Finishes qlog streaming serialization.
+    ///
+    /// The JSON-serialized output has remaining close delimiters added.
+    /// After this is called, no more serialization will occur.
+    pub fn log_finish(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
LPardue: naming things is hard

tools/qlog/src/lib.rs
LPardue: I only added this recently because I noticed a problem when that testing the C API. In that test program, it opened the file but no data was written, adding flush here resolved the problem. But actually I now realise that this indicates the log is getting buffered somewhere. This doesn't happen in rust natively. Any ideas? 

`log_finish()` can only be called once so it didn't seem to be too much overhead in ensuring all data was flushed at the end of a connection.

tools/qlog/src/lib.rs
LPardue: Maybe I'm missing something but AFIACT to use `serialize_seq()` I would need to define manual serialization code and find some way to keep the `SerliazeSeq` around so that it can be invoked as frames and when frames need to be streamed. 

The autogenerated serialization code for `EventData` today is ~4k lines. And the qlog specification for this datatype is iikely to churn.

Cargo.toml
```diff
@@ -62,6 +62,7 @@ libc = "0.2"
 ring = "0.16"
 criterion = { version = "0.3", optional = true }
 lazy_static = "1"
+qlog = { version = "0.2", path = "./tools/qlog", optional = true}
```
junhochoi:
```suggestion
qlog = { version = "0.2", path = "./tools/qlog", optional = true }
```

src/ffi.rs
```diff
@@ -29,6 +29,9 @@ use std::ptr;
 use std::slice;
 use std::sync::atomic;
 
+#[cfg(all(unix, feature = "qlog"))]
```
junhochoi: Does it mean qlog doesn't work on Windows?

src/frame.rs
```diff
@@ -689,6 +689,154 @@ impl Frame {
...
+            Frame::NewConnectionId {
+                seq_num,
+                retire_prior_to,
+                conn_id,
+                ..
+            } => qlog::QuicFrame::new_connection_id(
+                seq_num.to_string(),
+                retire_prior_to.to_string(),
+                conn_id.len() as u64,
+                "TODO:https://github.com/quiclog/internet-drafts/issues/36"
```
junhochoi:
```suggestion
                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
```

src/frame.rs
```diff
@@ -689,6 +689,154 @@ impl Frame {
...
+                    .to_string(),
+                "TODO:https://github.com/quiclog/internet-drafts/issues/36"
```
junhochoi:
```suggestion
                "TODO: https://github.com/quiclog/internet-drafts/issues/36"
```

tools/apps/Cargo.toml
```diff
@@ -5,18 +5,15 @@ authors = ["Lucas Pardue <lucaspardue.24.7@gmail.com>"]
...
 [dependencies]
 docopt = "1"
 env_logger = "0.6"
 mio = "0.6"
 url = "1"
 log = "0.4"
 ring = "0.16"
-quiche = { path = "../../" }
+quiche = { path = "../../", features=["qlog"]}
+qlog = { version = "0.2", path = "../qlog", optional = true}
```
junhochoi:
```suggestion
quiche = { path = "../../", features=["qlog"] }
qlog = { version = "0.2", path = "../qlog", optional = true }
```

tools/apps/src/lib.rs
```diff
@@ -207,6 +207,21 @@ fn autoindex(path: path::PathBuf, index: &str) -> path::PathBuf {
...
+    dir: &std::ffi::OsStr, role: &str, id: &str,
+) -> std::io::BufWriter<std::fs::File> {
+    let mut path = std::path::PathBuf::from(dir);
+    let filename = format!("{}-{}.qlog", role, id);
+    path.push(filename);
+
+    match std::fs::File::create(&path) {
+        Ok(f) => std::io::BufWriter::new(f),
+
+        Err(e) => panic!("Bad times: {}", e),
```
junhochoi: Probably need similar patch like 13d385b44a0a207c86d43834b89b85d946b253e1?

tools/qlog/src/lib.rs
```diff
@@ -233,8 +233,48 @@ use serde::{
...
+///
+/// This type is used throughout the public API for any operation that
+/// can produce an error.
+///
+/// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
+pub type Result<T> = std::result::Result<T, Error>;
+
 #[serde_with::skip_serializing_none]
 #[derive(Serialize, Clone)]
 pub struct Qlog {
```
junhochoi: Just curious... Qlog or QLog?

tools/qlog/src/lib.rs
junhochoi: I think buffers will be flushed depending on the timing when this is out of scope? Or it may help if you explicitly close the writer (drop?) here.

src/ffi.rs
LPardue: this means specifically that the ability to pass in a raw FD for use with qlog is only available on unix platforms, which is a constraint of the function https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html

tools/qlog/src/lib.rs
```diff
@@ -1148,6 +1465,37 @@ impl PacketHeader {
...
     }
+
+    /// Creates a new PacketHeader.
+    ///
+    /// Once a QUIC connection has formed, version, dcid and scid are stable, so
+    /// there are space benefits to not logging them in every packet, especially
+    /// PacketType::OneRtt.
+    pub fn with_type(
+        ty: PacketType, packet_number: u64, packet_size: Option<u64>,
+        payload_length: Option<u64>, version: Option<u32>, scid: Option<&[u8]>,
```
junhochoi: I think packet_size and payload_length is good enough with `usize` since it's all udp payload? Any reason to use u64 here?

src/ranges.rs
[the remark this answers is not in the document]
LPardue: caused by serde_json. Even though only qlog imports that crate, it seems to leak into the quiche library

tools/qlog/src/lib.rs
LPardue: The qlog schema defines this field type as `number` - see https://quiclog.github.io/internet-drafts/draft-marx-qlog-event-definitions-quic-h3.html#name-notational-conventions

When starting off, I took the very simple option of declaring any field type of `number` as u64. It wouldn't hurt to analyse all of these and figure out if we can shrink them.

LPardue: But also there is a bit of activity right now in qlog to rationalise the logging of numerical fields, which are sometimes written as strings.

rmarx: Yes, and it looks like draft-02 will say something like:

> just use numbers but be aware that some platforms (e.g., web-based tools) cannot cope with numbers > 2^53, so if you have -very- long running logs or start at a very high packet number offset, you can run into troubles. (note that this is a current limitation of the webbrowsers' JavaScript engines, not the JSON format itself)

For things like packets and QUIC-level frame lengths, 64-bit probably isn't needed. But in JSON/TypeScript there is only 1 "number" type. If we ever move qlog to a different schema format than TypeScript, we might be more specific in what's expected. 

In short: feel free to strip out numbers-as-strings and encode things as small as they need to be here.

tools/qlog/src/lib.rs
LPardue: The canonical name (from the I-Ds) is `qlog`, so I think `Qlog` is the appropriate name here in Rust

tools/apps/src/lib.rs
LPardue: good idea.

src/ffi.rs
LPardue: LOL. With applying your suggestions there is little that can co wrong with the method. I removed the return codes as part of simplifying

src/ffi.rs
```diff
@@ -231,6 +234,19 @@ pub extern fn quiche_config_set_cc_algorithm(
...
 
+#[no_mangle]
+#[cfg(all(unix, feature = "qlog"))]
+pub extern fn quiche_conn_set_qlog_fd(conn: &mut Connection, fd: c_int) {
+    let f = unsafe { std::fs::File::from_raw_fd(fd) };
+    let writer = std::io::BufWriter::new(f);
+
+    conn.set_qlog(
+        std::boxed::Box::new(writer),
+        "quiche-client qlog".to_string(),
```
junhochoi: Is this function is client only?

LPardue: good spot!

tools/qlog/src/lib.rs
LPardue: This PR is already pretty gargantuan. I'm going to do this in a separate PR that probably updates the qlog crate to draft -02 and bumps its version again.

tools/qlog/src/lib.rs
LPardue: done

tools/qlog/src/lib.rs
LPardue: done

tools/qlog/src/lib.rs
LPardue: added finish_frames, to ensure state transitions.

LPardue:
Okay... I've addressed a lot of the feedback comments, thanks! 

This is now ready for another review. In the meantime I'll start on updating the qlog documentation ready to publish the new v0.2 of the crate before this PR lands.

LPardue:
Docs updated and comments addressed, PTAL

ghedo:
@LPardue looks like this needs to be rebased on master as there is a conflict. While you are at it, can you also submit the new `qlog` crate to crates.io and squash the commits in the PR? So we can finally merge this.

PR #380, opened 2020-02-26
qlog: remove derived deserialize

LPardue:
Previously we derived qlog serde deserialize code
and this was not heavily tested. In some offline
tests, I found that the deserializer often failed
to construct the correct object, especially for
EventData which contains the most useful data.

A quick analysis puts this a few factors:
  * EventData is an enum that is untagged, which
    means serde has to do some "duck typing".
  * qlog is very flexible and allows omission of
    many fields, that makes guessing harder.
  * EventData variants can be quite similar in
    structure and serde geusses in first-to-last
    order. This causes, for example,
    EventData::PacketSent to be deserialized as
    EventData::PacketReceived. Probably the worst
    possible aliasing.

The deserialization code ends up quite large,
I counted ~12k LOC jus for EventData. Until a
robust solution can be found, there seems no
benefit in keeping this.

cc: @agrover

### DIFF
--- a/tools/qlog/src/lib.rs
+++ b/tools/qlog/src/lib.rs
@@ -236,7 +236,7 @@ use serde::{
 pub const QLOG_VERSION: &str = "draft-01";
 
 #[serde_with::skip_serializing_none]
-#[derive(Serialize, Deserialize, Clone)]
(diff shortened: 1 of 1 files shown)


agrover:
:+1:

PR #384, opened 2020-03-02
docker: small build time improvement

junhochoi:
* Make a debug build for docker images
  (non-docker build is already debug build).
  qns test doesn't need a high performance so it should be ok
* No need to run `update-ca-certificates` (it will run at installation time)
* Add `git log..` for logging quiche commit is built. Only for
  travis log.

Also, current total build time is half dominated by docker build (taking >10 min) at the end of master build. Probably better to make it parallel task?

### DIFF
--- a/extras/docker/Dockerfile
+++ b/extras/docker/Dockerfile
@@ -7,23 +7,18 @@ RUN apt-get update && apt-get install -y git cmake golang && \
 
 RUN git clone --recurse-submodules --depth 1 https://github.com/cloudflare/quiche
 RUN cd quiche && \
-    cargo build --release --examples
-RUN cd quiche && \
-    cargo build --manifest-path tools/apps/Cargo.toml --release
+    git log --oneline -n 1 && \
+    cargo build --manifest-path tools/apps/Cargo.toml
 
 ##
 ## quiche-base: base quiche image
 ##
 FROM debian:latest as quiche-base
+
 RUN apt-get update && apt-get install -y ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-RUN update-ca-certificates
-COPY --from=build /build/quiche/target/release/examples/http3-client \
-     /build/quiche/target/release/examples/http3-server \
-     /build/quiche/target/release/examples/client \
-     /build/quiche/target/release/examples/server \
-     /build/quiche/tools/apps/target/release/quiche-client \
-     /build/quiche/tools/apps/target/release/quiche-server \
+COPY --from=build /build/quiche/tools/apps/target/debug/quiche-client \
+     /build/quiche/tools/apps/target/debug/quiche-server \
      /usr/local/bin/
 ENV PATH="/usr/local/bin/:${PATH}"
 ENV RUST_LOG=info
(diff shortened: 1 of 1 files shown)


extras/docker/Dockerfile
```diff
@@ -7,23 +7,23 @@ RUN apt-get update && apt-get install -y git cmake golang && \
...
+    cargo build --manifest-path tools/apps/Cargo.toml
 
 ##
 ## quiche-base: base quiche image
 ##
 FROM debian:latest as quiche-base
+
 RUN apt-get update && apt-get install -y ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-RUN update-ca-certificates
```
LPardue: did you mean to remove `RUN update-ca-certificates`? 

The (perhaps mistaken) intent was to ensure the image's CA store was up to date to trust public sites. If that is not required, then there seems no need to install `ca-certificates` package above.

junhochoi: - `update-ca-certificates` will run when you install the package, so no need to run it twice.
- Also in quiche image, `ca-certificates` is not installed by default, it you need to give `--no-verify` for public websites (e.g. blog.cloudflare.com) every time, so I keep it.

ghedo:
>  qns test doesn't need a high performance so it should be ok

But the bandwidth and multiple handshake tests might benefit from release builds?

Also, to reduce build times we shoukd probably stop building the examples since we don't need them in the image.

extras/docker/Dockerfile
LPardue: makes sense

junhochoi:
> > qns test doesn't need a high performance so it should be ok
> 
> But the bandwidth and multiple handshake tests might benefit from release builds?

Good question. Here is some result on my MacOS laptop:

release build of quiche-server/quiche-client: ~1Gbps
debug build of quiche-server/quiche-client: ~180Mbps

debug build is a bit slower, but still can do more than 100Mbps. quic-interop test is limiting bandwidth to 10Mbps for goodput and crosstraffic test, so I think it's fast enough for tests. We may reconsider if quic-interop test try to more than 100Mbps but I think it's unlikely.

> 
> Also, to reduce build times we should probably stop building the examples since we don't need them in the image.

Currently we include http3-{client,server} and {client,server}. Do you mean only having quiche-{client,server} in cloudflare/quiche image? I'm ok with it.

junhochoi:
Here is a result of quic-interop handshake test between release/debug build:

- Release build (current)

```
+--------+--------+
|        | quiche |
+--------+--------+
| quiche |   H    |
|        |        |
|        |        |
+--------+--------+
+--------+---------------------+
|        |        quiche       |
+--------+---------------------+
| quiche | G: 9188 (± 94) kbps |
|        | C: 3044 (± 64) kbps |
+--------+---------------------+
```

- Debug build

```
 +--------+--------+
 |        | quiche |
 +--------+--------+
 | quiche |   H    |
 |        |        |
 |        |        |
 +--------+--------+
 +--------+----------------------+
 |        |        quiche        |
 +--------+----------------------+
 | quiche | G: 9104 (± 44) kbps  |
 |        | C: 3971 (± 211) kbps |
 +--------+----------------------+
```

No critical difference between release/debug build I can say.

junhochoi:
examples are removed from the image.

junhochoi:
build time: 23min -> 17min 👍

PR #388, opened 2020-03-03
ignore invalid unauthenticated packets

ghedo:
Follow-up to e739f554 and 2e04d446.

The idea is that invalid packets that haven't been authenticated
shouldn't cause a connection to be aborted, as they may be injected by
malicious man-in-the-middle or man-on-the-side attackers.

We were already doing this for undecryptable packets, but there were a
number of cases not covered.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3109,7 +3109,7 @@ impl Connection {
 /// noise, and it shouldn't keep resources occupied indefinitely.
 ///
 /// This function maps an error to `Error::Done` to ignore a packet failure
-/// without aborting the connection, except when no other pakcet was prevously
+/// without aborting the connection, except when no other packet was previously
 /// received, in which case the error itself is returned.
 ///
 /// This must only be used for errors preceding packet authentication. Failures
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3109,6 +3098,37 @@ impl Connection {
...
+/// happening after a packet has been authenticated should still cause the
+/// connection to be aborted.
+fn drop_pkt_on_err(e: Error, recv_count: usize, trace_id: &str) -> Error {
+    // If no other packet has been successflully processed, abort the connection
+    // to avoid keeping the connection open when only junk is received.
+    if recv_count == 0 {
+        return e;
+    }
+
+    trace!("{} dropped invalid packet", trace_id,);
```
junhochoi:
```suggestion
    trace!("{} dropped invalid packet", trace_id);
```

src/lib.rs
```diff
@@ -3109,6 +3098,37 @@ impl Connection {
...
+/// When a received packet that hasn't yet been authenticated triggers a failure
+/// it should, in most cases, be ignored, instead of raising a connection error,
+/// to avoid potential man-in-the-middle and man-on-the-side attacks.
+///
+/// However, if no other packet was previously received, the connection should
+/// indeed be closed as the received packet might just be network background
+/// noise, and it shouldn't keep resources occupied indefinitely.
+///
+/// This function maps an error to `Error::Done` to ignore a packet failure
+/// without aborting the connection, except when no other pakcet was prevously
```
```suggestion
/// without aborting the connection, except when no other packet was previously
```

PR #391, opened 2020-03-04
expose delivery rate in Stats + clean-up

ghedo:
My original objective was to expose the delivery rate via `Stats`, though while at it I made some clean-up to the recovery and rate estimation code.

### DIFF
--- a/src/recovery.rs
+++ b/src/recovery.rs
@@ -131,15 +131,15 @@ pub struct Recovery {
 
     app_limited: bool,
 
-    pub delivered: usize,
+    delivered: usize,
 
-    pub delivered_time: Option<Instant>,
+    delivered_time: Option<Instant>,
 
-    pub recent_delivered_packet_sent_time: Option<Instant>,
+    recent_delivered_packet_sent_time: Option<Instant>,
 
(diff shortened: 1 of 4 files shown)


src/recovery.rs
```diff
@@ -806,10 +794,10 @@ mod tests {
             is_app_limited: false,
         };
 
-        recovery.rate_on_packet_sent(&mut pkt_2, Instant::now(), TRACE_ID);
+        recovery.rate_on_packet_sent(&mut pkt_2, Instant::now());
         std::thread::sleep(Duration::new(0, 50000000));
```
LPardue: if we are changing the rate to bytes/s, lets change the sleep duration to seconds too? e.g. `Duration::new(5, 0)`

ghedo: The calculation doesn't change, just the unit, and it's `f64` so you should have mostly the same precision (if not better).

Though we should probably use `Duration::from_millis()` instead of `new()`, to make this code a bit cleaner.

ghedo: Also, 5 seconds to run each of these tests is a bit too much me thinks.

PR #392, opened 2020-03-04
deps: update BoringSSL

ghedo:
The newer version of BoringSSL has an incompatible API change: the read
and write secrets are now set from separate callbacks, instead of the
same one.

In pratice this is not much of a problem, as the code is already capable
of setting read and write secrets independently.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -84,11 +84,19 @@ struct X509(c_void);
 #[repr(C)]
 #[allow(non_camel_case_types)]
 struct SSL_QUIC_METHOD {
-    set_encryption_secrets: extern fn(
+    set_read_secret: extern fn(
         ssl: *mut SSL,
         level: crypto::Level,
-        read_secret: *const u8,
-        write_secret: *const u8,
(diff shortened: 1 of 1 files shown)


src/tls.rs
```diff
@@ -590,9 +596,48 @@ extern fn set_encryption_secrets(
...
+    trace!("{} set write secret lvl={:?}", conn.trace_id, level);
+
+    let space = match level {
+        crypto::Level::Initial => &mut conn.pkt_num_spaces[packet::EPOCH_INITIAL],
+        crypto::Level::ZeroRTT =>
+            &mut conn.pkt_num_spaces[packet::EPOCH_APPLICATION],
+        crypto::Level::Handshake =>
+            &mut conn.pkt_num_spaces[packet::EPOCH_HANDSHAKE],
+        crypto::Level::OneRTT =>
+            &mut conn.pkt_num_spaces[packet::EPOCH_APPLICATION],
```
LPardue: OCD trigger warning :)

Issue #305, opened 2019-12-17
Tracking a windowed rtt_min


junhochoi:
Not a part of spec, but many implementations use a windowed min_rtt. e.g. https://cs.chromium.org/chromium/src/net/third_party/quiche/src/quic/core/congestion_control/windowed_filter.h?g=0

PR #398, opened 2020-03-07
recovery: windowed min and max filter implementation

lohith-bellad:
Port google opensourced min-max filter code based on
Kathleen Nichols' algorithm for tracking the minimum
(or maximum) value of a data stream over some fixed
time interval.

### DIFF
--- a/src/minmax.rs
+++ b/src/minmax.rs
@@ -1,0 +1,310 @@
+// Copyright (C) 2020, Cloudflare, Inc.
+// Copyright (C) 2017, Google, Inc.
+//
+// Use of this source code is governed by the following BSD-style license:
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//    * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//    * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//
+//    * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// lib/minmax.c: windowed min/max tracker
+//
+// Kathleen Nichols' algorithm for tracking the minimum (or maximum)
+// value of a data stream over some fixed time interval.  (E.g.,
+// the minimum RTT over the past five minutes.) It uses constant
+// space and constant time per update yet almost always delivers
+// the same minimum as an implementation that has to keep all the
+// data in the window.
+//
+// The algorithm keeps track of the best, 2nd best & 3rd best min
+// values, maintaining an invariant that the measurement time of
+// the n'th best >= n-1'th best. It also makes sure that the three
+// values are widely separated in the time window since that bounds
+// the worse case error when that data is monotonically increasing
+// over the window.
+//
+// Upon getting a new min, we can forget everything earlier because
+// it has no value - the new min is <= everything else in the window
+// by definition and it's the most recent. So we restart fresh on
+// every new min and overwrites 2nd & 3rd choices. The same property
+// holds for 2nd & 3rd best.
+
+use std::time::Duration;
+use std::time::Instant;
+
+#[derive(Copy, Clone)]
+struct MinmaxSample {
+    time: Instant,
+    value: Duration,
+}
+
+pub struct Minmax {
+    estimate: [MinmaxSample; 3],
+}
+
+impl Minmax {
+    pub fn new() -> Self {
+        Minmax {
+            estimate: [MinmaxSample {
+                time: Instant::now(),
+                value: Duration::new(0, 0),
+            }; 3],
+        }
+    }
+
+    /// Resets the estimates to the given value.
+    pub fn reset(&mut self, time: Instant, meas: Duration) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        for i in self.estimate.iter_mut() {
+            *i = val;
+        }
+
+        self.estimate[0].value
+    }
+
+    /// Updates the min estimate based on the given measurement, and returns it.
+    pub fn running_min(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let delta_time = time.duration_since(self.estimate[2].time);
+
+        // Reset if there's nothing in the window or a new min value is found.
+        if val.value <= self.estimate[0].value || delta_time > win {
+            return self.reset(time, meas);
+        }
+
+        if val.value <= self.estimate[1].value {
+            self.estimate[2] = val;
+            self.estimate[1] = val;
+        } else if val.value <= self.estimate[2].value {
+            self.estimate[2] = val;
+        }
+
+        self.subwin_update(win, time, meas)
+    }
+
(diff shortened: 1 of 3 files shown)


src/recovery.rs
```diff
@@ -51,6 +52,8 @@ const INITIAL_RTT: Duration = Duration::from_millis(500);
 
 const PERSISTENT_CONGESTION_THRESHOLD: u32 = 3;
 
+const RTT_WINDOW: Duration = Duration::from_secs(10);
```
junhochoi: This default is too small. Let's start with 300sec (I think that's what Chrome uses)

src/recovery.rs
```diff
@@ -391,11 +403,19 @@ impl Recovery {
...
 
                 self.rttvar = latest_rtt / 2;
+
+                self.minmax_filter.reset(update_time, latest_rtt);
             },
 
             Some(srtt) => {
                 self.min_rtt = cmp::min(self.min_rtt, latest_rtt);
 
+                self.windowed_min_rtt = self.minmax_filter.running_min(
```
junhochoi: We can simply use this filter to update self.min_rtt. No need to keep separate value.

src/minmax.rs
```diff
@@ -0,0 +1,281 @@
...
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::time::Duration;
+use std::time::Instant;
+
+#[derive(Default, Copy, Clone)]
+struct MinmaxSample {
+    time: Option<Instant>,
```
junhochoi: Does `time` need to be `Option`?

src/minmax.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut f = Minmax::default();
+        let rtt_25 = Duration::new(0, 25000000);
+        let rtt_24 = Duration::new(0, 24000000);
+        let win = Duration::new(1, 0);
+
+        let mut rtt_min = f.reset(Instant::now(), rtt_25);
+        assert_eq!(rtt_min, rtt_25);
+
+        std::thread::sleep(Duration::new(0, 500000000));
+        rtt_min = f.running_min(win, Instant::now(), rtt_24);
```
junhochoi: Can you update not to use `sleep()`? We can use now + Duration(x) to simulate some time in the future.

lohith-bellad:
Addressed all the comments. Please review, thanks

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+        let temp = time.duration_since(self.estimate[2].time);
+
+        // if nothing in window or found new max value, reset!
+        if val.value.ge(&self.estimate[0].value) || temp.gt(&win) {
+            return self.reset(time, meas);
+        }
+
+        if val.value.ge(&self.estimate[1].value) {
+            self.estimate[2] = val;
+            self.estimate[2] = val;
```
LPardue: typo?

lohith-bellad: Ah, fixed it, thanks!

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
+// Copyright (C) 2018-2019, Cloudflare, Inc.
+// Copyright (C) 2017, Google, Inc.
```
ghedo: Where does this code come from?

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn new() -> Self {
+        Minmax {
+            estimate: [MinmaxSample {
+                time: Instant::now(),
+                value: Duration::new(0, 0),
+            }; 3],
+        }
+    }
+
+    // Reset estimates to measured value.
```
ghedo:
```suggestion
    /// Resets the estimates to the given value.
```

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn reset(&mut self, time: Instant, meas: Duration) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        for i in self.estimate.iter_mut() {
+            *i = val;
+        }
+        self.estimate[0].value
+    }
+
+    // Check if new measurement updates previous min estimates.
```
ghedo:
```suggestion
    /// Updates the min estimate based on the given measurement, and returns it.
```

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn running_min(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let temp = time.duration_since(self.estimate[2].time);
+
+        // if nothing in window or found new min value, reset!
```
ghedo:
```suggestion
        // Reset if there's nothing in the window or a new min value is found.
```

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+        self.estimate[0].value
+    }
+
+    // Check if new measurement updates previous min estimates.
+    pub fn running_min(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let temp = time.duration_since(self.estimate[2].time);
```
ghedo: What does "temp" mean?

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        // if nothing in window or found new min value, reset!
+        if temp.gt(&win) || val.value.le(&self.estimate[0].value) {
```
ghedo: Why are you not using `>` and `<=`? No other code in the whole project does this, so we shouldn't do this unless necessary (this also applies to the other cases in the whole file).

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+            self.estimate[2] = val;
+            self.estimate[1] = val;
+        } else if val.value.le(&self.estimate[2].value) {
+            self.estimate[2] = val;
+        }
+
+        self.subwin_update(win, time, meas)
+    }
+
+    // Check if new measurement updates previous max estimates.
```
ghedo:
```suggestion
    /// Updates the max estimate based on the given measurement, and returns it.
```

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn _running_max(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let temp = time.duration_since(self.estimate[2].time);
+
+        // if nothing in window or found new max value, reset!
```
ghedo:
```suggestion
        // Reset if there's nothing in the window or a new max value is found.
```

src/minmax.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn reset_filter() {
+        let mut f = Minmax::new();
+        let now = Instant::now();
+        let rtt = Duration::new(0, 50000000);
```
ghedo: Better use `Duration::from_millis()`, as it's easier to understand what the unit is without having to count the zeros :)

Same below.

src/recovery.rs
```diff
@@ -381,20 +388,25 @@ impl Recovery {
     }
 
     fn update_rtt(&mut self, latest_rtt: Duration, ack_delay: Duration) {
+        let update_time = Instant::now();
```
ghedo: We should pass the `now` value as parameter to `update_rtt()` to avoid calling `Instant::now()` all over the place as that's kind of expensive, and also makes it harder to mock time in tests. We already have `now` in `on_ack_received()` so we can just pass that forward.

src/minmax.rs
lohith-bellad: changed to dt => deltatime

src/minmax.rs
lohith-bellad: this is from Google open sources code from BSD license.
https://groups.google.com/forum/#!topic/bbr-dev/3RTgkzi5ZD8

junhochoi:
This will close #305

src/recovery.rs
```diff
@@ -380,21 +387,28 @@ impl Recovery {
         self.rate_sample.delivery_rate
     }
 
-    fn update_rtt(&mut self, latest_rtt: Duration, ack_delay: Duration) {
+    fn update_rtt(
+        &mut self, latest_rtt: Duration, ack_delay: Duration, now: Instant,
+    ) {
+        let update_time = now;
```
ghedo: This variable is not really needed at all, you can pass `now` directly into `reset()` and `running_min()` AFAICT.

src/minmax.rs
```diff
@@ -0,0 +1,272 @@
...
+        self.estimate[0].value
+    }
+
+    /// Updates the min estimate based on the given measurement, and returns it.
+    pub fn running_min(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let dt = time.duration_since(self.estimate[2].time);
```
ghedo: TBH this name is not much more helpful. The original code also used this name, and it's not quite clear to me what it's supposed to stand for so not sure if we can come up with something better....

src/minmax.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    fn subwin_update(
+        &mut self, win: Duration, time: Instant, meas: Duration,
+    ) -> Duration {
+        let val = MinmaxSample { time, value: meas };
+
+        let dt = time.duration_since(self.estimate[0].time);
+
+        if dt > win {
```
ghedo: The original code seems to have a fair amount of comments in this function which are pretty useful to understand what it's doing, Can we keep them?

src/recovery.rs
lohith-bellad: done!

src/minmax.rs
lohith-bellad: made it delta_time which explains what it is

src/minmax.rs
lohith-bellad: Sure, added them.

ghedo:
For whatever reason rustc nightly seems to be crashing which causes the Travis CI builds to fail. Though that's probably unrelated (the crash happens when building `dockopt`) so I merged this anyway.

ghedo:
FTR, I made https://github.com/rust-lang/rust/issues/70041 for the compiler crash.

PR #400, opened 2020-03-09
avoid some needless small allocations when creating packets

ghedo:
Instead of unconditionally cloning optional QUIC header values, check
that the output packet type requires them

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1710,10 +1710,27 @@ impl Connection {
             ty: pkt_type,
             version: self.version,
             dcid: self.dcid.clone(),
-            scid: self.scid.clone(),
+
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1710,10 +1710,27 @@ impl Connection {
...
             version: self.version,
             dcid: self.dcid.clone(),
-            scid: self.scid.clone(),
+
+            // Don't needlessly clone the source connection ID for 1-RTT packets
+            // as it is not used.
+            scid: if pkt_type != packet::Type::Short {
+                self.scid.clone()
+            } else {
+                Vec::new()
```
LPardue: TIL new() doesn't allocate

PR #407, opened 2020-03-10
Make sure there is enough space left in the packet before adding frames

junhochoi:
When we add frames in Connection.send(), some frames are missing
checking if there is enough space left in the packet (`left`),
may cause underflow of `left` in release build (debug build
it will crash runtime)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -977,6 +977,26 @@ pub fn version_is_supported(version: u32) -> bool {
     }
 }
 
+/// Pushes a frame to the output packet if there is enough space.
+///
+/// Returns `true` on success, `false` otherwise. In case of failure it means
+/// there is no room to add the frame in the packet. You may retry to add the
+/// frame later.
+macro_rules! push_frame_to_pkt {
+    ($frames:expr, $frame:expr, $payload_len: expr, $left:expr) => {{
+        if $frame.wire_len() <= $left {
+            $payload_len += $frame.wire_len();
+            $left -= $frame.wire_len();
+
+            $frames.push($frame);
+
+            true
+        } else {
+            false
+        }
+    }};
+}
+
 impl Connection {
     fn new(
         scid: &[u8], odcid: Option<&[u8]>, config: &mut Config, is_server: bool,
@@ -1772,13 +1792,8 @@ impl Connection {
                 ranges: self.pkt_num_spaces[epoch].recv_pkt_need_ack.clone(),
             };
 
-            if frame.wire_len() <= left {
+            if push_frame_to_pkt!(frames, frame, payload_len, left) {
                 self.pkt_num_spaces[epoch].ack_elicited = false;
-
-                payload_len += frame.wire_len();
-                left -= frame.wire_len();
-
-                frames.push(frame);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -977,6 +977,21 @@ pub fn version_is_supported(version: u32) -> bool {
     }
 }
 
+macro_rules! add_frame {
```
LPardue: can this just be a function rather than a macro?

src/lib.rs
```diff
@@ -977,6 +977,21 @@ pub fn version_is_supported(version: u32) -> bool {
...
+    {$frames:expr, $frame:expr, $payload_len: expr, $left:expr} => {{
```
ghedo: Why are you using `{ }` instead of `( )`? `()` is more consistent with other macros and you can call this more like a function: `add_frame!(...)`.

src/lib.rs
```diff
@@ -977,6 +977,21 @@ pub fn version_is_supported(version: u32) -> bool {
...
+        if $frame.wire_len() <= $left {
```
ghedo: It'd be easier to read inverting the condition:
```rust
if $frame.wire_len() > $left {
    return false;
}

<add frame here>

true
```

src/lib.rs
ghedo: This probably needs a more decruptive name, like `push_frame_to_pkt` or something like that, also would be good to have a bit of documentation.

src/lib.rs
```diff
@@ -1791,40 +1801,31 @@ impl Connection {
...
                 let frame = frame::Frame::MaxStreamsBidi { max };
 
-                payload_len += frame.wire_len();
-                left -= frame.wire_len();
-
-                frames.push(frame);
-
-                ack_eliciting = true;
-                in_flight = true;
+                if add_frame! {frames, frame, payload_len, left} {
```
ghedo: `update_max_streams_bidi()` is not idempotent, and if this fails we'll not try sending a MAX_STREAMS frame again. `update_max_streams_bidi()` needs to be split into two, with one function returning the new value, and the other updating the internal value.

src/lib.rs
```diff
@@ -1791,40 +1801,31 @@ impl Connection {
...
                 let frame = frame::Frame::MaxStreamsUni { max };
 
-                payload_len += frame.wire_len();
-                left -= frame.wire_len();
-
-                frames.push(frame);
-
-                ack_eliciting = true;
-                in_flight = true;
+                if add_frame! {frames, frame, payload_len, left} {
```
ghedo: Same as `MAX_STREAMS_BIDI`.

src/lib.rs
```diff
@@ -1999,18 +1974,15 @@ impl Connection {
...
+                    ack_eliciting = true;
+                    in_flight = true;
 
-                // If the stream is still flushable, push it to the back of the
-                // queue again.
-                if stream.is_flushable() {
-                    self.streams.push_flushable(stream_id);
+                    // If the stream is still flushable, push it to the back of
+                    // the queue again.
+                    if stream.is_flushable() {
```
ghedo: This needs to be outside of the `if`.

src/lib.rs
junhochoi: I prefer to use macro because when it is a function there is many side effects (`&mut`), such as updating payload_len, left etc.

src/lib.rs
junhochoi: Done. Somehow I had errors when I use `(` but it seems ok actually...

src/lib.rs
junhochoi: I tried this but I got this problem: https://stackoverflow.com/questions/24502282/why-does-removing-return-give-me-an-error-expected-type-but-found-type
Also can't use `return` because it will return from the function (`send()`) itself.

src/lib.rs
junhochoi: Changed to `push_frame_to_pkt` and add a comment.

src/lib.rs
junhochoi: Good point. I split into `update_max_streams_bidi/uni()` to return the next value and `max_streams_bidi/uni_next()` to get next value for creating a frame.

src/lib.rs
junhochoi: See above for update_max_streams_uni() case.

src/lib.rs
junhochoi: Done.

src/lib.rs
junhochoi: Also did the same for stream.max_data().

src/lib.rs
```diff
@@ -1861,22 +1867,17 @@ impl Connection {
...
                 }
 
-                payload_len += frame.wire_len();
-                left -= frame.wire_len();
-
-                frames.push(frame);
-
-                ack_eliciting = true;
-                in_flight = true;
+                self.streams.mark_almost_full(stream_id, false);
```
ghedo: This should be inside the `if`, otherwise we won't retry sending the frame if it failed this time.

ghedo:
Also tweaked the macro's docs a bit.

PR #408, opened 2020-03-10
Fix invalid_initial() and invalid_packet() tests

junhochoi:
When it tries to corrupt buf[], there is a 1/256 chance of test
failure when buf[written - 1] = 0. Try to have always different value
by inverting bits.


ghedo:
Nice! Though this should be done for the `invalid_packet()` test as well.

junhochoi:
> Nice! Though this should be done for the `invalid_packet()` test as well.

Sure. Fixed it as well.

PR #417, opened 2020-03-16
h3: prevent DATA before headers

LPardue:
We already have a few measures in place to prevent DATA before HEADERS, mainly on the client side. This change just enhances that by adding some tracking into the `h3::Stream` object for local initialization (this is special because `Stream` is generally geared around tracking state of the stream as data is read).

### DIFF
--- a/src/h3/stream.rs
+++ b/src/h3/stream.rs
@@ -132,6 +132,9 @@ pub struct Stream {
 
     /// Whether the stream has been initialized.
     initialized: bool,
+
+    /// Whether the stream has been locally initialized.
+    local_initialized: bool,
 }
 
 impl Stream {
@@ -167,6 +170,7 @@ impl Stream {
 
             is_local,
             initialized: false,
+            local_initialized: false,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/h3/stream.rs
```diff
@@ -132,6 +132,9 @@ pub struct Stream {
 
     /// Whether the stream has been initialized.
     initialized: bool,
+
+    /// Whether the stream has been locally initialized.
+    local_initialized: bool,
```
junhochoi: A little confused on the naming -- initialized vs local_initialized. Probably a better clarifying name such as 'header_sent' or something like that?

LPardue: since h3::Stream handles all stream types, having a `headers_sent` message would also seem weird.

I could rename `initialized` to `remote_intialized`...

junhochoi: Ok. In the diff `local_initialized` seems only used for checking HEADERS frame sent earlier so I thought `headers_sent' (or `is_headers_sent`) may be better naming. If you think it can be used for other purpose it will be fine.

ghedo: Yeah, renaming `initialized` seems like a good idea, it's kind of confusing right now. Or maybe call them `initial_frame_sent` / `initial_frame_received`? It seems that's what they are used to check for and they are a bit more descriptive.

LPardue: Not all streams take a frame. For example, Push streams start with a Push ID, we just don't have to code implemented to make that clear right now. 

I think local/remote is the appropriate path forward here.

PR #419, opened 2020-03-16
Fix debug msg

junhochoi:
- Add space after delivered_time
- Change delivery_rate display to u64 for easier read

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3176,7 +3176,7 @@ impl std::fmt::Debug for Stats {
             self.lost,
             self.rtt,
             self.cwnd,
-            self.delivery_rate
+            self.delivery_rate as u64
         )
     }
 }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3176,7 +3176,7 @@ impl std::fmt::Debug for Stats {
             self.lost,
             self.rtt,
             self.cwnd,
-            self.delivery_rate
+            self.delivery_rate as u64
```
ghedo: Maybe we should just make the field `u64`? Not quite sure how useuful it is to keep it as float since we just convert it to `u64` in every place where we use it (e.g. FFI does the same to mimic what `TCP_INFO` does).

junhochoi: Good idea. I opened #422 including this diff.

junhochoi:
Fixed in #422

PR #421, opened 2020-03-18
recovery: convert congestion control API to raw vtable

ghedo:
Using a trait for the congestion control operations means that different
algorithms will need to duplicate some of the logic due to not being
able to share common values between each other (e.g. congestion_window,
bytes_in_flight, ssthresh, ...). This also makes it hard to implement
more complex features (e.g. Hystart) that should really apply to all
algorithms.

This changes the congestion control API from a trait, where each
algorithm is implemented as a separate type that implements the common
trait, to a raw function vtabl, where each function is passed the entire
Recovery struct (so each algorithm don't have to define a separate type).

This allows all of the fields of Recovery to be available within each
congestion control implementation, so that they can be reused without
having to duplicate them, and without having to define awkward and
verbose trait methods to access them (another good side-effect being
that we don't need to pass many of these fields to the CC methods
anymore, as they can now be accessed directly).

This also makes it possible to mix and match the algorithm's methods
(e.g. in the future an implementation could reuse one of the existing
algorithms' functions), as well as wrap them and expand their logic
(commong logic between all algorithms can be implemented as part of
Recovery's on_packet_sent_cc(), on_packet_acked_cc() and
congestion_event() methods, rather than each algorithm's methods).

Due to the fact that Recovery's fields are private, the congestion
control implementations need to be submodule of recovery, rather than
having a separate cc submodule. One possibility would be to make a cc
module that is a submodule of recovery, but that doesn't seem to improve
things much, as the CC API is pretty small.


src/recovery/reno.rs
```diff
@@ -0,0 +1,181 @@
...
+            as usize;
+
+        r.congestion_window =
+            std::cmp::max(r.congestion_window, recovery::MINIMUM_WINDOW);
+
+        r.ssthresh = r.congestion_window;
+    }
+}
+
+// #[cfg(test)]
```
junhochoi: Please enable tests again.

ghedo: Done.

ghedo: Done.

src/recovery.rs
```diff
@@ -221,17 +252,20 @@ impl Recovery {
...
-            self.cc.on_packet_sent_cc(sent_bytes, now, trace_id);
+            self.on_packet_sent_cc(sent_bytes, now);
 
             self.set_loss_detection_timer(handshake_completed);
         }
 
         trace!("{} {:?}", trace_id, self);
     }
 
+    fn on_packet_sent_cc(&mut self, sent_bytes: usize, now: Instant) {
```
junhochoi: Do we need to make it as a separate function? It's a single line and pretty clear what it does so moving L266 to L257 is good enough I think.

src/recovery.rs
```diff
@@ -564,18 +593,28 @@ impl Recovery {
         false
     }
 
-    // TODO: move to Congestion Control and implement draft 24
+    fn on_packet_acked_cc(&mut self, packet: &Sent, now: Instant) {
```
junhochoi: Same comment as L265

src/recovery.rs
```diff
@@ -592,24 +631,30 @@ impl Recovery {
...
+            self.congestion_event(largest_lost_pkt.time, now);
 
             if self.in_persistent_congestion(&largest_lost_pkt) {
-                self.cc.collapse_cwnd();
+                self.collapse_cwnd();
             }
         }
     }
 
+    fn congestion_event(&mut self, time_sent: Instant, now: Instant) {
```
junhochoi: Same comment as L265

src/recovery.rs
```diff
@@ -679,12 +724,53 @@ impl Recovery {
...
+        match name {
+            "reno" => Ok(CongestionControlAlgorithm::Reno),
+
+            _ => Err(crate::Error::CongestionControl),
+        }
+    }
+}
+
+pub struct CongestionControlOps {
+    pub on_packet_sent: fn(r: &mut Recovery, sent_bytes: usize, now: Instant),
```
junhochoi: I think we can move `now` to a field of Recovery and remove `now` from all the hooks. `now` timestamp mostly coming from the caller of Recovery (e.g. `on_packet_sent()` called in `lib.rs`) and passed to most of calls inside, so it will be simpler if we update `recovery.now` only and all functions inside can use that value without additional parameters in those calls. Similar idea to how you removed `trace_id` I think.

src/recovery.rs
```diff
@@ -351,7 +385,7 @@ impl Recovery {
             unacked_bytes += p.size;
         }
 
-        self.cc.decrease_bytes_in_flight(unacked_bytes);
+        self.bytes_in_flight -= unacked_bytes;
```
junhochoi: not `saturating_sub()`? (it was)

src/recovery.rs
```diff
@@ -592,24 +631,30 @@ impl Recovery {
                 continue;
             }
 
-            self.cc.decrease_bytes_in_flight(p.size);
+            self.bytes_in_flight -= p.size;
```
junhochoi: Same as L388

src/recovery.rs
ghedo: `now` is not a property of `Recovery` so it doesn't belong there. Also, I think it would make it harder to see where the value is initialized (currently you just have to follow the call stack).

src/recovery.rs
ghedo: The idea was that common CC logic could be implemented within these methods before/after the specific CC-algorithm logic, if needed. But I don't care too much.

junhochoi: Currently the cases I have is to change reno ops itself, no much necessity to add before/after the hook caller.

src/recovery.rs
junhochoi: My question is why not? However when I look the usage of `now` across the code, this may not be simple. Let's move on.

Issue #418, opened 2020-03-16
Custom Roots of Trust?


ctiller:
I'm currently putting together something that needs to have a set of certs trusted that are not the OS defaults. Any chance of adding some call path to `quiche::Context` to call through to `SSL_CTX_load_verify_locations` instead of `SSL_CTX_set_default_verify_paths`?

I'm happy to type the code after checking the design.

I'd be thinking about adding to Context:

```
pub fn load_verify_locations_from_file(&mut self, file: &str);
```

LPardue:
would your use case be satisfied by setting `SSL_CERT_DIR` or `SSL_CERT_FILE` env vars?

ctiller:
Short term, probably yes. Long term - it's unlikely.
Right now we could force it at application startup with a
std::env::set_var, but longer term it's unclear that process-wide is the
right thing to do.

ghedo:
@ctiller yeah, that sounds good to me, though might want to have both `_from_file()` and `_from_directory()` variants.

ctiller:
The environment variable approach hasn't worked out well in the past few
days - we end up in situations in unit tests where we want >1 option set.

I'm going to put something together and send a PR.

PR #423, opened 2020-03-19
Allow custom roots of trust

ctiller:
Original issue: https://github.com/cloudflare/quiche/issues/418

Allows loading custom trusted certificates to perform certificate
validation with.

Adds a test that this works, which meant creating a cert chain we could
validate. Added a script to generate the certificate data on demand (so
that the process for this is at least marginally documented).


examples/gen-certs.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -ex
+openssl genrsa -out rootca.key 4096
+bash -c $(dirname $0)/rootca.exp
+openssl genrsa -out cert.key 2048
+bash -c $(dirname $0)/csr.exp
+openssl x509 -req -days 1000 -in cert.csr -CA rootca.crt -CAkey rootca.key -CAcreateserial -out cert.crt -sha1
```
ghedo: The above 3 command can be simplified as:
```
openssl req -new -batch -nodes -sha256 -keyout cert.key -out cert.csr -subj '/C=GB/CN=quic.tech'
openssl x509 -req -days 10000 -in cert.csr -CA rootca.crt -CAkey rootca.key -CAcreateserial -out cert.crt
```
So we don't need the `expect` script here either.

src/tls.rs
```diff
@@ -149,6 +149,28 @@ impl Context {
         }
     }
 
+    pub fn load_verify_locations(
```
ghedo: It would be better to also split this into two (`_from_file` and `_from_directory`) variants, so we don't have to deal with the `Option`s.

examples/cert.csr
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE REQUEST-----
```
ghedo: We shouldn't commit this into the repository as it's not needed and it can be easily regenerated if needed.

examples/rootca.crt
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
ghedo: This also shouldn't be committed.

examples/rootca.key
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
ghedo: This also shouldn't be committed.

examples/rootca.srl
```diff
@@ -0,0 +1 @@
+5324F08044C2BCB1081F5F527FF04D75AF9C760A
```
ghedo: This also shouldn't be committed.

LPardue:
Do we want to expose the API to FFI too?

examples/rootca.crt
ctiller: This one's used in tests for the new feature and so is necessary.

examples/cert.csr
ctiller: Fixed, and updated gen-certs.sh to follow this policy.

examples/rootca.key
ctiller: Fixed, and updated gen-certs.sh to follow this policy.

examples/rootca.srl
ctiller: Fixed, and updated gen-certs.sh to follow this policy.

examples/gen-certs.sh
ctiller: Done.

examples/gen-certs.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -ex
+openssl genrsa -out rootca.key 4096
+bash -c $(dirname $0)/rootca.exp
```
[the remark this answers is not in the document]
ctiller: Done.

ctiller:
> Do we want to expose the API to FFI too?

Probably?

I don't know much about the FFI or how to test it - my use of quiche so far has been solely from Rust.

LPardue:
To explain a bit, the FFI allows quiche to be called from C (for example), methods in `ffi.rs` handle effectively wrap the quiche library methods and allow the transfer of C-friendly types. The functions are defined in `quiche.h`. 

An example that is close to yours is `quiche_config_load_cert_chain_from_pem_file()` https://github.com/cloudflare/quiche/blob/master/src/ffi.rs#L90

src/tls.rs
ctiller: Done

ghedo:
Merged as https://github.com/cloudflare/quiche/commit/f8126019a893e9f94ada3d21ee5274e54b53e603, thanks @ctiller! I tweaked the commit message slightly, and also removed th `*.exp` files that are not needed anymore.

@LPardue good point, but I think this is good enough for now, we can add the FFI API separately.

PR #426, opened 2020-03-26
apps: handle partial HTTP/0.9 requests on the server

ghedo:
When a stream doesn't have enough capacity, we might receive a partial
HTTP/0.9 request. Before handling it we need to wait until the full
request is received.

### DIFF
--- a/tools/apps/src/bin/quiche-server.rs
+++ b/tools/apps/src/bin/quiche-server.rs
@@ -292,6 +292,7 @@ fn main() {
                 let client = Client {
                     conn,
                     http_conn: None,
+                    partial_requests: HashMap::new(),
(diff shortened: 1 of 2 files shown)


ghedo:
The partial requests are not removed from the map when they are completed, which is not great, but all memory is freed eventually when the connection ends. I'll try to make it so partial requests can be freed earlier, but it becomes a bit awkward given how the code is currently structured.

ghedo:
Actually nvm, this was easier than I thought. The partial requests are now freed, though there are some edge cases where we might keep buffering stuff (e.g. if the data never terminates with "\r\n"), though I think that's fine as these apps are not really designed with DDoS in mind anyway.

PR #427, opened 2020-03-28
apps: make 'qlog' feature really optional

ghedo:
Both the apps crate and quiche itself optionally depend on the qlog
crate, however the apps crate enables quiche's qlog feature
unconditionally. This means that even if the qlog feature of the apps
crate is disabled, the qlog crate will still be built.

However the apps crate doesn't need to depend on the qlog crate at all,
as it doesn't use it directly. This change removes the dependency, and
turns the qlog feature into an explicit feature that only enables
quiche's qlog feature when enabled.

---

Also, we might want to enable the `qlog` feature in the apps crate by default.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,9 +271,6 @@ use std::time;
 use std::pin::Pin;
 use std::str::FromStr;
 
-#[cfg(feature = "qlog")]
-use qlog::event::Event;
-
 /// The current QUIC wire version.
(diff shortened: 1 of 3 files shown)


junhochoi:
I think it's good to have some example how to run apps with qlog.
e.g.
```
$ env QLOGDIR=/tmp/qlog cargo run --features qlog --manifest-path=tools/apps/Cargo.toml --bin quiche-client -- https://quic.tech:8443/
```
Same with quiche-server.

PR #429, opened 2020-03-30
apps: make qlog feature default

LPardue:
The quic-interop-runner sets the `QLOGDIR` env var (https://github.com/marten-seemann/quic-interop-runner/blob/master/docker-compose.yml), so all we need to do is enable the feature in our quiche apps. Tested the interop runner locally and it seems to work fine, dumping the qlog output to `logs/qlog/{role}-{CID}.qlog`

~Note that I think this also enables the feature for both the `cloudflare/quiche` and `cloudflare/quiche-qns` images but I think that is an acceptable tradeoff. The alternative is probably a complicated branching dockerfile :nauseated_face:~

### DIFF
--- a/tools/apps/Cargo.toml
+++ b/tools/apps/Cargo.toml
@@ -12,6 +12,8 @@ fuzzing = ["quiche/fuzzing"]
 # Enable qlog support.
 qlog = ["quiche/qlog"]
 
+default = ["qlog"]
+
 [dependencies]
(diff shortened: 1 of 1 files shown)


ghedo:
Why not just enable the apps feature by default? So we would also end up testing the code in CI.

LPardue:
I've made this a default feature as suggested

Issue #32, opened 2019-04-08
Send Trait for Connection


sgessa:
I'm trying to use this library from a multi-threaded context.
Is there any particular reason the `Connection` struct can't implement `Send`?

ghedo:
One problem is that the TLS handshake state (which is a raw pointer to a BoringSSL `SSL` object) is not thread-safe, which makes the whole Connection not thread-safe as well. It might be possible to make it thread-safe by wrapping the TLS state in `Arc<Mutex<...>` but that would need to be tested to make sure it doesn't break any assumptions made by BoringSSL.

puckipedia:
I believe `impl`ing `Send` would be safe on both `Context` and `Handshake`, as that just means allowing the respective structs to move between threads, which is okayed by the `ssl.h` comments:

```c
// An |SSL| object represents a single TLS or DTLS connection. Although the
// shared |SSL_CTX| is thread-safe, an |SSL| is not thread-safe and may only be
// used on one thread at a time.
```

ctiller:
One would also need to deal with application data on streams - currently this is a `Box<dyn std::any::Any>` and it would need to be `Box<dyn Send + std::any::Any>`, which would break the current ffi interface, since it's taking a `*mut c_void` for application data, which is not Send.

PR #433, opened 2020-03-30
Make Connection Send

ctiller:
This PR addresses comments in #32 which taken together imply that Connection could be Send (but not Sync). Doing so would greatly ease integration in a project I'm working on, so I thought I'd see what the maintainers think.


ghedo:
Merged now, thanks @ctiller!

PR #435, opened 2020-03-31
WIP: stream: initial support for prioritization

ghedo:
This implements a very simple scheme for stream prioritization at the
transport layer. Each stream can be prioritized with a number from 0 to
255 (lower is more important).

Currently priority is only used when flushing streams to the wire, but
not when streams are reported as readable or writable to the application
(yet).

Internally, instead of keeping a single queue for all scheduled streams,
an ordered map of lists is used instead. Each list holds the streams
that are scheduled at a given priority. When flushing data, the
non-empty list with the lower priority is used first. Within the same
list, streams are round-robined (so the default behavior doesn't change).

Re-prioritization is handled by removing the stream from the queue
corresponding to the old priority, and adding it to the queue of the new
pririty. However the implementation is very inefficient, due to the fact
that finding and removing the stream from the old queue requires
scanning the whole list linearly, or shifting all the elements in the
queue in the worst case scenario, or a combination of both in most
cases. So until we have a better implementation (I'm looking into it) we
might want to just ignore re-prioritization.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1856,7 +1856,8 @@ impl Connection {
                     // Consider the stream flushable also when we are sending a
                     // zero-length frame that has the fin flag set.
                     if (stream.is_flushable() || empty_fin) && !was_flushable {
-                        self.streams.push_flushable(stream_id);
+                        let priority = stream.priority;
+                        self.streams.push_flushable(stream_id, priority);
                     }
                 },
 
@@ -2205,7 +2206,8 @@ impl Connection {
                 // If the stream is still flushable, push it to the back of the
                 // queue again.
                 if stream.is_flushable() {
-                    self.streams.push_flushable(stream_id);
+                    let priority = stream.priority;
+                    self.streams.push_flushable(stream_id, priority);
                 }
 
                 // When fuzzing, try to coalesce multiple STREAM frames in the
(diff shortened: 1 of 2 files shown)


ghedo:
The idea to resolve the re-prioritization problem is to use an intrusive linked list that allows elements to be remove from the middle. So `Stream` would hold the `next`/`prev` pointers, and could then be removed from the list by simply having a reference to it (since streams are held in a `HashMap` this should be sufficiently efficient), as you would do in a typical C linked list.

However this of course is very unsafe,  so will require some time to get all the details right (assuming it is even possible).

ghedo:
Changed priority value to `u64` so we can use stream IDs as priorities.

src/lib.rs
```diff
@@ -2558,6 +2562,44 @@ impl Connection {
...
+        &mut self, stream_id: u64, priority: u64,
+    ) -> Result<()> {
+        // Get existing stream.
+        let stream = self
+            .streams
+            .get_mut(stream_id)
+            .ok_or(Error::InvalidStreamState)?;
+
+        if stream.priority == priority {
+            return Err(Error::Done);
```
LPardue, line 2587: Error::Done doesn't seem super helpful for the calling app. 

With a test implementation, I'm calculating the new priority based on information in H3 layer and that might yield the same answer, but I don't store the old priority so can't check that before calling `stream_priority()`. That means instead of being able to do `conn.stream_priority(id, priority)?`, I need to capture Ok and Error::Done and throw them away.

ghedo:
Will re-open with updated implementation.

PR #436, opened 2020-04-01
remove duplicate qlog finish_frames() call

ghedo:


ghedo:
@LPardue I didn't test this, but the same call is done like 10 lines above this?

ghedo:
@LPardue it seems to work fine to me... Also, this uncovered another small problem, so added a commit to fix that.

LPardue:
Weird, I'm sure it was broken at one point without that extra call. Maybe then double it got fixed in code review before merging.  

I've tested your changes locally and the generated qlogs are all good.

PR #437, opened 2020-04-01
qlog: add HANDSHAKE_DONE frame

ghedo:
This adds the HANDSHAKE_DONE frame to qlog, and makes quiche log it.

This also removes the "unknown frame" case in the Frame::to_qlog()
method, as there can't be unknown frames, and it hides missing frame
types conversions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ log = { version = "0.4", features = ["std"] }
 libc = "0.2"
 ring = "0.16"
 lazy_static = "1"
-qlog = { version = "0.2", optional = true }
+qlog = { version = "0.2", path = "tools/qlog", optional = true }
 
 [target."cfg(windows)".dependencies]
 winapi = { version = "0.3", features = ["wincrypt"] }
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -834,7 +834,7 @@ impl Frame {
                            * frame type */
                 ),
 
-            _ => qlog::QuicFrame::unknown(std::u64::MAX),
+            Frame::HandshakeDone => qlog::QuicFrame::handshake_done(),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


LPardue:
So the change itself is fine but the problem is that qlog draft-01 lags behind the IETF drafts and represent a version of the protocol before draft 27 (https://tools.ietf.org/html/draft-marx-qlog-event-definitions-quic-h3-01).

The HandshakeDone event has been added to the editors copy (https://quiclog.github.io/internet-drafts/draft-marx-qlog-event-definitions-quic-h3.html#rfc.appendix.A.5.20) but so have a bunch of other changes. I'm not sure if/when a new draft will be cut. @rmarx has made some comments that they may look for a different versioning approach but I'm not sure the current status. AFAIK the tools are very Postel-y, and can handle frakenqlogs.

So, I think we should modify the QLOG_VERSION string to indicate this is not "pure" draft-01, and then continue to incorporate improvements as a series of PRs.

rmarx:
So yeah, I'm planning to push hard on draft-02 after April 10th (as I'm working towards a paper deadline for that date). Some smaller changes like the handshake_done frame are already in the editor's version, as Lucas mentions. I hope to incorporate all currently outstanding issues in draft-02.

At the moment though, if you do not-plain-draft-01, you MUST use `"qlog_version": "draft-02-wip"` according to the editor's draft of the main schema (https://quiclog.github.io/internet-drafts/draft-marx-qlog-main-schema.html) ;) 

I made the mistake of not having a separate `-wip` version for 01, which gave me some headaches supporting not-quite-01 in qvis. Hence the inbetween you can use.

But, as Lucas indicates, qvis should be well-prepared to handle funky qlogs

LPardue:
Thanks for the feedback @rmarx. 

In that case let's bump the QLOG_VERSION string to `draft-02-wip` and then land this change.

ghedo:
Ok, updated `QLOG_VERSION` as well.

ghedo:
Added a commit to fix formatting in the `qlog` crate, that was broken while the nightly builds were out.

PR #452, opened 2020-04-09
immediately mark connection as closed when receiving an invalid first packet

ghedo:
When an invalid first packet is received the connection is correctly
marked as closing, however due to the anti-amplification mechanism it is
prevented from actually sending anything.

Since the draining timeout is armed only when `CONNECTION_CLOSE` is
actually sent, and the idle timeout is armed when a packet is processed
successfully, the connection ends up never being closed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2871,6 +2871,11 @@ impl Connection {
             self.error = Some(err);
         }
 
+        // When no packet was successfully processed close connection immediately.
+        if self.recv_count == 0 {
+            self.closed = true;
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2871,6 +2871,11 @@ impl Connection {
             self.error = Some(err);
         }
 
+        // When no packet was successfully processed close connection immediately.
+        if self.recv_count == 0 {
+            self.closed = true;
```
junhochoi, line 2876: Is it better to do at `drop_pkt_on_err()`?

ghedo: This is not really possible as we can't reference `self` when `drop_pkt_on_err()` is called (which is why it is a separate function rather than a `Connection` method).

PR #453, opened 2020-04-09
close connection on first packet's payload length mismatch

ghedo:
Similarly to other cases, if the first packet on a connection has an
invalid payload length, we should close the connection immediately to
avoid keeping it open for packets that may never arrive.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1502,7 +1502,12 @@ impl Connection {
 
         if b.cap() < payload_len {
             trace!("{} payload length mismatch", self.trace_id);
-            return Err(Error::Done);
+
+            return Err(drop_pkt_on_err(
+                Error::InvalidPacket,
+                self.recv_count,
+                &self.trace_id,
+            ));
         }
 
         let header_len = b.off();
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1502,7 +1502,12 @@ impl Connection {
 
         if b.cap() < payload_len {
```
junhochoi, line 1503: better to be `!=`? since payload_len may use the value in the packet.

junhochoi: e.g. in the additional test below, if I set `payload_len = 20` (not 4096), `pipe.server.recv()` will return `CryptoFail`. Which is fine but if we change here to `!=` it will be caught here so we don't even need to try decrypt.

ghedo: `b.cap()` represents the amount of bytes remaining in the whole UDP datagram, which could potentially include additional coalesced packets or additional junk at the end (e.g. see padding mechanism in FIrefox).

Though, thinking about this more, we don't actually need an explicit check for this, as decryption will fail later anyway (with `BufferTooShort` which seems an appropriate error), so I'll just remove this check completely.

ghedo: See https://github.com/cloudflare/quiche/pull/460.

PR #454, opened 2020-04-09
fixed certificate and key path in .c examples

piano-man:
Fixed the path for the certificate and key files in server.c file in examples.
With the existing path, it leads to -
 **`quiche::tls: error:100000ae:SSL routines:OPENSSL_internal:NO_CERTIFICATE_SET`**

### DIFF
--- a/examples/http3-server.c
+++ b/examples/http3-server.c
@@ -499,8 +499,8 @@ int main(int argc, char *argv[]) {
         return -1;
     }
 
-    quiche_config_load_cert_chain_from_pem_file(config, "examples/cert.crt");
-    quiche_config_load_priv_key_from_pem_file(config, "examples/cert.key");
+    quiche_config_load_cert_chain_from_pem_file(config, "./cert.crt");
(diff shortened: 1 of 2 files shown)


piano-man:
Was the path left that way intentionally ? Just trying to figure out if this is an unnecessary change :)

ghedo:
@piano-man the value was indeed on purpose (and consistent with what the Rust examples also use). The idea was that a user would run the examples from the parent directory (again, like in the Rust examples case).

However the C examples are different as the `Makefile` to build them is actually inside the `examples/` directory, so it is likely that users would enter the directory to build them (which is also what the documentation says), so I think this change is actually okay.

Though you also need to update the `http3-server.c` example as well. Also please rebase the branch on top of the latest `master`.

piano-man:
Done!

ghedo:
Merged as https://github.com/cloudflare/quiche/commit/5974a02e3b083deff709cf729a1e8c6540de17e7 (with a slightly tweaked commit message), thanks!

PR #455, opened 2020-04-10
nginx: subrequests fix when parent request uses quic stream

morf:
ngx_http_v3_send_response() double call fix and aviod on this function call for subrequest
Run "posted" request (start subrequest) for quic request
Add qstream for subrequest for fix ngx_http_finalize_connection call and other functions with quic connection

### DIFF
--- a/extras/nginx/nginx-1.16.patch
+++ b/extras/nginx/nginx-1.16.patch
@@ -1,4 +1,4 @@
-From ab4917ce8797ff545723983d99803fa3689d3446 Mon Sep 17 00:00:00 2001
+From d8f9979b4e316fa89d374ef109cae5ff291de4c7 Mon Sep 17 00:00:00 2001
 From: Alessandro Ghedini <alessandro@cloudflare.com>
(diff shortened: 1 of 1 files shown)


extras/nginx/nginx-1.16.patch
```diff
@@ -3923,6 +3942,16 @@ index 00000000..5bbff860
 +        return ngx_http_next_header_filter(r);
 +    }
 +
++    if (r->header_sent) {
++        return NGX_OK;
++    }
++
++    r->header_sent = 1;
```
ghedo, line 3949: This can't be set before we actually send the headers, as it would break sending of responses that are blocked. It is already set inside `ngx_http_v3_send_response()` so there is no need to set it again here.

morf: Hi, I described in more detail (about r->header_sent) in the new pr https://github.com/cloudflare/quiche/pull/504 (https://github.com/cloudflare/quiche/pull/504/commits/938e9154dae0b0e761913f7f4fd5f7569dc68ae5)

PR #458, opened 2020-04-13
recovery: refactor delivery rate estimation

ghedo:
The delivery rate estimation code is fairly self-contained, so it can be
moved into its own submodule pretty easily.


ghedo:
@junhochoi yes, good point. I meant  to do move the tests as well but then apparently I forgot. Did that now.

PR #459, opened 2020-04-13
increase anti-amplification limit earlier

ghedo:
As per the comment, this needs to be done as early as possible, so that
we still increase the limit in case of errors and the like. This is
fine, as we only need to keep track of how many bytes the peer sent to
us, whether valid QUIC bytes or not.

This makes it possible to convert a couple other cases where incoming
packets are ignored to use `drop_pkt_on_err()` for consistency.

After this, we _might_ want to remove
db5e6acd59a4d15eb6f79a68d315a60bd729570e, though not sending
CONNECTION_CLOSE on the first invalid packet received might still be a
good idea (e.g. for DDoS prevention purposes).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1289,6 +1289,17 @@ impl Connection {
     pub fn recv(&mut self, buf: &mut [u8]) -> Result<usize> {
         let len = buf.len();
 
+        // Keep track of how many bytes we received from the client, so we
+        // can limit bytes sent back before address validation, to a multiple
+        // of this. The limit needs to be increased early on, so that if there
+        // is an error there is enough credit to send a CONNECTION_CLOSE.
+        //
+        // It doesn't matter if the packets received were valid or not, we only
+        // need to track the total amount of bytes received.
+        if !self.verified_peer_address {
+            self.max_send_bytes += len * MAX_AMPLIFICATION_FACTOR;
+        }
+
         let mut done = 0;
         let mut left = len;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1461,7 +1461,11 @@ impl Connection {
         //
         // TODO: buffer packets instead of discarding as an optimization.
         if hdr.ty == packet::Type::Short && !self.is_established() {
-            return Ok(b.len());
+            return Err(drop_pkt_on_err(
+                Error::InvalidPacket,
+                self.recv_count,
+                &self.trace_id,
+            ));
```
junhochoi, line 1468: Now it doesn't return the length. Is there any side effect?

ghedo: So, this is s a bit tricky. I have another PR coming later with additional changes related to this, though I need to do some more tests and I split them to make it easier to review them.

The problem here is that `Ok(b.len())` and `Err(Error::Done)` should mean the same thing in theory, that is "ignore the rest of the UDP datagram".

However, over time `Err(Error::Done)` has also interpreted by applications to mean "stop reading incoming packets from the socket", which is, I realized, the wrong thing to do, as the application shouldn't stop reading from the socket ever as it might lead to a deadlock as the socket might end up never being re-armed by the event loop.

So what I want to do is:
1. Unify how packets are ignored by `recv_single()` (that is, via `drop_pkt_on_err()` and returning `Error::Done`).
2. Change `recv()` to basically treat `Error::Done` from `recv_single()` as if `Ok(b.len())` is returned (and of course update the documentation to make this clearer).

The second part means that `recv()` will not return `Error::Done` anymore so applications don't need to be changed to ignore that return value (and it also simplifies the API a bit).

(In theory we could have gone the other way, that is, alsways return `Ok(b.len())` from `recv_single()`, but that's not always easy to do).

PR #461, opened 2020-04-14
don't return Error::Done from recv()

ghedo:
This changes the `recv()` public method to never return `Error::Done` to
applications, due to confusion of its meaning and misuse.

Many applications (including the quiche-apps and quiche examples) have
been interpreting the Done error as a signal to stop reading incoming
packets from a socket. However this is wrong, as applications shouldn't
stop reading incoming packets, as that might cause a deadlock where the
socket is not re-armed by the event loop and even if a connection is
closing, new packets should still be processed as the socket might be
shared among multiple connections.

Instead of `Error::Done`, `recv()` will now return an `OK(...)` value
with the full size of the incoming buffer, to mean that the entire
buffer has been processed (or ignored), which was the intended meaning
of Done.

---

This also updates the examples and quiche-apps to not stop processing incoming packets on error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,6 @@ loop {
     let read = match conn.recv(&mut buf[..read]) {
         Ok(v) => v,
 
-        Err(quiche::Error::Done) => {
-            // Done reading.
-            break;
-        },
-
         Err(e) => {
             // An error occurred, handle it.
             break;
--- a/examples/client.c
+++ b/examples/client.c
@@ -110,18 +110,15 @@ static void recv_cb(EV_P_ ev_io *w, int revents) {
 
         ssize_t done = quiche_conn_recv(conn_io->conn, buf, read);
 
-        if (done == QUICHE_ERR_DONE) {
-            fprintf(stderr, "done reading\n");
-            break;
-        }
-
         if (done < 0) {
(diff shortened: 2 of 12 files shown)


LPardue:
`tests::early_1rtt_packet` is causing the CI failure so needs a tweak

ghedo:
@LPardue yes, fixed now.

LPardue:
I really like this change because it makes things clearer for library users. 

My main question with landing it is, how much testing has been done to provide confidence it doesn't accidentally change some behaviour? This made me realise that the http3_test tool also needs to be updated: namely `tools/http3_test/src/runner.rs`

junhochoi:
LGTM. Just a quick question. I think `recv()` is able to return `Ok(0)`? (app can ignore it safely though)

ghedo:
@LPardue well, there is a behavior change, that's the whole point :P though the cases where this has any impact are fairly rare, so there is no change in normal operations.

@junhochoi yeah, that's fine. Applications generally don't need to check the actual return value if `Ok(...)` is returned. It was originally done to allow applications to process coalesced packets, but now we do it automatically internally. Maybe we can change the return type in the future, but it's not too important right now.

PR #462, opened 2020-04-14
Qlog data moved

LPardue:
Opening mainly to get some eyes on.

Qlog draft 02 has a data_moved event that allows logging when stream data is moved into or out of the transport. Logging this information allows us in the qvis multiplexing tool to see how the quiche-client reads data from the transport layer.

Our definition of data_moved is bit old, so this change tweaks it to contain the length. In testing I found that `SendBuf.off()` always returned `0` while the qvis expects the offset to move along at the stream offset grows. That is why I added the `cumulative_offset()` function as a hack. Happy to hear a better solution to this.

Right now data_moved is in the category HTTP/3 but that will change to be transport soon. So we can rename things in a separate followup PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2162,7 +2162,7 @@ impl Connection {
                     None => continue,
                 };
 
-                let off = stream.send.off();
+                let off = stream.send.off_front();
 
                 // Try to accurately account for the STREAM frame's overhead,
                 // such that we can fill as much of the packet buffer as
@@ -2426,6 +2426,9 @@ impl Connection {
             return Err(Error::Done);
         }
 
+        #[cfg(feature = "qlog")]
+        let offset = stream.recv.off_back();
+
         let (read, fin) = stream.recv.pop(out)?;
 
         self.max_rx_data_next = self.max_rx_data_next.saturating_add(read as u64);
@@ -2447,6 +2450,18 @@ impl Connection {
         if complete {
             self.streams.collect(stream_id, local);
         }
+
+        qlog_with!(self.qlog_streamer, q, {
+            let ev = qlog::event::Event::h3_data_moved(
+                stream_id.to_string(),
+                Some(offset.to_string()),
+                Some(read as u64),
+                Some(qlog::H3DataRecipient::Transport),
+                None,
+                None,
+            );
+            q.add_event(ev).ok();
+        });
 
         Ok((read, fin))
     }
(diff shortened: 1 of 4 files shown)


src/stream.rs
```diff
@@ -769,6 +769,11 @@ impl RecvBuf {
         Ok(())
     }
 
+    /// Returns the cumulative offset of data buffered.
+    pub fn cumulative_off(&self) -> u64 {
```
ghedo, line 773: This name sounds wrong. It's not really "cumulative" as there's just one offset, so maybe just call it `off()`?

src/stream.rs
```diff
@@ -1009,6 +1014,11 @@ impl SendBuf {
...
```
ghedo, line 1018: Same.

tools/qlog/src/lib.rs
```diff
@@ -1456,6 +1456,7 @@ pub enum EventData {
     H3DataMoved {
         stream_id: String,
         offset: Option<String>,
+        length: Option<u64>,
```
ghedo, line 1459: What's the logic behind using `String` in some places but `u64` in others? IIRC you said `String` is for big numbers that can't be encoded in JSON, but using both makes the API pretty confusing.

Not really related to this PR, but in the future we should just make all these fields `u64` and then tag the ones that should be encoded as string with the `#[serde(serialize_with = "u64_to_str")] ` attribute, where `u64_to_str` is a function we implement to automatically encode specific fields as string. This way at least the API exposed by the `qlog` crate doesn't leak limitations in the underlying encoding format.

LPardue: The qlog crate's datamodel follows closely the real qlog datamodel, which has inconsistencies like this throughout the schema. 

I believe these will be fixed across the board in qlog-02, so we can fix stuff up then and consider custom serialize/deserialize functions.

src/stream.rs
LPardue: as per discussion, renamed the old `off()` method to `off_front()` and `cumulative_off()` to `off_back()`.

src/lib.rs
```diff
@@ -2448,6 +2451,18 @@ impl Connection {
             self.streams.collect(stream_id, local);
         }
 
+        qlog_with!(self.qlog_streamer, q, {
```
ghedo, line 2454: Actually, wait, I just realized that this is in the transport code. I don't think we should be logging HTTP/3 events from here...

ghedo: Ah nvm, I read the PR description now.

PR #463, opened 2020-04-14
tools: improve HTTP/3 request processing

LPardue:
I wanted to add some tests that use querystring parameters and noticed that we actually fail to respond to anything that includes one. 

This change adds a few basic improvements to the area of request processing: it validates the scheme, it reconstructs and validates the URL of the object, and finally pulls out just path element of the URL which is used to retrieve the file.

A follow up PR can build on this by pulling out the querystring element now the groundwork is laid down.

### DIFF
--- a/tools/apps/src/lib.rs
+++ b/tools/apps/src/lib.rs
@@ -587,13 +587,16 @@ impl Http3Conn {
         let mut reqs = Vec::new();
         for url in urls {
             for i in 1..=reqs_cardinal {
+                let authority = match url.port() {
+                    Some(port) => format!("{}:{}", url.host_str().unwrap(), port),
+
+                    None => url.host_str().unwrap().to_string(),
+                };
+
                 let mut hdrs = vec![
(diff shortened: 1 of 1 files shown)


tools/apps/src/lib.rs
```diff
@@ -686,11 +699,24 @@ impl Http3Conn {
...
+                quiche::h3::Header::new("server", "quiche"),
+            ];
+
+            return (headers, b"Invalid scheme".to_vec());
+        }
+
+        let url = format!("{}://{}{}", scheme, host, path);
+        let url = url::Url::parse(&url).unwrap();
+
+        let mut pathbuf = path::PathBuf::from(url.path());
```
ghedo, line 714: Instead of making this mutable I think you can just do:
```rust
         let pathbuf = path::PathBuf::from(url.path());
         let pathbuf = autoindex(pathbuf, index);
```
?

LPardue: thanks, fixed

PR #464, opened 2020-04-15
delivery_rate: fmt

ghedo:


src/recovery/delivery_rate.rs
```diff
@@ -247,7 +255,9 @@ mod tests {
             is_app_limited: false,
         };
 
-        recvry.delivery_rate.on_packet_sent(&mut pkt_1, Instant::now());
+        recvry
```
LPardue, line 258: nit: typo in `recovry`

ghedo: :man_shrugging:  that's how the variable is called in the pre-existing test...

PR #465, opened 2020-04-16
WIP: H3 stream priority

LPardue:
This builds on top of the QUIC-level priority support WIP PR, to see the relevant changes to the H3 layer see commit https://github.com/cloudflare/quiche/commit/347b0a846320901474f1d9d2421acd238280aa55.

This is marked as WIP because the code and API would benefit from some input.

See the commit message for more context (and typos) but basically this change marks all requests streams with the same default priority. When responses have the same urgency it causes them to be served in FIFO order. A good simple way to test this is to download a large file and feed the qlog to qvis mutliplex view.

```
RUST_LOG=trace QLOGDIR=$PWD/logs ./target/debug/quiche-client --no-verify -n5 "https://localhost:4433/5MB.png"
```

It is more interesting to test different urgencies and incremental delivery. So to enable that, I borrowed Tatsuhiro Tsujikawa's idea of using query strings e.g `u=2,i=1`. 

We can repeat the above example and invert the order of response with something like:
```
RUST_LOG=trace QLOGDIR=$PWD/logs ./target/debug/quiche-client --no-verify "https://localhost:4433/5MB.png?u=6" "https://localhost:4433/5MB.png?u=5" "https://localhost:4433/5MB.png?u=4" "https://localhost:4433/5MB.png?u=3" "https://localhost:4433/5MB.png?u=2"
```

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -728,15 +728,33 @@ impl Connection {
             return Err(Error::FrameUnexpected);
         }
 
-        match self.streams.get(&stream_id) {
-            Some(s) =>
+        let http_priority = match self.streams.get(&stream_id) {
+            Some(s) => {
                 if !s.local_initialized() {
                     return Err(Error::FrameUnexpected);
-                },
+                }
+
+                &s.priority
+            },
 
             None => {
                 return Err(Error::FrameUnexpected);
             },
+        };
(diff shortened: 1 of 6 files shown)


LPardue:
Closing in favour of #518.

PR #466, opened 2020-04-18
Fix http3-server.c example early connection 'fin'

gfleury:
Fix http3-server.c example early connection 'fin', subsequent call to quiche_h3_send_body seems useless.


ghedo:
Merged (with a tweaked commit message), thanks!

PR #468, opened 2020-04-19
Recovery fixes

ghedo:
Just a few small recovery fixes and improvements.

PR #469, opened 2020-04-20
recovery: refactor sent packets tracking

ghedo:
Instead of using a BTreeMap, use a VecDeque to track sent packets, which
makes it easier to keep track of acked and lost packets for a short time
after they have been acked or declared lost instead of immediately
removing them from the list, which in turn will allow detecting things
like DSACKs and spurious losses.

Similarly to how the BTreeMap was used, sent packets are appended to the
back of the queue (so they end up being sorted by packet number / sent
time). Access now happen by iterating over the queue from the start
rather than looking up specific packets by packet number. When acked or
declared lost, the packets are simply marked accordingly _without_ being
removed from the queue.

At predefined times, the semt packets marked as acked or lost are then
removed from the start of the queue in contiguous blocks, in order to
avoid having to copy elements around to cover gaps in the queue caused
by removed elements. Later this can be changed to e.g. delay cleaning of
lost packets for 1 RTT after they were marked as lost.

This new strcture also allows us to follow the -recovery draft
pseudo-code somewhat more closely, and as a side-effect, there are no
more potentially expensive BTreeMap lookups in hot codepaths, which
might help reduce CPU usage when e.g. processing ACK frames (flamegraphs
generated on my laptop show that the time spent inside
Recovery::on_ack_received() is cut in half).

---

This is based on https://github.com/cloudflare/quiche/pull/468. Note that before merging we should do more testing in the lab to make sure this doesn't have any adverse side-effects.

### DIFF
--- a/src/recovery.rs
+++ b/src/recovery.rs
@@ -31,7 +31,7 @@ use std::str::FromStr;
 use std::time::Duration;
 use std::time::Instant;
 
-use std::collections::BTreeMap;
+use std::collections::VecDeque;
 
 use crate::Config;
 use crate::Error;
@@ -91,7 +91,7 @@ pub struct Recovery {
 
     loss_time: [Option<Instant>; packet::EPOCH_COUNT],
 
-    sent: [BTreeMap<u64, Sent>; packet::EPOCH_COUNT],
+    sent: [VecDeque<Sent>; packet::EPOCH_COUNT],
 
     pub lost: [Vec<frame::Frame>; packet::EPOCH_COUNT],
 
@@ -149,7 +149,7 @@ impl Recovery {
 
             loss_time: [None; packet::EPOCH_COUNT],
 
-            sent: [BTreeMap::new(), BTreeMap::new(), BTreeMap::new()],
+            sent: [VecDeque::new(), VecDeque::new(), VecDeque::new()],
 
             lost: [Vec::new(), Vec::new(), Vec::new()],
 
@@ -193,7 +193,7 @@ impl Recovery {
         self.largest_sent_pkt[epoch] =
             cmp::max(self.largest_sent_pkt[epoch], pkt_num);
 
-        self.sent[epoch].insert(pkt.pkt_num, pkt);
+        self.sent[epoch].push_back(pkt);
 
         if in_flight {
             if ack_eliciting {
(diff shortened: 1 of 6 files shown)


junhochoi:
> While the CC will not have access to the full acked packet information
> anymore, I believe that simply passing the cumulative acked bytes count
> and largest acked sent time should be enough for both Reno and CUBIC
> (and Hystart as well, though need to check that more carefully). Not
> sure about BBR though.

I think r/w access to the packet is required for various congestion control related implementation. For example CUBIC (#475) need `packet.time` and `packet.size` which can be covered in this PR but HyStart++ (#476) need a new `packet.pkt_num` of the packet to detect the RTT round. Also BBR need to access/update additional fields such as packet.{delivered,delivered_rate,rtt} etc - see ttps://tools.ietf.org/html/draft-cardwell-iccrg-bbr-congestion-control-00

ghedo:
> HyStart++ (#476) need a new packet.pkt_num of the packet to detect the RTT round

The new `on_packets_acked()` callback has access to the `largest_acked_pkt` field in `Recovery` like any other callback, is that not enough? We could also explicitly pass the largest acked number from the ACK frame itself, but that doesn't seem to be needed.

It also seems to me that Linux works the same way as this PR, that is, it first processes SACKs in a loop, and then calls `tcp_cong_control()` once per-ACK packet, so their Hystart implementation needs to work that way too, no? Is Hystart++ so different from Hystart that it cannot work like that?

Though, in the worst case scenario, each callback can still loop through the `r.sent[epoch]` map for packets that have `time_acked.is_some()`, though instead of doing that always like we do now, they could do that just when they actually need it.

> Also BBR need to access/update additional fields such as packet.{delivered,delivered_rate,rtt} etc

Why would it need to update information of acked packets? Like, it seems pointless to modify the `Sent` struct within `on_packet_acked()` / `on_packets_acked()` given that the packet is being removed from the sent list. Do you mean for delivery rate estimation? AFAICT that is done separately from `on_packet_acked()` / `on_packets_acked()` right now, will BBR need to change that?

But in any case I think the point from before applies, Linux seems to follow the same process as what is implemented in this PR, so if they can implement BBR like that why can't we?

junhochoi:
I think now I understand what you try to say.

First of all I don't think we want to follow what linux tcp does - different OS has different tcp implementation. However while I agree we may not need to access `Sent` structure of acked packet fully, it's still valuable to follow the spec first. For BBR, I need to correctly myself that we don't need to update fields on the acked packet.

I will add a comment in the code instead.

> It also seems to me that Linux works the same way as this PR, that is, it first processes SACKs in a loop, and then calls tcp_cong_control() once per-ACK packet, so their Hystart implementation needs to work that way too, no? Is Hystart++ so different from Hystart that it cannot work like that?

In linux tcp, it pass the number of packet acked by this ack additionally. Hystart++ need the same information but it's ok for now because current code (not this PR) iterates over acked packet so we can get acked packet count from there.

src/recovery.rs
```diff
@@ -277,12 +283,30 @@ impl Recovery {
...
+            };
+
+            self.update_rtt(latest_rtt, ack_delay, now);
+        }
+
+        // Detect and mark lost packets without removing them from the sent
+        // packets list.
         self.detect_lost_packets(epoch, now, trace_id);
 
+        self.on_packets_acked(acked_bytes, largest_newly_acked_sent_time, now);
```
junhochoi, line 302: So for me it doesn't match to current recovery pseudocode

    OnPacketsAcked(newly_acked_packets)

Because here we only pass the information of how many bytes acked and sent time of latest packet.

`newly_acked_packets` is a list of acked packet and later it is used to iterate over as follows:

https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-recovery.md#on-packet-acknowledgement

Also note that current quiche CC code assumes `on_packet_acked` hook is called for each acked packet, not only for the last one.

This will make a difference because CC state can be changed per acked packet.

junhochoi: For example newly_acked_packet has 5 packets, each ack'ing 1000 bytes, starting with cwnd = 1000 ssthresh = 3000 and assume Reno.
If we follow the current code, on_packets_acked(5000, <time>, now) will result in
cwnd = 6000 ssthresh = 3000
But in the current code, CC will switch to congestion avoidance when cwnd reaches to 3000, so the result will be
cwnd = 3908 ssthresh = 3000

src/recovery.rs
```diff
@@ -702,13 +741,18 @@ impl std::fmt::Debug for Recovery {
...
 
+#[derive(Clone)]
 pub struct Sent {
     pub pkt_num: u64,
 
     pub frames: Vec<frame::Frame>,
 
     pub time_sent: Instant,
 
+    pub time_acked: Option<Instant>,
```
junhochoi, line 752: Looks there is `Instant` is not used in the code. Is it good enough to be `bool`?

src/recovery.rs
```diff
@@ -702,13 +741,18 @@ impl std::fmt::Debug for Recovery {
...
+
+    pub time_lost: Option<Instant>,
```
junhochoi, line 754: Same to time_acked. Is it good enough to be bool?

src/recovery.rs
ghedo: Yes, you are right, I missed that part. Should be fixed now, though one of the CUBIC tests now fails, will need to investigate. Though if I understand RFC8312 right, the CUBIC congestion avoidance should only be invoked once per-ACK, rather than for each acked packet?

junhochoi: CUBIC (or reno or other cc) cwnd update is per acked packet in QUIC. I don't understand why each CC code need a change. I think simply following the pseudo code of on_packets_acked() (iterating over acked packet.. and we already have on_packet_acked() per packet so simply calling it multiple times) will be good enough for now?

ghedo: Because in that case we need to create temporary vectors with the acked packets which is a potential DoS vector.

In any case I changed the code to do just that now. We can reconsider this later on.

src/recovery.rs
ghedo: This is needed for adaptive thresholds.

src/recovery.rs
ghedo: This is mostly to be consistent with the `time_lost`, though as per the PR description this can later be used for other things. There's not much harm in leaving it like this I think.

src/recovery.rs
```diff
@@ -532,35 +577,43 @@ impl Recovery {
...
+        self.drain_packets(epoch);
+    }
 
-                self.delivery_rate.on_ack_received(p, now);
+    fn drain_packets(&mut self, epoch: packet::Epoch) {
+        let mut lowest_non_expired_pkt_index = self.sent[epoch].len();
+
+        // In order to avoid removing elements from the middle of the list
+        // (which would require copying other elements to compact the list),
+        // we only remove a contguous range of elements from the start of the
```
LPardue, line 592:
```suggestion
        // we only remove a contiguous range of elements from the start of the
```

src/recovery.rs
```diff
@@ -532,35 +577,43 @@ impl Recovery {
...
+        // list.
+        //
+        // This means that acked or lost elements coming after this will not
+        // be removed at this point, but their removal is delayed for a later
+        // time, once the gaps have been filled.
+
+        // First, find the first element that is neither acked not lost.
```
LPardue, line 599: nit: grammar of `neither acked not lost`

src/recovery.rs
```diff
@@ -1116,7 +1182,7 @@ mod tests {
             Ok(())
         );
 
-        assert_eq!(r.sent[packet::EPOCH_APPLICATION].len(), 1);
+        assert_eq!(r.sent[packet::EPOCH_APPLICATION].len(), 2);
```
junhochoi, line 1185: before calling on_ack_received() sent[] is [0, 1, 2, 3] (I only use packet number here, not ranges) and acked[] = [0, 1, 3]. Then after calling on_ack_received(), sent[] should be [2] I think? But when I checked sent[] at this line it contains [2, 3].

src/recovery.rs
ghedo: Fixed, thanks.

src/recovery.rs
ghedo: Fixed, thanks.

src/recovery.rs
ghedo: That's on purpose. From the commit message/PR description:
> At predefined times, the semt packets marked as acked or lost are then
removed from the start of the queue in contiguous blocks, in order to
avoid having to copy elements around to cover gaps in the queue caused
by removed elements.

ghedo: That is, we only remove packets from the start of the list, so if there are gaps like in this case the packets are not actually removed, but are still marked as lost or acked so they are ignored.

junhochoi: Ok. It would be great to add some comment there, not to be confused later. Thanks!

PR #470, opened 2020-04-20
Adaptive reordering thresholds

ghedo:
This implements adaptive reordering threshold as suggested in:
https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#section-5.1.1.

The algorithm is similar to what Chrome uses as well.

I did a quick test on my laptop with `netem delay 10ms 1ms 10%` and this seems to improve things a fair amount:

Before:
```
finished in 31.92s, 0.03 req/s, 321.44KB/s
requests: 1 total, 1 started, 1 done, 1 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 10.02MB (10506459) total, 97B (97) headers (space savings 47.57%), 10.00MB (10485760) data
```

After:
```
finished in 1.85s, 0.54 req/s, 5.40MB/s
requests: 1 total, 1 started, 1 done, 1 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 10.00MB (10488607) total, 97B (97) headers (space savings 47.57%), 10.00MB (10485760) data
```

Though we will need more rigorous tests in lab. In particular, I'm not super confident the code to undo the cwnd update is correct. We will also need a time reordering unit test.

This depends on https://github.com/cloudflare/quiche/pull/468 and https://github.com/cloudflare/quiche/pull/469.

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -43,9 +43,11 @@ use crate::packet;
 use crate::ranges;
 
 // Loss Recovery
-const PACKET_THRESHOLD: u64 = 3;
+const INITIAL_PACKET_THRESHOLD: u64 = 3;
 
-const TIME_THRESHOLD: f64 = 9.0 / 8.0;
+const MAX_PACKET_THRESHOLD: u64 = 20;
+
+const INITIAL_TIME_THRESHOLD: f64 = 9.0 / 8.0;
 
 const GRANULARITY: Duration = Duration::from_millis(1);
 
@@ -97,6 +99,8 @@ pub struct Recovery {
 
     pub lost_count: usize,
 
+    pub lost_spurious_count: usize,
+
     pub loss_probes: [usize; packet::EPOCH_COUNT],
 
     in_flight_count: [usize; packet::EPOCH_COUNT],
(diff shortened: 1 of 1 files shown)


src/recovery.rs
```diff
@@ -238,24 +255,50 @@ impl Recovery {
...
+
+                    // Calculate new time reordering threshold.
+                    let loss_delay = max_rtt.mul_f64(self.time_thresh);
+                    if now.duration_since(unacked.time_sent) > loss_delay {
+                        // TODO: do time threshold update
+                        self.time_thresh = 5_f64 / 4_f64;
+                    }
+
+                    // Undo congestion window update.
+                    self.congestion_window = self.congestion_window_prev;
```
goelvidhi, line 295: perhaps,
`self.congestion_window = max(self.congestion_window, self.congestion_window_prev)`

junhochoi:
> I did a quick test on my laptop with netem delay 10ms 1ms 10% and this seems to improve things a fair amount:

Have you tested with `netem reorder..`?

src/recovery.rs
ghedo: Yes, thank you.

ghedo:
Rebased and updated based on https://github.com/cloudflare/quiche/pull/893.

junhochoi (approved):
I think we still need to tune the algorithm (or threshold. ideally we need to increase a time/packet threshold based on the previous state) for some case (e.g. in a high bandwidth, the benefit is lower), but still better when there is a reodering happening. 👍

PR #475, opened 2020-04-21
CUBIC congestion control

junhochoi:
- Implemented based on RFC8312 "CUBIC for Fast Long-Distance Networks"
- CUBIC is the default CC algorithm.
- CUBIC specific variables is stored to cubic::State. I added
  this to Recovery structure itself. If you use Reno only
  this will be a waste of memory if you use Reno, but the overhead
  is small (4 fields) for now. It can be dynamically allocated later.
- `f64` is used for various calculations. Optimization using integer ops is a future task.
- Use libm::cbrt to avoid linking with -lm for C applications. Note that libm:cbrt
  is slower than C cbrt() which is asm-optimized (roughtly 2x when I
  test in MacOS) but the difference is small.
- For config API, to use Cubic explicitly for the connection, call the following:
  Rust: config.set_cc_algorithm(recovery::CongestionControlAlgorithm::Cubic);
  C: quiche_config_set_cc_algorithm(config, QUICHE_CC_CUBIC);
- `CongestionControlAlgorithm` exported as public. Also fixed the documentation.
- CongestionControlOps: add collapse_cwnd()

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ cmake = "0.1"
 [dependencies]
 log = { version = "0.4", features = ["std"] }
 libc = "0.2"
+libm = "0.2"
 ring = "0.16"
 lazy_static = "1"
 qlog = { version = "0.2", path = "tools/qlog", optional = true }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -257,7 +257,7 @@
 //! Note that the CC algorithm should be configured before calling [`connect()`]
 //! or [`accept()`]. Otherwise the connection will use a default CC algorithm.
 //!
-//! [`CongestionControlAlgorithm`]: cc/enum.Algorithm.html
+//! [`CongestionControlAlgorithm`]: enum.CongestionControlAlgorithm.html
 
 #![allow(improper_ctypes)]
 #![warn(missing_docs)]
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -55,6 +55,7 @@ cmake = "0.1"
 [dependencies]
 log = { version = "0.4", features = ["std"] }
 libc = "0.2"
+libm = "0.1"
```
ghedo, line 58: The latest version of `libm` is `0.2.1` so this should be `0.2`
```suggestion
libm = "0.2"
```

src/lib.rs
```diff
@@ -257,7 +257,8 @@
 //! Note that the CC algorithm should be configured before calling [`connect()`]
 //! or [`accept()`]. Otherwise the connection will use a default CC algorithm.
 //!
-//! [`CongestionControlAlgorithm`]: cc/enum.Algorithm.html
+//! [`CongestionControlAlgorithm`]:
+//! recovery/enum.CongestionControlAlgorithm.html
```
ghedo, line 261: This just ends up redirecting to the public alias anyway, so might as well point the docs to that.
```suggestion
//! enum.CongestionControlAlgorithm.html
```

ghedo:
I applied the nits manually since they are pretty minor. Will merge after CI is green.

PR #476, opened 2020-04-21
HyStart++

junhochoi:
**Note that it's based on #475 since this PR includes changes for CUBIC as well.**

Implements Hystart++ based on IETF draft:
  https://tools.ietf.org/html/draft-balasubramanian-tcpm-hystartplusplus-02

HyStart++ is an improved version of original Hystart. It can be used
with Reno and CUBIC (or possibly other loss based congestion control).
It's enabled by default. You can disable it by calling config API provided.

Config APIs:
- C: quiche_config_set_hystart()
- Rust: Config.set_hystart()

API Changes:
- CongestionControlOps API: `epoch` is added to on_packet_acked()
  and congestion_event()

Tools Update:
- tools/apps/quiche-{client,server}: added --disable-hystart
  command line option

Different from TCP: TCP implementation uses sequence number to
check a start of the round. In QUIC, current implementation uses
packet number instead. However packet number space is unique
per epoch but Initial and Handshake epoch is short, so only
applies HyStart++ to Application epoch.

In TCP, sequence number is reused for retransmitted packets
but QUIC will use a new packet number for retransmit so
this is not exactly same. But it should be fine because this is
only for the slow start episode and when QUIC connection sees a
packet loss, it enters into recovery episode and exits from slow
start, so retransmitted packet number is not used during HyStart++
slow start.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -172,6 +172,9 @@ enum quiche_cc_algorithm {
 // Sets the congestion control algorithm used.
 void quiche_config_set_cc_algorithm(quiche_config *config, enum quiche_cc_algorithm algo);
 
+// Configures whether to use HyStart++.
+void quiche_config_set_hystart(quiche_config *config, bool v);
+
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
 
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -173,7 +173,7 @@ enum quiche_cc_algorithm {
 void quiche_config_set_cc_algorithm(quiche_config *config, enum quiche_cc_algorithm algo);
 
 // Configures whether to use HyStart++.
-void quiche_config_set_hystart(quiche_config *config, bool v);
+void quiche_config_enable_hystart(quiche_config *config, bool v);
 
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -254,6 +254,11 @@ pub extern fn quiche_conn_set_qlog_fd(
 }
 
 #[no_mangle]
+pub extern fn quiche_config_set_hystart(config: &mut Config, v: bool) {
+    config.set_hystart(v);
+}
+
+#[no_mangle]
 pub extern fn quiche_config_free(config: *mut Config) {
     unsafe { Box::from_raw(config) };
 }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -254,8 +254,8 @@ pub extern fn quiche_conn_set_qlog_fd(
 }
 
 #[no_mangle]
-pub extern fn quiche_config_set_hystart(config: &mut Config, v: bool) {
-    config.set_hystart(v);
+pub extern fn quiche_config_enable_hystart(config: &mut Config, v: bool) {
+    config.enable_hystart(v);
 }
 
 #[no_mangle]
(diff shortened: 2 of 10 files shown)


junhochoi:
Rebase to mater.

src/lib.rs
```diff
@@ -722,6 +725,13 @@ impl Config {
     pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
         self.cc_algorithm = algo;
     }
+
+    /// Configures whether to enable HyStart++
```
LPardue, line 729:
```suggestion
    /// Configures whether to enable HyStart++.
```

junhochoi:
@ghedo let me know if there is additional comments/work needed. I rebased to master again today.

src/lib.rs
```diff
@@ -722,6 +725,13 @@ impl Config {
     pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
         self.cc_algorithm = algo;
     }
+
+    /// Configures whether to enable HyStart++.
+    ///
+    /// The default value is `true`.
+    pub fn set_hystart(&mut self, v: bool) {
```
ghedo, line 732:
```suggestion
    pub fn enable_hystart(&mut self, v: bool) {
```

src/ffi.rs
```diff
@@ -253,6 +253,11 @@ pub extern fn quiche_conn_set_qlog_fd(
     );
 }
 
+#[no_mangle]
+pub extern fn quiche_config_set_hystart(config: &mut Config, v: bool) {
```
ghedo, line 257:
```suggestion
pub extern fn quiche_config_enable_hystart(config: &mut Config, v: bool) {
```

src/ffi.rs
```diff
@@ -253,6 +253,11 @@ pub extern fn quiche_conn_set_qlog_fd(
...
+    config.set_hystart(v);
```
ghedo, line 258:
```suggestion
    config.enable_hystart(v);
```

include/quiche.h
```diff
@@ -172,6 +172,9 @@ enum quiche_cc_algorithm {
 // Sets the congestion control algorithm used.
 void quiche_config_set_cc_algorithm(quiche_config *config, enum quiche_cc_algorithm algo);
 
+// Configures whether to use HyStart++.
+void quiche_config_set_hystart(quiche_config *config, bool v);
```
ghedo, line 176:
```suggestion
void quiche_config_enable_hystart(quiche_config *config, bool v);
```

tools/apps/src/bin/quiche-client.rs
```diff
@@ -134,6 +135,10 @@ fn main() {
         config.log_keys();
     }
 
+    if conn_args.disable_hystart {
+        config.set_hystart(false);
```
ghedo, line 139:
```suggestion
        config.enable_hystart(false);
```

tools/apps/src/bin/quiche-server.rs
```diff
@@ -126,6 +127,10 @@ fn main() {
         .set_cc_algorithm_name(&conn_args.cc_algorithm)
         .unwrap();
 
+    if conn_args.disable_hystart {
+        config.set_hystart(false);
```
ghedo, line 131:
```suggestion
        config.enable_hystart(false);
```

junhochoi:
Cool. Thanks!

PR #482, opened 2020-04-27
chore:h3:correct doc typo

detailyang:
AFAIK, There is no pseudo-header like `:user-agent` in [draft-ietf-quic-http.html#section-4.1.1.1](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-4.1.1.1)


ghedo:
Merged, thanks!

PR #487, opened 2020-04-29
docker: build images from the local git checkout

ghedo:
Currently the Docker images are built from a subdirectory, which
prevents the Dockerfile from accessing files outside of the directory.

This in turn means that in order to build the quiche-apps, the
Dockerfile needs to clone a whole new copy of quiche via git and build
it from scratch.

This is very inefficient (slowing down CI builds), and it also prevents
building images from a non-master branch for local testing.

This change simply shuffles files around to allow the Docker images to
be built from the pre-existing checkout of the quiche repository.

---

I tested the base and qns images locally and they seem to work fine. Haven't tried to push them though since I don't have permissions.


junhochoi:
I think we still need to build it inside the docker container to get the same build environment not depending on host OS. e.g. If I build this on my MacOS laptop it will result copying MacOS binary into Docker images which won't work.

LPardue:
Maybe in the original dockerfile, the `build` image can take a copy of local files e.g. `COPY ../ /path/in/container` and then we call cargo build from there. In the ideal case the artifacts between host and container align and no extra work is required

junhochoi:
> Maybe in the original dockerfile, the `build` image can take a copy of local files e.g. `COPY ../ /path/in/container` and then we call cargo build from there. In the ideal case the artifacts between host and container align and no extra work is required

In such case (already checked out locally and want to run `cargo build` inside the container), we can use docker volume mount `docker -v src:/build/quiche/src' to map host directory into the container instead of COPY

ghedo:
@junhochoi AFAICT there's no way to mount a volume during build. `COPY` seems to work fine, so I added a commit to do that. Seems to work fine locally.

ghedo:
The downside is that building quiche-apps for yet another time within docker increases the build time by 4 minutes compared to reusing the already built apps.

junhochoi:
Hmm.. volume mount seems to be Enterprise only feature. https://github.com/travis-ci/worker/issues/171

junhochoi:
> The downside is that building quiche-apps for yet another time within docker increases the build time by 4 minutes compared to reusing the already built apps.

If this is a concern we can move docker image build to dockerhub, not in travis.

ghedo:
@junhochoi it's not a blocker for merging this PR I think given that it provides some benefits, though that doesn't prevent us from looking at improvements later on.

The main difference in terms of build times from this PR is that the docker images are now built as part of the _build_ jobs rather than only the _deploy_ ones. I made this change to ensure that docker builds are tested properly before merging PRs, though we might later decide to revert this part.

Makefile
```diff
@@ -0,0 +1,55 @@
...
+# build fuzzing image
+.PHONY: docker-fuzz
+docker-fuzz: build-fuzz
+	$(DOCKER) build --tag $(FUZZ_REPO):$(FUZZ_TAG) fuzz
+
+.PHONY: docker-fuzz-publish
+docker-fuzz-publish:
+	$(DOCKER) push $(FUZZ_REPO):$(FUZZ_TAG)
+
+.PHONY: clien
```
junhochoi, line 50:
```suggestion
.PHONY: clean
```

Makefile
```diff
@@ -0,0 +1,55 @@
...
+.PHONY: docker-qns
+docker-qns: Dockerfile tools/qns/run_endpoint.sh
+	$(DOCKER) build --target quiche-qns -t $(QNS_REPO):$(QNS_TAG) .
+
+.PHONY: docker-publish
+docker-publish:
+	$(DOCKER) push $(BASE_REPO):$(BASE_TAG)
+	$(DOCKER) push $(QNS_REPO):$(QNS_TAG)
+
+# build fuzzers
```
junhochoi, line 34: When will those fuzz targets run?

ghedo: I run these manually for now, since we don't really want to stop the fuzz runs for every commit.

PR #490, opened 2020-04-30
lib: fix app_limited to allow cwnd growth

junhochoi:
There is a problem when sender has some leftover data cannot be
sent as a QUIC packet, app_limited is true so cwnd doesn't
grow well, especially during congestion avoidance.

- max_packet_size=1200
- at send(): cwnd=1220, flight_size=0, send a full packet, 1200 bytes
- at send(): cwnd=1220, flight_size=1200
   we are only able to send 20 bytes, but it does't fit even for
   QUIC headers (usually larger than 20 bytes) and
   send() returns Error::Done.
   In this case, flight_size < cwnd so app_limited is still true.

However the app already send as much as possible to cwnd and
can't send more. This case should be considered
as app_limited=false.

The fix is to check and update app_limited when we return Error::Done
from send(). If still considered limited by cwnd, app_limited
will be false.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1889,8 +1889,9 @@ impl Connection {
         let pn_len = packet::pkt_num_len(pn)?;
 
         // The AEAD overhead at the current encryption level.
-        let overhead =
-            self.pkt_num_spaces[epoch].overhead().ok_or(Error::Done)?;
+        let crypto_overhead = self.pkt_num_spaces[epoch]
+            .crypto_overhead()
+            .ok_or(Error::Done)?;
 
         let hdr = Header {
             ty: pkt_type,
@@ -1923,16 +1924,30 @@ impl Connection {
 
         hdr.to_bytes(&mut b)?;
 
-        // Make sure we have enough space left for the header, the payload
-        // length, the packet number and the AEAD overhead.
-        left = left
-            .checked_sub(b.off() + pn_len + overhead)
-            .ok_or(Error::Done)?;
+        // Calculate the space required for the packet, including the header
+        // the payload length, the packet number and the AEAD overhead.
+        let mut overhead = b.off() + pn_len + crypto_overhead;
 
         // We assume that the payload length, which is only present in long
         // header packets, can always be encoded with a 2-byte varint.
         if pkt_type != packet::Type::Short {
-            left = left.checked_sub(2).ok_or(Error::Done)?;
+            overhead += 2;
+        }
+
+        // Make sure we have enough space left for the packet.
+        match left.checked_sub(overhead) {
+            Some(v) => left = v,
+
+            None => {
+                // We can't send more because there isn't enough space available
+                // in the output buffer.
+                //
+                // This usually happens when we try to send a new packet but
+                // failed because cwnd is almost full. In such case app_limited
+                // is set to false here to make cwnd grow when ACK is received.
+                self.recovery.update_app_limited(false);
+                return Err(Error::Done);
+            },
         }
 
         let mut frames: Vec<frame::Frame> = Vec::new();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1884,6 +1884,10 @@ impl Connection {
         // Limit output packet size to respect peer's max_packet_size limit.
         left = cmp::min(left, max_pkt_len);
 
+        // Update app_limited.
+        self.recovery
+            .app_limited(left <= self.recovery.cwnd_available());
```
ghedo, line 1889: The `left` value at this point simply represents the _maximum possible_ size of the output packet. It has nothing to do with how much data the application needs to send, so I don't understand how this represents the "app limited" state? I also don't think this takes flow control into consideration?

ghedo: At the very least this needs a much more detailed comment than "Update app_limited" as well as a test.

junhochoi: Good point. The reason why I put it here is to update app_limited even send() returns Error::Done. It covers most of the case but there are some exceptions. Let me think a little more.

junhochoi: Updated and also added test cases.

junhochoi: Recent recovery draft code need to check app limited and flow limited both, however this PR is to fix app limited. For flow control it's better to do in a separate PR.

ghedo: So, if I understand the new changes correctly the idea is to update `app_limited` when we exit early without sending a packet?

I think in both of the cases covered the only reason why we wouldn't send a packet is a too small cwnd, so the `hdr_len` checks seem redundant and we might just want to set `app_limited` to `false` in both cases?

junhochoi: I changed first case to simply set app_limited=false. However second case (when no frame is generated) it still need to check with cwnd_available(). (e.g. when we have nothing to send but cwnd has a room, app_limited is true). Also added one more test case.

src/lib.rs
```diff
@@ -1923,16 +1923,32 @@ impl Connection {
...
+            // When there is nothing to write, we stil need to update
+            // app_limited flag before return.
+            //
+            // This usually happens when we called send() multiple times
+            // in a single event to send a large data.
+            // When cwnd is almost full and the last send() can't send
+            // a new packet because the packet header length is bigger
+            // than the allowed length by cwnd, it's considered as a
+            // full of cwnd is sent and app_limited is set to
+            // false to make cwnd grow when ACK is received.
```
LPardue, line 1949: I'm finding it a bit hard to comprehend this comment

junhochoi: I updated the comment. Let me know if it works.

LPardue: much clearer, thanks!

src/lib.rs
ghedo: If there is nothing to write we will return much earlier, when `write_epoch()` is called, so we wouldn't reach this point, and I think the check always ends up evaluating to `false`.

src/lib.rs
```diff
@@ -1923,16 +1923,29 @@ impl Connection {
 
         hdr.to_bytes(&mut b)?;
 
-        // Make sure we have enough space left for the header, the payload
+        // Calculate the size of header including the payload
         // length, the packet number and the AEAD overhead.
-        left = left
-            .checked_sub(b.off() + pn_len + overhead)
-            .ok_or(Error::Done)?;
+        let mut hdr_len = b.off() + pn_len + overhead;
```
ghedo, line 1928: Also, btw, this is not a "header length". Overhead is not part of the packet header.

ghedo: Myabe rename `overhead` to `crypto_overhead`, and rename this to `overhead`?

src/lib.rs
junhochoi: Thanks. Fixed!

src/lib.rs
junhochoi: Updated.

ghedo:
I did some minor tweaks:
* Use `match` instead of `if`.
* Rename `PktNumSpace::overhead()` -> `crypto_overhead()` to be be consistent.
* Slightly tweaked some comments.
there is no functional change.

PR #493, opened 2020-05-01
allow sending 0.5 RTT data when 0-RTT is used

ghedo:
When accepting early data, the server needs to be able to send data in
the 0.5 RTT before the handshake is completed, otherwise there would be
no difference from normal 1-RTT data handling.

Note that this doesn't handle the HTTP/3 case right now. That is, hTTP/3
control streams are still only opened after the handshake is completed
when not in 0-RTT.

---

Will need some more testing to make sure it doesn't accidentally start sending 1-RTT data too early.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2474,7 +2474,13 @@ impl Connection {
     /// capacity for the operation to complete. The application should retry the
     /// operation once the stream is reported as writable again.
     ///
+    /// Applications should call this method only after the handshake is
+    /// completed (whenever [`is_established()`] returns `true`) or during
+    /// early data if enabled (whenever [`is_in_early_data()`] returns `true`).
+    ///
     /// [`Done`]: enum.Error.html#variant.Done
+    /// [`is_established()`]: struct.Connection.html#method.is_established
+    /// [`is_in_early_data()`]: struct.Connection.html#method.is_in_early_data
     ///
     /// ## Examples:
     ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3199,6 +3176,44 @@ impl Connection {
...
+
+                // Try to parse transport parameters as soon as the first flight
+                // of handshake data is processed.
+                //
+                // This is potentially dangerous as the handshake hasn't been
+                // completed yet, though it's required to be able to send data
+                // in 0.5 RTT.
+                let raw_params = self.handshake.quic_transport_params();
+
+                if !self.parsed_peer_transport_params && !raw_params.is_empty() {
```
ghedo, line 3188: Note to self, will need to check that transport parameters have actually been sent by the client after handshake completion.

src/lib.rs
```diff
@@ -2483,7 +2483,13 @@ impl Connection {
     /// capacity for the operation to complete. The application should retry the
     /// operation once the stream is reported as writable again.
     ///
+    /// Applications should call this method only after the handshake is
+    /// completed (whenever [`is_establushed()`] returns `true`) or during
```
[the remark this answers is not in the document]
ghedo: esta:blush:ed

src/lib.rs
ghedo: Actually, we didn't check before, so it's not a regression.

PR #494, opened 2020-05-01
Mutable and immutable variants of Octets

ghedo:
This changes the `octets` API into immutable and mutable variants of the same API. The main motivation for this is to avoid having to take mutable references to buffers when decoding read-only buffers (in most cases there's no reason to use mutable references other than the fact that the octets API forces it).

For example, when parsing transport parameters we get an immutable buffer from the BoringSSL API, so in order to use it with the octets API we currently need to clone it so that we can get a mutable reference. Though in general it's a good idea to use immutable references when the data doesn't actually need to be modified, which happens a lot.

The duplication in `octets.rs` is somewhat unfortunate, though I couldn't figure out a nice way to avoid it for now, but it can be improved later on.

### DIFF
--- a/src/octets.rs
+++ b/src/octets.rs
@@ -28,15 +28,15 @@
 use std::mem;
 use std::ptr;
 
-/// A specialized [`Result`] type for [`Octets`] operations.
+/// A specialized [`Result`] type for [`OctetsMut`] operations.
 ///
 /// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
-/// [`Octets`]: struct.Octets.html
+/// [`OctetsMut`]: struct.OctetsMut.html
 pub type Result<T> = std::result::Result<T, BufferTooShortError>;
 
-/// An error indicating that the provided [`Octets`] is not big enough.
+/// An error indicating that the provided [`OctetsMut`] is not big enough.
 ///
-/// [`Octets`]: struct.Octets.html
+/// [`OctetsMut`]: struct.OctetsMut.html
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct BufferTooShortError;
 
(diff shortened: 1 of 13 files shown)


src/octets.rs
```diff
@@ -109,28 +111,28 @@ macro_rules! put_u {
...
 }
 
-/// A zero-copy mutable byte buffer.
+/// A zero-copy immutable byte buffer.
 ///
 /// `Octets` wraps an in-memory buffer of bytes and provides utility functions
 /// for manipulating it. The underlying buffer is provided by the user and is
-/// not copied when creating an `Octets`. Operations are panic-free and will
+/// not copied when creating an `OctetsMut`. Operations are panic-free and will
 /// avoid indexing the buffer past its end.
```
LPardue, line 119: I think this might be an erroneous s/Octets/OctestsMut, and/or he block not reads a bit weird because it is describing an immutable byte buffer that can be manipulated.

src/octets.rs
```diff
@@ -109,28 +111,28 @@ macro_rules! put_u {
...
 #[derive(Debug, PartialEq)]
 pub struct Octets<'a> {
-    buf: &'a mut [u8],
+    buf: &'a [u8],
     off: usize,
 }
 
 impl<'a> Octets<'a> {
-    /// Creates an `Octets` from the given slice, without copying.
+    /// Creates an `OctetsMut` from the given slice, without copying.
```
LPardue, line 131: erroneous /s/Octets/OctetsMut ?

src/octets.rs
ghedo: Yeah, copy-pasta error...

src/octets.rs
ghedo: Yup, same.

PR #500, opened 2020-05-05
lib: allow incoming Initial packets lost/reordered

junhochoi:

0. Client sends Initial pn=0
1. Server sends Initial pn=0 -> Lost
2. Server sends Handshake pn=0

Currently when Initial#0 is lost, Client received Handshake#0 and
close the connection immediately because now it gets `Error::CryptoFail`
because Handshake#0 cannot be decrypted.

This can happen when 1 is lost, or reordered so 2 is delivered before 1.
Server doesn't know the connection is closed in the Client,
so keep trying to retransmit and timed out eventually.

The fix is not to immediately close the connection for invalid
incoming packets when Client has no valid received packets. To allow
loss recovery from Server or reordered delivery, now it has a threshold
to allow -- `recovery::PACKET_THRESHOLD` constant is used here because
it's already used for loss recovery/reordering detection. Within
the threshold, it simply discard the packet, waiting Server to
retransmit unless reordered and Client will retransmit its Initial
until the correct packet is received.

Server behavior is not changed (close the connection when
invalid packet is received).

Test Update:
- invalid_initial() is renamed to invalid_initial_server().
- invalid_initial_client() is added to check the client behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1347,8 +1347,15 @@ impl Connection {
 
         let mut b = octets::OctetsMut::with_slice(buf);
 
-        let mut hdr = Header::from_bytes(&mut b, self.scid.len())
-            .map_err(|e| drop_pkt_on_err(e, self.recv_count, &self.trace_id))?;
+        let mut hdr =
+            Header::from_bytes(&mut b, self.scid.len()).map_err(|e| {
+                drop_pkt_on_err(
+                    e,
+                    self.recv_count,
+                    self.is_server,
+                    &self.trace_id,
+                )
+            })?;
 
         if hdr.ty == packet::Type::VersionNegotiation {
             // Version negotiation packets can only be sent by the server.
@@ -1549,6 +1556,7 @@ impl Connection {
                     return Err(drop_pkt_on_err(
                         Error::CryptoFail,
                         self.recv_count,
+                        self.is_server,
                         &self.trace_id,
                     )),
             }
@@ -1556,8 +1564,9 @@ impl Connection {
 
         let aead_tag_len = aead.alg().tag_len();
 
-        packet::decrypt_hdr(&mut b, &mut hdr, &aead)
-            .map_err(|e| drop_pkt_on_err(e, self.recv_count, &self.trace_id))?;
+        packet::decrypt_hdr(&mut b, &mut hdr, &aead).map_err(|e| {
+            drop_pkt_on_err(e, self.recv_count, self.is_server, &self.trace_id)
+        })?;
 
         let pn = packet::decode_pkt_num(
             self.pkt_num_spaces[epoch].largest_rx_pkt_num,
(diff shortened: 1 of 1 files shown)


junhochoi:
When using qvalve, `< 1r drop` or `< 1r reor` can reproduce this.

junhochoi:
Note that this is a client side fix - the server is not impacted by this change I think.

ghedo:
> Note that this is a client side fix - the server is not impacted by this change I think.

The `invalid_initial` test is intended to test for the server behavior, so the above is not really true since the original test would fail with these changes. The server won't close the connection anymore, which kind of defeats the whole point of `drop_pkt_on_err()`.

I think a better fix would be to not do the `recv_count` check on the client-side at all, since the client already sent a packet then the idle timeout is already armed and the connection will time-out eventually.

ghedo:
Also should probably add a separate test specifically for the client behavior.

junhochoi:
> The `invalid_initial` test is intended to test for the server behavior, so the above is not really true since the original test would fail with these changes. The server won't close the connection anymore, which kind of defeats the whole point of `drop_pkt_on_err()`.

I bring back `invalid_packet()` test and renamed to `invalid_packet_server()` since it's server behavior test.
 
> I think a better fix would be to not do the `recv_count` check on the client-side at all, since the client already sent a packet then the idle timeout is already armed and the connection will time-out eventually.

Now I split the case using `is_server` flag -- existing behavior for server but client will ignore a first few invalid packets as described in the PR.

> Also should probably add a separate test specifically for the client behavior.

invalid_packet_client() to test the client behavior.

In summary, existing behavior (close the connection immediately when server received an invalid packet) is kept -- tested by invalid_packet_server() and client will have a new behavior (ignore a few packets in the beginning) -- tested by invalid_packet_client().

ghedo:
@junhochoi my point was that you should not check for `sent_count` at all. The client already has the idle timeout to make sure the connection is closed.

junhochoi:
> @junhochoi my point was that you should not check for `sent_count` at all. The client already has the idle timeout to make sure the connection is closed.

Idle timeout still kicks in when there is inactivity. Here I use sent_count to check if the client is still in the beginning and during that time it discards invalid packet instead of closing. After the threshold, client will close the connection by idle time out or invalid packets received as usual. recv_count can't be used because it doesn't change when receiving invalid packet.

Ideally it'd better if we can count ignored packets but it seems need more changes

ghedo:
So, maybe I wasn't clear. What I'm saying is that we should only change the check to:
```rust
     if is_server && recv_count == 0 {
     ...
     }
```
the rest is unnecessary and breaks the client-side idle timeout so it should be removed.

junhochoi:
> So, maybe I wasn't clear. What I'm saying is that we should only change the check to:
> 
> ```rust
>      if is_server && recv_count == 0 {
>      ...
>      }
> ```
> 
> the rest is unnecessary and breaks the client-side idle timeout (as it causes the connection to be closed before the timeout expires) so it should be removed.

I see. With the changes above now it becomes as follows:

- Client: always ignore invalid packet incoming, relying on the timeout.
- Server: same as before (invalid initial packet before receiving anything will close the connection)

Is it something you want? I'm ok with it. If so I will make changes accordingly including tests.

src/recovery.rs
```diff
@@ -43,7 +43,7 @@ use crate::packet;
 use crate::ranges;
 
 // Loss Recovery
-const PACKET_THRESHOLD: u64 = 3;
+pub const PACKET_THRESHOLD: u64 = 3;
```
ghedo, line 46:
```suggestion
const PACKET_THRESHOLD: u64 = 3;
```

src/lib.rs
```diff
@@ -5285,6 +5302,41 @@ mod tests {
...
+        // Corrupt the packets's last byte to make decryption fail (the last
+        // byte is part of the AEAD tag, so changing it means that the packet
+        // cannot be authenticated during decryption).
+        buf[written - 1] = !buf[written - 1];
+
+        // Client will ignore invalid packet.
+        assert_eq!(pipe.client.recv(&mut buf[..written]), Ok(68));
+
+        // The connection should be alive.
+        assert_eq!(pipe.client.is_closed(), false);
```
ghedo, line 5337: We should also check that the idle timeout is actually set, just to make sure.

Issue #445, opened 2020-04-05
Request HTTP3 timeout when nginx with fastcgi cache


kn007:
When i using this [http3check.net](https://http3check.net/?host=kn007.net) to test http3 with nginx have fastcgi cache:
```
HTTP/3 Check Internal Errors
Request timed out.
Request timed out.
```

nginx.log
```
2020/04/05 18:45:06 [alert] 19893#19893: *9 epoll_ctl(1, 41) failed (17: File exists), client: 159.65.253.210, server: kn007.net, request: "HEAD / HTTP/3", subrequest: "/index.php", host: "kn007.net"
2020/04/05 18:45:06 [info] 19893#19893: *9 shutdown() failed (107: Transport endpoint is not connected), client: 159.65.253.210, server: kn007.net, request: "HEAD / HTTP/3", subrequest: "/index.php", host: "kn007.net"
```

nginx with fastcgi without fastcgi cache test is fine.

```
nginx version: nginx/1.17.9
built by gcc 8.3.1 20190507 (Red Hat 8.3.1-4) (GCC)
built with OpenSSL 1.1.0 (compatible; BoringSSL) (running with BoringSSL)
TLS SNI support enabled

PHP 7.4.4 (cli) (built: Mar 19 2020 23:05:27) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.4, Copyright (c), by Zend Technologies
```

But I using cargo with http3-client and [gf.dev](https://gf.dev/http3-test) is passed.

I don't know where the problem...

kn007:
I using Firefox Nightly to test:
nginx with fastcgi without fastcgi cache return HTTP/3
nginx with fastcgi cache return HTTP/2

lizj3624:
I also encountered a similar bug, how to fix it？

LPardue:
As of https://github.com/cloudflare/quiche/commit/e434e42173f73bb533f6a39cd853eac752731fd9 the nginx patch was removed from quiche. Hence we're closing this issue.

PR #504, opened 2020-05-06
nginx: subrequests fix and optimization headers

morf:
This should fix https://github.com/cloudflare/quiche/issues/445 (`subrequest: "/index.php"` + `shutdown() failed` this happens when the subrequest does not have qstream)


ghedo:
Sorry for the delay @morf. Note that the first commit will cause NGINX to access invalid memory, as some of the headers sent in the response are simply allocated on the stack so you can't just save them for later as the pointrs will be invalid.

Still haven't found the time to test the other fix though :/

morf:
@ghedo fixed invalid memory access

PR #505, opened 2020-05-07
h3: add GOAWAY API

LPardue:
**_NB this is mostly done, I just need to have another sanity check of the API and error conditions that we CAN detect._**

Previously, quiche did some basic GOAWAY frame reception validation
internally but did not expose a method to send GOAWAY, nor an event
to inform apps that a GOAWAY had been received.

On the sending side, this change adds the `send_goaway()` method
that an application can use to send a GOAWAY frame that informs the
peer of the intention to close the connection.

On the receiving side, this change adds the `h3::Event::GoAway`
event which is returned by `poll()`.

The GOAWAY frame contains an identifier field, which affects how an
application should continue using the HTTP/3 connection. This change
is based on the draft-28 version of GOAWAY, which permits
both endpoints to send GOAWAY. If a client receives the frame, it
MUST NOT issue requests with a larger stream ID than indicated. If
a server receives the frame, it MUST NOT issue a push.

Note that applications, not the quiche library, are responsible for
processing request and response messages. Therefore the application
must ensure that GOAWAYS are issued in a semantically valid way. For
instance, a server that sends GOAWAY for stream ID 100 commits to
processing all requests up to that stream ID, unless it explicitly
cancels them or further reduces the limit. An application that
processes requests on stream IDs beyond the limit will cause
problems that cannot be detected by quiche.

### DIFF
--- a/src/h3/ffi.rs
+++ b/src/h3/ffi.rs
@@ -113,6 +113,8 @@ pub extern fn quiche_h3_event_type(ev: &h3::Event) -> u32 {
         h3::Event::Finished { .. } => 2,
 
         h3::Event::Datagram { .. } => 3,
+
+        h3::Event::GoAway { .. } => 4,
     }
 }
 
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -162,6 +162,10 @@
 //!
 //!         Ok((_flow_id, quiche::h3::Event::Datagram)) => (),
 //!
+//!         Ok((goaway_id, quiche::h3::Event::GoAway)) => {
+//!              // Peer signalled it is going away, handle it.
+//!         },
+//!
 //!         Err(quiche::h3::Error::Done) => {
 //!             // Done reading.
 //!             break;
@@ -210,6 +214,10 @@
 //!         },
 //!
 //!         Ok((_flow_id, quiche::h3::Event::Datagram)) => (),
+//!
+//!         Ok((goaway_id, quiche::h3::Event::GoAway)) => {
+//!              // Peer signalled it is going away, handle it.
+//!         },
 //!
 //!         Err(quiche::h3::Error::Done) => {
 //!             // Done reading.
@@ -514,6 +522,9 @@ pub enum Event {
 
     /// DATAGRAM was received.
     Datagram,
+
+    /// GOAWAY was received.
+    GoAway,
 }
 
 struct ConnectionSettings {
@@ -553,6 +564,9 @@ pub struct Connection {
     finished_streams: VecDeque<u64>,
 
     frames_greased: bool,
+
+    local_goaway_id: Option<u64>,
+    peer_goaway_id: Option<u64>,
 }
 
 impl Connection {
@@ -601,6 +615,9 @@ impl Connection {
             finished_streams: VecDeque::new(),
 
             frames_greased: false,
+
+            local_goaway_id: None,
+            peer_goaway_id: None,
         })
     }
 
@@ -648,6 +665,12 @@ impl Connection {
     pub fn send_request<T: NameValue>(
         &mut self, conn: &mut super::Connection, headers: &[T], fin: bool,
     ) -> Result<u64> {
+        // If we received a GOAWAY from the peer, MUST NOT initiate new
+        // requests.
+        if self.peer_goaway_id.is_some() {
+            return Err(Error::FrameUnexpected);
+        }
+
         let stream_id = self.next_request_stream_id;
 
         self.streams
@@ -995,6 +1018,10 @@ impl Connection {
     ///
     /// The event [`Datagram`] returns a flow ID.
     ///
+    /// The event [`GoAway`] returns an ID that depends on the connection role.
+    /// A client receives the largest processed stream ID. A server receives the
+    /// the largest permitted push ID.
+    ///
     /// If an error occurs while processing data, the connection is closed with
     /// the appropriate error code, using the transport's [`close()`] method.
     ///
@@ -1003,6 +1030,7 @@ impl Connection {
     /// [`Data`]: enum.Event.html#variant.Data
     /// [`Finished`]: enum.Event.html#variant.Finished
     /// [`Datagram`]: enum.Event.html#variant.Datagram
+    /// [`GoAway`]: enum.Event.html#variant.GoAWay
     /// [`recv_body()`]: struct.Connection.html#method.recv_body
     /// [`send_response()`]: struct.Connection.html#method.send_response
     /// [`send_body()`]: struct.Connection.html#method.send_body
@@ -1018,15 +1046,33 @@ impl Connection {
 
         // Process control streams first.
         if let Some(stream_id) = self.peer_control_stream_id {
-            self.process_control_stream(conn, stream_id)?;
+            match self.process_control_stream(conn, stream_id) {
+                Ok(ev) => return Ok(ev),
+
+                Err(Error::Done) => (),
+
+                Err(e) => return Err(e),
+            };
         }
 
         if let Some(stream_id) = self.peer_qpack_streams.encoder_stream_id {
-            self.process_control_stream(conn, stream_id)?;
+            match self.process_control_stream(conn, stream_id) {
+                Ok(ev) => return Ok(ev),
+
+                Err(Error::Done) => (),
+
+                Err(e) => return Err(e),
+            };
         }
 
         if let Some(stream_id) = self.peer_qpack_streams.decoder_stream_id {
-            self.process_control_stream(conn, stream_id)?;
+            match self.process_control_stream(conn, stream_id) {
+                Ok(ev) => return Ok(ev),
+
+                Err(Error::Done) => (),
+
+                Err(e) => return Err(e),
+            };
         }
 
         // Process finished streams list.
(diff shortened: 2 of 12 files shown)


LPardue:
Coming back to this after some time away, it look ok. So I've rebased on latest and removed the WIP marking

src/h3/mod.rs
```diff
@@ -474,6 +482,9 @@ pub enum Event {
 
     /// Stream was closed,
     Finished,
+
+    /// GOAWAY was received.
+    GoAway(u64),
```
ghedo, line 487: Instead of adding the "id" to the event, would it make sense to jusr use the `stream_id` we already return from `poll()` for this? It might be confusing to have both.

src/h3/mod.rs
```diff
@@ -2416,31 +2505,77 @@ mod tests {
...
+        s.advance().ok();
+
+        assert_eq!(
+            s.poll_client(),
+            Ok((s.server.control_stream_id.unwrap(), Event::GoAway(4000)))
+        );
+    }
+
+    #[test]
+    /// A client MUST NOT send a request after it receives GOAWAY
```
ghedo, line 2519:
```suggestion
    /// A client MUST NOT send a request after it receives GOAWAY.
```

src/h3/mod.rs
```diff
@@ -2416,31 +2505,77 @@ mod tests {
...
+        assert_eq!(
+            s.poll_client(),
+            Ok((s.server.control_stream_id.unwrap(), Event::GoAway(4000)))
+        );
+
+        assert_eq!(s.send_request(true), Err(Error::FrameUnexpected));
+    }
+
+    #[test]
+    /// Send a GOAWAY frame from the server, using an invalid goaway ID
```
ghedo, line 2537:
```suggestion
    /// Send a GOAWAY frame from the server, using an invalid goaway ID.
```

src/h3/mod.rs
```diff
@@ -2416,31 +2505,77 @@ mod tests {
...
-        assert_eq!(s.poll_client(), Err(Error::Done));
+        assert_eq!(s.poll_client(), Err(Error::IdError));
     }
 
     #[test]
-    /// Send a GOAWAY frame from the server, that references a request that does
-    /// not exist.
-    fn goaway_from_server_bad_id() {
+    /// Send a multiple GOAWAY frames from the server, that increase the goaway
+    /// ID
```
ghedo, line 2554:
```suggestion
    /// ID.
```

src/h3/mod.rs
```diff
@@ -2416,31 +2505,77 @@ mod tests {
...
 
-        assert_eq!(s.poll_client(), Err(Error::Done));
+        assert_eq!(s.poll_client(), Err(Error::IdError));
     }
 
     #[test]
-    /// Send a GOAWAY frame from the server, that references a request that does
-    /// not exist.
-    fn goaway_from_server_bad_id() {
+    /// Send a multiple GOAWAY frames from the server, that increase the goaway
```
ghedo, line 2553:
```suggestion
    /// Send multiple GOAWAY frames from the server, that increase the goaway
```

tools/apps/src/lib.rs
```diff
@@ -927,13 +934,15 @@ impl HttpConn for Http3Conn {
         loop {
             match self.h3_conn.poll(conn) {
                 Ok((stream_id, quiche::h3::Event::Headers { list, .. })) => {
-                    info!(
+                    error!(
```
ghedo, line 937: ?

tools/apps/src/lib.rs
```diff
@@ -927,13 +934,15 @@ impl HttpConn for Http3Conn {
...
                         "{} got request {:?} on stream id {}",
                         conn.trace_id(),
                         &list,
                         stream_id
                     );
 
+                    self.largest_processed_request = stream_id;
```
ghedo, line 944: Should this be somthing like `cmp::max(self.largest_processed_request, stream_id)` instead so if requests are reordered we don't decrease the value?

But also, do we need to do something to make sure reordered requests get processed correctly? Like, if we received request 4, then GOAWAY, then request 0, should we process request 0 before sending our GOAWAY?

src/h3/mod.rs
```diff
@@ -2393,21 +2484,19 @@ mod tests {
     }
 
     #[test]
-    /// Send a GOAWAY frame from the client, which is forbidden.
-    fn goaway_from_client() {
+    /// Send a GOAWAY frame from the client
```
ghedo, line 2487:
```suggestion
    /// Send a GOAWAY frame from the client.
```

src/h3/ffi.rs
```diff
@@ -109,6 +109,8 @@ pub extern fn quiche_h3_event_type(ev: &h3::Event) -> u32 {
         h3::Event::Data { .. } => 1,
 
         h3::Event::Finished { .. } => 2,
+
+        h3::Event::GoAway { .. } => 3,
```
ghedo, line 113: Need to update quiche.h too.

src/h3/mod.rs
```diff
@@ -954,6 +995,54 @@ impl Connection {
...
+    /// these conditions will return an error.
+    ///
+    /// This method does not close the QUIC connection. Applications are
+    /// required to call [`close`] themselves.
+    ///
+    /// [`close()`]: ../struct.Connection.html#method.close
+    pub fn send_goaway(
+        &mut self, conn: &mut super::Connection, id: u64,
+    ) -> Result<()> {
+        if self.is_server && id % 4 != 0 {
```
ghedo, line 1012: Since we don't support push, should we prevent quiche clients from calling this?

src/h3/mod.rs
```diff
@@ -954,6 +995,54 @@ impl Connection {
         Err(Error::Done)
     }
 
+    /// Sends a GOAWAY frame to support graceful connection closure.
```
ghedo, line 998:
```suggestion
    /// Sends a GOAWAY frame to initiate graceful connection closure.
```

tools/apps/src/lib.rs
ghedo: Also, because of how we process streams, a goaway event will be returned before any other events, so even if there are requests buffered the application will receive that one first... is this ok? Should we not return other events if we received or sent goaway?

src/h3/mod.rs
```diff
@@ -608,6 +625,12 @@ impl Connection {
     pub fn send_request(
         &mut self, conn: &mut super::Connection, headers: &[Header], fin: bool,
     ) -> Result<u64> {
+        // If we received a GOAWAY from the peer, MUST NOT initiate new
```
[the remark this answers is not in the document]
LPardue: In draft 28+, GOAWAY is available to client and server but kinda asymmetric.

> Endpoints MUST NOT initiate new requests or promise new pushes on the
   connection after receipt of a GOAWAY frame from the peer.

The requirement are also looser for the server, if a client sends GOAWAY, it will reject pushes with Push IDs greater than or equal to this value (not detect a protocol error).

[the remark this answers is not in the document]
LPardue: - oh wait, I think that is just reflecting the race condition between client GOAWAY and push. Think it is a MUST NOT for the server to generate pushes too.

src/h3/mod.rs
LPardue: good point, it would also make the C API easier

`poll()` currently says 

    /// On success it returns an [`Event`] as well as the event's source stream
    /// ID. The stream ID can be used when calling [`send_response()`] and
    /// [`send_body()`] when responding to incoming requests. On error the
    /// connection will be closed by calling [`close()`] with the appropriate
    /// error code.

With GOAWAY, it won't be `source stream ID` but I think we can accomodate that without breaking the current API. Let me make a commit that augments that description for review.

tools/apps/src/lib.rs
LPardue: To clarify, this is the quiche-apps code so whatever we do is an application decision and not common to users of quiche library.

GOAWAY is asymmetric: a server that receives GOAWAY is being told not to initiate any new pushes. A client that receives GOAWAY is being told not to initiate any new requests. Both SHOULD cancel anything with a higher ID than included in GOAWAY (but quiche doesn't support that at the moment). 

quiche-server is being dumb and only generating a GOAWAY when it receives a GOAWAY. Ideally it should do something smarter but we don't have many places to gracefully close the connection. We typically close a connection when the client timeouts or there is some low-level error. We could add some special thing (drop the connection after 1000 requests? ;) ) but it adds complexity

The code here keeps track of the highest stream ID that was "processed" as soon as we extracted the headers via `poll()`. It doesn't matter if the server got the GOAWAY event before processing any headers - for the server it is actually useful because now we know the limit of pushes.

On the client side, if it receives GOAWAY before responses, it knows the limit of what the server processed and could decide to spin up another connection to deal with retrying requests, while finishing off reception of responses on the current connection.

src/h3/ffi.rs
LPardue: done

src/h3/mod.rs
LPardue: I added an early exit and a TODO.

ghedo (approved):

PR #506, opened 2020-05-08
Added C ffi function to make set_qlog compatible with Windows

xanathar:
Adds a ffi function `quiche_conn_set_qlog` to call `Connection::set_qlog` from C.

Work heavily based on the implementation of `quiche_conn_set_qlog_fd`, but taking a file path as an input instead of a file descriptor, and being, in this way, compatible with non posix OSs (i.e. Windows).

Tested on Windows, Linux and MacOS.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -218,6 +218,10 @@ quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
                                       quiche_config *config, void *ssl,
                                       bool is_server);
 
+// Enables qlog to the specified file path. Returns true on success.
+bool quiche_conn_set_qlog(quiche_conn *conn, const char * path,
+                          const char * log_title, const char * log_desc);
+
 // Enables qlog to the specified file descriptor. Unix only.
 void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char * log_title,
                              const char * log_desc);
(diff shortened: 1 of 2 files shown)


include/quiche.h
```diff
@@ -218,6 +218,10 @@ quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
                                       quiche_config *config, void *ssl,
                                       bool is_server);
 
+// Enables qlog to the specified file path. Returns true on success.
+bool quiche_conn_set_qlog(quiche_conn *conn, const char * path,
```
ghedo, line 222:
```suggestion
bool quiche_conn_set_qlog_path(quiche_conn *conn, const char *path,
```

include/quiche.h
```diff
@@ -218,6 +218,10 @@ quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
...
+                          const char * log_title, const char * log_desc);
```
ghedo, line 223:
```suggestion
                          const char *log_title, const char *log_desc);
```

src/ffi.rs
```diff
@@ -234,6 +234,36 @@ pub extern fn quiche_config_set_cc_algorithm(
     config.set_cc_algorithm(algo);
 }
 
+#[no_mangle]
+#[cfg(feature = "qlog")]
+pub extern fn quiche_conn_set_qlog(
```
ghedo, line 239: We should call this something like `quiche_conn_set_qlog_path` to be more specific.
```suggestion
pub extern fn quiche_conn_set_qlog_path(
```

xanathar:
I've added a commit accepting the changes.
Thanks!

ghedo:
Merged (just tweaked commit message a bit), thanks @xanathar!

Issue #450, opened 2020-04-08
Quiche use in windows and relevant  errors


foroughmajidi:
I built quiche as following steps:
 1- install rust language
2- install perl 
3-install go
4-install NASM
5- git clone --recursive https://github.com/cloudflare/quiche
6-cargo build
and I got quiche.lib, quiche.dll.lib and quiche.dll.
But when I used quiche in my project in visual studio 2019 I got these errors:
![Capture11](https://user-images.githubusercontent.com/51900859/78747951-9d67e400-791f-11ea-9d96-f3503656df24.PNG)

foroughmajidi:
Could I add these lines  to solve errors in quiche.h?
#if SIZE_MAX == ULONG_MAX
    typedef long ssize_t;       /* linux 64 bits */
#elif SIZE_MAX == ULLONG_MAX
    typedef long long ssize_t;  /* windows 64 bits */
#endif

ghedo:
I _think_ a better solution would be to do something like:
```c
#ifdef _MSC_VER
#include <BaseTsd.h>
#define ssize_t SSIZE_T
#endif
```
though I don't have a Windows machine to try this on.

foroughmajidi:
> I _think_ a better solution would be to do something like:
> 
> ```c
> #ifdef _MSC_VER
> #include <BaseTsd.h>
> #define ssize_t SSIZE_T
> #endif
> ```
> 
> though I don't have a Windows machine to try this on.

Thanks it was ok.

PR #507, opened 2020-05-09
quiche.h: ensure size_t and ssize_t are defined

Lekensteyn:
Ensure that check_symbol_exists calls in CMake do not fail due to
missing headers. The minimal reproducer can be stripped down to:

    echo '#include <quiche.h>' | gcc -Iinclude -fsyntax-only -xc -

A similar (untested) fix for Windows was taken from
https://github.com/cloudflare/quiche/issues/450

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -33,6 +33,14 @@ extern "C" {
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stddef.h>
+#ifdef __unix__
+#include <sys/types.h>
+#endif
+#ifdef _MSC_VER
+#include <BaseTsd.h>
+#define ssize_t SSIZE_T
+#endif
 
 // QUIC transport API.
 //


Lekensteyn:
The examples in `examples/` might not be directly usable with Windows, and the Makefile will definitely need changes. I can provide a tiny CMake check to catch regressions on this fix, that will automatically work for all platform (Linux, macOS, Windows).

Alternatively, what do you think of a CMakeLists.txt file in the `examples` directory that replaces the current examples/Makefile?

junhochoi:
> Alternatively, what do you think of a CMakeLists.txt file in the examples directory that replaces the current examples/Makefile?

👍

include/quiche.h
```diff
@@ -33,6 +33,14 @@ extern "C" {
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stddef.h>
+#ifdef __unix__
+#include <sys/types.h>
```
ghedo, line 38: Why is this needed? Shouldn't `stddef.h` be enough?

Lekensteyn: `size_t` is part of the C standard, but `ssize_t` is part of POSIX
http://man7.org/linux/man-pages/man0/stddef.h.0p.html
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html

For POSIX systems inclusion of sys/types.h without stddef.h might be sufficient, but that might not work for Windows.

ghedo:
Merged, thanks @Lekensteyn.

PR #508, opened 2020-05-09
cubic: fix HyStart++ cwnd growth during LSS

junhochoi:
During LSS, cwnd didn't grow because we reset congestion
recovery start time and `on_packet_acked()` exit early because
it thinks it's in a recovery mode. But there was no congestion
event happened, so LSS should not be considered as a recovery
and cwnd will increase in LSS.

- LSS doesn't initialize congestion recovery start time
  and use `lss_start_time` instead. `lss_start_time` is initialized
  when LSS started.
- hystart.in_lss() is replaced by lss_start_time().
- `now` is added to recovery.hystart_on_packet_acked().

Note that Reno is not impacted.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -163,7 +163,7 @@ fn on_packet_acked(
     if r.congestion_window < r.ssthresh {
         // Slow start.
         if r.hystart.enabled() && epoch == packet::EPOCH_APPLICATION {
-            let (cwnd, ssthresh) = r.hystart_on_packet_acked(packet);
+            let (cwnd, ssthresh) = r.hystart_on_packet_acked(packet, now);
 
             r.congestion_window = cwnd;
             r.ssthresh = ssthresh;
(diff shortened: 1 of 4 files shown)


src/recovery/cubic.rs
```diff
@@ -173,22 +173,31 @@ fn on_packet_acked(
...
+            // When we come here without congestion_event() triggered,
+            // initialize congestion_recovery_start_time, w_max and k.
+            ca_start_time = Some(now);
+            r.congestion_recovery_start_time = ca_start_time;
             cubic.w_max = r.congestion_window as f64;
+            cubic.k = 0.0;
         }
 
-        let t = now - r.congestion_recovery_start_time.unwrap();
+        let t = now - ca_start_time.unwrap();
```
LPardue, line 200: Compared to the old code, there seems to be a greater risk that the subtracting component is None.

Would this alternative work? 

* At L176, instead of declaring `ca_start_time` declare `let mut t = now;`
* Inside the LSS case, `if let Some(st) = r.hystart.lss_start_time() {t -= st};`
* In the else case,

```
else if r.congestion_recovery_start_time.is_none {
  r.congestion_recovery_start_time = Some(now);
  cubic.w_max = r.congestion_window as f64;
  cubic.k = 0.0;
  t -= now;
}

Comment (junhochoi) [reply src/recovery/cubic.rs:200]: > Inside the LSS case, if let Some(st) = r.hystart.lss_start_time() {t -= st};

L182 in_lss() = lss_start_time.is_some(). So inside if at L184, lss_start_time() is not None. So at the end of if (L200) ca_start_time is never None.

Comment (junhochoi) [reply src/recovery/cubic.rs]: @LPardue I tried to rewrite the logic a little cleaner to cover all the cases. PTAL

**Comment by junhochoi** [comment]
Test added.

**Comment by LPardue** [comment]
The code is clearer with the changes thanks.

But I wonder if the `in_lss()` method is remaining helpful. I ask because since you changed to an Option<Instant> it's Some or None states reflect what is happening and it is simple enough enough for a caller to do `is_some()` or `is_none()` at the place that is using the value. Whereas `in_lss()` does that internally and so the caller needs to understand the relationship between the method and the Option object, which makes the function brittle. If we replaced the is_lss for is_some, it would be clear in the code that the `unwrap()` will always succeed.

**Comment by junhochoi** [comment]
> The code is clearer with the changes thanks.
> 
> But I wonder if the `in_lss()` method is remaining helpful. I ask because since you changed to an Option it's Some or None states reflect what is happening and it is simple enough enough for a caller to do `is_some()` or `is_none()` at the place that is using the value. Whereas `in_lss()` does that internally and so the caller needs to understand the relationship between the method and the Option object, which makes the function brittle. If we replaced the is_lss for is_some, it would be clear in the code that the `unwrap()` will always succeed.

Ok. in_lss() is replaced by lss_start_time() which returns Option and necessary change is done.
PTAL

File: src/recovery/hystart.rs [hunk 2]
```diff
@@ -103,67 +104,70 @@ impl Hystart {
...
     // Returns a new (ssthresh, cwnd) during slow start.
     pub fn on_packet_acked(
         &mut self, packet: &recovery::Sent, rtt: Duration, cwnd: usize,
-        ssthresh: usize,
+        ssthresh: usize, now: Instant,
     ) -> (usize, usize) {
         let mut ssthresh = ssthresh;
         let mut cwnd = cwnd;
 
-        if !self.lss {
```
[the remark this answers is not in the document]
junhochoi: Ok. Updated.

PR #517, opened 2020-05-12
qlog: fix flappy test

LPardue:


junhochoi:
How is different from #503  ?

LPardue:
Ghedo reported test failures even with #503, so this change unwinds the other fix that didn't seem to work and removes the timing race conditions from all qlog events by forcing them to the same relative offset 0

PR #518, opened 2020-05-12
H3 stream priority

LPardue:
This replaces #465.

It builds on top of #513, adding APIs to the H3 layer and setting the priority of the stream prior to sending any data on it. I've included the quiche commit from #465 so that the CI can run properly and allow anyone interested to test it out.

The quiche-client and quiche-server applications have been updated
to test this out. quiche-client can be run with e.g.
`-H "Priority:u=4,i"` to send a priority signal. quiche-server
process the Priority header (or assumes the default if none is
provided). It also supports a special query string syntax that
takes precedence over the header. For example, a client can send
`https://example.com?u=4&i=1` to indicate the priority as `u=1,i`. In this way,
multiple requests can be issued with different priorities, which makes for more
interesting test cases.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -260,6 +260,9 @@ use crate::octets;
 /// ../struct.Config.html#method.set_application_protos
 pub const APPLICATION_PROTOCOL: &[u8] = b"\x05h3-29\x05h3-28\x05h3-27";
 
+// The offset used when converting HTTP/3 urgency to quiche urgency.
+const PRIORITY_URGENCY_OFFSET: u8 = 124;
+
 /// A specialized [`Result`] type for quiche HTTP/3 operations.
 ///
 /// This type is used throughout quiche's HTTP/3 public API for any operation
@@ -628,7 +631,7 @@ impl Connection {
         Ok(stream_id)
     }
 
-    /// Sends an HTTP/3 response on the specified stream.
+    /// Sends an HTTP/3 response on the specified stream with default priority.
     ///
     /// This method sends the provided `headers` without a body. To include a
     /// body, set `fin` as `false` and subsequently call [`send_body()`] with
@@ -645,6 +648,66 @@ impl Connection {
         &mut self, conn: &mut super::Connection, stream_id: u64,
         headers: &[Header], fin: bool,
     ) -> Result<()> {
+        let priority = "u=3";
+
+        self.send_response_with_priority(
+            conn, stream_id, headers, priority, fin,
+        )?;
+
+        Ok(())
+    }
+
+    /// Sends an HTTP/3 response on the specified stream with specified
+    /// priority.
+    ///
+    /// The [`StreamBlocked`] error is returned when the underlying QUIC stream
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application should retry the operation once the stream is
+    /// reported as writable again.
+    ///
+    /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
+    pub fn send_response_with_priority(
+        &mut self, conn: &mut super::Connection, stream_id: u64,
+        headers: &[Header], priority: &str, fin: bool,
+    ) -> Result<()> {
+        if !self.streams.contains_key(&stream_id) {
+            return Err(Error::FrameUnexpected);
+        }
+
+        let mut urgency = 3;
+        let mut incremental = false;
+
+        for param in priority.split(',') {
+            if param.trim() == "i" {
+                incremental = true;
+                continue;
+            }
+
+            if param.trim().starts_with("u=") {
+                // u is an sh-integer (an i64) but it has a constrained range of
+                // 0-7. So detect anything outside that range and clamp it to
+                // the lowest urgency in order to avoid it interfering with
+                // valid items.
+                //
+                // TODO: this also detects when u is not an sh-integer and
+                // clamps it in the same way. A real structured header parser
+                // would actually fail to parse.
+                let mut u =
+                    i64::from_str_radix(param.rsplit('=').next().unwrap(), 10)
+                        .unwrap_or(7);
+
+                if u < 0 || u > 7 {
+                    u = 7;
+                }
+
+                // The HTTP/3 urgency needs to be shifted into the quiche
+                // urgency range.
+                urgency = (u as u8).saturating_add(PRIORITY_URGENCY_OFFSET);
+            }
+        }
+
+        conn.stream_priority(stream_id, urgency, incremental)?;
+
         self.send_headers(conn, stream_id, headers, fin)?;
 
         Ok(())
@@ -751,7 +814,7 @@ impl Connection {
             None => {
                 return Err(Error::FrameUnexpected);
             },
-        }
+        };
 
         let overhead = octets::varint_len(frame::DATA_FRAME_TYPE_ID) +
             octets::varint_len(body.len() as u64);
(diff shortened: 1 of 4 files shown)


src/h3/mod.rs
```diff
@@ -891,6 +905,8 @@ impl Connection {
         let mut d = [0; 8];
         let mut b = octets::OctetsMut::with_slice(&mut d);
 
+        conn.stream_priority(stream_id, 0, true)?;
```
junhochoi, line 908: `urgency` need to be a default value, not 0?

src/h3/priority.rs
```diff
@@ -0,0 +1,92 @@
...
+                // header parser would actually fail to parse.
+                let mut u =
+                    i64::from_str_radix(param.rsplit('=').next().unwrap(), 10)
+                        .unwrap_or(7);
+
+                if u < 0 || u > 7 {
+                    u = 7
+                };
+
+                priority.urgency = u as u8;
```
junhochoi, line 76: In #513 default urgency is 127. In Priority header draft the range of urgency is [0, 7]. How does quiche stream urgency map to Priority header urgency? It looks like simply use the same number. Or u=0 need map to stream.priority=127?

src/h3/mod.rs
```diff
@@ -637,6 +637,20 @@ impl Connection {
         &mut self, conn: &mut super::Connection, stream_id: u64,
         headers: &[Header], fin: bool,
     ) -> Result<()> {
+        match self.streams.get_mut(&stream_id) {
+            Some(s) => {
+                conn.stream_priority(
+                    stream_id,
+                    s.priority.urgency,
```
junhochoi, line 644: Do we need to map Priority header value to quiche stream priority? It looks like Priority header urgency default is 1 but stream priority default is 127 (https://github.com/cloudflare/quiche/blob/57d448cb6bf28cc5feba347c9fb99d418b26125a/src/stream.rs#L41)

src/h3/mod.rs
LPardue: Thanks for commenting on this, I realise my intention was not obvious. quiche's priority space is 0-255, while HTTP `urgency` is 0-7. In this PR it's mapped 1:1, but I want to boost unidirectional control and qpack streams priority above all else. So I picked the lowest value allowed, 0, (We'll ignore push streams for now because quiche does not support them yet). By picking 0, this means there is an overlap and these streams could get interfered with by requests on lower stream IDs and urgency=0.

One solution I was contemplating was to not map the spaces 1:1 but to right shift urgency so that there was space in the low end (along with all the spare space in the top end). I had something like that in the old branch but think it got lost in this one. So I'll add a followup commit that does this and adds some comments to explain things better.

src/h3/priority.rs
```diff
@@ -0,0 +1,92 @@
...
+                // u is an sh-integer (an i64) but it has a constrained range of
+                // 0-7. So detect anything outside that range and clamp it to
+                // the lowest priority in order to avoid it interfering with
+                // valid items.
+                //
+                // TODO: this also detects when u is not an
+                // sh-integer and clamps it in the same way. A real structured
+                // header parser would actually fail to parse.
+                let mut u =
+                    i64::from_str_radix(param.rsplit('=').next().unwrap(), 10)
```
junhochoi, line 69:
```suggestion
                    u8::from_str_radix(param.rsplit('=').next().unwrap(), 10)
```

src/h3/priority.rs
```diff
@@ -0,0 +1,92 @@
...
+                        .unwrap_or(7);
+
+                if u < 0 || u > 7 {
```
junhochoi, line 72:
```suggestion
                if u > 7 {
```

src/h3/priority.rs
junhochoi, line 76:
```suggestion
                priority.urgency = u;
```

src/h3/mod.rs
LPardue: The spaces are different so have different rules. quiche's default is "halfway" in the space while HTTP priority default is "close to most urgent" because most HTTP resources will fit that bucket and it means that no header needs to be sent.

In the HTTP/3 application mapping, no stream should ever operate at quiche's default. 

In order to make the code flexible and robust, we should create a conversion methods between the two spaces.

src/h3/priority.rs
LPardue: The problem with doing this is that structured-headers defines the complete allowed range for sh-integer https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-17#section-3.3.1. When parsing the header, we we can't reject it if it the urgency  falls beyond the additional constrained scope that `urgency` defines.

junhochoi: I think changing from i64 to u8 result the same? Negative values will be 7 due to unwrap_or(7).

LPardue: The ideal solution is to have a structured header parser, but this code is an attempt to take some shortcuts. I'm pretty happy with the TODO for now and think changing to u8 will make the TODO intent less clear

ghedo:
> Maybe we can just avoid answering this for now and just pass through the Priority header if a client provided it.

Yeah, for the time being we can just assume the application included the header, in the future we can do whatever we want.

LPardue:
OK, I added nuked the priority module, added `send_response_with_priority` and exposed it to the C API. I retested and it all works the same. PTAL.

junhochoi:
New API looks good but not sure quiche parsing the priority string value in *_with_priority() API. I think parsing the value is up to the application and API is good enough to have a priority and urgency parameter as integer?

LPardue:
Passing the Structured headers formatted Extensible priorities field value means we can easily handle extension parameters. If we make priority parameters explicit in the API, then it makes things trickier to handle in the future. The QUIC API doesn't have this problem because it doesn't need to implement any standardized scheme.

I'm not stuck on using a string but it seems to make some things easier. What is the downside?

ghedo:
The other PR is merged, so can be rebased on master and the target of the PR also changed to master.

@junhochoi I kind of agree that using a string is not a great API, though without the parsing of the priority header, `send_response_with_priority()` would basically just be `stream_priority()`+`send_response()`, which an application can already do.

I guess maybe we can hold off on adding `send_response_with_priority()` for now, and let applications deal with calling `stream_priority()` on their own? Though having to reimplement the parsing of the priority string in every implementation will become tedious...

LPardue:
> I guess maybe we can hold off on adding send_response_with_priority() for now, and let applications deal with calling stream_priority() on their own? Though having to reimplement the parsing of the priority string in every implementation will become tedious...

Extensible priorities can only express priority of response streams. The code in this PR also shifts H3 urgency levels around in order to carve out space for control streams; it is important to not block that behind response data. So I don't really like the idea of people calling stream_priority() themselves. 

When quiche is used for H3, the application mapping takes responsibility for using streams. Today we rely on people understanding not to do mess around with streams directly once the H3 connection is created (Ideally the quiche API could explicitly prevent stream operations methods this happens, but I think that ship has sailed). Recommending the use of `stream_priority()` seems inconsistent with the current design philosophy.

tools/apps/src/lib.rs
```diff
@@ -744,9 +770,10 @@ impl Http3Conn {
             quiche::h3::Header::new(":status", &status.to_string()),
             quiche::h3::Header::new("server", "quiche"),
             quiche::h3::Header::new("content-length", &body.len().to_string()),
+            quiche::h3::Header::new("priority", &priority.to_string()),
```
ghedo, line 773: No, need to convert this.
```suggestion
            quiche::h3::Header::new("priority", &priority),
```

tools/apps/src/lib.rs
```diff
@@ -918,18 +945,17 @@ impl HttpConn for Http3Conn {
...
-                    {
+                    match self.h3_conn.send_response_with_priority(
+                        conn, stream_id, &headers, &priority, false,
+                    ) {
                         Ok(v) => v,
 
                         Err(quiche::h3::Error::StreamBlocked) => {
                             let response = PartialResponse {
-                                headers: Some(headers),
+                                headers: Some(headers.to_vec()),
```
ghedo, line 958: No need to copy this AFAICT.
```suggestion
                                headers: Some(headers),
```

src/h3/mod.rs
```diff
@@ -645,6 +648,65 @@ impl Connection {
...
+        if !self.streams.contains_key(&stream_id) {
+            return Err(Error::FrameUnexpected);
+        }
+
+        let mut urgency = 3;
+        let mut incremental = false;
+
+        for param in priority.split(',') {
+            if param.trim() == "i" {
+                incremental = true;
```
ghedo, line 682: Maybe you can do
```rust
continue;
```
after this? So we don't try to decode this parameter again.

weiwei99:
Priority do not work as expected.
Several requests with different priority start at the same time, but the lowest priority request did not finish first.

LPardue:
Please open a new issue with more details of  and a reproduction case.

weiwei99:
@LPardue Thanks for your reply. It's caused by call of send_body(send_response) in handle_writable, which could unexpected to set priority to default value. 
If push_flushable is called before send_body, then the priority has already written to the flushable map, but if not, the wrong priority(default value) is saved in flushable map.

LPardue:
Please open a new issue so we can track this independently of an old merge PR.

PR #519, opened 2020-05-12
Fixed app_limited always true when using datagrams

xanathar:


xanathar:
Sent here by mistake, sorry.

PR #526, opened 2020-05-15
qlog 0.3.0

LPardue:


ghedo:
Probably want to bump the version in quiche Cargo.toml as well, due to API changes.

ghedo:
Oh, I think this is the same problem we had before... qlog needs to be uploaded first before quiche is updated :/

The PR looks good to me though, so you can upload qlog now and then we can rerun the builds here.

PR #527, opened 2020-05-15
retransmit MAX_DATA when lost

junhochoi:
Same idea of #525 but we need a little change because
after we normally send MAX_DATA the next call of
should_update_max_data() returns false so no way to
trigger retransmit.

The change is to have a similar way of updating
MAX_STREAM_DATA. The connection is marked as full and
trigger to send MAX_DATA when self.almost_full is true.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -783,6 +783,9 @@ pub struct Connection {
     /// trigger sending MAX_DATA frames after a certain threshold.
     max_rx_data_next: u64,
 
+    /// Whether we send MAX_DATA frame.
+    almost_full: bool,
+
     /// Total number of bytes sent to the peer.
     tx_data: u64,
 
@@ -1111,6 +1114,7 @@ impl Connection {
             rx_data: 0,
             max_rx_data,
             max_rx_data_next: max_rx_data,
+            almost_full: false,
 
             tx_data: 0,
             max_tx_data: 0,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3416,6 +3429,11 @@ impl Connection {
             self.max_rx_data_next / 2 > self.max_rx_data - self.rx_data
     }
 
+    /// Commits the new max_rx_data limit.
+    pub fn update_max_data(&mut self) {
+        self.max_rx_data = self.max_rx_data_next;
```
ghedo, line 3434: Nitpick, but do we really need a separate method for this? It's only called once and the field is not hidden so can just move this line to where `self.almost_full = false;` is.

src/lib.rs
```diff
@@ -2048,13 +2056,14 @@ impl Connection {
             }
 
             // Create MAX_DATA frame as needed.
-            if self.should_update_max_data() {
+            if self.almost_full {
```
ghedo, line 2059: If this is the signal that triggers sending the frame, you also need to replace the `self.should_update_max_data()` call in the `write_epoch()` method, otherwise it won't trigger sending a packet if there nothing else to send.

src/lib.rs
junhochoi: Ok. I inlined the function to where we send MAX_DATA.

src/lib.rs
junhochoi: Good point. Updated.

PR #529, opened 2020-05-16
update the receiver flow control

junhochoi:
This PR is to adopt [Chromium's approach](https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit?usp=sharing) for the receiver flow control.

### Problem

Currently we tend to send too many `MAX_DATA` and `MAX_STREAM_DATA` after
receiving multiple megabytes.
We use `should_update_max_data()` for deciding whether to send `MAX_DATA`.
However when `max_rx_data_next` is getting bigger, `max_rx_data - rx_data`
is close to BDP (constant), so this function will return true
for most of the time later in the connection. `MAX_STREAM_DATA` has
same problem. You can compare the behavior with other implementations
[here](https://qlog.edm.uhasselt.be/epiq/).

### Fix

`flowcontrol.rs` contains a common framework for flow control updates.

Basically, the concept is to keep a `window` / 2 between the consumed
data offset and `max_data`. When the gap is smaller than window / 2,
we update `MAX_DATA` frame. Same applies to the stream updating
`MAX_STREAM_DATA`.

Auto-tuning of the window is also implemented. When `window`
update keeps happening within 2 x rtt, double `window` value.
By this `window` will be close to the actual BDP. Note that
the window only grows, not shrinks.

Also, make sure the connection window is always 1.5x higher
than any of the stream window. This is for making sure that
the connection flow limit is not a blocker for the stream
to grow. This is from [Chromium code](https://source.chromium.org/chromium/chromium/src/+/master:net/third_party/quiche/src/quic/core/quic_flow_controller.h;bpv=1;bpt=1;l=142?q=flowcontrol&ss=chromium%2Fchromium%2Fsrc&originalUrl=https:%2F%2Fcs.chromium.org%2F).

`RecvBuf.push()` now returns the increased size of the last offset,
which will be used for updating `rx_data`. This is actually a bug fix.
Previously we increased `rx_data` by the size of the data received,
however when the data has an overlap (e.g. duplicate packet) or
a hole in the receive buffer (server send "12345" and client received "5" and "2")
we should not increase `rx_data`. Also `rx_data_consumed` is added for
tracking `rx_data` offset consumed by the application.

#### 5 Mbytes x 10 requests

While no error in flow control, here is the number of
`MAX_DATA` and `MAX_STREAM_DATA` frame sent during the connection with
5 Mbytes x 10 requests (`quiche-client -n 10 https://.../5MB.png`):

- master branch
  - MAX_DATA 9955
  - MAX_STREAM_DATA 31695
- This branch
  - MAX_DATA 21
  - MAX_STREAM_DATA 385

This PR can reduce the flow control updates -- almost 1/100 of master branch.

#### 16 Mbytes x 1 request

Also when you take qvis, you can see the difference of the flow control update. This is 16Mbytes test in the [blog post](https://blog.cloudflare.com/cubic-and-hystart-support-in-quiche/).

Frames sent:
- master branch
  - MAX_DATA 6484
  - MAX_STREAM_DATA 15195
- This branch
  - MAX_DATA 19
  - MAX_STREAM_DATA 73

qvis:
master branch: `MAX_DATA` grows too fast and `MAX_STREAM_DATA` updates more often.
cubic-hs-master

This branch: `MAX_DATA` and `MAX_STREAM_DATA` only start to grow when passing its default value and follows the data received. Also it increases like the staircase.
cubic-hs-829

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -822,7 +822,7 @@ impl Connection {
             conn.trace_id(),
             stream_id,
             header_block.len(),
-            fin
+            fin,
         );
 
         b.put_varint(frame::HEADERS_FRAME_TYPE_ID)?;
@@ -3098,12 +3098,22 @@ mod tests {
             Header::new(":path", "/test"),
         ];
 
+        let ev_headers = Event::Headers {
+            list: req.clone(),
+            has_body: false,
+        };
+
         assert_eq!(s.client.send_request(&mut s.pipe.client, &req, true), Ok(0));
 
         assert_eq!(
             s.client.send_request(&mut s.pipe.client, &req, true),
             Err(Error::StreamBlocked)
         );
+
+        s.advance().ok();
+
+        // The server should consume the request to update flow control credit.
+        assert_eq!(s.server.poll(&mut s.pipe.server), Ok((0, ev_headers)));
 
         s.advance().ok();
 
@@ -3144,6 +3154,11 @@ mod tests {
             Header::new(":path", "/test"),
         ];
 
+        let ev_headers = Event::Headers {
+            list: req.clone(),
+            has_body: true,
+        };
+
         assert_eq!(
             s.client.send_request(&mut s.pipe.client, &req, false),
             Ok(0)
@@ -3153,6 +3168,11 @@ mod tests {
             s.client.send_body(&mut s.pipe.client, 0, b"", true),
             Err(Error::Done)
         );
+
+        s.advance().ok();
+
+        // The server should consume the request to update flow control credit.
+        assert_eq!(s.server.poll(&mut s.pipe.server), Ok((0, ev_headers)));
 
         s.advance().ok();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -313,6 +313,16 @@ const DEFAULT_MAX_DGRAM_QUEUE_LEN: usize = 0;
 // frames size. We enforce the recommendation for forward compatibility.
 const MAX_DGRAM_FRAME_SIZE: u64 = 65536;
 
+// The default size of the receiver connection flow control window.
+const DEFAULT_CONNECTION_WINDOW: u64 = 48 * 1024;
+
+// The maximum size of the receiver connection flow control window.
+const MAX_CONNECTION_WINDOW: u64 = 24 * 1024 * 1024;
+
+// How much larger the connection flow control window need to be larger than
+// the stream flow control window.
+const CONNECTION_WINDOW_FACTOR: f64 = 1.5;
+
 /// A specialized [`Result`] type for quiche operations.
 ///
 /// This type is used throughout quiche's public API for any operation that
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -301,6 +301,33 @@ const MAX_AMPLIFICATION_FACTOR: usize = 3;
...
+        MIN_RX_WINDOW_UNIT;
+
+/// The maximum size of the receiver stream flow control window.
+pub const MAX_RX_STREAM_WINDOW: u64 = 16 * 1024 * 1024;
+
+// The minimum size of the receiver connection flow control window.
+const MIN_RX_CONNECTION_WINDOW: u64 = MIN_RX_STREAM_WINDOW * 2;
+
+// The maximum size of the receiver connection flow control window.
+const MAX_RX_CONNECTION_WINDOW: u64 = 24 * 1024 * 1024;
```
LPardue, line 329: where does this value come from?

src/lib.rs
```diff
@@ -3296,13 +3344,18 @@ impl Connection {
...
                     self.streams.mark_readable(stream_id, true);
                 }
 
-                self.rx_data += data_len;
+                // Check for flow control limits.
+                if self.rx_data + rx_inc > self.max_rx_data {
+                    return Err(Error::FlowControl);
+                }
+
+                self.rx_data = self.rx_data.saturating_add(rx_inc);
```
LPardue, line 3358: `stream::push()` describes itself with "This also takes care of enforcing stream flow control limits", so is there a reason not to do the check inside push? It seems like it would be easier

src/h3/mod.rs
```diff
@@ -2764,6 +2769,11 @@ mod tests {
 
         s.advance().ok();
 
+        // The server should consume the request to update flow control credit.
+        assert_eq!(s.server.poll(&mut s.pipe.server), Ok((0, ev_headers)));
+
+        s.advance().ok();
```
LPardue, line 2775: It's not clear to me why these steps need to be added in the H3 unit tests, they don't seem to test any aspect of the flow control and Event::Headers already has pretty good coverage.

LPardue: Sorry I misread. So actually what is happening is that the flow control is only updated once the application reads stream data out of the library. In this case that happens when the app calls `poll()`

src/lib.rs
junhochoi: I used the const from Chromium

src/lib.rs
junhochoi: This part is checking "connection level FC limit". stream::push() checks "stream level FC limit" as described.

LPardue: Makes sense. I think the comment was confusing me so maybe better to just say "check the connection flow control limit" or something

src/h3/mod.rs
junhochoi: Exactly. In this PR receiving packet into the receiver buffer will check the flow limit against MAX_DATA but doesn't update MAX_DATA (which make sense because updating MAX_DATA without the data consumed by the app only increases receiver buffer too much) Previously test was passed because receiving the packet will trigger MAX_DATA update.

src/lib.rs
junhochoi: Updated the comment.

junhochoi:
Refactored the code a little and added `flowcontrol.rs` containing the common routine for the flow control updates.

PR #535, opened 2020-05-19
tls: write SSLKEYLOGFILE line in one syscall

Lekensteyn:
* Write the key log line and LF in one go to avoid problems with
   multiple interleaving applications.
 * Drop the "QUIC_" prefix, since draft-ietf-quic-tls-17 the same HKDF
   label from TLS 1.3 is in use. Wireshark v3.1.0rc0-836-gcc50ec3634
   supports both, but the "QUIC_" variant is deprecated.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -758,10 +758,10 @@ extern fn keylog(_: *mut SSL, line: *const c_char) {
 
         if let Ok(mut file) = file {
             let data = unsafe { ffi::CStr::from_ptr(line).to_bytes() };
-
-            file.write_all(b"QUIC_").unwrap_or(());
-            file.write_all(data).unwrap_or(());
-            file.write_all(b"\n").unwrap_or(());
+            let mut full_line = Vec::with_capacity(data.len() + 1);
+            full_line.extend_from_slice(data);
+            full_line.push(b'\n');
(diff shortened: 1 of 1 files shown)


LPardue:
WRT SSLKEYLOGFILE, this invokes a similar comment to the one I made on #531. The examples are intended to demo transport aspects of the API and remain simple. We just did a tranche of work to remove a lot of their complexity, which amusingly removed SSLKEYLOGFILE.  

We're now instead investing features in quiche-client and quiche-server, which have purpose beyond API demo.

Other changes look good though.

Lekensteyn:
Ok, would you prefer the SSLKEYLOGFILE calls to be removed from the all examples? I added it since it can help with debugging, and allowed me to validate these changes. I would be happy either way.

LPardue:
It's only 3 lines of code per file, really not the end of the world and I can live with the answer either way. However, it raises the question of scope (and scope creep) in examples which is why I wanted to highlight it in review.

Let's see what others think before making extra work for you.

Lekensteyn:
Removed the unrelated changes to the examples, now this PR is limited to the prime concern that prompted me to start contributing (observing two write()s for SSLKEYLOGFILE in curl while built with quiche).

PR #536, opened 2020-05-19
apps: handling h3::Error::Done

junhochoi:
When running quiche-server I got the following message pretty often
when the session hit the bandwidth limit.

```
[2020-05-19T06:34:28.655009788Z ERROR quiche_apps] 8e01440c13c8788bacc6b0c12265cc4210bafc05 stream send failed Done
```

However, all transfer was normal and this seems not an actual error,
so I think it's safe to ignore this error(`quiche::h3::Error::Done`).

### DIFF
--- a/tools/apps/src/lib.rs
+++ b/tools/apps/src/lib.rs
@@ -955,6 +955,8 @@ impl HttpConn for Http3Conn {
                     {
                         Ok(v) => v,
 
+                        Err(quiche::h3::Error::Done) => 0,
+
                         Err(e) => {
                             error!(
                                 "{} stream send failed {:?}",
@@ -1035,6 +1037,10 @@ impl HttpConn for Http3Conn {
 
         let written = match self.h3_conn.send_body(conn, stream_id, body, true) {
             Ok(v) => v,
+
+            Err(quiche::h3::Error::Done) => {
+                return;
+            },
 
             Err(e) => {
                 error!("{} stream send failed {:?}", conn.trace_id(), e);


junhochoi:
> AFAICT the only method that can return `Done` is `send_body()` so the other change are not necessary (and we shouldn't hide errors if not needed for them IMO).

Thanks. I removed other changes. In `send_body()` `Error:Done` returns when a sender buffer is full, so it's not a real error and indicating that it need a simple retry.

tools/apps/src/lib.rs
```diff
@@ -955,6 +955,10 @@ impl HttpConn for Http3Conn {
                     {
                         Ok(v) => v,
 
+                        Err(quiche::h3::Error::Done) => {
```
ghedo, line 958: Thinking more about this, I think this actually needs to "return" the value 0, instead of skipping the loop, otherwise the partial response object is not created below.
```suggestion
                        Err(quiche::h3::Error::Done) => 0,
```

tools/apps/src/lib.rs
```diff
@@ -955,6 +955,10 @@ impl HttpConn for Http3Conn {
...
+                            continue;
```
ghedo, line 959:
```suggestion
```

tools/apps/src/lib.rs
```diff
@@ -955,6 +955,10 @@ impl HttpConn for Http3Conn {
...
+                        },
```
ghedo, line 960:
```suggestion
```

Issue #530, opened 2020-05-17
h3::Connection::send_response only sends headers


keepsimple1:
This fn name `send_response` is confusing as it only sends headers:
https://github.com/cloudflare/quiche/blob/98757caf1f36ac4236174d9ebae09ac027e78df6/src/h3/mod.rs#L636-L643

At the very least, the doc of `send_response` should clearly say it only sends headers.

And It seems clearer for users to call `send_headers` and then `send_body` (if needed). Is it possible to make `send_headers` public?

LPardue:
The HTTP/3 specification is pretty clear on what comprises a request or response message, see https://tools.ietf.org/html/draft-ietf-quic-http-27#section-4.1.

An HTTP message by definition has to begin with headers, following elements (body or trailers) are optional message elements all depend on the the applicationation setting fin or not. The documentation describes how a server can respond to requests at https://github.com/cloudflare/quiche/blob/98757caf1f36ac4236174d9ebae09ac027e78df6/src/h3/mod.rs#L121

Exposing `send_headers()` is not something I'd want to do because it exposes a primitive that risks applications confusing the message ordering rules.

keepsimple1:
Yes, my understanding is that by definition the HTTP response consists of headers _and_ optional body. 

The current `send_response` API in quiche does not send the whole response. It sends only the headers.  The user has to call `send_body` explicitly after calling `send_response`. 

IMO, either of the following options will be clearer: 

1. call `send_response` that can send headers and optional body, trailers.
or
2. call `send_headers` followed by `send_body` if needed. 

I have no problem with keeping `send_headers` private if `send_response` can be updated to actually send the whole response.

LPardue:
The main benefit of the current design is it supports chunk-based streaming output.

In order to support your first option, the library would need to take and buffer all of the response body. However, there are scenarios where that is not possible such as running quiche in a reverse proxy that receives response data piece by piece. The current design is flexible in supporting different use cases, and has better performance by avoiding buffering.

send_response is analogous to send_headers, with the addition that internally we can do extra work needed to manage and validate aspects of the H3 connection.

keepsimple1:
I think it's possible to have the `send_response` to send the whole response and support chunk-based streaming output. 

[...]

PR #537, opened 2020-05-19
update doc for h3 Connection::send_response

keepsimple1:
This is to address issue #530 . Please review. Thanks.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -627,11 +627,16 @@ impl Connection {
 
     /// Sends an HTTP/3 response on the specified stream.
     ///
+    /// This method sends the provided `headers` without a body. To include a body,
+    /// set `fin` as `false` and subsequently call [`send_body`] with the same
+    /// connection and stream.
+    ///
     /// The [`StreamBlocked`] error is returned when the underlying QUIC stream
     /// doesn't have enough capacity for the operation to complete. When this
     /// happens the application should retry the operation once the stream is
     /// reported as writable again.
     ///
+    /// [`send_body`]: struct.Connection.html#method.send_body
     /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
     pub fn send_response(
         &mut self, conn: &mut super::Connection, stream_id: u64,
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -588,8 +588,10 @@ impl Connection {
 
     /// Sends an HTTP/3 request.
     ///
-    /// The request is encoded from the provided list of headers, and sent on
-    /// a newly allocated stream.
+    /// The request is encoded from the provided list of headers without a
+    /// body, and sent on a newly allocated stream. To include a body,
+    /// set `fin` as `false` and subsequently call [`send_body`] with the
+    /// same connection and the stream_id returned from this method.
     ///
     /// On success the newly allocated stream ID is returned.
     ///
@@ -598,6 +600,7 @@ impl Connection {
     /// happens the application should retry the operation once the stream is
     /// reported as writable again.
     ///
+    /// [`send_body`]: struct.Connection.html#method.send_body
     /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
     pub fn send_request(
         &mut self, conn: &mut super::Connection, headers: &[Header], fin: bool,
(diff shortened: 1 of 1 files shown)


LPardue:
Generally looks good, thanks.

The CI is having a bit of trouble right now but I think it would fail due to rustfmt triggering errors due to line length.

I've made a suggestion of some slight tweaks in https://github.com/cloudflare/quiche/commit/1c0510617025cd4ed0d310fd6c5b4f8561273926. It addresses the above problem and also adds the additional comments to the send_request() method that behaves similarly.

keepsimple1:
Addressed review comments and rebased.

src/h3/mod.rs
```diff
@@ -588,8 +588,10 @@ impl Connection {
 
     /// Sends an HTTP/3 request.
     ///
-    /// The request is encoded from the provided list of headers, and sent on
-    /// a newly allocated stream.
+    /// The request is encoded from the provided list of headers without a
+    /// body, and sent on a newly allocated stream. To include a body,
+    /// set `fin` as `false` and subsequently call [`send_body`] with the
+    /// same connection and the stream_id returned from this method.
```
LPardue, line 594:
```suggestion
    /// same `conn` and the `stream_id` returned from this method.
```

src/h3/mod.rs
```diff
@@ -627,11 +630,16 @@ impl Connection {
 
     /// Sends an HTTP/3 response on the specified stream.
     ///
+    /// This method sends the provided `headers` without a body. To include a
+    /// body, set `fin` as `false` and subsequently call [`send_body`] with the
+    /// same connection and stream_id.
```
LPardue, line 635:
```suggestion
    /// same `conn` and `stream_id`.
```

src/h3/mod.rs
```diff
@@ -588,8 +588,10 @@ impl Connection {
 
     /// Sends an HTTP/3 request.
     ///
-    /// The request is encoded from the provided list of headers, and sent on
-    /// a newly allocated stream.
+    /// The request is encoded from the provided list of headers without a
+    /// body, and sent on a newly allocated stream. To include a body,
+    /// set `fin` as `false` and subsequently call [`send_body`] with the
```
LPardue, line 593:
```suggestion
    /// set `fin` as `false` and subsequently call [`send_body()`] with the
```

src/h3/mod.rs
```diff
@@ -627,11 +630,16 @@ impl Connection {
 
     /// Sends an HTTP/3 response on the specified stream.
     ///
+    /// This method sends the provided `headers` without a body. To include a
+    /// body, set `fin` as `false` and subsequently call [`send_body`] with the
```
LPardue, line 634:
```suggestion
    /// body, set `fin` as `false` and subsequently call [`send_body()`] with the
```

src/h3/mod.rs
```diff
@@ -627,11 +630,16 @@ impl Connection {
...
+    /// same connection and stream_id.
+    ///
     /// The [`StreamBlocked`] error is returned when the underlying QUIC stream
     /// doesn't have enough capacity for the operation to complete. When this
     /// happens the application should retry the operation once the stream is
     /// reported as writable again.
     ///
+    /// [`send_body`]: struct.Connection.html#method.send_body
```
LPardue, line 642:
```suggestion
    /// [`send_body()`]: struct.Connection.html#method.send_body
```

keepsimple1:
Travis CI failed with one error:
```
$ rustup component add rustfmt
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
The command "rustup component add rustfmt" failed and exited with 1 during .
```

Is it a reliable test? How can I re-run the Travis CI?

LPardue:
Its mostly reliable. All this test does is run the code linter, rustfmt, over the code. From time to time the package gets broken.

In this case you're just adding some comments, so as long as the column width is within bounds we should be fine to go ahead and merge.

junhochoi:
> Travis CI failed with one error:

In CI we use nightly `rustfmt` but it's broken last few days : https://rust-lang.github.io/rustup-components-history/
That's why all recent PR show a broken status. Usually it will be resolved in a few days.
If all other build (stable and other platforms) is ok you are fine.

keepsimple1:
I don't have permissions to merge the PR.  If you (reviewers) think it's good, could you please help merge it? 

And I have a question: why not to use the stable `rustfmt` instead of the nightly?

ghedo:
Note that the nightly build for this PR failed because the code needs to be reformatted, see https://travis-ci.com/github/cloudflare/quiche/jobs/338131105#L1757

keepsimple1:
I've updated the diff to fix the `rustfmt` error. Please review again. Thanks.

ghedo:
Merged, thanks @keepsimple1!

PR #541, opened 2020-05-26
CI: disable nightly rustfmt (tmp)

junhochoi:
Since nightly rustfmt is not coming back a week, disable it
temporarily.

https://rust-lang.github.io/rustup-components-history/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,28 +62,28 @@ matrix:
       - eval "$(gimme stable)"
       - gimme --list
      install:
-      - rustup component add rustfmt
+      #- rustup component add rustfmt
       - cargo install cargo-fuzz
      script:
       - RUSTFLAGS="-D warnings" cargo test --verbose
       - RUSTFLAGS="-D warnings" cargo package --verbose --allow-dirty
-      - cargo fmt -- --check
+      #- cargo fmt -- --check
       - cargo doc --no-deps
       - make -C examples
       # fuzzers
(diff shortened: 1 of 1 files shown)


ghedo:
It looks like rustfmt will be fixed soon (https://github.com/rust-lang/rust/pull/72671) so this is not needed anymore I think.

Though for the future, instead of disabling rustfmt, a better fix would be to pin to an older nightly version that has the required components, like so: 

```patch
commit e3718528cb9ad9c2550c6eeef564ac3d0ad8d7a3 (HEAD -> travis-nightly, origin/travis-nightly)
Author: Alessandro Ghedini <alessandro@ghedini.me>
Date:   Fri May 29 15:15:33 2020 +0100

    travis: pin rust nightly to latest with rustfmt

diff --git a/.travis.yml b/.travis.yml
index 15c6d8ab..6a1994bb 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
       - make docker-build
    - name: "nightly Linux x86_64"
      language: rust
-     rust: nightly
+     rust: nightly-2020-05-14
      addons:
        apt:
          packages:
```

junhochoi:
Sure. Your patch should work too. I'll close this one once rustfmt is back today or tomorrow.

junhochoi:
`rustfmt` is back in nightly. Closing.

PR #544, opened 2020-06-01
Draft 28

ghedo:


src/lib.rs
```diff
@@ -3225,10 +3268,62 @@ impl Connection {
...
+
+                        // Validate retry_source_connection_id.
+                        if let Some(rscid) = &self.rscid {
+                            match &peer_params.retry_source_connection_id {
+                                Some(v) if v != rscid =>
+                                    return Err(Error::InvalidTransportParam),
+
+                                Some(_) => (),
+
+                                // initial_source_connection_id must be sent by
```
LPardue, line 3310: nit: retry_source_connection_id ?

src/lib.rs
```diff
@@ -3817,7 +3934,7 @@ impl TransportParams {
 
         // TODO: encode preferred_address
 
-        if tp.active_conn_id_limit != 0 {
+        if tp.active_conn_id_limit != 2 {
```
LPardue, line 3937: Does this need to be `less than 2`?

ghedo: No, this is to decide whether to encode the TP or not (i.e. if the value is the default -> don't encode as it is implicit), and the default value is "2".

PR #546, opened 2020-06-02
tls: configure early data context

ghedo:
This is required to fix accepting early data, that was broken by the
last BoringSSL update (but only for some implementations for some
reasons, so this wasn't caught at the time). BoringSSL also needs to be
updated as well, because the logic from the previous update was wrong.

The "early data context" is used by BoringSSL to decide whether early
data should be accepted or not. In theory this should include the
transport parameters and HTTP/3 SETTINGS in wire format, so that early
data can be accepted on a connection only if they match the ones from
the original connection.

However this is slightly annoying as it would require re-encoding
transport parameters _excluding_ the ones that shouldn't be remembered
when using 0-RTT, as well as exposing some kind of API to allow the
HTTP/3 layer to pass the SETTINGS blob, and in practice these parameters
don't really change that often, so leave that as a TODO item.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -35,7 +35,6 @@ use libc::c_int;
 use libc::c_long;
 use libc::c_uint;
 use libc::c_void;
-use libc::size_t;
 
 use crate::Error;
 use crate::Result;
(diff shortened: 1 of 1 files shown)


ghedo:
So, I think the reason this worked before was that I tested before draft-28 support was merged. That added the new transport params that change for each connection, so the BoringSSL check started failing as it was automatically checking TPs. This is all fixed with these changes.

PR #550, opened 2020-06-04
ChaCha20 tests

ghedo:
Test vectors from https://github.com/quicwg/base-drafts/pull/3712. While at it I also did some small refactorings.


ghedo:
While at it I also synce the Initial packet encryption/decryption test vectors as we haven't update them in a while.

PR #551, opened 2020-06-04
Make minmax filter implementation generic

lohith-bellad:
Making this minmax filter implementation generic so that it can be used to get filtered minmax bandwidth for BBR implementation.

### DIFF
--- a/src/minmax.rs
+++ b/src/minmax.rs
@@ -305,7 +305,6 @@ mod tests {
         assert_eq!(f.estimate[2].value, bw_200);
     }
 
-
     #[test]
     fn get_windowed_min_estimates_rtt() {
         let mut f = Minmax::new(Duration::new(0, 0));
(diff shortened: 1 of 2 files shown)


src/minmax.rs
```diff
@@ -56,27 +56,27 @@ use std::time::Duration;
...
 
-pub struct Minmax {
-    estimate: [MinmaxSample; 3],
+pub struct Minmax<T> {
+    estimate: [MinmaxSample<T>; 3],
 }
 
-impl Minmax {
-    pub fn new() -> Self {
+impl<T: std::cmp::PartialOrd + std::marker::Copy> Minmax<T> {
```
ghedo, line 68:
```suggestion
impl<T: PartialOrd + Copy> Minmax<T> {
```

lohith-bellad: Thanks, taken care!

PR #553, opened 2020-06-05
Draft 29

ghedo:
Based on the latest consensus call, the only real change is the need to support multiple initial salts, as well as updating all the test vectors.

Will merge once the draft-29 I-Ds are submitted.


src/packet.rs
```diff
@@ -636,12 +637,12 @@ fn compute_retry_integrity_tag(
...
     const RETRY_INTEGRITY_KEY: [u8; 16] = [
-        0x4d, 0x32, 0xec, 0xdb, 0x2a, 0x21, 0x33, 0xc8, 0x41, 0xe4, 0x04, 0x3d,
-        0xf2, 0x7d, 0x44, 0x30,
+        0xcc, 0xce, 0x18, 0x7e, 0xd0, 0x9a, 0x09, 0xd0, 0x57, 0x28, 0x15, 0x5a,
+        0x6c, 0xb9, 0x6b, 0xe1,
     ];
 
     const RETRY_INTEGRITY_NONCE: [u8; aead::NONCE_LEN] = [
-        0x4d, 0x16, 0x11, 0xd0, 0x55, 0x13, 0xa5, 0x52, 0xc5, 0x87, 0xd5, 0x75,
+        0xe5, 0x49, 0x30, 0xf9, 0x7f, 0x21, 0x36, 0xf0, 0x53, 0x0a, 0x8c, 0x1c,
```
LPardue, line 645: are the same across draft 27/28 and draft 29?

ghedo: See the commit message :)
> Unlike the initial salt, we only support the keys from the latest
version, as supporting Retry for multiple versions would require
changing the public API.

LPardue: makes sense, thanks

ghedo: I guess in the future (e.g. for QUICv2) we'll probably need to support multiple keys anyway, but we can do those changes separately from this PR I think as we don't really use Retry that much anyway.

PR #562, opened 2020-06-15
try to retransmit old frames on PTO

ghedo:
Currently when sending a PTO probe we either include new data frames or
a PING frame in the packet to make sure that the probe is ack-eliciting.
However if no new data is availabe, we should try to include old data
that has not been marked as lost yet if there is any, instead of just
sending a PING.

This changes that by iterating over old sent packets (that have not been
acked or lost yet) and copying CRYPTO and STREAM frames into the outgoing
packet when creating PTO probes.

This requires cloning the frames, which is somewhat unfortunate, but it
should only happen when PTO expires, which has exponential backoff so it
shouldn't be possible to use this as a DoS vector.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2340,7 +2340,46 @@ impl Connection {
             }
         }
 
-        // Create PING for PTO probe.
+        // Try to retransmit old frames on PTO, if there is space left.
+        if self.recovery.loss_probes[epoch] > 0 &&
+            left > cmp::min(
+                frame::MAX_CRYPTO_OVERHEAD,
+                frame::MAX_STREAM_OVERHEAD,
+            ) &&
+            !is_closing
+        {
+            let unacked_iter = self.recovery.sent[epoch]
+                .iter_mut()
+                // Skip packets that have already been acked or lost and packets
+                // that are not in-flight.
+                .filter(|p| p.in_flight && p.time_acked.is_none() && p.time_lost.is_none());
+
+            'unacked_iter: for unacked in unacked_iter {
+                for frame in &unacked.frames {
+                    // Only retransmit CRYPTO and STREAM frames.
+                    match frame {
+                        frame::Frame::Crypto { .. } |
+                        frame::Frame::Stream { .. } => (),
+
+                        _ => continue,
+                    }
+
+                    // Skip frame if it's too big.
+                    if left < frame.wire_len() {
+                        break 'unacked_iter;
+                    }
+
+                    push_frame_to_pkt!(frames, frame.clone(), payload_len, left);
+
+                    ack_eliciting = true;
+                    in_flight = true;
+
+                    break 'unacked_iter;
+                }
+            }
+        }
+
+        // Create PING for PTO probe if no other ack-elicitng frame is sent.
         if self.recovery.loss_probes[epoch] > 0 &&
             !ack_eliciting &&
             left >= 1 &&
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2340,7 +2340,40 @@ impl Connection {
             }
         }
 
-        // Create PING for PTO probe.
+        // Try to retransmit old frames on PTO, if there is space left.
+        if self.recovery.loss_probes[epoch] > 0 && left >= 1 && !is_closing {
```
LPardue, line 2344: is `left >= 1` the right thing to do here? Maybe take into account CRYPTO ¦ STREAM frame overheads?

junhochoi: I think it's a copy of L2377 and just to fail early when there is no space to add a frame.

src/lib.rs
```diff
@@ -2340,7 +2340,40 @@ impl Connection {
...
+
+                        _ => continue,
+                    }
+
+                    // Skip frame if it's too big.
+                    if left < frame.wire_len() {
+                        break 'unacked_iter;
+                    }
+
+                    push_frame_to_pkt!(frames, frame.clone(), payload_len, left);
```
junhochoi, line 2366: I think we need to check the return value of push_frame_to_pkt!()

src/lib.rs
```diff
@@ -2340,7 +2340,40 @@ impl Connection {
...
         }
 
-        // Create PING for PTO probe.
+        // Try to retransmit old frames on PTO, if there is space left.
+        if self.recovery.loss_probes[epoch] > 0 && left >= 1 && !is_closing {
+            let unacked_iter = self.recovery.sent[epoch]
+                .iter_mut()
+                // Skip packets that have already been acked or lost and packets
+                // that are not in-flight.
+                .filter(|p| p.in_flight && p.time_acked.is_none() && p.time_lost.is_none());
```
junhochoi, line 2349: If I follow correctly it will send a first (lowest packet number) lost packet. Even though the recovery draft doesn't specify what to send as a probe, but I think it's better to send highest packet number lost packet first when there is no new data to send, according to the definition of TCP TLP (https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-rack-08#section-7.3), saying "The probe data segment can either be new, previously unsent data, or a retransmission of previously sent data just below SND.NXT".

With QUIC we use monotonic increasing packet number so this rule have less importance than TCP, but I think it's still valuable to send a highest lost packet as a probe to recovery from a tail loss quickly.

ghedo: > but I think it's still valuable to send a highest lost packet as a probe to recovery from a tail loss quickly.

Why? I'd say the opposite is true. If we are sending contiguous STREAM or CRYPTO data we should prioritize earlier data to try to reduce head-of-line blocking, and let the receiver start processing the data.

junhochoi: AFAICU the main role of the probe is to trigger a loss recovery, not to recover it by itself. 

PTO is expired relatively long time after, so the network already went wrong (e.g. lost a consecutive packets or black out for a short amount of time), so we should not assume what is lost or not. e.g. lowest packet number one is already delivered but we lost the ACK. In this case sending a lowest packet number as a probe is a waste and better to wait to get a ACK range report from the peer triggered by this probe to know the situation correctly. That's why it works even with PING frame.

The draft (and RACK) says we send a new or unsent data first because those are free from the previous loss, so can be sent safely without worrying if it's already delivered. When we don't have those options, 2nd choice will be choosing from currently know lost packets.

In this case sending (actually it's guess but to have a better guess) a highest packet number is better because this PTO expiry might be caused more by last packet (whether last sent data packet is lost, ACK of last sent packet is lost), not by the earlier packet.

src/lib.rs
ghedo: We already check that the frame can fit in the packet above this, so we don't need to check for it again.

src/lib.rs
ghedo: We can probably check for min(CRYPTO_OVERHEAD, STREAM_OVERHEAD), or something like that.

src/lib.rs
ghedo: Well, TBH that still doesn't really explain why we should do it... you first say that we shouldn't assume which packets are lost, then say that in one specific case it's better to resend newer unacked packets.

Looking at other implementations (e.g. Google QUIC, mvfst, quic-go), they all retransmit from older packets (if I understand the code correctly), though that may not have been on purpose, so I asked the question on Slack to see what other people think.

src/lib.rs
ghedo: Done.

src/lib.rs
ghedo: So, for the time being I think we should just go with the current implementation, and then we can change it later when we have a better answer.

junhochoi: Sure.

PR #567, opened 2020-06-18
reject invalid STREAMS_BLOCKED frames

### LINKED ISSUE #565: Quiche drops the packet with OOB stream id instead of sending CONNECTION_CLOSE
When we inject a MAX_STREAMS frame with a value greater than 2^60 . The QUIC library must close the connection immediately with a connection error STREAM_LIMIT_ERROR.

Example,
MAX_STREAMS_BIDI[id=1152921504606846977]

Similar test for injecting OOB STREAMS_BLOCKED has same result, i.e. the packet is dropped without connection close.

ghedo:
Also slightly refactor existing check to avoid repeating the same
constant over and over.

Fixes #565.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -303,6 +303,9 @@ const MAX_AMPLIFICATION_FACTOR: usize = 3;
 // This represents more or less how many ack blocks can fit in a typical packet.
 const MAX_ACK_RANGES: usize = 68;
 
+// The highest possible stream ID allowed.
+const MAX_STREAM_ID: u64 = 1 << 60;
+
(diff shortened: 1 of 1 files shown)


ghedo:
@LPardue ah, yes, good point. I this these checks could also just be moved to the frame decoding code itself since they don't need any additional information about the connection.

ghedo:
Updated to use FRAME_ENCODING_ERROR. Moving the checks is slightly messier than I thought, so might look into that separately.

ghedo:
While at it, I also changed the error for the related transport params.

PR #571, opened 2020-06-26
recovery: retransmit frames from 2 oldest unacked packets on PTO

ghedo:
In 39905083 we implemented support for retransmitting unacked data in
PTO probes if no new data is available.

However after discussion it seems that prioritizing new data is not
actually necessary, and simply retransmitting old data first might be
the better strategy (if anything it's easier to implement).

This reverts the changes from the previous commit, and simply implements
logic to reschedule the frames from the 2 oldest sent packets when the
PTO expires.

Besides having a much simpler implementation, this also has the
advantage of allowing other frames to be sent in the probe packet (e.g.
ACKs), without risking that the original data frame wouldn't fit in the
output packet, since we won't retransmit CRYPTO and STREAM frames as-is
but will go through the normal packetization procedure (including e.g.
respecting STREAM priorities).

This also means that we won't retransmit the same data in both probe
packets. Though if there aren't enough packets with data frames
available we might still end-up sending a PING frame only in the second
probe packet.

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -55,6 +55,8 @@ const PERSISTENT_CONGESTION_THRESHOLD: u32 = 3;
 
 const RTT_WINDOW: Duration = Duration::from_secs(300);
 
+const PTO_PROBES_COUNT: usize = 2;
+
 // Congestion Control
 const INITIAL_WINDOW_PACKETS: usize = 10;
 
@@ -91,7 +93,7 @@ pub struct Recovery {
 
     loss_time: [Option<Instant>; packet::EPOCH_COUNT],
 
-    pub sent: [VecDeque<Sent>; packet::EPOCH_COUNT],
+    sent: [VecDeque<Sent>; packet::EPOCH_COUNT],
 
     pub lost: [Vec<frame::Frame>; packet::EPOCH_COUNT],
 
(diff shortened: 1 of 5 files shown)


junhochoi:
> This reverts the changes from the previous commit, and simply implements
logic to reschedule the frames from the 2 oldest sent packets when the
PTO expires.

recovery 6.2.4 has the following text:
```An endpoint MAY send up to two full-sized datagrams containing ack-eliciting packets, to avoid an expensive consecutive PTO expiration due to a single lost datagram or transmit data from multiple packet number spaces```

It reads for me the main reason to send 2 PTO packet is worrying about lost of PTO packet delaying a recovery.

But also this can be an waste where packet loss is pretty small and there is only 1 lost data to recover. I think we start from sending 1 PTO and dynamically decide to send 2 PTO if packet loss rate is high or when we experienced PTO loss before in the same connection.

**Comment by ghedo** [comment]
@junhochoi yes, but we are already sending 2 probes on PTO so that's outside of the scope of this PR, so we should do that separately.

File: src/recovery/mod.rs [hunk 1]
```diff
@@ -354,7 +356,27 @@ impl Recovery {
             handshake_completed,
         );
 
-        self.loss_probes[epoch] = 2;
+        let unacked_iter = self.sent[epoch]
+            .iter_mut()
+            // Skip packets that have already been acked or lost, and packets
+            // that don't contain either CRYPTO or STREAM frames.
+            .filter(|p| p.has_data && p.time_acked.is_none() && p.time_lost.is_none())
```
junhochoi, line 363: It has been same from previous commit, but I am just thinking of `p.time_lost.is_none()` condition is necessary or not. This will skip packets declared as lost for a Probe packet.

However when a Probe packet is scheduled, I expect there is already some data frame marked as lost (by nature) and those unacked and lost data frames may need to be retransmitted first, since those are usually older than unacked & non-lost ones. Also when 2nd PTO timeout all send buffer is already lost, then we may end up sending PING instead of retransmitting oldest unacked one.

junhochoi: This is interesting topic for me but not a main topic of this PR, so let me dig into a little bit and create a new PR if needed.

ghedo: One thing to note is that when a packet is marked as lost, its frames are moved out of the `Sent` object (as opposed to just being cloned as in the PTO case), so we can't actually schedule them again here.

Though I guess the main point is whether we should re-schedule lost data again here. If the packet is known as lost than that data is already re-scheduled, are you saying we should re-schedule it again?

junhochoi: > If the packet is known as lost than that data is already re-scheduled, are you saying we should re-schedule it again?
Assuming we lost all the ack and let's say we send data 1, after some time we mark data 1 as lost, retransmit, still getting no ack, and then PTO is fired. My point is that in this case this lost data 1 need to be retransmit in the PTO, if rescheduled already but not sent yet.

PR #579, opened 2020-07-01
lib: finish qlog events even on frame error

LPardue:
There might be occasions when the frame parsing fails and then we end up leaving the qlog streamer in a state that causes the next packet's frames to be recorded in the wrong event.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1727,7 +1727,14 @@ impl Connection {
                 ack_elicited = true;
             }
 
-            self.process_frame(frame, epoch, now)?;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1727,7 +1727,18 @@ impl Connection {
                 ack_elicited = true;
             }
 
-            self.process_frame(frame, epoch, now)?;
+            match self.process_frame(frame, epoch, now) {
```
ghedo, line 1730: Given that you don't care about `Ok()` you can just do:
```rust
if let Err(e) = ... {
}
```

PR #586, opened 2020-07-10
Fixed Windows build when used as cargo dependency or building for release without debug symbols

xanathar:
This PR fixes an issue with `--release` Windows builds of rust packages using quiche as a dependency (and quiche stand-alone, if built with custom configurations).

What happens (at least in my case) is that cargo builds the dependencies without debug symbols for release builds of a Rust package (i.e. builds with symbols only the root package).
If `quiche` is one of those deps, it gets built `--release` but without the debug symbols, and cmake-rs consistently puts boringssl in a `<somepath>/Release` directory, instead of the `<somepath>/RelWithDebInfo` directory expected by `build.rs` which then fails with the dreaded `could not find native static library 'crypto', perhaps an -L flag is missing?`.

This code synchronizes the behaviour of `build.rs` with [the one included in cmake-rs](https://github.com/alexcrichton/cmake-rs/blob/master/src/lib.rs#L553); in this way it should supposedly work with all the configurations currently supported by cmake-rs on Windows using the msvc compiler chain.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -47,15 +47,15 @@ const CMAKE_PARAMS_IOS: &[(&str, &[(&str, &str)])] = &[
 fn get_boringssl_platform_output_path(lib: &str) -> String {
     if cfg!(windows) {
         // Code under this branch should match the logic in cmake-rs
-        let debug_env_var = std::env::var("DEBUG")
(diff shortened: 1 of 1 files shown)


ghedo:
Merged, thanks @xanathar!

PR #594, opened 2020-07-16
build: simplify Android build

junhochoi:
We only need Android NDK to build quiche for Android platform.
Simply use rust image from Travis and install NDK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,26 +107,14 @@ matrix:
       - RUSTFLAGS="-D warnings" cargo build --verbose --target=$TARGET_32
       - RUSTFLAGS="-D warnings" cargo test --verbose --target=$TARGET_32
    - name: "stable Android"
-     language: android
-     dist: trusty
+     language: rust
+     rust: stable
      env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -117,31 +107,19 @@ matrix:
...
      env:
        NDK_VER_OLD=r13b
        NDK_VER=r21
-       CMAKE_VER=3.6.4111459
-     android:
-       components:
-        - build-tools-26.0.1
-        # Minimum API level supported
-        - android-21
+       NDK_URL=https://dl.google.com/android/repository/android-ndk-%s-linux-x86_64.zip
```
ghedo, line 115: I think it's better to leave this in the `script` section, so it doesn't show up in the build configuration https://travis-ci.com/github/cloudflare/quiche/jobs/361909618 which is not very useful.

junhochoi: Done.

junhochoi: And rebased.

junhochoi:
It saved 30-40sec in android build time in Travis.

Issue #165, opened 2019-09-13
soname versioning


candrews:
Currently, there's just the library file `libquiche.so`

It would be better if soname versioning was implemented so `libquiche.so` is a symlink to, for example, the actual library file named `libquiche.so.0.0.0`
The library should also have `DT_SONAME` set appropriately.

Could you please implement soname versioning?

ghedo:
Again something that cargo doesn't support https://github.com/rust-lang/cargo/issues/5045

ghedo:
in the cargo ticket above it was suggested to use the [cdylib-link-lines](https://crates.io/crates/cdylib-link-lines) crate, which supposedly would set the SONAME in the dynamic library. Though I don't know if it also renames the .so file accordingly.

I'll have a look once I have a bit of time.

Issue #598, opened 2020-07-17
How to add soname into libquiche.so


coolcih:
Hello,
     I am new for RUST and quic. Hope it is a proper way to raise the question I met. I managed to compile the quiche on Android platform. For static build(libquiche.a), my app can use quic protocol perfectly through quiche. Now I am working on a dynamic lib(libquiche.so) version.  I notice the libquiche.so compile out with below commands line lack "soname" filed.

cargo ndk --target aarch64-linux-android --android-platform 21 -- build

$readelf -a target/aarch64-linux-android/debug/libquiche.so |grep soname
## Output nothing, libquiche.so didn't contain "soname" field.

The "soname" field can be set by "Wl,soname libquiche.so". My question is how to set this "Wl,soname libquiche.so" compile option into the RUST cross compile system?

VMatrix1900:
I am also working on Quiche on Android. Can you share some details on how you use Quiche in your Android app?

junhochoi:
@coolcih is this work for you? https://github.com/rust-lang/rust/issues/43190
or you can patch to add soname later: https://stackoverflow.com/questions/45392319/set-soname-after-linkage

coolcih:
@junhochoi , thanks for the info! You are the best. It works after I add below line in src/build.rs - main function:
println!("cargo:rustc-cdylib-link-arg=-Wl,-soname,libquiche.so");

I suggest add similar line in official release. From what I met, if without soname in the so lib, android build system will hardcode the native dir path into the bin file which use the libquiche.so. For example:
1. Built libquiche.so for Android.
2. Use  libquiche.so as a prebuilt lib i.e. write a prebuilt Android.mk or Android.bp.
3. In app, let's say "myapp",  in myapp's android.mk reference "libquiche" as a shared libaray.
4. After compile myapp, readelf myapp, below result show:

$readelf -a myapp|grep NEEDED
shared lib: [out/target/product/delta/obj/SHARED_LIBRARIES/libquiche_intermediates/libquiche.so]

You can see a native build path hardcode into the myapp bin file. This can't work on Android device.
After I set "cargo:rustc-cdylib-link-arg=-Wl,-soname,libquiche.so", readelf output as below:

$readelf -a myapp|grep NEEDED
shared lib: [libquiche.so]

In this way myapp can find libquiche.so during runtime base on the android system configuration.

coolcih:
@VMatrix1900 , it is not a proper thread to share for what I did. Briefly:
[...]

PR #599, opened 2020-07-17
build: add soname to libquiche.so in android build

junhochoi:
Suggested by #598. Probably beneficial for other platform
build but currently only for android build.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -222,6 +222,10 @@ fn main() {
         println!("cargo:rustc-cdylib-link-arg=-Wl,-undefined,dynamic_lookup");
     }
 
+    if cfg!(target_os = "android") {
+        println!("cargo:rustc-cdylib-link-arg=-Wl,-soname,libquiche.so");
+    }
+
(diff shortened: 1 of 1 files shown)


lu-zero:
Different os have different requirements for the soname.

https://crates.io/crates/cdylib-link-lines might help, a PRs to add android is welcome.

junhochoi:
> Different os have different requirements for the soname.
> 
> https://crates.io/crates/cdylib-link-lines might help, a PRs to add android is welcome.

Sounds good but it will add one more dependency. @ghedo how do you think?

ghedo:
I think we can gate it behind an optional feature at build time, so the people interested can enable it. This would probably also fix https://github.com/cloudflare/quiche/issues/165 btw, if we don't limit it to just Android.

lu-zero:
I suggested already to use `cargo-c` to avoid adding dependencies. But currently Android is not among the targets supported so some patches would be welcomed :)

junhochoi:
Closing in favor of #1769

PR #630, opened 2020-09-05
fix bug:  Probe packets be blocked by the congestion controller.

TOPbuaa:
# Problem
When all packets in flight lost and available congestion window became zero, sender can't send probe packets, and connection will closed because of IDLE timeout.  
As **7.7.  Probe Timeout** in [draft](https://tools.ietf.org/html/draft-ietf-quic-recovery-29) say, **Probe packets MUST NOT be blocked by the congestion controller ... Note that sending probe packets might cause the sender's bytes in flight to exceed the congestion window.** 

# Fix
When sender need to send probe packets, cwnd_available not limit output packet size.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2012,7 +2012,9 @@ impl Connection {
         left = cmp::min(left, max_pkt_len);
 
         // Limit output packet size by congestion window size.
-        left = cmp::min(left, self.recovery.cwnd_available());
+        if self.recovery.loss_probes[epoch] <= 0 {
+            left = cmp::min(left, self.recovery.cwnd_available());
+        }
 
         // Limit data sent by the server based on the amount of data received
         // from the client before its address is validated.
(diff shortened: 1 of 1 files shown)


ghedo:
Thanks for the patch. Would you be able to provide a test case for this? Or in alternative a way to reproduce the problem. I ask because this case should already be handled by the `cwnd_available()` method itself, see https://github.com/cloudflare/quiche/blob/master/src/recovery/mod.rs#L452 but of course the existing code could be buggy.

TOPbuaa:
@ghedo Thanks for the reply. I saw the code you posted and, as you said, this case has already been handled.
I will close the request. Thanks!

more details: 
I found this problem when I used a test app based on C examples of quiche. In the beginning, I checked the log and saw probe packets can't be sent because of [left](https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L2015)==0, so I mistakenly assumed that there was  a problem with the assignment of left. Finally, I realized on_loss_detection_timeout has not been call. 
After I gave [timer.repeat](https://github.com/cloudflare/quiche/blob/master/examples/server.c#L114) a minimum value greater than 0, the test app works well. But the false asssumption were kept until saw your comment.

Issue #632, opened 2020-09-10
RecvBuf +  SendBuf Memory Usage


victorstewart:
I saw that `quiche_config_set_initial_max_data` controls the amount of unacknowledged data allowed to buffer, and that data fills into `RecvBuf`'s and `SendBuf`'s `data: BinaryHeap<RangeBuf>`.

I'm concerned about overall memory usage of the server overtime.

It seems to me that in the worst case, the limits set in `quiche_config_set_initial_max_data`, could cause both buffers to grow to their capacities and that memory is never reclaimed?

Even at a modest 512KB value for the send and receive buffers, with 100,000 connections, we're talking 100GB.

I haven't had a chance to learn rust yet, so maybe I'm totally misreading this or missing something, but in C my approach would be to `mmap(..., MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE,...)` in the memory and then `MADVISE_COLD` the pages after consuming them.

Otherwise we're in a position where a MASSIVE amount of unused memory is consumed. (And in my case this is untenable because I don't have that much memory per connection available on my machines).

Maybe a simple solution here is to just add some `shrink_to_fit` or `shrink_to` calls?

And on this subject, are there any other parts of the implementation that would suffer from similar memory usage problems at such a high connection count?

victorstewart:
since Rust has native mmap and madvise support, i suggest we replace the BinaryHeap object with a raw memory buffer that we track the head and tail of, shift when needed and madvise cold once consumed.

just requires rounding up the supplied max data config values to the nearest page boundary.

whether this above direction or not, i can work on this PR once a direction is agreed upon.

ghedo:
> It seems to me that in the worst case, the limits set in quiche_config_set_initial_max_data quiche_config_set_initial_max_stream_data_bidi_local etc, could cause both buffers to grow to their capacities and that memory is never reclaimed?

Based on what are you saying that? The data is stored as a list of buffers. Those buffers actually own the chunks of application data sent/received (i.e. what initial_max_data and initial_max_stream_data limit) and those are moved/freed after they are read by the application (in case of incoming data), or when they are acked (in case or outgoing data), so what you are saying is not really true (unless of course there is a bug in the code).

The lists themselves are not reduced in size, but the overall memory consumed by them should be pretty low, even in cases where they have been grown a lot.

ghedo:
[...]

PR #643, opened 2020-10-01
stream: allow RangeBuf to be partially consumed without splitting

ghedo:
When "popping" a RangeBuf that exceeded the maximum capacity of the
output buffer, we would simply split-off the internal buffer up to the
maximum cap, and then push a new RangeBuf back into the send or recv
buffer.

This is inefficient for two reasons:

1. Splitting off a Vec means duplicating the remainder of the buffer
   into a new Vec. Depending on the size of the original buffer this
   could be very expensive, which is why we had a hack to split-off
   buffers received by the application before creating the original
   RangeBuf.

2. Pushing back the new RangeBuf into the send or recv buffer is
   potentially expensive as a binary heap is used.

The new popping implementations allow for a RangeBuf to be partially
consumed at its start, without the need for splitting it off.

The downside is that the "consumed" bytes will only be freed when the
whole RangeBuf is consumed, but that should be fine.

---

Before:
```
finished in 74.20s, 1.35 req/s, 134.79MB/s
requests: 100 total, 100 started, 100 done, 100 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 9.77GB (10487377385) total, 9.67KB (9900) headers (space savings 47.06%), 9.77GB (10485760000) data
```
After:
```
finished in 65.57s, 1.53 req/s, 152.54MB/s
requests: 100 total, 100 started, 100 done, 100 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 9.77GB (10487377412) total, 9.67KB (9900) headers (space savings 47.06%), 9.77GB (10485760000) data
```

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -40,8 +40,6 @@ use crate::ranges;
 
 const DEFAULT_URGENCY: u8 = 127;
 
-const MAX_WRITE_SIZE: usize = 1000;
-
 /// Keeps track of QUIC streams and enforces stream limits.
 #[derive(Default)]
 pub struct StreamMap {
@@ -596,7 +594,7 @@ impl ExactSizeIterator for StreamIter {
 /// Stream data received by the peer is buffered in a list of data chunks
 /// ordered by offset in ascending order. Contiguous data can then be read
 /// into a slice.
-#[derive(Default)]
+#[derive(Debug, Default)]
 pub struct RecvBuf {
     /// Chunks of data received from the peer that have not yet been read by
     /// the application, ordered by offset.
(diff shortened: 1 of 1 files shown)


LPardue:
does this relate to https://github.com/cloudflare/quiche/issues/632?

src/stream.rs
```diff
@@ -1177,18 +1176,24 @@ impl RangeBuf {
...
-        self.data.len()
+        self.data.len() - self.pos
     }
 
     /// Returns true if `self` has a length of zero bytes.
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
 
+    /// Consumes the starting `count` butes of `self`.
```
LPardue, line 1187:
```suggestion
    /// Consumes the starting `count` bytes of `self`.
```

LPardue (changes requested):

ghedo:
> does this relate to #632?

Nope, that seems like a potentially invalid ticket.

LPardue (approved):
LGTM

PR #645, opened 2020-10-02
h3: allow passing headers by reference

ghedo:
This makes it possible to pass headers to quiche, without having to clone them (see commit message for more details). This requires moving the conversion to lower-case for header names directly into the QPACK encoder, so we don't need to modify the original string (again, see commit message for details).

Before:
```
finished in 18.03s, 55471.30 req/s, 4.76MB/s
requests: 1000000 total, 1000000 started, 1000000 done, 1000000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1000000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 85.83MB (90000088) total, 82.97MB (87000000) headers (space savings 49.71%), 0B (0) data
```
After:
```
finished in 15.52s, 64426.74 req/s, 5.53MB/s
requests: 1000000 total, 1000000 started, 1000000 done, 1000000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1000000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 85.83MB (90000088) total, 82.97MB (87000000) headers (space savings 49.71%), 0B (0) data
```


ghedo:
The `qpack_decode` fuzzer was unhappy because it expects a round-trip decode -> encode -> decode to produce the same headers, but this was not happening anymore because the QPACK decoder wasn't lower-casing header names anymore as it depended on `Header::new()` to do that.

I added a commit to fix this, which incidentally also removes a bunch of unnecessary copies that were happening when decoding literal headers, so we might get a speed bump from that as well, but I haven't benchmarked it.

I'm not quite sure whether we should actually lower-case header names when decoding though, as e.g. nghttp3 returns an error when receiving a non-lower-case header name, so maybe we should do the same? Checking whether a string is lowercase is about as expensive as turning a string into lower-case so I doubt we would see any performance improvement there.

ghedo:
Turns out we don't need the `unicase` crate at all, since the standard library already provides ASCII case-insensitive equal operation.

LPardue:
I like the overall changes to here, good work.

```
Instead of immediately converting header names to lower-case when
creating a `Header`, defer the conversion to the huffman encoding phase
as in many cases we don't actually need to convert. This also makes it
possible to avoid modifying the headers.

The main annoyance of this is that the static table lookups now need to
be case insensitive, otherwise we could potentially miss some of the
headers if the applications passes them as upper-case.
```

The casing insensitivity sounded quite drastic to start. But then I realised this is just an implementation detail. If we convert_to_lower_and_compare or compare_insensitively, there's little difference. (The opposite of this would be to remove the to_lower() and compare sensitively, and that could cause things to be sent as Literals, which would be a shame.).

So given that our API remains consistent with this change, I think its GTG.

LPardue (approved):
These changes look good. But just to confirm, does the addition of `quiche::h3::NameValue` need to be reflected into any test apps or examples?

junhochoi:
What is our expectation when the request headers have same name with different case?
e.g. when a request headers are: `hello: world`, `Hello: world`, `hEllO: world'
Do we leave only one, or merge, or keep all 3 with lowercase? I think we need to have a test case for that.

LPardue:
Let's never merge field lines in quiche

ghedo:
> does the addition of quiche::h3::NameValue need to be reflected into any test apps or examples?

It needs to be if the application calls `name()` and/or `value()` methods on the header, which is likely. I have an idea on how to improve this, but I think I'll do that in a separate PR.

> What is our expectation when the request headers have same name with different case?

Define "our"? quiche just returns a list of headers, merging them is up to the application. It is not any different from receiving the same header name.

> I think we need to have a test case for that.

There is really nothing to test.

PR #656, opened 2020-10-08
use matches! macro

ghedo:
As per the new clippy warnings.


ghedo:
For whatever reason rustfmt formats the two cases differently :shrug:

LPardue (dismissed):

ghedo:
qlog had a bunch of cases too.

PR #665, opened 2020-10-13
apps: add quiche-client JSON dump option

LPardue:
This change introduces the `--dump-json` option for quiche-client.
When enabled HTTP/3 request headers, response headers and
response body (up to a hard-coded 10KB limit) is written to stdout
once all requests complete or the client times out.

Timed-out requests contain empty values.

The format is HAR-like but simplified.


LPardue:
Note that this uses some hand-rolled serializer code rather than serde. The main benefit of doing it this way is to avoid needing to pollute the quiche library with anything, the code is short and straightforward too.

inikulin (approved):

ghedo (approved):

Issue #585, opened 2020-07-08
STOP_SENDING frame


keepsimple1:
Currently quiche seems not really processing STOP_SENDING frame [except the error case](https://github.com/cloudflare/quiche/blob/c25a59543b2778afe70a467b22b34b10a0d60021/src/lib.rs#L3387-L3394
).   Is there existing or planned work to bubble up this as a H3 event or something like that? I might missed that.

One use case is to let server stop streaming data earlier based on a client request (i.e. STOP_SENDING).

LPardue:
This is still a TODO item but implementing it is low on the priority list at the moment.

PR #669, opened 2020-10-17
bubble up STOP_SENDING frame to H3 event

keepsimple1:
This is trying to address issue #585. The QUIC frame STOP_SENDING triggers a H3 event of `StopSending`, so that the application can react accordingly.
 
Update:

- Handled `TODO` items in `stream_shutdown` to send out `STOP_SENDING` and `RESET_STREAM` frames.
- Reply `RESET_STREAM` frame when received `STOP_SENDING` frame.
- Bubble up `StopSending` and `ResetStream` H3 events only for user request streams.

### DIFF
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -265,6 +265,27 @@ fn main() {
                         info!("GOAWAY id={}", goaway_id);
                     },
 
+                    Ok((
+                        stream_id,
+                        quiche::h3::Event::StopSending { error_code },
+                    )) => {
+                        info!(
+                            "StopSending received for stream {}, error_code {}",
+                            stream_id, error_code
+                        );
+                    },
+
+                    Ok((
+                        stream_id,
+                        quiche::h3::Event::ResetStream {
+                            error_code,
+                            final_size,
+                        },
+                    )) => {
+                        info!("ResetStream received for stream {}, error_code {}, final_size {}",
+                              stream_id, error_code, final_size);
+                    },
+
                     Err(quiche::h3::Error::Done) => {
                         break;
                     },
--- a/examples/http3-server.rs
+++ b/examples/http3-server.rs
@@ -362,6 +362,25 @@ fn main() {
 
                         Ok((_goaway_id, quiche::h3::Event::GoAway)) => (),
 
+                        Ok((
+                            stream_id,
+                            quiche::h3::Event::StopSending { error_code },
+                        )) => {
+                            info!("StopSending received for stream {}, error_code {}",
+                                  stream_id, error_code);
+                        },
+
+                        Ok((
+                            stream_id,
+                            quiche::h3::Event::ResetStream {
+                                error_code,
+                                final_size,
+                            },
+                        )) => {
+                            info!("ResetStream received for stream {}, error_code {}, final_size {}",
+                                  stream_id, error_code, final_size);
+                        },
+
                         Err(quiche::h3::Error::Done) => {
                             break;
                         },
--- a/src/h3/ffi.rs
+++ b/src/h3/ffi.rs
@@ -115,6 +115,10 @@ pub extern fn quiche_h3_event_type(ev: &h3::Event) -> u32 {
         h3::Event::Datagram { .. } => 3,
 
         h3::Event::GoAway { .. } => 4,
+
+        h3::Event::StopSending { .. } => 5,
+
+        h3::Event::ResetStream { .. } => 6,
     }
 }
 
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -166,6 +166,14 @@
 //!              // Peer signalled it is going away, handle it.
 //!         },
 //!
+//!         Ok((stream_id, quiche::h3::Event::StopSending {error_code})) => {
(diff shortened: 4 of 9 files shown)


LPardue:
Thanks for the PR. This is a decent start but it's quite a "pass-through" style of change.

If we're touching code for STOP_SENDING then it should leverage `stream_shutdown()` and address the TODO at https://github.com/cloudflare/quiche/blob/a6200650cc626fb34cb00225bbb9a8e7d5e169e1/src/lib.rs#L2889. 

Rather than creating a collection of frames, we should probably be modifying the stream state and be thinking further about if quiche transport wants to do anything like start to drain the stream or, preventing the reciever of STOP_SENDING from further writes to the stream state.

We also have to process any occurence on STOP_SENDING in the H3 layer first before passing it to the appication. For example, if we get a STOP_SENDING on a control stream then we need to handle that inside quiche; not just punt the problem to the application.

junhochoi:
I wonder what is actual use case for http/3 app to handle stop_sending event?

keepsimple1:
> I wonder what is actual use case for http/3 app to handle stop_sending event?

I have an iOS client that retrieves videos from http/3 server and plays back.  A video file can be very big. The client would send STOP_SENDING frame if it decides no longer need to continue reading the current stream and to request a different chunk of the video file. I would like to have the server stops sending ASAP otherwise it wastes the bandwidth.

The client uses the standard iOS AVPlayer and uses Cronet from Chromium for QUIC/H3 support.

I'm already using quiche (thanks!) and this patch in my http/3 server, it has been helpful.

LPardue:
@keepsimple1's use case is a good example. But we need to think about how this gets handled. Today `stream_shutdown()` also has a todo for RESET_STREAM, so a change in this area should consider addressing that too and decide if quiche generates a RESET_STREAM automatically, and whether we create a new H3 API to wrap around the transport function.

keepsimple1:
> @keepsimple1's use case is a good example. But we need to think about how this gets handled. Today `stream_shutdown()` also has a todo for RESET_STREAM, so a change in this area should consider addressing that too and decide if quiche generates a RESET_STREAM automatically, and whether we create a new H3 API to wrap around the transport function.

Per QUIC spec (draft 29 and latest draft 32), it seems quiche needs to generate a RESET_STREAM automatically, quote from [spec](https://tools.ietf.org/html/draft-ietf-quic-transport-32): 
```
   A STOP_SENDING frame requests that the receiving endpoint send a
   RESET_STREAM frame.  An endpoint that receives a STOP_SENDING frame
   MUST send a RESET_STREAM frame if the stream is in the Ready or Send
   state.  If the stream is in the "Data Sent" state, the endpoint MAY
   defer sending the RESET_STREAM frame until the packets containing
   outstanding data are acknowledged or declared lost.  If any
   outstanding data is declared lost, the endpoint SHOULD send a
   RESET_STREAM frame instead of retransmitting the data.
```

There are also other points you mentioned I need to look into, and try to update the diff.

keepsimple1:
I've rebased to the latest, and updated the diff to address previous review comments. (see "Update" in the PR description). Please review. 

Also I have a question:  do we use rustfmt? If yes, how do we use it when submitting PR?  Thanks.

LPardue:
thanks @keepsimple1! This looks more comprehensive a solution now and seems to be on the right track. Gvien the scope of changes we'll need some time to review. In the meantime, there are some cases of `info` or `println` inside src/lib.rs that can be tidied up.

Wrt to rustfmt, this is something that we run manually locally before pushing. It requires the nightly toolchain, and that can be a bit flaky. See https://github.com/rust-lang/rustfmt#on-the-nightly-toolchain for more details. e.g. `cargo +nightly fmt`.

ghedo:
rustfmt is run automatically as part of CI.

keepsimple1:
I've fixed the `info!`, `println!` statements and `cargo fmt` errors, thanks!

keepsimple1:
Rebased to the latest master at commit `d6e650bfd068b194c86013e44b4ed03e06aefef1` .

keepsimple1:
Updated the diff to address an issue when STOP_SENDING was received after a stream is collected. A new test case is also added.

LPardue:
thanks @keepsimple1! We're still working through some other things but reviewing this is still on our radar, apologies it's taking a while.

ghedo:
First off, thank you for looking at this issue and proposing a patch @keepsimple1.

However, I don't quite agree with the general direction here, and there are a few issues with these changes that would require a fair amount of work to address (this is totally understandable considering this is your first big contribution to the project!), so I made https://github.com/cloudflare/quiche/pull/771 with an alternative proposal to implement the processing of incoming `STOP_SENDING` frames, which also tries to handle the more finicky complications that go along with it (these are part of why this functionality wasn't implemented before).

Additional changes to better handle incoming `RESET_STREAM` frames (including at the HTTP/3 layer), and to support sending `RESET_STREAM` / `STOP_SENDING` via `stream_shutdown()` can be done separately, to make it easier to review them.

Anyway, thank you again for the contribution, we appreciate it!

keepsimple1:
Thanks for letting me know, @ghedo.

PR #674, opened 2020-10-20
apps: return non-zero status code for incomplete requests

ghedo:
This should make it easier to use quiche-client in scripts.

### DIFF
--- a/tools/apps/src/bin/quiche-client.rs
+++ b/tools/apps/src/bin/quiche-client.rs
@@ -37,6 +37,9 @@ use quiche_apps::*;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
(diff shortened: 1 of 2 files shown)


LPardue:
This change is a good one to make.

I think it would help to have a way to identify clearly if the connection handshake did not succeed at all vs. Connection opened and only some requests did not succeed. Can we have different status codes?

Siduck doesn't strictly care if the number of acks doesn't match the number of quacks. It is a lossy protocol after all. Perhaps to avoid noise by not marking it as fail?

ghedo:
@LPardue ok, that's actually slightly nicer to implement, PTAL.

LPardue (approved):

ghedo:
Yeah, good point. Added error message, will merge once green.

PR #676, opened 2020-10-23
stream: fix name and documentation of method returning lowest recv offset

ghedo:
The documentation of this method doesn't match what it actually does.

Also rename it to be consistent with SendBuf.


LPardue:
Only called for qlog? :grin:

LPardue (approved):

PR #679, opened 2020-10-26
lib: send MAX_DATA when MAX_STREAM_DATA is sent

junhochoi:

- starting default of stream and connection flow limit

stream#1 limit=100, want to write 150 bytes in total
stream#2 limit=100, want to write 150 bytes in total
stream#3 limit=100, want to write 150 bytes in total
connection limit=300

And let's say server want to write 150 bytes each in the stream.
And all stream has a same priority: each stream data need to be delivered FIFO, which means
we should deliver 150 bytes in stream#1 full and 150 bytes in stream#2, and so on.

- Current behavior

server stream_send() stream#1 100 bytes (stream#1 reach limit = 100)
server stream_send() stream#2 100 bytes (stream#2 reach limit = 100)
server stream_send() stream#3 100 bytes (stream#1 reach limit = 100) (connection reached limit = 300)

server send() stream#1 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_STREAM_DATA(stream#1, 200)
**client: MAX_DATA is not sent because should_update_max_data() returns false at this point**

_server receives ACK and update stream#1 limit to 200_
_server want to write more on stream#1, but cannot due to connection limit (300)_ --> this is a problem breaking FIFO

server send stream#2 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_DATA(500)
server send stream#2 50 bytes + fin
...

Eventually MAX_DATA will be sent but it arrives later, so FIFO delivery start to break and there is a race condition
where the sender can't write more until MAX_DATA arrives. To solve the issue, when MAX_STREAM_DATA is sent,
and MAX_DATA if increased from the previously sent value. This will unblock the sender to write on the right stream
early.

- This PR's behavior

server stream_send() stream#1 100 bytes (stream#1 reach limit = 100)
server stream_send() stream#2 100 bytes (stream#2 reach limit = 100)
server stream_send() stream#3 100 bytes (stream#1 reach limit = 100) (connection reached limit = 300)

server send() stream#1 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_STREAM_DATA(stream#1, 200) + **MAX_DATA(400)**
**client send MAX_DATA with MAX_STREAM_DATA, regardless of should_update_max_data()**

_server receives ACK and updates connection limit to 400 and stream#1 limit to 200_

server send() stream#1 50 bytes + fin --> FIFO works
client recv() 50 bytes from stream#1 and send() ACK + MAX_DATA(450)

server send stream#1 100 bytes
...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2255,23 +2255,6 @@ impl Connection {
                 }
             }
 
-            // Create MAX_DATA frame as needed.
-            if self.almost_full {
-                let frame = frame::Frame::MaxData {
-                    max: self.max_rx_data_next,
-                };
-
-                if push_frame_to_pkt!(b, frames, frame, left) {
-                    self.almost_full = false;
-
-                    // Commits the new max_rx_data limit.
-                    self.max_rx_data = self.max_rx_data_next;
-
-                    ack_eliciting = true;
-                    in_flight = true;
-                }
-            }
-
             // Create DATA_BLOCKED frame.
             if let Some(limit) = self.blocked_limit {
                 let frame = frame::Frame::DataBlocked { limit };
@@ -2306,6 +2289,27 @@ impl Connection {
                     stream.recv.update_max_data();
 
                     self.streams.mark_almost_full(stream_id, false);
+
+                    ack_eliciting = true;
+                    in_flight = true;
+
+                    // Also send MAX_DATA when MAX_STREAM_DATA is sent, to avoid a
+                    // potential race condition.
+                    self.almost_full = true;
+                }
+            }
+
+            // Create MAX_DATA frame as needed.
+            if self.almost_full && self.max_rx_data < self.max_rx_data_next {
+                let frame = frame::Frame::MaxData {
+                    max: self.max_rx_data_next,
+                };
+
+                if push_frame_to_pkt!(b, frames, frame, left) {
+                    self.almost_full = false;
+
+                    // Commits the new max_rx_data limit.
+                    self.max_rx_data = self.max_rx_data_next;
 
                     ack_eliciting = true;
                     in_flight = true;
@@ -5464,6 +5468,13 @@ mod tests {
         // Ignore ACK.
         iter.next().unwrap();
 
+        assert_eq!(
+            iter.next(),
+            Some(&frame::Frame::MaxStreamData {
+                stream_id: 4,
+                max: 30
+            })
+        );
         assert_eq!(iter.next(), Some(&frame::Frame::MaxData { max: 46 }));
     }
 


LPardue:
I don't know if this is the right solution to the problem. In the example case why doesn't the MAX_DATA get sent earlier? Is there are problem with `should_update_max_data()`?

junhochoi:
> I don't know if this is the right solution to the problem. In the example case why doesn't the MAX_DATA get sent earlier? Is there are problem with `should_update_max_data()`?

Eventually MAX_DATA will be sent, however before receiving it (because it's from the client, there is a time gap of rtt), the sender continues to schedule what's available in next stream. It's some type of race condition -- can be avoided by updating MAX_STREAM_DATA and MAX_DATA in the same packet (usually bundled with ACK).

src/lib.rs
```diff
@@ -2239,6 +2222,26 @@ impl Connection {
 
                     ack_eliciting = true;
                     in_flight = true;
+
+                    // if MAX_STREAM_DATA is sent, send MAX_DATA as well.
```
ghedo, line 2226:
```suggestion
                    // Also send MAX_DATA when MAX_STREAM_DATA is sent, to avoid a potential race condition.
```

ghedo (changes requested):
LGTM, added a nit to slightly improve the comment.

LPardue:
This change seems to be marking the connection as almost full (`self.almost_full = true`) when a stream is full, that seems partially wrong to me. It appears we "correct" this by changing the creation of MAX_DATA to be conditional on two things

```
            if self.almost_full && self.max_rx_data < self.max_rx_data_next {
```

So why not just inside the MAX_STREAM_FRAME block do `self.almost_full = should_update_max_data();` and then we can just check for `self.almost_full` when it comes to the MAX_DATA block?

junhochoi:
> This change seems to be marking the connection as almost full (`self.almost_full = true`) when a stream is full, that seems partially wrong to me. It appears we "correct" this by changing the creation of MAX_DATA to be conditional on two things
> 
> ```
> // Create MAX_DATA frame as needed.
> if self.almost_full && self.max_rx_data < self.max_rx_data_next {
> ```
> 
> So why not just inside the MAX_STREAM_FRAME block do `self.almost_full = should_update_max_data();` and then we can just check for `self.almost_full` when it comes to the MAX_DATA block?

- The intention of this PR to to send MAX_DATA ignoring should_update_max_data() check when MAX_STREAM_DATA is sent, so this is correct.
- self.max_rx_data < self.max_rx_data_next check is needed to prevent to send MAX_DATA again with the same value, if we sent it already.

LPardue:
I still don't get it, sorry. If for the connection to work properly the connections flow control needs to be increased, then surely the function named should_update_max_data() needs to return the right result.

junhochoi:
Let's give a more detailed scenario. Actually I think this is better one, need to update the PR description.

- starting default of stream and connection flow limit

stream#1 limit=100, want to write 150 bytes in total
stream#2 limit=100, want to write 150 bytes in total
stream#3 limit=100, want to write 150 bytes in total
connection limit=300

And let's say server want to write 150 bytes each in the stream.
And all stream has a same priority: each stream data need to be delivered FIFO, which means
we should deliver 150 bytes in stream#1 full and 150 bytes in stream#2, and so on.

- Current behavior

server stream_send() stream#1 100 bytes (stream#1 reach limit = 100)
server stream_send() stream#2 100 bytes (stream#2 reach limit = 100)
server stream_send() stream#3 100 bytes (stream#1 reach limit = 100) (connection reached limit = 300)

server send() stream#1 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_STREAM_DATA(stream#1, 200)
**client: MAX_DATA is not sent (should_update_max_data()=false)**

_server want to write more on stream#1, but cannot due to connection limit (300)_ --> this is a problem breaking FIFO

server send stream#2 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_DATA(500)
server send stream#2 50 bytes + fin
...

- This PR behavior

server stream_send() stream#1 100 bytes (stream#1 reach limit = 100)
server stream_send() stream#2 100 bytes (stream#2 reach limit = 100)
server stream_send() stream#3 100 bytes (stream#1 reach limit = 100) (connection reached limit = 300)

server send() stream#1 100 bytes
client recv() 100 bytes from stream#1 and send() ACK + MAX_STREAM_DATA(stream#1, 200) + MAX_DATA(400)
**client send MAX_DATA with MAX_STREAM_DATA**

_server updates connection limit to 400 and stream#1 limit to 200_

server send() stream#1 50 bytes + fin --> FIFO works
client recv() 50 bytes from stream#1 and send() ACK + MAX_DATA(450)

server send stream#1 100 bytes
...

LPardue:
Thanks for the explanation @junhochoi. 

I think _my_ problem really is that we're overloading the variable name `almost_full` when it is not the case. So how about, for future clarity, we do something like

```
let conn_max_with_stream = false;

// Create MAX_STREAM_DATA frames as needed.
loop over streams {
  ...
  // if MAX_STREAM_DATA is sent, send MAX_DATA as well.
  conn_max_with_stream = self.max_rx_data < self.max_rx_data_next;
}

if self.almost_full || conn_max_with_stream {
  ...
}
```

junhochoi:
> Thanks for the explanation @junhochoi.
> 
> I think _my_ problem really is that we're overloading the variable name `almost_full` when it is not the case. So how about, for future clarity, we do something like
> 
> ```
> let mut conn_max_with_stream = false;
> 
> // Create MAX_STREAM_DATA frames as needed.
> loop over streams {
>   ...
>   // if MAX_STREAM_DATA is sent, send MAX_DATA as well.
>   conn_max_with_stream = self.max_rx_data < self.max_rx_data_next;
> }
> 
> // Create MAX_DATA frame as needed.
> if self.almost_full || conn_max_with_stream {
>   ...
> }
> ```

Currently the meaning of `almost_full` flag is simply whether we send MAX_DATA frame:
https://github.com/cloudflare/quiche/blob/875012ec90b1f409b647944176c607281a01412c/src/lib.rs#L825
And this technique (setting the flag) is used in other places such as retransmitting MAX_DATA frame:
https://github.com/cloudflare/quiche/blob/875012ec90b1f409b647944176c607281a01412c/src/lib.rs#L2020

TBH I think it's ok to set/use the flag in this way, but your logic work too.

LPardue:
Fair point, I forgot about the use in retransmission. That makes me lean towards consistency of variable name. Let's ship this as is!

LPardue (dismissed):

src/lib.rs
ghedo: It looks like this change was removed in the latest version, was this on purpose?

junhochoi: Oops. Updated. Thanks

ghedo (approved):

PR #680, opened 2020-10-26
build: rustfmt missing since 2020-10-26

junhochoi:
https://rust-lang.github.io/rustup-components-history/
pin to 2020-10-25.


ghedo (approved):

ghedo:
FTR, fixing rustfmt is in progress (https://github.com/rust-lang/rust/issues/78341), but the process is going to take some time I think.

PR #681, opened 2020-10-27
Fix the compiling for rust 1.39.0

bachue:
Rust 1.39.0 has no `matches!` macro in std lib.


ghedo:
Hello, thank you for the pull request! However, I think it'd be better to simply bump the minimum required version rather than add a new dependency to the crate. I made https://github.com/cloudflare/quiche/pull/691 to update the documentation.

PR #684, opened 2020-10-30
lib: limiting a burst of outgoing packets

junhochoi:
"QUIC draft-32 7.7" says:
>     Sending multiple packets into the network without any delay between
>     them creates a packet burst that might cause short-term congestion
>     and losses.  Senders MUST either use pacing or limit such bursts.
>     Senders SHOULD limit bursts to the initial congestion window; see
>     Section 7.2.  A sender with knowledge that the network path to the
>     receiver can absorb larger bursts MAY use a higher limit.

When I test quiche in a high bandwidth/low rtt environment such as localhost,
cwnd can grow very quickly and can create a lot of packet sent
when 1) many sender packets are acked on once 2) try to recover many losses.

This PR implements limiting outgoing packet burst as recommended in the
draft. Here we limit total number of outgoing packets in a single send() loop
to the initial congestion window. When send() reaches a burst limit,
now `Error::Again` will be returned, so it's up to application to flush
current packets and check if there is incoming packets, or can continue
to allow bursting.

Also note that I added the limit on the send(). Such limiting burst
can be done in the application itself, but it provides more general
support and all quiche application can use same burst limiting.

### DIFF
--- a/examples/http3-server.c
+++ b/examples/http3-server.c
@@ -92,6 +92,11 @@ static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
     while (1) {
         ssize_t written = quiche_conn_send(conn_io->conn, out, sizeof(out));
 
+        if (written == QUICHE_ERR_AGAIN) {
+            fprintf(stderr, "try again writing\n");
+            continue;
+        }
+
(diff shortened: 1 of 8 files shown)


junhochoi:
Closing - this is better to be implemented in the application

PR #685, opened 2020-10-30
recovery: configurable max_datagram_size

junhochoi:
In the recovery module we use MAX_DATAGRAM_SIZE constants (1452)
and when the actual payload size (max_udp_payload_size) parameter
and the actual max size of the packet doesn't match, the congestion
control might work inefficiently. For example quiche-server
sets max_udp_payload_size and max udp packet size to 1350, but
since MAX_DATAGRAM_SIZE is 1452, the congestion window grows
too fast then expected, resulting unnecessary packet loss due to
overtransfer.

This PR changes MAX_DATAGRAM_SIZE as a configurable parameter
and adds set_max_datagram_size() API for the application.

When tested, this can reduce errors such as timeout a lot when
running in very high bandwidth/rtt environment.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -145,6 +145,9 @@ void quiche_config_set_max_idle_timeout(quiche_config *config, uint64_t v);
 // Sets the `max_udp_payload_size transport` parameter.
 void quiche_config_set_max_recv_udp_payload_size(quiche_config *config, size_t v);
 
+// Sets the maximum outgoing UDP payload size.
+void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
+
 // Sets the `initial_max_data` transport parameter.
 void quiche_config_set_initial_max_data(quiche_config *config, uint64_t v);
 
@@ -187,9 +190,6 @@ void quiche_config_enable_hystart(quiche_config *config, bool v);
 void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
-
-// Sets the maximum outgoing UDP payload size used in congestion control.
-void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
 
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -143,7 +143,7 @@ int quiche_config_set_application_protos(quiche_config *config,
 void quiche_config_set_max_idle_timeout(quiche_config *config, uint64_t v);
 
 // Sets the `max_udp_payload_size transport` parameter.
-void quiche_config_set_max_udp_payload_size(quiche_config *config, uint64_t v);
+void quiche_config_set_max_recv_udp_payload_size(quiche_config *config, uint64_t v);
 
 // Sets the `initial_max_data` transport parameter.
 void quiche_config_set_initial_max_data(quiche_config *config, uint64_t v);
@@ -187,6 +187,9 @@ void quiche_config_enable_hystart(quiche_config *config, bool v);
 void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
+
+// Sets the maximum datagram size used by the congestion control.
+void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
 
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -143,7 +143,7 @@ int quiche_config_set_application_protos(quiche_config *config,
 void quiche_config_set_max_idle_timeout(quiche_config *config, uint64_t v);
 
 // Sets the `max_udp_payload_size transport` parameter.
-void quiche_config_set_max_recv_udp_payload_size(quiche_config *config, uint64_t v);
+void quiche_config_set_max_recv_udp_payload_size(quiche_config *config, size_t v);
 
 // Sets the `initial_max_data` transport parameter.
 void quiche_config_set_initial_max_data(quiche_config *config, uint64_t v);
@@ -188,7 +188,7 @@ void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
 
-// Sets the maximum datagram size used by the congestion control.
+// Sets the maximum outgoing UDP payload size used in congestion control.
 void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
 
 // Frees the config object.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -306,8 +306,8 @@ const MAX_ACK_RANGES: usize = 68;
 // The highest possible stream ID allowed.
 const MAX_STREAM_ID: u64 = 1 << 60;
 
-/// The default max_datagram_size in the congestion control.
-pub const MAX_DATAGRAM_SIZE: usize = 1200;
+/// The default max_datagram_size used in congestion control.
+const MAX_SEND_UDP_PAYLOAD_SIZE: usize = 1200;
 
 // The default length of DATAGRAM queues.
 const DEFAULT_MAX_DGRAM_QUEUE_LEN: usize = 0;
@@ -447,7 +447,7 @@ pub struct Config {
     dgram_recv_max_queue_len: usize,
     dgram_send_max_queue_len: usize,
 
-    max_datagram_size: usize,
+    max_send_udp_payload_size: usize,
 }
 
 impl Config {
@@ -474,7 +474,7 @@ impl Config {
             dgram_recv_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
             dgram_send_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
 
-            max_datagram_size: MAX_DATAGRAM_SIZE,
+            max_send_udp_payload_size: MAX_SEND_UDP_PAYLOAD_SIZE,
         })
     }
 
@@ -619,8 +619,8 @@ impl Config {
     /// Sets the `max_udp_payload_size transport` parameter.
     ///
     /// The default value is `65527`.
-    pub fn set_max_recv_udp_payload_size(&mut self, v: u64) {
-        self.local_transport_params.max_udp_payload_size = v;
+    pub fn set_max_recv_udp_payload_size(&mut self, v: usize) {
+        self.local_transport_params.max_udp_payload_size = v as u64;
     }
 
     /// Sets the `initial_max_data` transport parameter.
@@ -780,11 +780,11 @@ impl Config {
         self.dgram_send_max_queue_len = send_queue_len;
     }
 
-    /// Sets the maximum datagram size used by the congestion control.
+    /// Sets the maximum outgoing UDP payload size used in congestion control.
     ///
     /// The default and minimum value is `1200`.
     pub fn set_max_send_udp_payload_size(&mut self, v: usize) {
-        self.max_datagram_size = cmp::max(v, MAX_DATAGRAM_SIZE);
+        self.max_send_udp_payload_size = cmp::max(v, MAX_SEND_UDP_PAYLOAD_SIZE);
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -306,6 +306,9 @@ const MAX_ACK_RANGES: usize = 68;
 // The highest possible stream ID allowed.
 const MAX_STREAM_ID: u64 = 1 << 60;
 
+/// The default max_datagram_size in the congestion control.
+pub const MAX_DATAGRAM_SIZE: usize = 1200;
+
 // The default length of DATAGRAM queues.
 const DEFAULT_MAX_DGRAM_QUEUE_LEN: usize = 0;
 
@@ -443,6 +446,8 @@ pub struct Config {
 
     dgram_recv_max_queue_len: usize,
     dgram_send_max_queue_len: usize,
+
+    max_datagram_size: usize,
 }
 
 impl Config {
@@ -468,6 +473,8 @@ impl Config {
 
             dgram_recv_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
             dgram_send_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
+
+            max_datagram_size: MAX_DATAGRAM_SIZE,
         })
     }
 
@@ -612,7 +619,7 @@ impl Config {
     /// Sets the `max_udp_payload_size transport` parameter.
     ///
     /// The default value is `65527`.
-    pub fn set_max_udp_payload_size(&mut self, v: u64) {
+    pub fn set_max_recv_udp_payload_size(&mut self, v: u64) {
         self.local_transport_params.max_udp_payload_size = v;
     }
 
@@ -771,6 +778,13 @@ impl Config {
         };
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
+    }
+
+    /// Sets the maximum datagram size used by the congestion control.
+    ///
+    /// The default and minimum value is `1200`.
+    pub fn set_max_send_udp_payload_size(&mut self, v: usize) {
+        self.max_datagram_size = cmp::max(v, MAX_DATAGRAM_SIZE);
     }
 }
 
@@ -6795,7 +6809,7 @@ mod tests {
         config.set_initial_max_data(50000);
         config.set_initial_max_stream_data_bidi_local(50000);
         config.set_initial_max_stream_data_bidi_remote(50000);
-        config.set_max_udp_payload_size(1200);
+        config.set_max_recv_udp_payload_size(1200);
         config.verify_peer(false);
 
         let mut pipe = testing::Pipe::with_client_config(&mut config).unwrap();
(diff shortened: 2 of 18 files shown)


src/lib.rs
```diff
@@ -769,6 +773,13 @@ impl Config {
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
     }
+
+    /// Sets the maximum datagram size used by the congestion control.
+    ///
+    /// The default and minimum value is `1200`.
+    pub fn set_max_datagram_size(&mut self, v: usize) {
```
ghedo, line 780: Why do we need an additional configuration in addition to `max_udp_payload_size`? The name is also really confusing IMO, considering the name of `max_udp_payload_size` and the fact that we have `DATAGRAM` support now.

Also the documentation seems to be wrong, as the default value is `1452` AFAICT, and we don't enforce any minimum value.

src/lib.rs
```diff
@@ -769,6 +773,13 @@ impl Config {
...
+        self.max_datagram_size = cmp::min(v, recovery::MAX_DATAGRAM_SIZE);
```
LPardue, line 781: I don't think this does what you intend? If the specification minimum is 1200 then don't we want something like this?

```
self.max_datagram_size = cmp::max(v, 1200);
``

File: src/recovery/mod.rs [hunk 3]
```diff
@@ -181,6 +183,8 @@ impl Recovery {
 
             congestion_recovery_start_time: None,
 
+            max_datagram_size: config.max_datagram_size,
```
ghedo, line 186: The initial and minimum congestion window values should also change according to the configured value I think.

That is, instead of being constants, they should change dynamically. i.e. `INITIAL_WINDOW` should be replaced by `INITIAL_WINDOW_PACKETS * config.max_datagram_size` and `MINIMUM_WINDOW` should instead be `MINIMUM_WINDOW_PACKETS * r.max_datagram_size` whenever they are used.

src/recovery/mod.rs
```diff
@@ -64,7 +64,7 @@ const INITIAL_WINDOW: usize = INITIAL_WINDOW_PACKETS * MAX_DATAGRAM_SIZE;
 
 const MINIMUM_WINDOW: usize = 2 * MAX_DATAGRAM_SIZE;
 
-const MAX_DATAGRAM_SIZE: usize = 1452;
+pub const MAX_DATAGRAM_SIZE: usize = 1452;
```
ghedo, line 67: Once `INITIAL_WINDOW` and `MINIMUM_WINDOW` are removed, this should be moved to `lib.rs` instead.

ghedo (changes requested):

src/lib.rs
junhochoi: My understanding between `max_udp_payload_size` and `max_datagram_size` is two different thing:
- `max_udp_payload_size`: max udp packet size the endpoint can **receive**, telling the peer not to send a packet bigger than this. https://tools.ietf.org/html/draft-ietf-quic-transport-32#section-18.2
- `max_datagram_size`: max packet size considered when the packet is **sent**, once peer's `max_udp_payload_size` is known, it can be additionally limited by that. https://tools.ietf.org/html/draft-ietf-quic-transport-32#section-14
`max_datagram_size` itself comes from the recovery draft: https://tools.ietf.org/html/draft-ietf-quic-recovery-32#section-7.2

Regarding DATAGRAM support, it looks like it's called `max_datagram_frame_size`. Yes those are confusing names so I'm open to the recommendation.

ghedo: But my question still stands, why do we need 2 options? I would expect that in most cases the maximum size we can send and receive is the same, there's not much point in configuring them separately given that in most cases the same value is configured.

junhochoi: Path MTU can be asymmetric. Also note that max_udp_payload_size's default is 65527 and max_datagram_size is 1200. For example the sender can send 1200 bytes packets but can receive 1500 bytes packets from the peer. If we combine two api (or meaning of the value) into one, it would result in rejecting incoming 1500 bytes one.

src/lib.rs
junhochoi: Fixed. Thanks

src/recovery/mod.rs
junhochoi: Thanks. Now INTIAL_WINDOW and MINIMUM_WINDOW is gone and calculated when Recovery is initialized based on the Config max_datagram_size value.

src/recovery/mod.rs
junhochoi: Moved.

src/lib.rs
junhochoi: Also, I changed the default (MAX_DATAGRAM_SIZE) to 1200 and updated tests.

ghedo: I see what you mean. We could rename the existing `max_udp_payload_size` and call them:
* `max_recv_udp_payload_size` for the existing option (i.e. the transport aprameter)
* `max_send_udp_payload_size` for the new option
?

ghedo:
@junhochoi can you also rebase on master so this can run on the new CI pipeline?

src/lib.rs
junhochoi: Ok, API renamed as follows. I leave internal variable names unchanged.
- `set_max_udp_payload_size()` -> `set_max_recv_udp_payload_size()`
- `set_max_datagram_size()` -> `set_max_send_udp_payload_size()`

junhochoi:
> @junhochoi can you also rebase on master so this can run on the new CI pipeline?

rebased.

junhochoi:
set_max_recv_udp_payload_size() parameter is u64 but I think `usize` is enough. set_max_send_udp_payload_size() use `usize` (size_t). For consistency I think it's good to use `usize` as well there.

src/lib.rs
```diff
@@ -306,6 +306,9 @@ const MAX_ACK_RANGES: usize = 68;
 // The highest possible stream ID allowed.
 const MAX_STREAM_ID: u64 = 1 << 60;
 
+/// The default max_datagram_size in the congestion control.
+pub const MAX_DATAGRAM_SIZE: usize = 1200;
```
ghedo, line 310: This shouldn't be `pub`, as it shouldn't be exported in the quiche public API I think. It's probably also a good idea to rename this to `MAX_SEND_UDP_PAYLOAD_SIZE` to avoid confusion with DATAGRAM as well (it's fine to leave the `recovery` code as is though, since that's what the variable is called in the recovery draft).
```suggestion
const MAX_SEND_UDP_PAYLOAD_SIZE: usize = 1200;
```

src/lib.rs
```diff
@@ -306,6 +306,9 @@ const MAX_ACK_RANGES: usize = 68;
 // The highest possible stream ID allowed.
 const MAX_STREAM_ID: u64 = 1 << 60;
 
+/// The default max_datagram_size in the congestion control.
```
ghedo, line 309:
```suggestion
/// The default max_datagram_size used in congestion control.
```

src/lib.rs
```diff
@@ -772,6 +779,13 @@ impl Config {
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
     }
+
+    /// Sets the maximum datagram size used by the congestion control.
```
ghedo, line 783:
```suggestion
    /// Sets the maximum outgoing UDP payload size used in congestion control.
```

include/quiche.h
```diff
@@ -188,6 +188,9 @@ void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
 
+// Sets the maximum datagram size used by the congestion control.
```
ghedo, line 191:
```suggestion
// Sets the maximum outgoing UDP payload size used in congestion control.
```

src/lib.rs
```diff
@@ -443,6 +446,8 @@ pub struct Config {
 
     dgram_recv_max_queue_len: usize,
     dgram_send_max_queue_len: usize,
+
+    max_datagram_size: usize,
```
ghedo, line 450:
```suggestion
    max_send_udp_payload_size: usize,
```

src/lib.rs
```diff
@@ -468,6 +473,8 @@ impl Config {
 
             dgram_recv_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
             dgram_send_max_queue_len: DEFAULT_MAX_DGRAM_QUEUE_LEN,
+
+            max_datagram_size: MAX_DATAGRAM_SIZE,
```
ghedo, line 477:
```suggestion
            max_send_udp_payload_size: MAX_SEND_UDP_PAYLOAD_SIZE,
```

src/lib.rs
```diff
@@ -772,6 +779,13 @@ impl Config {
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
     }
+
+    /// Sets the maximum datagram size used by the congestion control.
+    ///
+    /// The default and minimum value is `1200`.
+    pub fn set_max_send_udp_payload_size(&mut self, v: usize) {
+        self.max_datagram_size = cmp::max(v, MAX_DATAGRAM_SIZE);
```
ghedo, line 787:
```suggestion
        self.max_send_udp_payload_size = cmp::max(v, MAX_SEND_UDP_PAYLOAD_SIZE);
```

src/recovery/mod.rs
```diff
@@ -171,7 +169,7 @@ impl Recovery {
 
             in_flight_count: [0; packet::EPOCH_COUNT],
 
-            congestion_window: INITIAL_WINDOW,
+            congestion_window: config.max_datagram_size * INITIAL_WINDOW_PACKETS,
```
ghedo, line 172:
```suggestion
            congestion_window: config.max_send_udp_payload_size * INITIAL_WINDOW_PACKETS,
```

src/recovery/mod.rs
```diff
@@ -181,6 +179,8 @@ impl Recovery {
 
             congestion_recovery_start_time: None,
 
+            max_datagram_size: config.max_datagram_size,
```
ghedo, line 182:
```suggestion
            max_datagram_size: config.max_send_udp_payload_size,
```

ghedo (changes requested):
I added a few more nits (mostly related to the rename), but looks good generally.

Though I have a question, given that we are basing all the congestion control calculations on this value configured by the application, should we also make sure that we are not actually generating packets bigger than the limit?

That is, if the application configures `max_send_udp_payload_size` to, say, 1200, but then passes a 1500 bytes buffer to `send()`, should we limit what we write to 1200? It seems like an easy thing to enforce, and would remove a potential footgun for applications.

junhochoi:
> That is, if the application configures `max_send_udp_payload_size` to, say, 1200, but then passes a 1500 bytes buffer to `send()`, should we limit what we write to 1200? It seems like an easy thing to enforce, and would remove a potential footgun for applications.

Ok. 1ff2ac9 will address the issue. PTAL

ghedo:
I applied the following small changes (to avoid going through another round of discussion) after the latest update:
```diff
diff --git a/examples/client.c b/examples/client.c
index 80224c78..ca5ff6e8 100644
--- a/examples/client.c
+++ b/examples/client.c
@@ -244,6 +244,7 @@ int main(int argc, char *argv[]) {
 
     quiche_config_set_max_idle_timeout(config, 5000);
     quiche_config_set_max_recv_udp_payload_size(config, MAX_DATAGRAM_SIZE);
+    quiche_config_set_max_send_udp_payload_size(config, MAX_DATAGRAM_SIZE);
     quiche_config_set_initial_max_data(config, 10000000);
     quiche_config_set_initial_max_stream_data_bidi_local(config, 1000000);
     quiche_config_set_initial_max_stream_data_uni(config, 1000000);
diff --git a/examples/client.rs b/examples/client.rs
index 4e0c361f..ed7af055 100644
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -92,6 +92,7 @@ fn main() {
 
     config.set_max_idle_timeout(5000);
     config.set_max_recv_udp_payload_size(MAX_DATAGRAM_SIZE);
+    config.set_max_send_udp_payload_size(MAX_DATAGRAM_SIZE);
     config.set_initial_max_data(10_000_000);
     config.set_initial_max_stream_data_bidi_local(1_000_000);
     config.set_initial_max_stream_data_bidi_remote(1_000_000);
diff --git a/examples/http3-client.c b/examples/http3-client.c
index b3f632d5..44a862c7 100644
--- a/examples/http3-client.c
+++ b/examples/http3-client.c
@@ -339,6 +339,7 @@ int main(int argc, char *argv[]) {
 
     quiche_config_set_max_idle_timeout(config, 5000);
     quiche_config_set_max_recv_udp_payload_size(config, MAX_DATAGRAM_SIZE);
+    quiche_config_set_max_send_udp_payload_size(config, MAX_DATAGRAM_SIZE);
     quiche_config_set_initial_max_data(config, 10000000);
     quiche_config_set_initial_max_stream_data_bidi_local(config, 1000000);
     quiche_config_set_initial_max_stream_data_bidi_remote(config, 1000000);
diff --git a/examples/http3-client.rs b/examples/http3-client.rs
index d819867f..925b6fd0 100644
--- a/examples/http3-client.rs
+++ b/examples/http3-client.rs
@@ -90,6 +90,7 @@ fn main() {
 
     config.set_max_idle_timeout(5000);
     config.set_max_recv_udp_payload_size(MAX_DATAGRAM_SIZE);
+    config.set_max_send_udp_payload_size(MAX_DATAGRAM_SIZE);
     config.set_initial_max_data(10_000_000);
     config.set_initial_max_stream_data_bidi_local(1_000_000);
     config.set_initial_max_stream_data_bidi_remote(1_000_000);
diff --git a/include/quiche.h b/include/quiche.h
index 6619f022..1c9b7039 100644
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -145,6 +145,9 @@ void quiche_config_set_max_idle_timeout(quiche_config *config, uint64_t v);
 // Sets the `max_udp_payload_size transport` parameter.
 void quiche_config_set_max_recv_udp_payload_size(quiche_config *config, size_t v);
 
+// Sets the maximum outgoing UDP payload size.
+void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
+
 // Sets the `initial_max_data` transport parameter.
 void quiche_config_set_initial_max_data(quiche_config *config, uint64_t v);
 
@@ -188,9 +191,6 @@ void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
 
-// Sets the maximum outgoing UDP payload size used in congestion control.
-void quiche_config_set_max_send_udp_payload_size(quiche_config *config, size_t v);
-
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
 
diff --git a/src/lib.rs b/src/lib.rs
index 73e74727..a03e643b 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -623,6 +623,13 @@ impl Config {
         self.local_transport_params.max_udp_payload_size = v as u64;
     }
 
+    /// Sets the maximum outgoing UDP payload size.
+    ///
+    /// The default and minimum value is `1200`.
+    pub fn set_max_send_udp_payload_size(&mut self, v: usize) {
+        self.max_send_udp_payload_size = cmp::max(v, MAX_SEND_UDP_PAYLOAD_SIZE);
+    }
+
     /// Sets the `initial_max_data` transport parameter.
     ///
     /// When set to a non-zero value quiche will only allow at most `v` bytes
@@ -779,13 +786,6 @@ impl Config {
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
     }
-
-    /// Sets the maximum outgoing UDP payload size used in congestion control.
-    ///
-    /// The default and minimum value is `1200`.
-    pub fn set_max_send_udp_payload_size(&mut self, v: usize) {
-        self.max_send_udp_payload_size = cmp::max(v, MAX_SEND_UDP_PAYLOAD_SIZE);
-    }
 }
 
 /// A QUIC connection.
@@ -2041,7 +2041,7 @@ impl Connection {
         let mut left = b.cap();
 
         // Limit output packet size to respect the sender and receiver's
-        // maximum udp payload size limit.
+        // maximum UDP payload size limit.
         left = cmp::min(left, self.max_send_udp_payload_len());
 
         // Limit output packet size by congestion window size.
@@ -2621,10 +2621,11 @@ impl Connection {
         Ok(written)
     }
 
-    // Returns the maximum len of a packet to be sent. This is a minimum of
-    // the sender and the receiver's maximum udp payload size, except during
-    // the handshake when we haven't parsed transport parameters yet,
-    // so use a default value then.
+    // Returns the maximum size of a packet to be sent.
+    //
+    // This is a minimum of the sender's and the receiver's maximum UDP payload
+    // sizes, except during the handshake when we haven't parsed transport
+    // parameters yet, so use a default value then.
     fn max_send_udp_payload_len(&self) -> usize {
         if self.is_established() {
             // We cap the maximum packet size to 16KB or so, so that it can be
diff --git a/tools/apps/src/bin/quiche-client.rs b/tools/apps/src/bin/quiche-client.rs
index 6b152217..4b7078b9 100644
--- a/tools/apps/src/bin/quiche-client.rs
+++ b/tools/apps/src/bin/quiche-client.rs
@@ -128,6 +128,7 @@ fn main() {
 
     config.set_max_idle_timeout(conn_args.idle_timeout);
     config.set_max_recv_udp_payload_size(MAX_DATAGRAM_SIZE);
+    config.set_max_send_udp_payload_size(MAX_DATAGRAM_SIZE);
     config.set_initial_max_data(conn_args.max_data);
     config.set_initial_max_stream_data_bidi_local(conn_args.max_stream_data);
     config.set_initial_max_stream_data_bidi_remote(conn_args.max_stream_data);
```

ghedo (approved):

PR #693, opened 2020-11-03
initial GitHub Actions workflows

ghedo:
This adds some GitHub Actions workflows. It's not quite as complete as the Travis CI configuration yet (e.g. missing Windows, macOS, Android builds), and I couldn't test the deploy workflow yet (only one way to find out if it works...). Also, we first need to create the secret for the Docker Hub credentials (though we could also just publish the image on the GitHub registry instead which might be easier).

One advantage of Actions vs Travis CI is that we seem to have a bigger number of concurrent jobs available, so the build overall should take less time than on Travis. Not quite sure yet whether we'll end up hitting the monthly limits yet though.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,173 @@
+on: [push, pull_request]
+
+name: Nightly
+
+env:
+  RUSTFLAGS: "-D warnings"
+  TOOLCHAIN: "nightly-2020-10-25"
+
+jobs:
+  quiche:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
+      - name: Install nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ env.TOOLCHAIN }}
+          components: rustfmt
+          override: true
+
+      - name: Install dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+jobs:
+  quiche:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
+      - name: Install stable toolchain
```
LPardue, line 18: don't we want nightly?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+
+  quiche_x86:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
+      - name: Install stable toolchain
```
LPardue, line 64: same

LPardue (approved):
LGTM modulo typos

.github/workflows/nightly.yml
ghedo: Fixed, thanks.

.github/workflows/nightly.yml
ghedo: Fixed, thanks. At some point we should probably find a way to avoid the duplication as well...

PR #694, opened 2020-11-03
write frames into output packet upon creation

ghedo:
Instead of waiting to write frames into the packet to after all frames
are generated, write frames immediately after creation.

For now this doesn't really make much of a difference, but in the future
this will allow writing stream data directly into the packet without
having to copy it into an intermediate buffer first, which should also
decrease the number of allocations.

This requires changing how the payload length is encoded into the packet
as well. Since we don't know the length in advance, we simply reserve a
fixed number of bytes, and later fill this space once all frames have
been encoded.

For the time being we still maintain a vector of all frames. This is
mostly only for logging purposes to avoid logging frames before packet
headers, as the latter are logged alongside the payload length.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2107,7 +2107,7 @@ impl Connection {
         let header_offset = b.off();
 
         // Reserve space for payload length in advance. Since we don't yet know
-        // what the final length will be, we reserver 2 bytes in all cases.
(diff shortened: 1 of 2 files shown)


LPardue (approved):
LGTM modulo nit

PR #695, opened 2020-11-03
stream: use VecDeque to store SendBuf data

ghedo:
This changes the internal storage of SendBuf to a VecDeque, but it
doesn't change the underlying logic. In the future this will allow us to
change how SendBuf data is emitted to avoid the allocation and
additional copy we currently perform.

In most cases insertions to the SendBuf happen at the end of the buffer,
however, due to the fact that buffers need to be stored in order, in
case of retransmissions they might happen at any index in the vector.

While at it, the SendBuf::push_slice() method is also renamed to
SendBuf::write() to be consistent with RecvBuf.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -634,7 +634,7 @@ impl RecvBuf {
     /// This also takes care of enforcing stream flow control limits, as well
     /// as handling incoming data that overlaps data that is already in the
     /// buffer.
-    pub fn push(&mut self, buf: RangeBuf) -> Result<()> {
+    pub fn write(&mut self, buf: RangeBuf) -> Result<()> {
         if buf.max_off() > self.max_data {
             return Err(Error::FlowControl);
         }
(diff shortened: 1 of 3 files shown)


src/stream.rs
```diff
@@ -985,7 +984,33 @@ impl SendBuf {
...
+                    // Slow path: the new data needs to be inserted at a
+                    // specific index in the send buffer.
+                    let mut insert_at = None;
+
+                    for i in 0..self.data.len() {
+                        if buf.off < self.data[i].off {
+                            insert_at = Some(i);
+                            break;
+                        }
+                    }
```
LPardue, line 1005: Maybe a chance to improve this further once binary_search lands? https://github.com/rust-lang/rust/issues/78021

LPardue (approved):

src/stream.rs
ghedo: The implementation looks pretty simple, so we can probably do it without waiting for that PR. Though note that after the next set of changes (once I get them to work) we won't need the slow path anymore as we will always appending to the end, so might end up being wasted work.

PR #696, opened 2020-11-03
ci: fix Docker deployment

ghedo:
* Clone submodules as well
* Don't set name explicitly (to be consistent with other jobs)


ghedo:
We still don't have the secrets setup though, so this will still fail. But at least the workflow configuration should be correct now.

LPardue (approved):

PR #697, opened 2020-11-03
octets: simplify get_varint()

junhochoi:
get_u64() will get in a network order and it passes all tests. Let me know if there is any concerns.

### DIFF
--- a/src/octets.rs
+++ b/src/octets.rs
@@ -186,26 +186,11 @@ impl<'a> Octets<'a> {
         let out = match len {
             1 => u64::from(self.get_u8()?),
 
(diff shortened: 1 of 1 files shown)


lohith-bellad (approved):

ghedo:
I guess this might break quiche on big endian architectures? Not sure if we even work there at all right now though, and it's probably not too important.

junhochoi:
I tried aarch64/arm32 linux but it's all little endian too (arm supports both, but Raspberry Pi linux seems to be configured to use a little endian). If I can find a little endian machine I will test.

PR #700, opened 2020-11-04
SendBuf emit API

ghedo:
This updates the `SendBuf` API so that data can be written directly to the output packet buffer, rather than have to allocate an intermediate buffer, copy the data to that, and then copy the intermediate buffer to the output packet. Refer to the specific commits for more details.

The overall changes are somewhat intricated, so they need to be reviewed and tested carefully...

The next step will be allocating `RangeBuf`s with consistent sizes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1836,21 +1836,26 @@ impl Connection {
                     }
                 },
 
-                frame::Frame::Crypto { data } => {
+                frame::Frame::CryptoHeader { offset, length } => {
                     self.pkt_num_spaces[epoch]
                         .crypto_stream
                         .send
-                        .ack(data.off(), data.len());
+                        .ack_and_drop(offset, length);
                 },
 
-                frame::Frame::Stream { stream_id, data } => {
+                frame::Frame::StreamHeader {
+                    stream_id,
+                    offset,
+                    length,
+                    ..
+                } => {
                     let stream = match self.streams.get_mut(stream_id) {
                         Some(v) => v,
 
                         None => continue,
                     };
 
-                    stream.send.ack(data.off(), data.len());
+                    stream.send.ack_and_drop(offset, length);
 
                     if stream.is_complete() {
                         let local = stream.local;
@@ -1983,11 +1988,19 @@ impl Connection {
         // Process lost frames.
         for lost in self.recovery.lost[epoch].drain(..) {
             match lost {
-                frame::Frame::Crypto { data } => {
-                    self.pkt_num_spaces[epoch].crypto_stream.send.push(data)?;
+                frame::Frame::CryptoHeader { offset, length } => {
+                    self.pkt_num_spaces[epoch]
+                        .crypto_stream
+                        .send
+                        .retransmit(offset, length);
                 },
 
-                frame::Frame::Stream { stream_id, data } => {
+                frame::Frame::StreamHeader {
+                    stream_id,
+                    offset,
+                    length,
+                    fin,
+                } => {
                     let stream = match self.streams.get_mut(stream_id) {
                         Some(v) => v,
 
@@ -1996,9 +2009,9 @@ impl Connection {
 
                     let was_flushable = stream.is_flushable();
 
-                    let empty_fin = data.is_empty() && data.fin();
+                    let empty_fin = length == 0 && fin;
 
-                    stream.send.push(data)?;
+                    stream.send.retransmit(offset, length);
 
                     // If the stream is now flushable push it to the flushable
                     // queue, but only if it wasn't already queued.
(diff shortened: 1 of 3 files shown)


src/stream.rs
```diff
@@ -1222,12 +1316,16 @@ impl RangeBuf {
     /// Splits the buffer into two at the given index.
     pub fn split_off(&mut self, at: usize) -> RangeBuf {
         let buf = RangeBuf {
-            data: self.data.split_off(at),
-            pos: 0,
+            data: self.data.clone(),
+            start: self.start + at,
+            pos: cmp::max(self.pos, self.start + at),
+            len: self.len - at,
```
LPardue, line 1322: do you want a `saturating_sub` here?

src/lib.rs
```diff
@@ -2312,13 +2325,50 @@ impl Connection {
...
+
+            // Write stream data into the packet buffer.
+            let (len, _) = self.pkt_num_spaces[epoch]
                 .crypto_stream
                 .send
-                .pop(crypto_len)?;
+                .emit(&mut crypto_payload.as_mut()[..max_len])?;
+
+            // Encode the frame's header.
+            crypto_hdr.skip(hdr_off)?;
```
LPardue, line 2361: This took a couple of read-overs. I'm still a bit unclear why the skip is required. Is it because `crypto_hdr` contains something before the space for the crypto header?

src/stream.rs
```diff
@@ -1246,13 +1248,22 @@ impl SendBuf {
...
     /// The internal buffer holding the data.
-    data: Vec<u8>,
+    ///
+    /// To avoid neeless allocations when a RangeBuf is split, this can be
+    /// shared between multiple RangeBuf objects, and sliced using the `start`
+    /// and `len` values.
+    data: Arc<Vec<u8>>,
+
+    /// The initial offset within the internal buffer.
+    start: usize,
```
junhochoi, line 1258: Explained individually, but I think it's good to have a short description about the relationship among start, pos, len and off, which is sometimes confusing to read.

src/lib.rs
```diff
@@ -2370,31 +2420,60 @@ impl Connection {
...
+                //
+                // Instead of creating a `frame::Frame` object, encode the frame
+                // directly into the packet buffer.
+                //
+                // First we reserve some space in the output buffer for writing
+                // the frame header (we assume the length field is
+                // always a 2-byte varint as we don't know the
+                // value yet).
+                //
+                // Then we emit the data from the crypto stream's send buffer.
```
junhochoi, line 2435:
```suggestion
                // Then we emit the data from the stream's send buffer.
```

src/stream.rs
```diff
@@ -1182,8 +1274,10 @@ impl RangeBuf {
     /// Creates a new `RangeBuf` from the given slice.
     pub(crate) fn from(buf: &[u8], off: u64, fin: bool) -> RangeBuf {
         RangeBuf {
-            data: Vec::from(buf),
+            data: Arc::new(Vec::from(buf)),
```
junhochoi, line 1277: Just curious why `Arc` is needed now.

LPardue: Arc usage means there is only one underlying data store copy. 

When the split method is called later and `data` is cloned, it only increments the ref count. The new RangeBuf just has a different view onto the underlying data.

When all RangBufs for this data get tidied up, the ref count hits zero and the memory freed at the right time automatically.

src/lib.rs
ghedo: The `split_at()` method is kind of weird, because it splits the buffer from its beginning, rather than the current offset (this IIRC is required for the packet encryption/decryption). So `crypto_hdr` will actually start at the very beginning of the packet and needs to be advanced to the actual offset of the frame. I can probably add a comment to clarify.

src/stream.rs
ghedo: The idea is that `at` should never be grater than `self.len`, otherwise it's a bug. So crashing here might not be a bad thing I think.

ghedo: Though we could add an explicit check for that.

src/lib.rs
LPardue: yeah a comment or a slight rename of the variable would be fine.

src/stream.rs
LPardue: Yeah the std::Vec implementation does this

```
fn assert_failed(at: usize, len: usize) -> ! {
            panic!("`at` split index (is {}) should be <= len (is {})", at, len);
        }

        if at > self.len() {
            assert_failed(at, self.len());
        }
```

junhochoi (approved):

src/stream.rs
ghedo: Fixed.

src/lib.rs
ghedo: I added some more commentary, hopefully it improves things.

src/stream.rs
ghedo: I expanded the `RangeBuf` description with an overviews of these fields, hopefully it doesn't make things more confusing :)

src/lib.rs

PR #704, opened 2020-11-05
h3: add event yielding to poll()

LPardue:
When DATAGRAM support was added, poll() was updated to return
Event::Datagram. We made the decision to prefer returning such
events ahead of stream events (Event::Headers, Event::Data).

There can be cases when an application wants to handle streams
ahead of DATAGRAMS. However, just reordering things inside poll()
wouldn't work for applications that prefer to read DATAGRAMs.

This change introduces a basic yield mechanism into poll().
Poll() starts by processing DATAGRAMs. Once a threshold
count is hit, it moves to processing request
streams. Once that threshold is hit, it moves back to DATAGRAMs.

The thresholds are configured using quiche::h3::Config. By
default there are no thresholds applied, this maintains prior
behavior.

Since this changes poll(), I've added several unit tests for both
simple DATAGRAM sending/receiving, and DATAGRAM poll yielding.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -480,6 +480,12 @@ void quiche_h3_config_set_qpack_max_table_capacity(quiche_h3_config *config, uin
 // Sets the `SETTINGS_QPACK_BLOCKED_STREAMS` setting.
 void quiche_h3_config_set_qpack_blocked_streams(quiche_h3_config *config, uint64_t v);
 
+// Sets the `quiche_h3_conn_poll()` repetition threshold for DATAGRAM events.
+void quiche_h3_config_set_dgram_poll_threshold(quiche_h3_config *config, uint64_t v);
+
+// Sets the `quiche_h3_conn_poll()` repetition threshold for stream events.
+void quiche_h3_config_set_dgram_poll_threshold(quiche_h3_config *config, uint64_t v);
+
 // Frees the HTTP/3 config object.
 void quiche_h3_config_free(quiche_h3_config *config);
 
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -444,6 +444,8 @@ pub struct Config {
     max_header_list_size: Option<u64>,
     qpack_max_table_capacity: Option<u64>,
     qpack_blocked_streams: Option<u64>,
+    dgram_poll_threshold: u64,
+    stream_poll_threshold: u64,
 }
 
 impl Config {
@@ -453,6 +455,8 @@ impl Config {
             max_header_list_size: None,
             qpack_max_table_capacity: None,
             qpack_blocked_streams: None,
+            dgram_poll_threshold: std::u64::MAX,
+            stream_poll_threshold: std::u64::MAX,
         })
     }
 
@@ -475,6 +479,46 @@ impl Config {
     /// The default value is `0`.
     pub fn set_qpack_blocked_streams(&mut self, v: u64) {
         self.qpack_blocked_streams = Some(v);
+    }
+
+    /// Sets the [`poll()`] repetition threshold for DATAGRAM events.
+    ///
+    /// When there is at least one pending DATAGRAM, [`poll()`] can return an
+    /// [`Event::Datagram`] and the application should use [`recv_dgram()`] to
+    /// remove it from the pending queue. If all DATAGRAMs are not removed,
+    /// repeat [`poll()`] calls can continue to return events. By default there
+    /// is no threshold to the number of times this can happen.
+    ///
+    /// This parameter adds a threshold for the number of times a DATAGRAM event
+    /// will be returned by [`poll()`]. Once reached, stream events will be
+    /// returned.
+    ///
+    /// [`poll()`]: struct.Connection.html#method.poll
+    /// [`recv_dgram()`]: struct.Connection.html#method.recv_dgram
+    /// [`Event::Datagram`]: enum.Event.html#variant.Datagram
+    pub fn set_dgram_poll_threshold(&mut self, v: u64) {
+        self.dgram_poll_threshold = v;
+    }
+
+    /// Sets the [`poll()`] repetition threshold for stream events.
+    ///
+    /// When there is at least one readable stream, [`poll()`] can return an
+    /// [`Event::Headers`], [`Event::Data`] or [`Event::Finished`]. While the
+    /// stream remains readable repeat [`poll()`] calls can continue to return
+    /// stream events. By default there is no threshold to the number of times
+    /// this can happen.
+    ///
+    /// This parameter adds a threshold for the number of times a stream event
+    /// of any type will be returned by [`poll()`]. Once reached, datagram
+    /// events will be returned.
+    ///
+    /// [`poll()`]: struct.Connection.html#method.poll
+    /// [`recv_body)`]: struct.Connection.html#method.body
+    /// [`Event::Headers`]: enum.Event.html#variant.Headers
+    /// [`Event::Data`]: enum.Event.html#variant.Data
+    /// [`Event::Finished`]: enum.Event.html#variant.Finished
+    pub fn set_stream_poll_threshold(&mut self, v: u64) {
+        self.stream_poll_threshold = v;
     }
 }
 
@@ -550,15 +594,25 @@ pub enum Event {
     /// to retrieve the data from the stream.
     ///
     /// This event will keep being reported until all the available data is
-    /// retrieved by the application.
+    /// retrieved by the application or the event [threshold] is reached.
     ///
     /// [`recv_body()`]: struct.Connection.html#method.recv_body
+    /// [threshold]: struct.Config.html#method.set_stream_poll_threshold
     Data,
 
     /// Stream was closed,
     Finished,
 
     /// DATAGRAM was received.
+    ///
+    /// This indicates that the application can use the [`recv_dgram()`] method
+    /// to retrieve the HTTP/3 DATAGRAM.
+    ///
+    /// This event will keep being reported until all the available data is
+    /// retrieved by the application or the event [threshold] is reached.
+    ///
+    /// [`recv_dgram()`]: struct.Connection.html#method.recv_dgram
+    /// [threshold]: struct.Config.html#method.set_dgram_poll_threshold
     Datagram,
 
     /// GOAWAY was received.
@@ -606,6 +660,11 @@ pub struct Connection {
 
     local_goaway_id: Option<u64>,
     peer_goaway_id: Option<u64>,
+
+    dgram_event_count: u64,
+    dgram_poll_threshold: u64,
+    stream_event_count: u64,
+    stream_poll_threshold: u64,
 }
 
 impl Connection {
@@ -662,6 +721,11 @@ impl Connection {
 
             local_goaway_id: None,
             peer_goaway_id: None,
+
+            dgram_event_count: 0,
+            dgram_poll_threshold: config.dgram_poll_threshold,
+            stream_event_count: 0,
+            stream_poll_threshold: config.stream_poll_threshold,
         })
     }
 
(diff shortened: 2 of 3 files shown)


LPardue:
I need to do a bit more tweaking on the poll() internals.

LPardue:
The tweaks actually ended up changing the API too :) But I think what's there now is conceptually easier to follow.

src/h3/mod.rs
```diff
@@ -415,6 +417,8 @@ impl Config {
             max_header_list_size: None,
             qpack_max_table_capacity: None,
             qpack_blocked_streams: None,
+            dgram_poll_threshold: None,
```
ghedo, line 420: Instead of using `Option` here, we could just set the default value to `std::usize::MAX` I think, as it seems like it would simplify things as we wouldn't need to deal with `None` anymore later on.

src/h3/mod.rs
```diff
@@ -1080,22 +1108,45 @@ impl Connection {
...
         // Process HTTP/3 data from readable streams.
+        if let Some(thresh) = self.stream_poll_threshold {
+            if self.stream_poll_count > thresh {
+                self.dgram_poll_count = 0;
+                self.stream_poll_count = 0;
+
+                return Err(Error::Done);
+            }
+
+            self.stream_poll_count += 1;
```
ghedo, line 1147: I think this should be inside the loop, and only increased if we actually return a stream event. Otherwise it seems like it would be increased even when there are no stream events and `Done` is returned?

src/h3/mod.rs
```diff
@@ -1080,22 +1108,45 @@ impl Connection {
...
+
+                Err(crate::Error::Done) => (),
+
+                Err(e) => return Err(Error::TransportError(e)),
+            };
+        }
 
         // Process HTTP/3 data from readable streams.
+        if let Some(thresh) = self.stream_poll_threshold {
+            if self.stream_poll_count > thresh {
```
ghedo, line 1140: Should probably also move this inside the loop, to after `stream_poll_count` is incremented (see other comment), and change the condition to `>=`. Then we can also avoid returning `Done` here.

src/h3/mod.rs
ghedo: Also, type should probably be `u64` as this doesn't really refer to any "memory-related quantities".

include/quiche.h
```diff
@@ -443,6 +443,12 @@ void quiche_h3_config_set_qpack_max_table_capacity(quiche_h3_config *config, uin
 // Sets the `SETTINGS_QPACK_BLOCKED_STREAMS` setting.
 void quiche_h3_config_set_qpack_blocked_streams(quiche_h3_config *config, uint64_t v);
 
+// Sets the `quiche_h3_conn_poll()` repetition threshold for DATAGRAMs.
```
ghedo, line 446:
```suggestion
// Sets the `quiche_h3_conn_poll()` repetition threshold for DATAGRAM events.
```

include/quiche.h
```diff
@@ -443,6 +443,12 @@ void quiche_h3_config_set_qpack_max_table_capacity(quiche_h3_config *config, uin
...
+void quiche_h3_config_set_dgram_poll_threshold(quiche_h3_config *config, size_t v);
+
+// Sets the `quiche_h3_conn_poll()` repetition threshold for streams.
```
ghedo, line 449:
```suggestion
// Sets the `quiche_h3_conn_poll()` repetition threshold for stream events.
```

src/h3/mod.rs
```diff
@@ -438,6 +442,20 @@ impl Config {
     pub fn set_qpack_blocked_streams(&mut self, v: u64) {
         self.qpack_blocked_streams = Some(v);
     }
+
+    /// Sets the `poll()` repetition threshold for DATAGRAMs.
```
ghedo, line 446:
```suggestion
    /// Sets the `poll()` repetition threshold for DATAGRAM events.
```

src/h3/mod.rs
```diff
@@ -438,6 +442,20 @@ impl Config {
...
+    ///
+    /// The default is no threshold.
+    pub fn set_dgram_poll_threshold(&mut self, v: usize) {
+        self.dgram_poll_threshold = Some(v);
+    }
+
+    /// Sets the `poll()` repetition threshold for streams.
```
ghedo, line 453:
```suggestion
    /// Sets the `poll()` repetition threshold for stream events.
```

ghedo (changes requested):
I'm somewhat confused by the count reset logic, so some of the comments I made might not work, but I think there are some opportunities to simplify things a bit there.

src/h3/mod.rs
LPardue: Yeah this seems a bit wonky but during testing the problem I hit is that once you're into processing streams, you have to have a stream event in order to get back to datagrams - but that might never happen e,g, a basic MASQUE session.

The current code is intended to say "I'll try to process only stream events for N calls of poll, after that I'll go back to trying datagrams". It's not a count of the number of events, its the poll count.

src/h3/mod.rs
LPardue: I was trying to be clever, let me try the KISS option.

src/h3/mod.rs
ghedo: Normally when `poll()` returns `Done` the application will stop calling it until, e.g. new packets arrive or something. So this behavior might end up delaying the delivery of DATAGRAM events for several RTTs I think.

Instead of doing that, when checking if DATAGRAM can be sent, we can also call `conn.streams.has_readable()` to see if there even are any streams that are readable. so e.g. `if !conn.streams.has_readable() || below dgram threshold {` or something like that.

LPardue:
@ghedo I've added a commit that simplifies things and avoids returning Error::Done. Is the intent clearer now?

src/h3/mod.rs
```diff
@@ -406,6 +406,8 @@ pub struct Config {
     max_header_list_size: Option<u64>,
     qpack_max_table_capacity: Option<u64>,
     qpack_blocked_streams: Option<u64>,
+    dgram_poll_threshold: u64,
+    stream_poll_threshold: u64,
```
junhochoi, line 410: Does it need to be `u64`?

LPardue: That's what @ghedo suggested in https://github.com/cloudflare/quiche/pull/704#discussion_r519893116

src/h3/mod.rs
LPardue: I hadn't appreciated the importance of `Done` before. This causes a mild quandry because there is no outright correct order. Ideally, while there is any DATAGRAM or stream data buffered and ready to be processed, we want poll to return an event even if that "bends" the yield thresholds. We really don't want to have call `poll()` multiple times and have it return `Done` in order to get over the threshold to start processing the data we care about. 

I've tried to address this in https://github.com/cloudflare/quiche/pull/704/commits/91eab5f0dc1ac461e20c84a7e0baf48d329d073e. With that change, the intent is that `poll()` will jump into processing whatever type the connection-global threshold prefers. If there is nothing to do for a given type, switch to the other one. It only returns `Done` when there is nothing to do for either type.

src/h3/mod.rs
```diff
@@ -476,6 +480,20 @@ impl Config {
     pub fn set_qpack_blocked_streams(&mut self, v: u64) {
         self.qpack_blocked_streams = Some(v);
     }
+
+    /// Sets the `poll()` repetition threshold for DATAGRAM events.
```
ghedo, line 484: We might want to go into a bit more detail on what these config options do, and why one might want to use them, as it's not immediately evident IMO.

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
...
-        match conn.dgram_recv_peek(&mut d, 8) {
-            Ok(_) => {
-                let mut b = octets::Octets::with_slice(&d);
-                let flow_id = b.get_varint()?;
-                return Ok((flow_id, Event::Datagram));
-            },
+        if self.dgram_event_count < self.dgram_poll_threshold &&
+            conn.dgram_recv_front_len().is_some()
+        {
+            match conn.dgram_recv_peek(&mut d, 8) {
```
ghedo, line 1167: Might be a good idea to refactor this into a separate method, so it can be reused later rather than duplicating the code, though I'm not too sure it would actually simplify things, as you'd still need the `match` here anyway.

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
...
+        }
         let mut d = [0; 8];
 
-        match conn.dgram_recv_peek(&mut d, 8) {
-            Ok(_) => {
-                let mut b = octets::Octets::with_slice(&d);
-                let flow_id = b.get_varint()?;
-                return Ok((flow_id, Event::Datagram));
-            },
+        if self.dgram_event_count < self.dgram_poll_threshold &&
```
ghedo, line 1164:
```suggestion
        // Process queued DATAGRAMs if the poll threshold allows it.
        if self.dgram_event_count < self.dgram_poll_threshold &&
```

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
...
 
-            Err(e) => return Err(Error::TransportError(e)),
-        };
+                Err(crate::Error::Done) => (),
+
+                Err(e) => return Err(Error::TransportError(e)),
+            };
+        }
 
-        // Process HTTP/3 data from readable streams.
```
ghedo, line 1143: This comment should stay IMO.

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
...
+                // TODO: check if stream is completed so it can be freed
+                if let Some(ev) = ev {
+                    self.stream_event_count += 1;
+                    return Ok(ev);
+                }
             }
+        }
 
-            // TODO: check if stream is completed so it can be freed
+        // no event generated at this point, so try dgram again
```
ghedo, line 1208:
```suggestion
        // No event generated at this point, so try dgram again.
```

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
...
+                    return Ok((flow_id, Event::Datagram));
+                },
+
+                Err(crate::Error::Done) => (),
+
+                Err(e) => return Err(Error::TransportError(e)),
+            };
         }
 
         // Process finished streams list once again, to make sure `Finished`
```
ghedo, line 1225: I _think_ this should go before the second DATAGRAM block, as finished events are prioritized over dgrams in the previous block, but it's probably not super important..

src/h3/mod.rs
```diff
@@ -1125,42 +1153,73 @@ impl Connection {
             return Ok((finished, Event::Finished));
         }
 
-        // Process DATAGRAMs
+        if self.dgram_event_count == self.dgram_poll_threshold &&
```
ghedo, line 1156:
```suggestion
        // Both stream and DATAGRAM event limits have been reached,
        // so reset and start again.
        if self.dgram_event_count == self.dgram_poll_threshold &&
```

ghedo (changes requested):
Thanks, the latest logic looks a lot more clear! I added a few comments, but the overall implementation looks good to me.

src/h3/mod.rs
LPardue: Good idea, see `process_dgrams() which still returns a Result but handles the boring mapping between error types, which makes the code in `poll()` a little tidier.

src/h3/mod.rs

LPardue:
Thanks for the comments. 

I've incorporated them all, and added more description to `Event::Datagram` itself to describe the usage expections in tandem with the new thresholds.

I've also squashed all the commits down.

src/h3/mod.rs
```diff
@@ -476,6 +480,46 @@ impl Config {
...
         self.qpack_blocked_streams = Some(v);
     }
+
+    /// Sets the [`poll()`] repetition threshold for DATAGRAM events.
+    ///
+    /// When there is at least one pending DATAGRAM, [`poll()`] can return
+    /// an [`Event::Datagram`] and the application should use [`recv_dgram()`]
+    /// to remove it from the pending queue. If all DATAGRAMs are not removed,
+    /// repeat [`poll()`] calls can continue to return events. By
+    /// default there is no threshold to the number times this can happen.
```
ghedo, line 490:
```suggestion
    /// default there is no threshold to the number of times this can happen.
```

src/h3/mod.rs
```diff
@@ -476,6 +480,46 @@ impl Config {
...
+        self.dgram_poll_threshold = v;
+    }
+
+    /// Sets the [`poll()`] repetition threshold for stream events.
+    ///
+    /// When there is at least one readable stream, [`poll()`] can return
+    /// an [`Event::Headers`], [`Event::Data`] or [`Event::Finished`]. While
+    /// the stream remains readable
+    /// repeat [`poll()`] calls can continue to return stream events. By
+    /// default there is no threshold to the number times this can happen.
```
ghedo, line 509:
```suggestion
    /// default there is no threshold to the number of times this can happen.
```

ghedo (approved):
LGTM, added a couple of nits, but I will fixe those myself.

ghedo (approved):

PR #705, opened 2020-11-05
lib: add is_readable() helper method

LPardue:


inikulin (approved):

src/lib.rs
```diff
@@ -3233,6 +3233,11 @@ impl Connection {
             .is_none()
     }
 
+    /// Returns whether there is stream or DATAGRAM data available to read.
+    pub fn is_app_data_readable(&self) -> bool {
+        self.readable().next().is_some() || self.dgram_recv_front_len().is_some()
```
ghedo, line 3238: You can just do `self.readable().len() > 0` which is slightly simpler, but more generally creating the iterator is not free, and then it gets discarded so we are basically paying the price for nothing here.

A better alternative would be to add a `has_readable()` method like https://github.com/cloudflare/quiche/blob/master/src/stream.rs#L431 and then call that from here, so this method becomes effectively free.

src/lib.rs
```diff
@@ -3233,6 +3233,11 @@ impl Connection {
             .is_none()
     }
 
+    /// Returns whether there is stream or DATAGRAM data available to read.
+    pub fn is_app_data_readable(&self) -> bool {
```
ghedo, line 3237: I think we can just call this `is_readable()` to be consistent with `readable()` (and also app data is really the only thing that can be read from the connectio anyway).

ghedo (changes requested):

src/lib.rs
inikulin: You can't get length of the iterator without collecting it. Also, `readable` doesn't give information about pending dgrams

inikulin: But I agree that it would be even better without dealing with the iterator if that's possible.

ghedo: Just FTR, `StreamIter` implements `ExactSizeIterator` https://doc.rust-lang.org/nightly/core/iter/trait.ExactSizeIterator.html so you can get the length without collecting I think.

ghedo: (also, I meant use `len()` for the `self.readable()` part, the dgram check would still be needed).

LPardue:
Ok, I've tweaked the name and added a test to prove it works as intended.

ghedo:
Just FTR, I applied the following diff myself (no point in going through another round of comment/fix since it's pretty trivial, basically just moving the new method to where the other `is_*` methods are, and sync up the FFI documentation):
```diff
diff --git a/include/quiche.h b/include/quiche.h
index aa7b05df..e773bc1c 100644
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -287,9 +287,6 @@ quiche_stream_iter *quiche_conn_readable(quiche_conn *conn);
 // Returns an iterator over streams that can be written to.
 quiche_stream_iter *quiche_conn_writable(quiche_conn *conn);
 
-// Return true if there is stream or DATAGRAM data available to read.
-bool quiche_conn_is_readable(quiche_conn *conn);
-
 // Returns the amount of time until the next timeout event, in nanoseconds.
 uint64_t quiche_conn_timeout_as_nanos(quiche_conn *conn);
 
@@ -314,6 +311,9 @@ bool quiche_conn_is_established(quiche_conn *conn);
 // enough to send or receive early data.
 bool quiche_conn_is_in_early_data(quiche_conn *conn);
 
+// Returns whether there is stream or DATAGRAM data available to read.
+bool quiche_conn_is_readable(quiche_conn *conn);
+
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
diff --git a/src/lib.rs b/src/lib.rs
index 9f28e52b..2b1a3c2f 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3255,11 +3255,6 @@ impl Connection {
             .is_none()
     }
 
-    /// Returns whether there is stream or DATAGRAM data available to read.
-    pub fn is_readable(&self) -> bool {
-        self.streams.has_readable() || self.dgram_recv_front_len().is_some()
-    }
-
     /// Returns the amount of time until the next timeout event.
     ///
     /// Once the given duration has elapsed, the [`on_timeout()`] method should
@@ -3431,6 +3426,11 @@ impl Connection {
         self.handshake.is_in_early_data()
     }
 
+    /// Returns whether there is stream or DATAGRAM data available to read.
+    pub fn is_readable(&self) -> bool {
+        self.streams.has_readable() || self.dgram_recv_front_len().is_some()
+    }
+
     /// Returns true if the connection is closed.
     ///
     /// If this returns true, the connection object can be dropped.
```

ghedo (approved):

PR #707, opened 2020-11-09
Extract quiche-client into a lib that can be used programmatically

inikulin:
These are just minimal refactorings required to turn `quiche-apps` and namely the client into a lib that can be used in Rust unit tests by other crates. This currently fails the formatting tests because of the non-standard settings, @LPardue generously offered to run rust fmt with proper settings in a follow up commit.


tools/apps/src/client.rs
```diff
@@ -0,0 +1,487 @@
+use crate::common::*;
```
LPardue, line 1: Missing license

tools/apps/src/bin/quiche-client.rs
```diff
@@ -83,430 +68,10 @@ fn main() {
...
-            dump_json,
-            urls,
-            req_headers,
-            reqs_cardinal,
-            no_verify,
-            body,
-            method,
-            connect_to,
-        }
+    match request(args, conn_args, stdout_sink) {
```
LPardue, line 71: `request` is a bit of an overly generic name, especially since the client is able to initiate many requests (depending in the args). Maybe something like `run` and to compliment `RunError`?

inikulin: I don't have any preferences here. @ghedo  is `run` fine by you?

inikulin:
@LPardue thanks for the follow-ups!

tools/apps/src/bin/quiche-client.rs
junhochoi: `make_requests`?

LPardue: I have modes that open QUIC connections and don't make requests e.g. siduck or a WebTransport branch that I have.

inikulin: How about `client::send`?

LPardue: How about "connect"? It communicates the intent while abstracting over some notions of sending or requesting

inikulin: fine by me

inikulin: updated the PR

LPardue (dismissed):

LPardue (approved):

PR #710, opened 2020-11-10
CI: Use matrix for building multiple targets

junhochoi:
iOS and Windows: use `matrix` for building multiple targets in parallel.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -101,39 +101,31 @@ jobs:
 
   quiche_ios:
     runs-on: macos-latest
+    strategy:
+      matrix:
+        target: ["x86_64-apple-ios", "aarch64-apple-ios"]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
         with:
           submodules: 'recursive'
 
-      - name: Install stable toolchain x86_64
+      - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: ${{ env.TOOLCHAIN }}
-          target: x86_64-apple-ios
+          target: ${{ matrix.target }}
           override: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/stable.yml
```diff
@@ -101,39 +101,36 @@ jobs:
...
+        uses: actions-rs/install@v0.1
+        with:
+          crate: cargo-lipo
 
       - name: Run cargo lipo
-        uses: actions-rs/cargo@v1
-        with:
-          command: lipo
-          args: --verbose
+        run: cargo lipo --verbose --targets ${{ matrix.target }}
```
ghedo, line 127: Isn't the whole point of lipo that it builds a universal binary with all the architectures? Otherwise if we are just building each architecture one by one we could just use `cargo build`, no?

junhochoi: Right. We can go back to the previous state but I removed cargo-lipo now. It doesn't build a universal binary any more, but probably enough for testing a build.

.github/workflows/stable.yml
```diff
@@ -145,20 +115,17 @@ jobs:
...
-      - name: Install dependencies
-        run: choco install nasm
-
-      - name: Run cargo test
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --target=x86_64-pc-windows-msvc --verbose
+      - name: Run cargo build
+        run: cargo build --verbose --target ${{ matrix.target }}
```
ghedo, line 122: To be consistent with the other builds this should use the proper action:
```suggestion
        uses: actions-rs/cargo@v1
        with:
          command: build
          args: --target=${{ matrix.target }} --verbose
```

ghedo (changes requested):
One nitpick, but looks good otherwise.

PR #715, opened 2020-11-10
make Config and Connection Send + Sync

ghedo:
This makes both Config and Connection Sync-able (in addition to
Send-able which they were already).

The biggest hurdle is the fact that the BoringSSL SSL_CTX and SSL
structures are not thread-safe, so they need to be wrapped in Mutex.

In the case of Config this probably has no real downside, since an
application is not supposed to keep changing it once the initial
configuration is applied.

For Connection, this might affect performance slightly, but mostly just
during the handshake. However some of Connection's methods (like
`is_established()`) could be called repeatedly by the application even
after the handshake, so we "cache" the return value (since they won't
change after the handshake is complete) so we won't need to go through
the mutex anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,6 +271,8 @@ use std::time;
 use std::pin::Pin;
 use std::str::FromStr;
 
+use std::sync::Mutex;
+
 /// The current QUIC wire version.
 pub const PROTOCOL_VERSION: u32 = PROTOCOL_VERSION_DRAFT29;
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -434,7 +436,11 @@ pub struct Config {
 
     version: u32,
 
-    tls_ctx: tls::Context,
+    // BoringSSL's SSL_CTX structure is technically safe to share across threads
+    // but once shared, functions tha modify it can't be used anymore. We can't
```
LPardue, line 440:
```suggestion
    // but once shared, functions that modify it can't be used any more. We can't
```

src/lib.rs
```diff
@@ -434,7 +436,11 @@ pub struct Config {
...
+    // encode that in Rust, so just make it Send+Sync with a mutex to fullfill
```
LPardue, line 441:
```suggestion
    // encode that in Rust, so just make it Send+Sync with a mutex to fulfil
```

ghedo:
FTR, I ended up implementing the "caching" for `is_established()` and `application_proto()` in the end (for the latter, this is needed to make FFI work properly). Commit and PR description updated accordingly.

LPardue (approved):

ghedo:
Rebased.

LPardue (approved):

PR #727, opened 2020-11-17
ci: remove quiche_x86 build from nightly

ghedo:
There isn't much point to build this on both stable and nightly.


junhochoi (approved):
:shipit:

PR #728, opened 2020-11-17
ensure HANDSHAKE_DONE frame is sent

ghedo:
Currently we depend on the fact that a 1-RTT packet needs sending for
other reasons, in order to send that frame. Usually that happens because
we need to send TLS session tickets, or initialize an HTTP/3 connection,
but that's not guaranteed.


LPardue (approved):

ghedo:
Fixed format, derp.

LPardue (approved):

PR #735, opened 2020-11-20
add Connection::is_draining() method

ghedo:
This can be useful to applications to know when to stop sending /
expecting new data to / from a connection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1489,7 +1489,7 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        if self.is_closed() || self.draining_timer.is_some() {
+        if self.is_closed() || self.is_draining() {
             return Err(Error::Done);
         }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3561,6 +3563,23 @@ impl Connection {
         self.streams.has_readable() || self.dgram_recv_front_len().is_some()
     }
 
+    /// Returns true if the connection is draining.
+    ///
+    /// If this returns true, the connection object cannot yet be dropped, but
+    /// no new application data can be sent or received. An application should
```
LPardue, line 3569: Can we describe what to expect if stream_send/dgram_send is called when `is_draining` is true?

I'm  not so worried about documenting receive, although someone not so familiar with QUIC might be confused about whether exisitng buffered data would be availalable when stream_recv() is called

ghedo: I think that would need actually deciding what we want to do in those cases. Right now both `stream_send()` and `stream_recv()` kind of ignore this state (for no real reason), but of course if you try sending stream data it's not actually sent on the wire.

Maybe we should return `Error::InvalidState` from `stream_send()` in this case? Though that's probably for a separate PR.

LPardue: Agree this change doesn't change what we already do, and what we already do or don't document. Happy to punt that to a separate follow up as long as we remember to get back to it.

LPardue (approved):

PR #736, opened 2020-11-20
use ConnectionId more widely

ghedo:
Basically this replaces instances of Vec<u8> used to represent
connection IDs, with ConnectionId.

This also implements a few additional traits for ConnectionId to make
its use a bit easier.

---

This doesn't yet change the public API (well, no more than e0394a87cd75f30caa4ce1863bed3017ef0f7e94 does), but I think we should update `accept()`, `connect()`, `retry()`, etc... at some point to use `ConnectionId` instead of `&[u8]`.

Also, in the server examples/app,  there are a couple more clones of IDs than there were before, because of some lifetime issues I ran into. Will probably need more sleep to figure out a better way around that though :)

### DIFF
--- a/src/packet.rs
+++ b/src/packet.rs
@@ -130,10 +130,8 @@ impl Type {
 }
 
 /// A QUIC connection ID.
-#[derive(Clone)]
 pub struct ConnectionId<'a>(ConnectionIdInner<'a>);
 
-#[derive(Clone)]
 enum ConnectionIdInner<'a> {
     Vec(Vec<u8>),
     Ref(&'a [u8]),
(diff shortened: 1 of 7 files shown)


LPardue:
RIP `hex_dump`. Printable ConnectionId is nice usability win.

src/packet.rs
```diff
@@ -166,6 +178,12 @@ impl<'a> AsRef<[u8]> for ConnectionId<'a> {
     }
 }
 
+impl<'a> std::hash::Hash for ConnectionId<'a> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
inikulin, line 182: Nit: use `#[inline]` for small functions in public API. Rust doesn't inline functions without it during cross-crate compilation even if LTO is enabled.

LPardue (dismissed):
lgtm

src/packet.rs
ghedo: We never do `#[inline]` right now, so I'll go through over the public API and do this in a separate PR.

ghedo:
I merged this now, but I still want to try to figure out the clone thing in the server examples/apps later on.

src/packet.rs
ghedo: https://github.com/cloudflare/quiche/pull/751

PR #737, opened 2020-11-21
cc: RFC3465 ABC Slow Start

junhochoi:
Implements https://tools.ietf.org/html/rfc3465#section-2.2

During slow start, when there is an ACK frame ack'ing
too many packet at once, cwnd will go too high. For example
there is an single ACK ack'ing 5 packets, cwnd will be increased
by 5 x MSS. This can lead to unexpected burst of cwnd, such as
ACK for multiple received STREAM frames, or recovering from a loss.

To prevent from such burst, during slow start an single ACK
will not increase cwnd more than MSS x ABC_L constant. ABC_L is
defined as 2 in RFC3465, and QUIC transport draft also recommends
ACK'ing every two packets (transport-32 section 13.2.2), so this looks
a good number to start.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -170,15 +170,26 @@ fn on_packet_acked(
 
     if r.congestion_window < r.ssthresh {
         // Slow start.
+        let new_cwnd;
+
         if r.hystart.enabled() && epoch == packet::EPOCH_APPLICATION {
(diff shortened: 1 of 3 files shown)


ghedo:
I _think_ this is good, but need to read the RFC3465 section again to understand it better.

ghedo (approved):

Issue #740, opened 2020-11-22
Possible problem with empty FIN packets


paneu:
I don't know whether this problem I am seeing is a bug in quiche or a problem in the way the API is being used. I have noticed this with using `curl` as the client and `h2o` as the server. Intermittently, curl is hanging at the end of a download, i.e. the entire response body is transferred but curl is not terminating. I have narrowed this down to the following condition:

- the client is receiving data from the socket and handling this properly as far as I can tell, i.e. `recv`, `quiche_conn_recv`, `quiche_h3_conn_poll` & `quiche_h3_recv_body`
- at some point the entire body is being transferred and handled by the client. Even though the entire body has been transferred, the STREAM is not being FINinshed by the server yet
- when the final 0-length STREAM frame with the FIN bit sent, the client (in this case curl), goes through `recv`, `quiche_conn_recv` & `quiche_h3_conn_poll` again. However, the poll function is not raising a `QUICHE_H3_EVENT_FINISHED` but is returning `QUICHE_ERR_DONE`. As no more data is being sent by the server, `quiche_conn_recv` is not being called anymore, effectively leading to a deadlock.
- in `h3::poll`, the stream is added to the list of finished stream, though it is not until the next invocation of `h3::poll` that the stream is being reported finished to the API consumer

I have tried to express this behaviour into a test case, though, as I am not familiar with neither quiche nor rust in general, I might have missed something. If I did, please let me know. If necessary, I can provide a PCAP of the problem (enabling quiche debugging leads to the error disappearing as curl is then taking long enough to process everything for the final FIN to arrive in time to be handled in `quiche_h3_recv_body`).
You can find the test here: paneu/quiche@18755024305839aa2d9ee13169c740458c5252e9. On master, this test is failing. I have tried to implement a fix that is not breaking any other test case: paneu/quiche@5dc3153d78e4dd6d14110ccfee9be25b35ce754a

As I said, I am not sure whether this is a bug in quiche or whether the API is supposed to be used in a different way.

ghedo:
Thanks for looking into this and reporting! Could you please open a PR with your changes (after squashing the two commits together) so we can better review and discuss them?

paneu:
Done: #741

PR #741, opened 2020-11-23
h3: Ensure Event::Finished is raised in poll

paneu:
See also #740

If the server is finishing the stream with sending an empty frame with
the FIN bit set after the client has recv'd the body of the response,
the subsequent poll is returning Error::Done instead of raising
Event::Finished. If there is no more data to be received, the client
might not call poll again. Thus, make sure to raise the event properly
in such cases.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1163,6 +1163,11 @@ impl Connection {
             }
         }
 
+        // Process finished streams list once again.
+        if let Some(finished) = self.finished_streams.pop_front() {
+            return Ok((finished, Event::Finished));
+        }
+
         Err(Error::Done)
     }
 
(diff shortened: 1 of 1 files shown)


src/h3/mod.rs
```diff
@@ -1163,6 +1163,11 @@ impl Connection {
             }
         }
 
+        // Process finished streams list once again.
```
ghedo, line 1166:
```suggestion
        // Process finished streams list once again, to make sure `Finished`
        // events are returned when receiving empty stream frames with the fin
        // flag set.
```

ghedo (changes requested):
Thank you @paneu! I added a suggestion to expand a comment a bit to better explain why we are processing the finished streams list twice, but otherwise this looks good to me.

ghedo (approved):

ghedo:
I applied the change myself and merged, thanks again @paneu!

PR #742, opened 2020-11-23
lib: defer collecting streams when ack'ing empty frames

LPardue:
When processing ACK frames, quiche collects streams if the conditions
are suitable. Previously, endpoints exchanging 0-length STREAM frames
could confuse things, causing the stream to be collected while still
readable(). Applications that attempted to call stream_recv() would
encounter a InvalidStreamState error.

This change adds a check for is_readable() to the ACK processing, and
a test case to demonstrate the sequence of events that previously
triggered the error.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1893,7 +1893,10 @@ impl Connection {
 
                     stream.send.ack_and_drop(offset, length);
 
-                    if stream.is_complete() {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@LPardue thanks a lot!

ghedo (approved):

PR #743, opened 2020-11-23
lib: update max_datagram_size if smaller than peer's max_udp_payload_size

junhochoi:
When peer's max_udp_payload_size is smaller than the sender's
max_datagram_size, update max_datagram_size to match to the
peer's max_udp_payload_size for accurate congestion control.

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -474,6 +474,20 @@ impl Recovery {
         self.max_datagram_size
     }
 
+    pub fn update_max_datagram_size(&mut self, new_max_datagram_size: usize) {
+        let max_datagram_size =
+            cmp::min(self.max_datagram_size, new_max_datagram_size);
+
+        // Congestion Window is updated only when it's not updated already.
+        if self.congestion_window ==
+            self.max_datagram_size * INITIAL_WINDOW_PACKETS
+        {
+            self.congestion_window = max_datagram_size * INITIAL_WINDOW_PACKETS;
+        }
+
+        self.max_datagram_size = max_datagram_size;
+    }
+
     fn update_rtt(
         &mut self, latest_rtt: Duration, ack_delay: Duration, now: Instant,
     ) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2767,13 +2767,7 @@ impl Connection {
         if self.is_established() {
             // We cap the maximum packet size to 16KB or so, so that it can be
             // always encoded with a 2-byte varint.
-            cmp::min(
-                16383,
-                cmp::min(
-                    self.recovery.max_datagram_size(),
-                    self.peer_transport_params.max_udp_payload_size as usize,
-                ),
-            )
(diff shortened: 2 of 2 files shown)


src/recovery/mod.rs
```diff
@@ -458,6 +458,10 @@ impl Recovery {
         self.congestion_window.saturating_sub(self.bytes_in_flight)
     }
 
+    pub fn reset_initcwnd(&mut self) {
+        self.congestion_window = self.max_datagram_size * INITIAL_WINDOW_PACKETS;
```
ghedo, line 462: This should just be part of `update_max_datagram_size()` I think, I don't see much point in having a separate method right now.

ghedo (changes requested):
LGTM, but added a nit.

src/recovery/mod.rs
ghedo: Also, probably worth noting that resetting the cwnd like this means that e.g. the client's initial flight ends up not being counted towards cwnd, as the reset happens _after_ those packets are sent, so the client could later potentially send more data than what would be allowed by congestion control without the reset.

 In the future this will also apply to client's 0-RTT data as well I think.

ghedo: Actually, wait, I think I got confused here. Just changing `congestion_window` is probably fine, so nevermind.

ghedo: Hmmm, thinking some more about this. What happens when the client's Initial packet is lost, retransmitted then acked? It seems to me that if the server encodes its Initial packet with an ACK frame and then a CRYPTO frame:
1. The ACK frame is processed.
2. The original packet is declared lost due to the ACK for the successive packet (which was sent due to PTO).
3. The cwnd is reduced due to the loss.
4. The CRYPTO frame is processed.
5. The transport parameters from the server are parsed.
6. The cwnd is reset, which means that the reduction caused by the previous loss is reverted.

While this is probably not a huge problem, it seems unfortunate that we would have this side effect. Could we somehow propagate the cwnd decrease after the reset?

junhochoi: Thanks. Updated as follows:
- merge `reset_initcwnd()` into `update_max_datagram_size()`
- update cwnd only when it's not changed. I think this addresses your concern if cwnd is already changed by packet loss or other event during the handshake.

ghedo (approved):

src/recovery/mod.rs
ghedo: Yeah, or we could get the difference between `max_datagram_size * INITIAL_WINDOW_PACKETS` and the current cwnd, and then remove that from the new cwnd, but this is good too.

junhochoi: I think cwnd should be changed via CC, so this is a kind of exceptional case. In the future we may want to call CC initialize function instead, rather than modifying cwnd directly here, since we don't know if there is any CC internal state changes. But right now this is good enough.

PR #745, opened 2020-11-23
travis: remove configuration

ghedo:
The migration to GitHub Actions is complete. so we can remove this.

Also update the badge in the README. Technically GitHub Actions provides
its own badge, but the one from shields.io is more consistent with the
other badges we have.


junhochoi (approved):
🥇

PR #758, opened 2020-11-27
lib: document the error codes that are only generated internally

LPardue:


LPardue:
Closing as overtaken by events.

PR #763, opened 2020-11-30
RFC: h3: poll with closure API

ghedo:
This adds a `poll_with()` method that, instead of returning single
events like `poll()` does, passes each event to an application-provided
closure. Notably this means that all available events are passed to the
application, rather than just the first one like `poll()`.

This allows applications to potentially skip particular events while
still processing others, or stop processing before all events are
issued, rather than always getting the same event over and over until it
is processed.

The poll() method is also re-implemented using poll_with(). The stream
and DATAGRAM thresholds only apply to the poll() method, as an
application can more easily decide to skip particular events when using
poll_with(), so additional options to do that are not required.

For DATAGRAMs, due to the fact that each Datagram event needs to be
associated with a flow ID, we can't simply use the `dgram_recv_peek()`
method provided by the transport connection anymore, so the internal
dgram recv queue is accessed directly to ensure all required events are
issued (instead of just issuing the same event for the first DATAGRAM
over and over).

---

A somewhat more idiomatic alternative to using a closure would be to
instead have an event iterator type returned by a `poll_iter()` method
or something like that. A typical use-case would be something like this:

```rust
for ev in http3_conn.poll_iter(&mut conn) {
    match ev {
        (stream_id, quiche::h3::Event::Data) => {
            http3_conn.recv_body(stream_id, &mut buf)?;
        },

        ...
    }
}
```

However, due to the fact that polling for events requires mutable references to
both the transport and HTTP/3 connection objects, the above would require the
application to `collect()` the iterator, which could be expensive if many events
are generated. In any case, this could also be implementated on top of the poll
closure API.

It's not quite clear to me if there is a way to implement the above safely and without
requiring `collect()`. Will need to think about this some more, but if anyone has a
better idea, do tell.

Also, thinking about the whole DATAGRAM situation again, it's not quite clear
how useful it is to provide the flow ID to applications, as they will need to
read DATAGRAMs in order regardless of the flow ID, unlike stream events where
the application can read data for specific streams.

So an alternative could be to simply issue a single Datagram event if there is
at least one DATAGRAM queued, and use some kind of fake ID (e.g. could just
hardcode to "0" or something like that), which would simplify the logic to a
certain extent, as the actual DATAGRAM would not need to be touched at all.

The alternative to make the flow ID somewhat more useful I think would be to
make `recv_dgram()` only return DATAGRAMs that match a specifc ID provided by
the application. This could be used to e.g. avoid processing flows that the
application deems to be blocked, while still processing others, though it's
unclear in what cases this would be useful, so it might end up being a waste of
time.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1086,25 +1086,6 @@ impl Connection {
         }
     }
 
-    // A helper function for determining if there is a DATAGRAM event.
-    fn process_dgrams(
-        &mut self, conn: &mut super::Connection,
-    ) -> Result<(u64, Event)> {
-        let mut d = [0; 8];
-
-        match conn.dgram_recv_peek(&mut d, 8) {
-            Ok(_) => {
-                let mut b = octets::Octets::with_slice(&d);
-                let flow_id = b.get_varint()?;
-                Ok((flow_id, Event::Datagram))
-            },
-
-            Err(crate::Error::Done) => Err(Error::Done),
-
-            Err(e) => Err(Error::TransportError(e)),
-        }
-    }
-
     /// Reads request or response body data into the provided buffer.
     ///
     /// Applications should call this method whenever the [`poll()`] method
@@ -1146,9 +1127,9 @@ impl Connection {
     ///
     /// The event [`Datagram`] returns a flow ID.
     ///
-    /// The event [`GoAway`] returns an ID that depends on the connection role.
-    /// A client receives the largest processed stream ID. A server receives the
-    /// the largest permitted push ID.
+    /// The event [`GoAway`] returns an ID that depends on the connection's
+    /// role. A client receives the largest processed stream ID. A server
+    /// receives the the largest permitted push ID.
     ///
     /// If an error occurs while processing data, the connection is closed with
     /// the appropriate error code, using the transport's [`close()`] method.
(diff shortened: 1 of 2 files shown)


src/h3/mod.rs
```diff
@@ -1165,6 +1146,124 @@ impl Connection {
...
+                        h3.dgram_event_count += 1;
+                    } else if fallback_event.is_none() {
+                        fallback_event = Some((s, ev));
+                    }
+                },
+
+                _ => event = Some((s, ev)),
+            }
+
+            if event.is_some() {
```
LPardue, line 1194: is this supposed to be is_none()?

LPardue:
Thanks. I think this a step in the right direction but I'm not sure how the closure thing works across all potential cases. Some events like "Headers" are consuming - they eat the data when calling process_readable_stream(), so what happens if the closure doesn't handle them? I think this isn't too far off today's situatation with `poll()` if the caller just ignores the Headers event but the comment

```
    /// Events that are not immediately processed
    /// (e.g. by calling [`recv_body()`] / [`recv_dgram()`]) will be issued
    /// again in the following calls to `poll_with()` until they are processed.
```

is a little ambiguous.

If you have an example of an closure for an application that is mainly only interested in getting Data events from one stream ID,  that would help my understanding.

Regarding datagram flow ID. The main constraint/benefit was around varint generation and parsing. If we punt this to the application it means they need to do some work and we should probably expose Octets to help them do it right.

src/h3/mod.rs
ghedo: This is on purpose. The idea being that we can return early as soon as we got an event.

ghedo:
@LPardue The fact that some events are edge-triggered and some are level-triggered is indeed pretty confusing. I made https://github.com/cloudflare/quiche/pull/765 as an alternative to this, to make all event edge-triggered instead.

> Regarding datagram flow ID. The main constraint/benefit was around varint generation and parsing

My understanding is that `recv_dgram()` returns the full DATAGRAM to the application, including flow ID, so I thing this would be needed regardless, no?

LPardue:
> My understanding is that recv_dgram() returns the full DATAGRAM to the application, including flow ID, so I thing this would be needed regardless, no?

recv_dgram() follows the ethos of recv_body(). So it reads the entire transport DATAGRAM payload into the output buffer, finds the flow ID and returns it along with the varint-length of the flow ID. So an app needs to do something like

```
let mut buf = [0; 65535];
let (len, flow_id, flow_id_len) = h3_conn.recv_dgram(conn, buf).unwrap();

print!(
  "Received H3 DATAGRAM length={}, flow_id={}, payload={}",
  len,
  flow_id,
  buf[flow_id_len..len]
);
```

It is not the most user-friendly API...

ghedo:
I think it's safe to say we are not going in this direction, so closing.

PR #765, opened 2020-11-30
h3: make all events edge-triggered

ghedo:
Currently all HTTP/3 events are edge-triggered with the exception of
`Data` and `Datagram`, which get reported until the underlying data is
read by the application.

Besides being inconsistent, this is problematic because an application
can't decide to ignore those events in particular circumstances. For
example an application might want to apply backpressure to a particular
stream on which the Data event has been reported, but still process
other streams normally.

In the current implementation, a Data event for the "blocked" stream
would keep being fired until the application decides to actually read
the body data from that stream, which means that, if the application is
not ready to do that yet, no other events are reported, thus blocking
all other streams as well.

This changes those events to only be triggered once, until re-armed.

Data events for a particular stream are re-armed after all queued data
for that stream is read by the application (including data from multiple
DATA frames already buffered). A side-effect of this change is that now
data from consecutive DATA frames can be returned in the same
recv_body() call, while before an application had to call that method
once per every frame buffered.

Datagram events are re-armed once all queued DATAGRAMs for a specific
flow ID are read by the application (that is, if multiple DATAGRAMs are
queued next to each other, only one Datagram event for that flow ID is
fired), or when the incoming DATAGRAM queue is empty.

This also makes the poll count thresholds superfluous as each event is
only reported once rather than forever.

Note that applications will now need to make sure to read all of the
data from a stream, otherwise they might get stuck (see examples and
apps updates), so this is a breaking change.

---

This is in alternative to https://github.com/cloudflare/quiche/pull/763. From an API and usability perspective, this seems the better option, but it also breaks backwards compatibility (and also probably needs to be tested some more).

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -628,7 +628,7 @@ pub struct Connection {
     local_goaway_id: Option<u64>,
     peer_goaway_id: Option<u64>,
 
-    last_dgram_flow_id: u64,
+    dgram_event_triggered: bool,
 }
 
 impl Connection {
@@ -687,7 +687,7 @@ impl Connection {
             local_goaway_id: None,
             peer_goaway_id: None,
 
-            last_dgram_flow_id: std::u64::MAX,
+            dgram_event_triggered: false,
         })
     }
 
@@ -1092,22 +1092,19 @@ impl Connection {
 
         match conn.dgram_recv_peek(&mut d, 8) {
             Ok(_) => {
-                let mut b = octets::Octets::with_slice(&d);
-                let flow_id = b.get_varint()?;
-
-                if self.last_dgram_flow_id == flow_id {
+                if self.dgram_event_triggered {
                     return Err(Error::Done);
                 }
 
-                self.last_dgram_flow_id = flow_id;
+                self.dgram_event_triggered = true;
 
-                Ok((flow_id, Event::Datagram))
+                Ok((0, Event::Datagram))
             },
 
             Err(crate::Error::Done) => {
                 // The dgram recv queue is empty, so re-arm the Datagram event
                 // so it is issued next time a DATAGRAM is received.
-                self.last_dgram_flow_id = std::u64::MAX;
+                self.dgram_event_triggered = false;
 
                 Err(Error::Done)
             },
@@ -1200,7 +1197,8 @@ impl Connection {
     /// which is used in methods [`recv_body()`], [`send_response()`] or
     /// [`send_body()`].
     ///
-    /// The event [`Datagram`] returns a flow ID.
+    /// The event [`Datagram`] returns a dummy value of `0`, this should be
+    /// ignored by the application.
     ///
     /// The event [`GoAway`] returns an ID that depends on the connection role.
     /// A client receives the largest processed stream ID. A server receives the
@@ -4033,7 +4031,7 @@ mod tests {
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_0_result));
         assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_0_result));
-        assert_eq!(s.poll_server(), Ok((2, Event::Datagram)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_2_result));
         assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_2_result));
@@ -4092,7 +4090,7 @@ mod tests {
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(flow_0_result));
         assert_eq!(s.poll_client(), Err(Error::Done));
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(flow_0_result));
-        assert_eq!(s.poll_client(), Ok((2, Event::Datagram)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(flow_2_result));
         assert_eq!(s.poll_client(), Err(Error::Done));
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(flow_2_result));
@@ -4339,7 +4337,7 @@ mod tests {
         assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_0_result));
 
-        assert_eq!(s.poll_server(), Ok((2, Event::Datagram)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_2_result));
 
         assert_eq!(s.poll_server(), Err(Error::Done));
@@ -4354,12 +4352,9 @@ mod tests {
         assert_eq!(s.poll_server(), Err(Error::Done));
 
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_0_result));
-
-        assert_eq!(s.poll_server(), Ok((2, Event::Datagram)));
         assert_eq!(s.poll_server(), Err(Error::Done));
 
         assert_eq!(s.recv_dgram_server(&mut buf), Ok(flow_2_result));
-
         assert_eq!(s.poll_server(), Err(Error::Done));
 
         assert_eq!(s.recv_body_server(stream, &mut recv_buf), Ok(body.len()));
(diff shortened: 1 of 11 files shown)


src/h3/mod.rs
```diff
@@ -3554,23 +3498,28 @@ mod tests {
...
 
         s.send_dgram_server(0).unwrap();
         s.send_dgram_server(0).unwrap();
         s.send_dgram_server(0).unwrap();
 
         assert_eq!(s.poll_client(), Ok((0, Event::Datagram)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(result));
-        assert_eq!(s.poll_client(), Ok((0, Event::Datagram)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
```
LPardue, line 3516:
```suggestion
        assert_eq!(s.poll_client(), Err(Error::Done));
```

change the error but not the endpoint :)

src/h3/mod.rs
```diff
@@ -3554,23 +3498,28 @@ mod tests {
...
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(result));
-        assert_eq!(s.poll_client(), Ok((0, Event::Datagram)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
```
LPardue, line 3518:
```suggestion
        assert_eq!(s.poll_client(), Err(Error::Done));
```

src/h3/mod.rs
```diff
@@ -3554,23 +3498,28 @@ mod tests {
...
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(result));
         assert_eq!(s.poll_client(), Err(Error::Done));
+        assert_eq!(s.recv_dgram_client(&mut buf), Err(Error::Done));
+        assert_eq!(s.poll_server(), Err(Error::Done));
```
LPardue, line 3522:
```suggestion
        assert_eq!(s.poll_client(), Err(Error::Done));
```

src/h3/mod.rs
```diff
@@ -3762,21 +3673,16 @@ mod tests {
...
-        assert_eq!(s.poll_client(), Ok((stream, Event::Data)));
-        assert_eq!(s.poll_client(), Ok((stream, Event::Data)));
 
-        assert_eq!(s.poll_client(), Ok((0, Event::Datagram)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+
         assert_eq!(s.recv_dgram_client(&mut buf), Ok(result));
 
-        assert_eq!(s.poll_client(), Ok((stream, Event::Data)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
```
LPardue, line 3684:
```suggestion
        assert_eq!(s.poll_client(), Err(Error::Done));
```

LPardue (changes requested):

ghedo:
A potentially simpler alternative implementation that of course I only thought of after implementing this whole thing, could be to also make `Connection::readable()` edge-triggered. That is, only report streams that are newly readable, rather than all readable streams. Will look into that another time though.

src/h3/mod.rs
ghedo: Oooops, lol. Let's see if these tests still pass...

ghedo:
I updated the NGINX patch to work with the new behavior. There is still quite a bit of room for improvement in how NGINX handles HTTP/3 request bodies, but that is a separate issue.

I also changed how `recv_body()` works, to make it return data from consecutive DATA frames buffered in the stream, rather than make the application call it once for each frame if the data didn't fill the output buffer. This is slightly more efficient, though doesn't make a huge difference in performance, but the new behavior is more consistent to how `stream_recv()` (or even `recv()`) works, so that an application could in theory check whether the buffer was completely filled or not, to know if it needs to keep calling it, rather than always wait for `Done` (though that still works).

ghedo:
Rebased, and updated the NGINX patch to improve the overall handling of HTTP/3 request bodies, which simplifies the overall implementation and should work better (e.g. now request buffering is also supported).

ehaydenr:
I found what I believe to be a couple bugs in this PR

The first issue I noticed is `quiche::h3::Connection::recv_dgram` calls `dgram_recv`, but _doesn’t_ re-arm the trigger. When the application calls poll and an event is triggered, the trigger will be disarmed. If we then recv all the datagrams available _and receive more on the connection prior to calling poll_, the trigger won’t have been re-armed and the no event will be fired.
The second issue I noticed very similar to the first, but for streams. When a `stream::State::QpackInstruction` is encountered and `conn.stream_recv` is called, it’s possible that the `Done` is returned and the trigger needs to be re-armed.
Fixing these two seems to have mostly fixed things for me, but I’m still seeing some timeouts for loss detection during body streaming which makes me suspicious I may not have caught them all.

junhochoi (dismissed):

LPardue:
> Datagram events are re-armed once all queued DATAGRAMs for a specific
flow ID are read by the application (that is, if multiple DATAGRAMs are
queued next to each other, only one Datagram event for that flow ID is
fired), or when the incoming DATAGRAM queue is empty.

I don't quite follow the intent here, which is probably related to the issues @ehaydenr has described.

The internal datagram queue is a single entity, so there is no way to fire off different events for different flow IDs and have the flow IDs read independently (causing the triggers to reset independently). I read the PR's intent like so:

1. Client sends datagrams on flows in the sequence 0,0,0,2,2,2,2.
1. Server gets all of these and populates the queue in the same order
1. Server calls `poll()`, gets `Ok((0, Event::Datagram))`
1. Server calls `poll()`, gets `Err(Error::Done))`
1. Server calls `recv_dgram()` gets `(11, 0, 1)`
1. Server calls `recv_dgram()` gets `(11, 0, 1)`
1. Server calls `recv_dgram()` gets `(11, 0, 1)`
1. Server calls `recv_dgram()` `Err(Error::Done))`
1. Server calls `poll()`, gets `Ok((2, Event::Datagram))`
1. Server calls `poll()`, gets `Err(Error::Done))`
1. Server calls `recv_dgram()` gets `(11, 2, 1)`
1. Server calls `recv_dgram()` gets `(11, 2, 1)`
1. Server calls `recv_dgram()` gets `(11, 2, 1)`
1. Server calls `recv_dgram()` `Err(Error::Done))`

but in practice what happens is that the server is able to read all datagrams independent of any trigger.

There's a way to fix this, in `recv_dgram` add some code at the top to similarly check if the next dgram-in-queue matches the trigger flow ID

```
        match conn.dgram_recv_peek(buf, 8) {
            Ok(_) => {
                let mut b = octets::Octets::with_slice(&buf);
                let flow_id = b.get_varint()?;

                if self.last_dgram_flow_id != flow_id {
                    self.last_dgram_flow_id = std::u64::MAX;
                    return Err(Error::Done);
                }
            },

            Err(crate::Error::Done) => {
                self.last_dgram_flow_id = std::u64::MAX;
                return Err(Error::Done);
            },

            Err(e) => return Err(Error::TransportError(e)),
        }
```

That seems to work fine in the simple case but I think it has the potential to cause problems in actual usage due to dgram's unreliable nature and lack of flow control. We use the queue to accommodate bursts and it might happen that `poll()` returns a trigger event for datagrams with one flow ID, but that after processing some packets. the old dgram get dropped off the queue in favour of new dgrams with a diferent flow ID. 

Given that there is a single datagram queue, I wonder if its simpler to remove flow ID from the trigger calculations. I.e. just say "there are datagrams to read, until there are not. When that happens the event is rearmed."

ghedo:
>  I read the PR's intent like so:

No, the intention was always to still allow applications to read as many datagrams they wanted, just like they can read from any stream (e.g. body) even if no event was returned first by poll().

> I wonder if its simpler to remove flow ID from the trigger calculations

Yeah, that's a possibility too. I just tried to somewhat keep backwards compatibility for the value returned by poll().

I guess if we do this the "ID" return value for datagram events would be a dummy value?

ghedo:
I implemented the above (i.e. make Datagram event trigger not flow id specific) as an additional commit. Will probably need to add some more unit tests though, but should be enough for some initial testing.

ghedo:
Found a small bug in the NGINX integration, but haven't seen any other problem in quiche itself.

@ehaydenr would you be able to try the latest changes, in particular the `h3: make Datagram event's trigger not flow ID specific` commit? That might help with the Datagram events problem you were seeing.

As for the `QpackInstruction` one, I'm not sure I follow, as Data events shouldn't be triggered on QPACK streams, and the event is re-armed per-stream, so the QPACK stream shouldn't affect other streams anyway I think (we currently don't really do anything with QPACK streams other than just reading the data and discarding it).

LPardue (approved):

ehaydenr:
@ghedo I've tested it prior to the latest rebases and so far it's all looked great. I haven't seen the QpackInstruction issue since one of your rebases a couple weeks ago. Thank you!

ghedo:
Great, thanks for testing @ehaydenr!

PR #770, opened 2020-12-04
Pacing: Pacing of egress QUIC packets

lohith-bellad:
Currently pacing is enabled for Cubic and Reno, since BBR is under development.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -273,8 +273,12 @@ ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len,
                          const quiche_recv_info *info);
 
 typedef struct {
+    // The address the packet should be sent to.
     struct sockaddr_storage to;
     socklen_t to_len;
+
+    // The time to send the packet out.
+    struct timespec at;
 } quiche_send_info;
 
 // Writes a single QUIC packet to be sent to the peer.
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -40,6 +40,7 @@ use libc::c_void;
 use libc::size_t;
 use libc::sockaddr;
 use libc::ssize_t;
+use libc::timespec;
 
 #[cfg(not(windows))]
 use libc::sockaddr_in;
@@ -634,6 +635,8 @@ pub extern fn quiche_conn_recv(
 pub struct SendInfo {
     to: sockaddr_storage,
     to_len: socklen_t,
+
+    at: timespec,
 }
 
 #[no_mangle]
@@ -649,6 +652,8 @@ pub extern fn quiche_conn_send(
     match conn.send(out) {
         Ok((v, info)) => {
             out_info.to_len = std_addr_to_c(&info.to, &mut out_info.to);
+
+            std_time_to_c(&info.at, &mut out_info.at);
 
             v as ssize_t
         },
@@ -1091,4 +1096,18 @@ fn std_addr_to_c(addr: &SocketAddr, out: &mut sockaddr_storage) -> socklen_t {
             },
         }
     }
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "windows")))]
+fn std_time_to_c(time: &std::time::Instant, out: &mut timespec) {
+    unsafe {
+        ptr::copy_nonoverlapping(time as *const _ as *const timespec, out, 1)
+    }
+}
+
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "windows"))]
+fn std_time_to_c(_time: &std::time::Instant, out: &mut timespec) {
+    // TODO: implement Instant conversion for systems that don't use timespec.
+    out.tv_sec = 0;
+    out.tv_nsec = 0;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -490,6 +490,9 @@ pub struct RecvInfo {
 pub struct SendInfo {
     /// The address the packet should be sent to.
     pub to: SocketAddr,
+
+    /// The time to send the packet out.
+    pub at: time::Instant,
 }
 
 /// Represents information carried by `CONNECTION_CLOSE` frames.
@@ -2390,7 +2393,14 @@ impl Connection {
             done += pad_len;
         }
 
-        let info = SendInfo { to: self.peer_addr };
+        let info = SendInfo {
+            to: self.peer_addr,
+
+            at: self
+                .recovery
+                .get_packet_send_time()
+                .unwrap_or_else(time::Instant::now),
+        };
 
         Ok((done, info))
     }
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -121,6 +121,8 @@ pub struct Recovery {
 
     bytes_acked_ca: usize,
 
+    bytes_sent: usize,
+
     congestion_recovery_start_time: Option<Instant>,
 
     max_datagram_size: usize,
@@ -129,6 +131,11 @@ pub struct Recovery {
 
     // HyStart++.
     hystart: hystart::Hystart,
+
+    // Pacing.
+    pacing_rate: u64,
+
+    last_packet_scheduled_time: Option<Instant>,
 }
 
 impl Recovery {
@@ -185,6 +192,8 @@ impl Recovery {
 
             bytes_acked_ca: 0,
 
+            bytes_sent: 0,
+
             congestion_recovery_start_time: None,
 
             max_datagram_size: config.max_send_udp_payload_size,
@@ -198,6 +207,10 @@ impl Recovery {
             app_limited: false,
 
             hystart: hystart::Hystart::new(config.hystart),
+
+            pacing_rate: 0,
+
+            last_packet_scheduled_time: None,
         }
     }
 
@@ -240,11 +253,62 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if CC doesn't do its own.
+        if !(self.cc_ops.has_custom_pacing)() {
+            if let Some(srtt) = self.smoothed_rtt {
+                let rate = (self.congestion_window as u64 * 1000000) /
+                    srtt.as_micros() as u64;
+                self.set_pacing_rate(rate);
+            }
+        }
+
+        self.schedule_next_packet(epoch, now, sent_bytes);
+
+        self.bytes_sent += sent_bytes;
         trace!("{} {:?}", trace_id, self);
     }
 
     fn on_packet_sent_cc(&mut self, sent_bytes: usize, now: Instant) {
         (self.cc_ops.on_packet_sent)(self, sent_bytes, now);
+    }
+
+    pub fn set_pacing_rate(&mut self, rate: u64) {
+        if rate != 0 {
+            self.pacing_rate = rate;
+        }
+    }
+
+    pub fn get_packet_send_time(&self) -> Option<Instant> {
+        self.last_packet_scheduled_time
+    }
+
+    fn schedule_next_packet(
+        &mut self, epoch: packet::Epoch, now: Instant, packet_size: usize,
+    ) {
+        // Don't pace in any of these cases:
+        //   * Packet epoch is not EPOCH_APPLICATION.
+        //   * Packet contains only ACK frames.
+        //   * The start of the connection.
+        if epoch != packet::EPOCH_APPLICATION ||
+            packet_size == 0 ||
+            self.bytes_sent <= self.congestion_window ||
+            self.pacing_rate == 0
+        {
+            self.last_packet_scheduled_time = Some(now);
+            return;
+        }
+
+        self.last_packet_scheduled_time = match self.last_packet_scheduled_time {
+            Some(last_scheduled_time) => {
+                let interval: u64 =
+                    (packet_size as u64 * 1000000) / self.pacing_rate;
+                let interval = Duration::from_micros(interval);
+                let next_schedule_time = last_scheduled_time + interval;
+                Some(cmp::max(now, next_schedule_time))
+            },
+
+            None => Some(now),
+        };
     }
 
     pub fn on_ack_received(
@@ -847,6 +911,8 @@ pub struct CongestionControlOps {
     pub checkpoint: fn(r: &mut Recovery),
 
     pub rollback: fn(r: &mut Recovery),
+
+    pub has_custom_pacing: fn() -> bool,
 }
 
 impl From<CongestionControlAlgorithm> for &'static CongestionControlOps {
@@ -893,6 +959,12 @@ impl std::fmt::Debug for Recovery {
             self.congestion_recovery_start_time
         )?;
         write!(f, "{:?} ", self.delivery_rate)?;
+        write!(f, "pacing_rate={:?}", self.pacing_rate)?;
+        write!(
+            f,
+            "last_packet_scheduled_time={:?}",
+            self.last_packet_scheduled_time
+        )?;
 
         if self.hystart.enabled() {
             write!(f, "hystart={:?} ", self.hystart)?;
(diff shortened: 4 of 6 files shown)


src/recovery/mod.rs
```diff
@@ -233,13 +254,71 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if BBR is not used
+        if self.pacing_enabled &&
+            self.cc_algo == CongestionControlAlgorithm::CUBIC
```
junhochoi, line 259: Is it cubic specific? I think it can be used with Reno as well.

src/ffi.rs
```diff
@@ -38,6 +38,9 @@ use libc::c_void;
 use libc::size_t;
 use libc::ssize_t;
 
+#[cfg(unix)]
```
junhochoi, line 41: I think this is ok to be specific to linux? (I don't know other unix-like OS has similar support of SO_TXTIME)

junhochoi:
Please add some test for pacing.

src/ffi.rs
lohith-bellad: Yes, I think it should be good for now.

src/recovery/mod.rs
lohith-bellad: Added Reno also.

lohith-bellad:
Added some tests for pacing.

src/recovery/mod.rs
```diff
@@ -233,13 +254,73 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if BBR is not used
+        if self.pacing_enabled &&
+            (self.cc_algo == CongestionControlAlgorithm::CUBIC ||
+                self.cc_algo == CongestionControlAlgorithm::Reno)
```
junhochoi, line 260: Sorry I need to clarify a little more. I think we simply remove checking cc algorithm names here (and then you can remove cc_algo too). Is there any issue with it? Currently we only have Reno and Cubic, but considering in the future we may add more, I think it's good to make pacing simply an option.

src/lib.rs
```diff
@@ -2831,6 +2841,72 @@ impl Connection {
...
+    /// # Ok::<(), quiche::Error>(())
+    /// ```
+    pub fn send_at(
+        &mut self, out: &mut [u8], send_time: &mut time::Instant,
+    ) -> Result<usize> {
+        let written = self.send(out);
+        if written.is_ok() {
+            *send_time = self.recovery.get_packet_send_time().unwrap();
+        }
+        written
```
LPardue, line 2907: I think it would be cleaner to have this return a `Result<(usize, time::Instant)>` so then you'd avoid having to create a mutable Instant to pass in here. e.g.

```
pub fn send_at(&mut self, out: &mut [u8]) -> Result<(usize, time::Instant)> {
        match self.send(out) {
            Ok(v) => Ok((
                v,
                self.recovery
                    .get_packet_send_time()
                    .unwrap_or(time::Instant::now()),
            )),

            Err(e) => Err(e),
        }
    }
```

src/recovery/mod.rs
LPardue: AFAICT `set_pacing_rate()` is only called in this module, so I think it would be simpler to put all this code inside that method and drop the parameter.

lohith-bellad: set_pacing_rate() will be called by congestion control algos like BBR to set the pacing rate. Since BBR is not yet implemented, and current implemented congestion control algos don't set it, its used only here.

I am adding Cubic and Reno check here because, BBR is not there yet.
Logic is, if cc_algo is BBR then it will set the pacing rate we dont need to set it here. If other algos are used we need to set the pacing rate here.

When BBR is implemented we need to add !BBR check here.

src/recovery/mod.rs
```diff
@@ -233,13 +254,73 @@ impl Recovery {
...
+        {
+            let rate = match self.smoothed_rtt {
+                Some(srtt) =>
+                    ((self.congestion_window as u128 * 1000000) /
+                        srtt.as_micros()) as usize,
+                None => 0,
+            };
+            self.set_pacing_rate(rate);
+        }
```
LPardue, line 269: If you want to keep this block here, then since `set_pacing_rate()` ignores 0 values you can probably rewrite the internals to something like 

```
if let Some(srtt) = self.smoothed_rtt {
 let rate = ((self.congestion_window as u128 * 1000000) /
                 srtt.as_micros()) as usize;
 self.set_pacing_rate(rate);
}
```

lohith-bellad: done!

src/lib.rs
lohith-bellad: done, thanks for pointing out!

src/recovery/mod.rs
junhochoi: I understand your concern but I still don't like the idea of hard coding CC names. Can we move this to each CC module (e.g. Cubic module calls set_pacing_rate() when pacing_enabled)?

src/lib.rs
```diff
@@ -825,6 +828,13 @@ impl Config {
         self.dgram_recv_max_queue_len = recv_queue_len;
         self.dgram_send_max_queue_len = send_queue_len;
     }
+
+    /// Configures whether to enable Packet Pacing.
+    ///
+    /// The default is `true`.
+    pub fn enable_pacing(&mut self, v: bool) {
```
ghedo, line 835: Why do we need a configuration for this? The application can decide to ignore the timestamp value (or just not call `send_at()` at all) even if this is enabled by default.

src/recovery/mod.rs
ghedo: I agree that we should avoid hardcoding names here. An alternative would be to add a new function to `CongestionControlOps` like `use_generic_pacing()`, which will return `true` in the case of CUBIC and Reno, and `false` for BBR.

But generally, I also don't think this is needed at all until we actually implement BBR, so we can just remove the whole thing IMO. The comment is also confusing, as we don't even implement BBR right now.

ghedo (changes requested):

src/ffi.rs
ghedo: This will break on macOS, as `std::time::Instant` doesn't use `struct timespec`, so it needs to be excluded I think, see https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/time.rs#L274

src/ffi.rs
```diff
@@ -841,3 +871,36 @@ pub extern fn quiche_conn_dgram_purge_outgoing(
...
+        fn write_u64(&mut self, i: u64) {
+            self.0.push(i as _);
+        }
+
+        fn write_u32(&mut self, i: u32) {
+            self.0.push(i as _);
+        }
+    }
+    let mut hasher = ToTimeSpec(vec![]);
+    instant.hash(&mut hasher);
```
ghedo, line 901: TBH, I kind of prefer the `std::ptr::copy_nonoverlapping()` alternative to this, as it is clearer what it does, and is also less code.

src/recovery/mod.rs
```diff
@@ -233,13 +254,71 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if BBR is not used
+        if self.pacing_enabled &&
+            (self.cc_algo == CongestionControlAlgorithm::CUBIC ||
+                self.cc_algo == CongestionControlAlgorithm::Reno)
+        {
+            if let Some(srtt) = self.smoothed_rtt {
+                let rate = ((self.congestion_window as u128 * 1000000) /
```
ghedo, line 263: Why use `u128` and microseconds?

ghedo: (alternative would be to use https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64 and truncate the result, but I see why `u128` is used here).

src/recovery/mod.rs
```diff
@@ -233,13 +254,71 @@ impl Recovery {
...
+        let interval: u64 =
+            (packet_size as u64 * 1000000) / self.pacing_rate as u64;
+        let next_send_interval = Duration::from_micros(interval);
+
+        let next_schedule_time = match self.last_packet_scheduled_time {
+            Some(last_scheduled_time) => last_scheduled_time + next_send_interval,
+            None => now,
+        };
+
+        self.last_packet_scheduled_time = if next_schedule_time <= now {
```
ghedo, line 315: I think you can just do:
```rust
self.last_packet_scheduled_time = cmp::max(now, next_schedule_time);
```

src/recovery/mod.rs
ghedo: As opposed to truncate micros to `u64` as it will probably be enough, or use https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64.

src/recovery/mod.rs
```diff
@@ -124,6 +128,13 @@ pub struct Recovery {
 
     // HyStart++.
     hystart: hystart::Hystart,
+
+    // Pacing
+    pacing_enabled: bool,
+
+    pacing_rate: usize,
```
ghedo, line 135: Why `usize`? It seems like `u64` would be more appropriate.

src/lib.rs
```diff
@@ -2831,6 +2841,76 @@ impl Connection {
...
+    ///             // An error occurred, handle it.
+    ///             break;
+    ///         },
+    ///     };
+    ///
+    ///     socket.send(&out[..write]).unwrap();
+    /// }
+    /// # Ok::<(), quiche::Error>(())
+    /// ```
+    pub fn send_at(&mut self, out: &mut [u8]) -> Result<(usize, time::Instant)> {
```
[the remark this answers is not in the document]
lohith-bellad: how will C see this struct SendInfo as, it does not know time::Instant ?

src/recovery/mod.rs
lohith-bellad: Thought about each CC algo sending flag to set pacing_rate external or not, but each new algo has to add code to send the flag even though it does not anything about pacing.

But anyways above idea of each CC algo sending flag is to set pacing rate or not is cleaner and comes handy for BBR too.

src/lib.rs
lohith-bellad: Sure, this is not needed clearly.

src/ffi.rs
lohith-bellad: thought hash,hasher method was safe and predictable but anyways, will go with copy_nonoverlapping as it is clearer.

src/recovery/mod.rs
lohith-bellad: I was being too cautious I believe, clearly u64 should be fine, thanks.

src/recovery/mod.rs
lohith-bellad: sure, much cleaner.

src/recovery/mod.rs
lohith-bellad: will take care

junhochoi: what's the unit of pacing_rate?

lohith-bellad: bytes/sec

src/lib.rs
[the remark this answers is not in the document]
lohith-bellad: Cool, makes sense if we plan in the future like this.

src/ffi.rs
lohith-bellad: this is to include libc::timespec struct. I added unix check to differentiate between unix family and windows family. Windows does not have timespec
https://doc.rust-lang.org/reference/conditional-compilation.html#target_family

include/quiche.h
```diff
@@ -257,6 +257,16 @@ ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len);
 // Writes a single QUIC packet to be sent to the peer.
 ssize_t quiche_conn_send(quiche_conn *conn, uint8_t *out, size_t out_len);
 
+typedef struct {
+    // Time to send the packet out.
+    struct timespec send_time;
+} quiche_send_info;
```
junhochoi, line 263: Looks like `typedef struct SendInfo quiche_send_info` is good enough

junhochoi (dismissed):

include/quiche.h
lohith-bellad: hmm, ok, but send_time is of different type.

lohith-bellad: Without defining struct quiche_send_info, nginx build with quiche fails, because of unknown type.
`error: field ‘packet_send_info’ has incomplete type`. So this definition is needed here.

junhochoi: Oh I see. I thought it's an opaque. LGTM

src/ffi.rs
junhochoi: I think we still need to handle MacOS/iOS case as @ghedo mentioned above.

src/ffi.rs
```diff
@@ -543,6 +546,40 @@ pub extern fn quiche_conn_send(
...
+        panic!("The provided buffer is too large");
+    }
+
+    let out = unsafe { slice::from_raw_parts_mut(out, out_len) };
+
+    match conn.send_with_info(out) {
+        Ok((v, info)) => {
+            unsafe {
+                ptr::copy_nonoverlapping(
+                    &info.send_time as *const _ as *const timespec,
```
junhochoi, line 571: Related to the comment above on `#[cfg(unix)]` -- this should be done only for the platform which timespec is used.
Also this is more like a question, how do we make sure we can safely convert `Instant` to `timespec`? Is `cfg()` check is enough?

lohith-bellad: yes, it should be done where timespec is valid. Changed to include in Linux platforms only.
There is no safe/proper way. With 'linux only' check this should be ok I believe.

src/ffi.rs
lohith-bellad: How about adding 'linux only' check here

junhochoi: sounds good to me.

junhochoi (approved):

lohith-bellad:
@ghedo @LPardue Can you guys review!

lohith-bellad:
@ghedo @LPardue can you guys review this!

src/recovery/mod.rs
```diff
@@ -236,13 +249,62 @@ impl Recovery {
...
+        self.bytes_sent += sent_bytes;
         trace!("{} {:?}", trace_id, self);
     }
 
     fn on_packet_sent_cc(&mut self, sent_bytes: usize, now: Instant) {
         (self.cc_ops.on_packet_sent)(self, sent_bytes, now);
     }
 
+    pub fn set_pacing_rate(&mut self, rate: u64) {
+        if rate > 0 {
```
LPardue, line 272:
```suggestion
        if rate != 0 {
```

since this is a u64 we only care about that one special 0 value

src/recovery/mod.rs
```diff
@@ -236,13 +249,62 @@ impl Recovery {
...
+        self.last_packet_scheduled_time
+    }
+
+    fn schedule_next_packet(
+        &mut self, epoch: packet::Epoch, now: Instant, packet_size: usize,
+    ) {
+        // Pacing is not done for following cases,
+        // 1. Packet epoch is not EPOCH_APPLICATION.
+        // 2. If packet has only ACK frames.
+        // 3. Start of the connection.
```
LPardue, line 287:
```suggestion
        // Don't pace in any of these cases:
        //   * Packet epoch is not EPOCH_APPLICATION.
        //   * Packet contains only ACK frames.
        //   * The start of the connection.
```

src/recovery/mod.rs
```diff
@@ -236,13 +249,62 @@ impl Recovery {
...
+            return;
+        }
+
+        let interval: u64 = (packet_size as u64 * 1000000) / self.pacing_rate;
+        let next_send_interval = Duration::from_micros(interval);
+
+        let next_schedule_time = match self.last_packet_scheduled_time {
+            Some(last_scheduled_time) => last_scheduled_time + next_send_interval,
+            None => now,
+        };
```
LPardue, line 303: Since you only use the interval in the `Some` case you could probably rewrite this as 

```
 self.last_packet_scheduled_time = match self.last_packet_scheduled_time {
    Some(last_scheduled_time) => {
        let interval: u64 = (packet_size as u64 * 1000000) / self.pacing_rate;
        let interval = Duration::from_micros(interval);
        let next_schedule_time = last_scheduled_time + interval;
        Some(cmp::max(now, next_schedule_time))
    }.

    None => Some(now),
 };
```

src/lib.rs
```diff
@@ -1986,10 +1986,12 @@ impl Connection {
         Ok(read)
     }
 
-    /// Writes a single QUIC packet to be sent to the peer.
+    /// Writes a single QUIC packet to be sent to the peer along with send_info
+    /// like to time to send the packet out.
```
LPardue, line 1990: This will be available in the public crate docs. I think generally it would help to expand the documentation here a bit. Can you link to the `SendInfo` object rather referring to it as `send_info`?

src/lib.rs
```diff
@@ -1986,10 +1986,12 @@ impl Connection {
...
     ///
     /// On success the number of bytes written to the output buffer is
-    /// returned, or [`Done`] if there was nothing to write.
+    /// returned along with send_info, or [`Done`] if there was nothing to
```
LPardue, line 1993: same comment here about `send_info`

include/quiche.h
```diff
@@ -257,6 +257,16 @@ ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len);
...
+
+// Writes a single QUIC packet to be sent to the peer and fills in the send_info
+// struct.
```
LPardue, line 266: Noted in a different comment but I'll repeat here. It's not clear if this feature is only available on Linux etc, we should state that. There's already precedent 

```
// Enables qlog to the specified file descriptor. Unix only.
void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char *log_title,
                             const char *log_desc);
```

LPardue (changes requested):
I made some quality of life suggestions. Please also rebase.

src/recovery/mod.rs
lohith-bellad: done

src/lib.rs
lohith-bellad: made the changes.

src/lib.rs
```diff
@@ -2041,8 +2044,11 @@ impl Connection {
     /// # let scid = quiche::ConnectionId::from_ref(&[0xba; 16]);
     /// # let mut conn = quiche::accept(&scid, None, &mut config)?;
     /// loop {
-    ///     let write = match conn.send(&mut out) {
-    ///         Ok(v) => v,
+    ///     let write = match conn.send_with_info(&mut out) {
+    ///         Ok((v, send_info)) => {
+    ///             // Use send_info
+    ///             v
+    ///         },
```
LPardue, line 2051:
```suggestion
    ///     let (written, send_info) = match conn.send_with_info(&mut out) {
    ///         Ok(v) => {
    ///             v
    ///         },
```

As far as examples go, it would probably be clearer here to just return the 2-tuple here and then later, before the socket send, give an example of using send_info. Which actually identifies to me a bit of confusion - what exactly am I supposed to do with SendInfo.send_time? Maybe we should update the examples?

src/recovery/mod.rs
```diff
@@ -129,6 +131,11 @@ pub struct Recovery {
 
     // HyStart++.
     hystart: hystart::Hystart,
+
+    // Pacing
```
LPardue, line 135:
```suggestion
    // Pacing.
```

src/recovery/mod.rs
```diff
@@ -240,13 +253,64 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if BBR is not used
```
LPardue, line 256:
```suggestion
        // Pacing: Set the pacing rate if BBR is not used.
```

rafibaum:
I've been using this branch since March and it's been working excellently for me. The only comment I have would be that paced congestion controllers like BBR may have "send quantum" as an output control signal and future work may want to allow congestion controllers to set a quantum value as well, which would cap the size of generated QUIC datagrams at that size.

include/quiche.h
```diff
@@ -258,6 +258,16 @@ ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len);
...
+typedef struct {
+    // Time to send the packet out.
+    struct timespec send_time;
+} quiche_send_info;
+
+// Writes a single QUIC packet to be sent to the peer and fills in the send_info
+// struct. Linux only.
+ssize_t quiche_conn_send_with_info(quiche_conn *conn, uint8_t *out, size_t out_len,
+                                   quiche_send_info *out_info);
+
```
junhochoi, line 270: I think we also enclose this block (`quiche_send_info` and `quiche_conn_send_with_info()`) with `#ifdef linux` since those are linux-only api for now.

src/recovery/mod.rs
```diff
@@ -240,13 +253,64 @@ impl Recovery {
             self.hystart.start_round(pkt_num);
         }
 
+        // Pacing: Set the pacing rate if BBR is not used.
```
junhochoi, line 256:
```suggestion
        // Pacing: Set the pacing rate if CC doesn't do its own.
```

junhochoi (dismissed):

ghedo (approved):

ghedo:
FTR, I renamed the `SendInfo` field to `at`, to be somewhat more consistent. Also, the conversion from `Instant` to `timespec` in FFI is now only implemented when OS != macOS, iOS, Windows, in the other cases the field is just set to 0. In the future we can implement more OS-specific conversion functions.

PR #775, opened 2020-12-07
lib: fix DATAGRAM fram size calculation in  dgram_max_writable_len()

LPardue:
It seems we forgot to account for the 1-byte of frame type in dgram_max_writable_len(). Since this went undetected, I've added an explicit assertion of the return value in an existing test. The actual packet overhead varies due to variable DCID length, but the maximum overhead is 44 bytes with this change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3480,7 +3480,8 @@ impl Connection {
                 // ...clamp to what peer can support...
                 max_len = cmp::min(peer_frame_len as usize, max_len);
                 // ...subtract frame overhead, checked for underflow.
-                max_len.checked_sub(frame::MAX_DGRAM_OVERHEAD)
+                // (1 byte of frame type + len of length )
+                max_len.checked_sub(1 + frame::MAX_DGRAM_OVERHEAD)
             },
         }
     }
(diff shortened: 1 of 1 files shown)


src/frame.rs
```diff
@@ -723,7 +723,7 @@ impl Frame {
 
             Frame::Datagram { data } => {
                 1 + // frame type
-                octets::varint_len(data.len() as u64) + // length
+                2 + // length, always encode as 2-byte varint
```
ghedo, line 726: I think this was correct before, as the length varint is not fixed length:
https://github.com/cloudflare/quiche/blob/master/src/frame.rs#L1148

Compare to e.g. https://github.com/cloudflare/quiche/blob/master/src/frame.rs#L1148

ghedo (changes requested):

src/frame.rs
ghedo: (fixed the first link as I pasted the same thing twice...)

LPardue: yeah I  think you're correct

ghedo (approved):

PR #778, opened 2020-12-07
h3: add method to expose if H3 DATAGRAMS were enabled

LPardue:
We previously landed support for the H3_DATAGRAM setting. This
change adds a method that applications can use to determine if
the peer has sent the SETTING and corresponding Transport Parameter.

Since SETTINGS are processed with poll(), this method will always
return false until a peer has sent a SETTINGS frame with
H3_DATAGRAM=1, and the application has called poll() after the
frame has arrived.


ghedo (changes requested):
Looks good other than a tiny nit.

ghedo (approved):

PR #783, opened 2020-12-11
Correctly handle write_epoch() when an application close was done and…

normanmaurer:
… no streams are left to process

Motivation:

If all streams are closed and an application close is done we need ensure we still return an non error result from write_epoch as otherwise we will never send the application close to the remote peer

Modifications:

Change if branch to also respect app_error and so application close

Result:

Correctly send APPLICATION_CLOSE in all cases

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3779,7 +3779,7 @@ impl Connection {
     fn write_epoch(&self) -> Result<packet::Epoch> {
         // On error send packet in the latest epoch available, but only send
         // 1-RTT ones when the handshake is completed.
-        if self.error.is_some() {
+        if self.error.is_some() || self.app_error.is_some() {
             let epoch = match self.handshake.lock().unwrap().write_level() {
                 crypto::Level::Initial => packet::EPOCH_INITIAL,
                 crypto::Level::ZeroRTT => unreachable!(),


normanmaurer:
If you give me a bit of a hand here I can also try to write a unit test for it :) .. I found it while doing some test within https://github.com/netty/netty-incubator-codec-quic

ghedo:
This will cause application close frames to be sent in Initial and Handshake packets, which is a violation of §12.5 of the transport draft:
> CONNECTION_CLOSE frames signaling application errors (type 0x1d) MUST only appear in the application data packet number space.

Also, we already do trigger sending a 1-RTT packet when an application error is set:
https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L3828
...and we even have a test for this case:
https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L8479

But of course there could be a bug somewhere, so it would be helpful to understand your case better.

normanmaurer:
Let me try to adjust the test case as I only have some code that shows the issue with my java integration

ghedo:
Thanks for the contribution. The old `write_epoch()` close-state machinery was replaced as part of commit `8736dbec4`. Current `write_pkt_type()` explicitly considers a local application error and emits an `APPLICATION_CLOSE` frame even when no stream data is pending. Current `master` therefore already handles this case, so closing this PR as already fixed.

normanmaurer:
@ghedo thanks a lot! Sorry for dropping the ball here.

Issue #785, opened 2020-12-11
qlog output is malformed for packet_sent events when raw_encrypted is not None 


hawkinsw:
Hello!

First, thanks for the community contribution of the qlog library! 

I was investigating how qlog would handle events that include raw data and noticed a problem. When raw_encrypted or raw_decrypted are not None in the object to be serialized, the resulting qlog output is corrupt. 

I have a fix for this issue and the PR is incoming. 

Thanks again for the library!
Will

ghedo:
Fixed by https://github.com/cloudflare/quiche/commit/be23b6fcc3ce3bd595bef91931d248268b1da4c5.

PR #786, opened 2020-12-11
qlog: Always serialize frames field last

hawkinsw:
In order to preserve the behavior where multiple frames can be appended on a stream event, the frames field must serialized last. If that is not the case, it is not possible to keep the frames array "open" for appending.

This commit fixes Issue #785, specifically. It uses the preserve_order feature of the serde json library. This feature tells the library to serialize all structure fields in the order that they are inserted.

### DIFF
--- a/tools/qlog/Cargo.toml
+++ b/tools/qlog/Cargo.toml
@@ -12,6 +12,6 @@ license = "BSD-2-Clause"
 
 [dependencies]
 serde = { version = "1", features = ["derive"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", features = ["preserve_order"] }
 serde_derive = "1.0"
 serde_with = "1.3.1"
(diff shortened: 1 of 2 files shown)


LPardue (changes requested):
Thanks @hawkinsw! 

You identified an unfortunate bug in qlog when it is being run in sreaming serialization mode. This went undiscovered in the tests because we didn't have coverage of logging this event type with raw data.

So in addition to the code change, could you please enhance the `serialization_states()` unit test? What's needed is to define a new event after event2, e.g.

```
let event3 = event::Event::packet_sent(
            PacketType::Initial,
            pkt_hdr,
            Some(Vec::new()),
            None,
            Some("encrypted_foo".to_string()),
            Some("foo".to_string()),
        );
```

then before `finish_log()` is called, add that event (and its frames) via calls like   `s.add_event(event3.clone())`

Finally, the log string will need to be updated to check your raw values get serliazed out in the expected place.

LPardue (approved):
LGTM, thanks. The CI failure seems to be a transient issue caused by a problem contacting "www.nasm,us"

PR #789, opened 2020-12-14
qlog: bump version

LPardue:
So that we can cut a new release


LPardue:
I think the builds are failing for the chicken-egg dependency problem

ghedo (approved):
Looks good now.

PR #791, opened 2020-12-16
use ConnectionId in public functions

ghedo:
This changes the remaining public APIs to use the ConnectionId type for
connection ID values, instead of slices.

The FFI API is left as-is, as there isn't much point to introduce a new
type there.


LPardue (approved):
LGTM

PR #793, opened 2020-12-17
apps: implementing limiting send burst

junhochoi:
Similar to #684, but this is only for quiche-server in apps.
Should be able to help stable rtt and cwnd for high speed transfer.

### DIFF
--- a/tools/apps/src/bin/quiche-server.rs
+++ b/tools/apps/src/bin/quiche-server.rs
@@ -45,6 +45,8 @@ use quiche_apps::common::*;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
+const MAX_SEND_BURST_LIMIT: usize = MAX_DATAGRAM_SIZE * 10;
+
 fn main() {
     let mut buf = [0; 65535];
     let mut out = [0; MAX_DATAGRAM_SIZE];
@@ -137,11 +139,17 @@ fn main() {
 
     let mut pkt_count = 0;
 
+    let mut continue_write = false;
+
     loop {
         // Find the shorter timeout from all the active connections.
         //
         // TODO: use event loop that properly supports timers
-        let timeout = clients.values().filter_map(|c| c.conn.timeout()).min();
+        let timeout = match continue_write {
+            true => Some(std::time::Duration::from_secs(0)),
+
+            false => clients.values().filter_map(|c| c.conn.timeout()).min(),
+        };
 
         poll.poll(&mut events, timeout).unwrap();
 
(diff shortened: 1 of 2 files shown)


LPardue (approved):

lohith-bellad (approved):

tools/apps/src/bin/quiche-server.rs
```diff
@@ -472,6 +475,19 @@ fn main() {
...
+                client.bytes_sent += write;
+
+                if client.bytes_sent >= MAX_SEND_BURST_LIMIT {
+                    trace!(
+                        "{} pause writing at {}",
+                        client.conn.trace_id(),
+                        client.bytes_sent
+                    );
+                    client.bytes_sent = 0;
+                    break;
```
ghedo, line 489: My understanding is that after the break, the server will block again on the event loop (epoll/...). So in order for packets to be sent out again from a connection that reached the limit, a new I/O event (or a timeout) would need to happen, which seems like it could potentially stall a connection, or even completely hang it (e.g. if no event happens anymore, due to idk, packet loss on the other side of the connection).

So I'm not sure this is the proper way to resolve the problem, but I also can't think of an alternative right now.

ghedo (changes requested):

tools/apps/src/bin/quiche-server.rs
junhochoi: Right it can happen. I added some trick to set poll timeout to 0 when we need to continue writing, not to wait for incoming packet.

normanmaurer:
@junhochoi why was it considered to be better to not implement this in the lib itself (as you first did) ? To make things more flexible ?

junhochoi:
> @junhochoi why was it considered to be better to not implement this in the lib itself (as you first did) ? To make things more flexible ?
I think there is 2 things - not to break the API (adding one more return value to handle) impacting existing apps and also this can be implemented from the app without much effort (and flexible).

normanmaurer:
@junhochoi thanks a lot ... that was what I suspected. Just wanted to double check before I roll my own as well.

As a side note how do you do the "throughput" testing for QUIC atm ? Is there any tool like tcpkali ?

normanmaurer:
Do you have any before / after numbers ? I tried to implement something like this but didn't see a lot of difference

junhochoi:
> Do you have any before / after numbers ? I tried to implement something like this but didn't see a lot of difference

You can see the difference if you run with multiple concurrent connections.
Here is a short test with `quiche-server` and `h2load` (with `ngtcp2`), 1gbps link.

cmd: `h2load -N 30s -t 5 -n 50 -c 10 -m 1 --npn-list h3-29 https://buster:8443/10m.dat`

master:
```
finished in 15.12s, 3.31 req/s, 33.08MB/s
requests: 50 total, 50 started, 50 done, 50 succeeded, 0 failed, 0 errored, 0 timeout
```

this branch:
```
finished in 10.36s, 4.83 req/s, 48.27MB/s
requests: 50 total, 50 started, 50 done, 50 succeeded, 0 failed, 0 errored, 0 timeout
```

ghedo (approved):

PR #798, opened 2021-01-05
Remove delivery rate in the connection level stats logs

lohith-bellad:
Delivery rate is instanteous and is calculated at the end of every ack received. Logging it with entire connection level stats is misleading.


junhochoi (commented):
LGTM but I think it's ok to remove printing from fmt(), not in Stats structure, if the confusing part is printing stats at the end? quiche_conn_stats() can be called in the middle of the connection to get delivery rate and other stats such as cwnd.

lohith-bellad:
Agree on that. Made required changes, thanks

junhochoi (approved):

PR #801, opened 2021-01-11
ffi: don't create slice if odcid_len is 0

schien:


ghedo:
LGTM, but I think this also needs to be fixed in the `quiche_accept()` function as well.

ghedo (approved):

PR #802, opened 2021-01-11
pad server's ack-eliciting Initial packets

ghedo:
From section 14.1 of -transport:

    Similarly, a server MUST expand the payload of all UDP datagrams
    carrying ack-eliciting Initial packets to at least the smallest
    allowed maximum datagram size of 1200 bytes.


LPardue (approved):

ghedo:
Thinking about this some more, we will probably want to coalesce the first server's flight, to avoid wasting these bytes since we are limited by the anti-amplification mechanism.

Also, some tests are failing, so need to look at them.

ghedo:
Closing this, will go with the other approach instead.

PR #803, opened 2021-01-13
h3: forbid DATA frames before HEADERS

ghedo:
This is forbidden, so we should be strict about this. A couple of tests
depended on this behavior being allowed, so they need to be updated.

### DIFF
--- a/src/h3/stream.rs
+++ b/src/h3/stream.rs
@@ -261,6 +261,9 @@ impl Stream {
                         (frame::HEADERS_FRAME_TYPE_ID, false) =>
                             self.remote_initialized = true,
 
+                        (frame::DATA_FRAME_TYPE_ID, false) =>
+                            return Err(Error::FrameUnexpected),
+
                         (frame::CANCEL_PUSH_FRAME_TYPE_ID, _) =>
                             return Err(Error::FrameUnexpected),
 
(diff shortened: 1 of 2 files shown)


src/h3/stream.rs
```diff
@@ -261,6 +261,9 @@ impl Stream {
                         (frame::HEADERS_FRAME_TYPE_ID, false) =>
                             self.remote_initialized = true,
 
+                        (frame::DATA_FRAME_TYPE_ID, false) =>
```
LPardue, line 264: It would be good to add a unit test to stream.rs too. E.g.

```
#[test]
    fn data_before_headers() {
        let mut stream = Stream::new(0, false);

        let mut d = vec![42; 128];
        let mut b = octets::OctetsMut::with_slice(&mut d);


        let payload = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];

        let data = frame::Frame::Data {
            payload: payload.clone(),
        };


        data.to_bytes(&mut b).unwrap();

        let mut cursor = std::io::Cursor::new(d);

        // Parse the DATA frame type.
        stream.try_fill_buffer_for_tests(&mut cursor).unwrap();

        let frame_ty = stream.try_consume_varint().unwrap();
        assert_eq!(frame_ty, frame::DATA_FRAME_TYPE_ID);

        assert_eq!(stream.set_frame_type(frame_ty), Err(Error::FrameUnexpected));
    }
```

ghedo: Yeah, good idea. Copy-pasted your test in :) (I also checked that it fails without the fix).

LPardue (approved):
LGTM

PR #804, opened 2021-01-14
Fix duplicate declaration in quiche.h

Jaapp-:
When compiling a C program which includes `quiche.h`, the compiler noticed that `quiche_h3_config_set_dgram_poll_threshold` is defined twice.
It seems like changing `dgram` to `stream` was overlooked when this declaration was copied.


ghedo (approved):

ghedo:
Nice catch, thanks!

PR #807, opened 2021-01-14
Add quiche_conn_dgram_recv_front_len

normanmaurer:
Motivation:

It is useful to know the len of the datagram in the recv queue so we can allocate a buffer with the correct size. This is already possible in rust but wasn't exposed in the c layer.

Modifications:

Add quiche_conn_dgram_recv_front_len which expose this functionality in the c layer.

Result:

Be able to obtain the len of the first datagram in the recv queue.


LPardue (approved):
Lgtm, thanks

PR #809, opened 2021-01-15
ci: use cargo ndk 2

junhochoi:
- `cargo ndk` command line option are changed in v2.
- Since triple target doesn't work, a list of (triple, target) is added in the build matrix.
- arm-linux-androideabi is replaced with x86_64-linux-android


junhochoi:
FYI - the CI check status is yellow but this is because this pr updates list of builds, no build failures.

ghedo (approved):

PR #810, opened 2021-01-15
h3: close connection when processing HEADERS fails

ghedo:
Instead of delegating this to the application, make sure the connection
is closed with the appropriate error code, like in the other error
cases.


LPardue (approved):
LGTM

PR #814, opened 2021-01-20
http3_test: add some negative test cases

LPardue:
Previously, http3_test was focused on successful tests where the server
under test was expected to handle requests and provide complete
responses. If an error occurred the test runner would panic and throw
up to the invoker to deal with.

With this change, instead of panicking, the Result is returned directly
to the invoker. This allows for checking success or failure
explicitly.

Several new test cases have been added that ensure the server can handle
clients that violate rules set out in the HTTP/3 spec.

### DIFF
--- a/tools/http3_test/src/lib.rs
+++ b/tools/http3_test/src/lib.rs
@@ -241,7 +241,7 @@ pub const USER_AGENT: &str = "quiche-http3-integration-client";
 pub struct Http3Req {
     pub url: url::Url,
     pub hdrs: Vec<Header>,
-    body: Option<Vec<u8>>,
+    pub body: Option<Vec<u8>>,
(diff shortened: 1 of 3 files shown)


tools/http3_test/src/lib.rs
```diff
@@ -377,6 +408,22 @@ impl Http3Test {
...
+            for d in stream_data {
+                match conn.stream_send(d.stream_id, &d.data, d.fin) {
+                    Ok(_) => (),
+
+                    Err(e) => {
+                        error!(
+                            "failed to send data on stream {}: {:?}",
+                            d.stream_id, e
+                        );
+                        return Err(From::from(e));
```
junhochoi, line 421: No need to use `return` I think.

junhochoi (approved):
LGTM

PR #821, opened 2021-01-26
ci: use HTTP/1.1 to download Android NDK

ghedo:
Android builds keep failing randomly when downloading the Android NDK
from Google's servers with error "Error in the HTTP2 framing layer", so
disable HTTP/2 when downloading.


LPardue (approved):

junhochoi:
Interesting. Thanks for the fix. Just wondering if it's Google or curl issue.

PR #826, opened 2021-01-28
QUIC v1

ghedo:
This adds support for QUIC v1, while keeping support for current draft versions.

We shouldn't merge/deploy this yet, but can be useful to start testing interop.

(I had the test vectors change already made to test https://github.com/quicwg/base-drafts/pull/4431, so might as well finish this up)


ghedo:
An alternative to not merging could be adding a feature to optionally enable v1 at build time, though might end up being too messy...

LPardue (approved):

junhochoi (approved):

PR #829, opened 2021-01-28
Use CRYPTO_BUFFER to retrieve the peer certificate

normanmaurer:
Motivation:

BoringSSL uses CRYPTO_BUFFER to store the raw DER certificate already so we can make use of this and so allow to retrieve the DER certificate without the need to decode the X509 representation on each call.

Motiviation:

Use the CRYPTO_BUFFER via SSL_get0_peer_certificates

Result:

Less overhead when calling Connection::peer_cert()

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -498,21 +498,13 @@ impl Handshake {
 
     pub fn peer_cert(&self) -> Option<Vec<u8>> {
         let peer_cert = unsafe {
-            let chain = SSL_get0_peer_certificates(self.as_ptr());
-            if chain.is_null() {
-                return None;
-            }
-            let len = sk_num(chain);
-            if len <= 0 {
+            let chain = map_result_ptr(SSL_get0_peer_certificates(self.as_ptr())).ok()?;
+            if sk_num(chain) <= 0 {
                 return None;
             }
 
-            let buffer = sk_value(chain, 0) as *const CRYPTO_BUFFER;
-            if buffer.is_null() {
-                return None;
-            }
+            let buffer =  map_result_ptr(sk_value(chain, 0) as *const CRYPTO_BUFFER).ok()?;
             let out_len = CRYPTO_BUFFER_len(buffer);
-
             if out_len == 0 {
                 return None;
             }
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -1020,11 +1020,11 @@ extern {
     #[cfg(windows)]
     fn d2i_X509(px: *mut X509, input: *const *const u8, len: c_int) -> *mut X509;
 
-    // Seems like its not possible to expose sk_CRYPTO_BUFFER_num and
-    // sk_CRYPTO_BUFFER_value macros directly.
+    // STACK_OF
     fn sk_num(stack: *const STACK_OF) -> c_int;
     fn sk_value(stack: *const STACK_OF, idx: c_int) -> *mut c_void;
 
+    // CRYPTO_BUFFER
     fn CRYPTO_BUFFER_len(buffer: *const CRYPTO_BUFFER) -> usize;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@LPardue @ghedo can you help me to understand why it fails for windows ?

normanmaurer:
Ok I have no idea why it fails on windows.... I will need a bit of help from others that have more experience here.

normanmaurer:
Also it fails for `fuzz`.. no idea what exactly this is doing tho. All I know at the moment is that it compiles and works for me locally on a Mac :/

normanmaurer:
Seems like I was able to get it built without using the "proper macros": https://github.com/cloudflare/quiche/pull/829/commits/14b383fea74b421e154b67258121adc1945a11a6

I would be interested if there is any other way to retain the "original code" and still have it work on all platforms etc.

ghedo:
Yeah, you can't use C macros in Rust FFI, only symbols that can be linked against.

normanmaurer:
@ghedo TIL... Ok then it should be ok as it is.

src/tls.rs
```diff
@@ -489,24 +498,28 @@ impl Handshake {
 
     pub fn peer_cert(&self) -> Option<Vec<u8>> {
         let peer_cert = unsafe {
-            let mut out: *mut libc::c_uchar = ptr::null_mut();
+            let chain = SSL_get0_peer_certificates(self.as_ptr());
+            if chain.is_null() {
```
ghedo, line 502: Rather than have all these `is_null()` checks, it's probably better to use `map_result_ptr()`, e.g.
```rust
             let chain = map_result_ptr(SSL_get0_peer_certificates(self.as_ptr())).ok()?;
```

src/tls.rs
```diff
@@ -489,24 +498,28 @@ impl Handshake {
...
+                return None;
+            }
+            let len = sk_num(chain);
```
ghedo, line 505: Don't think this variable is used anywhere other than the check? So can just do:
```rust
if sk_num(chain) <= 0 {
```

src/tls.rs
```diff
@@ -1007,13 +1020,16 @@ extern {
     #[cfg(windows)]
     fn d2i_X509(px: *mut X509, input: *const *const u8, len: c_int) -> *mut X509;
 
-    fn i2d_X509(px: *const X509, out: *mut *mut u8) -> c_int;
+    // Seems like its not possible to expose sk_CRYPTO_BUFFER_num and
```
ghedo, line 1023:
```suggestion
```

src/tls.rs
```diff
@@ -1007,13 +1020,16 @@ extern {
...
+    // sk_CRYPTO_BUFFER_value macros directly.
```
ghedo, line 1024:
```suggestion
    // STACK_OF
```

src/tls.rs
```diff
@@ -1007,13 +1020,16 @@ extern {
...
+    fn sk_num(stack: *const STACK_OF) -> c_int;
+    fn sk_value(stack: *const STACK_OF, idx: c_int) -> *mut c_void;
+
+    fn CRYPTO_BUFFER_len(buffer: *const CRYPTO_BUFFER) -> usize;
```
ghedo, line 1028:
```suggestion
    // CRYPTO_BUFFER
    fn CRYPTO_BUFFER_len(buffer: *const CRYPTO_BUFFER) -> usize;
```

ghedo (changes requested):
Looks good overall, but added some nits, thanks!

src/tls.rs
normanmaurer: sure why not ....

normanmaurer:
@ghedo thanks for the review... addressed all of the comment.s PTAL again

ghedo (approved):

ghedo:
Merged, thanks!

PR #830, opened 2021-01-28
Remove quiche_rangebuf as it is not used anymore

normanmaurer:
Motivation:

quiche_rangebuf is not used anymore, we should remove it from the header file.

Modifications:

Remove unused typedef

Result:

Code cleanup


junhochoi (approved):
Nice catch!

ghedo:
Ah nice, thanks!

PR #831, opened 2021-01-28
recovery: update app_limited flag after ack received

junhochoi:
Currently we update app_limited flag on sent, but
this flag also need to be updated after ack
processing is done, especially when we get multiple
ack packets (contains multiple ack range) back to back.

Current problem is that it may cause to increase cwnd
more than expected, resulting overshooting cwnd.

e.g.
let's assume cwnd=10, ssthresh=inf, bytes_in_flight=0, datagram_size=1
sender send packet pn=0 to 9
    -> cwnd=10, bytes_in_flight=10, app_limited=false
sender received ack (0, 1)
    -> cwnd=12, bytes_in_flight=8, app_limited=false -> BAD
sender received ack (2, 3)
    -> cwnd=14, bytes_in_flight=8, app_limited=false -> BAD
sender send packet pn=10
    -> cwnd=14, bytes_in_flight=9, app_limited=true

After this fix:

sender send packet pn=0 to 9
    -> cwnd=10, bytes_in_flight=10, app_limited=false
sender received ack (0, 1)
    -> cwnd=12, bytes_in_flight=8, app_limited=true
sender received ack (2, 3)
    -> cwnd=12, bytes_in_flight=8, app_limited=true
sender send packet pn=10
    -> cwnd=12, bytes_in_flight=9, app_limited=true

You may wonder it's correct to increase cwnd when we receive ack,
but currently we use RFC3465 and considering that regardless of ack "packet"
(because ack frequency can be different depending the receiver e.g. ack
frequency=10) but actual acked range in a batch is more important.
This change will make sure when we get ack pn=0, 1, 2, 3 whether in 2 ack packets
or in 1 ack packet, we only increase cwnd 2 (L) x max_datagram_size at most.

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -345,6 +345,8 @@ impl Recovery {
 
         self.on_packets_acked(newly_acked, epoch, now);
 
+        self.app_limited = self.bytes_in_flight < self.congestion_window;
+
         self.pto_count = 0;
 
         self.set_loss_detection_timer(handshake_status, now);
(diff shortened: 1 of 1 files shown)


junhochoi:
Closing - it needs a better solution.

Issue #817, opened 2021-01-22
Connection::recv(...) semantics with empty buffers


normanmaurer:
While working with quiche I somehow missed that `Connection::recv(...)` doesn't return `Err::Done` when you pass in an empty buffer but just return 0. This did lead to an infinite loop as I basically followed the example in the docs:

```
loop {
    let read = socket.recv(&mut buf).unwrap();

    let read = match conn.recv(&mut buf[..read]) {
        Ok(v) => v,

        Err(e) => {
            // An error occurred, handle it.
            break;
        },
    };
}
```

So I wonder if we should at least adjust the docs to explain that you should never call it with an empty buffer or if we should return `Err:Done` in tis case. I guess updating the docs would be better.  Just wanted to see what the preference is and I can provide a PR .

ghedo:
I don't think `Done` makes sense for `recv()` to be honest, as it would mean the application should stop calling it as there is no more stuff to do, however quiche can't know if there are going to be more packets received, so it is up to the application to decide. I think a better error for this case would be `BufferTooShort`.

normanmaurer:
@ghedo yeah `BufferTooShort` sounds better I agree...  I can do a PR if you think that makes sense.

ghedo:
Sure, go ahead, thanks.

normanmaurer:
Fixed by https://github.com/cloudflare/quiche/pull/832

ghedo:
Fixed by https://github.com/cloudflare/quiche/commit/d1e704b041fcbebab0abbc7e148c118586f16705.

PR #832, opened 2021-01-29
Return BufferTooShort when Connection::recv(...) is called with an em…

normanmaurer:
…pty buffer

Motivation:

We should return BufferTooShort if recv is called with an empty buffer to guard against missuse and possible endless loops in users code.

Modifications:

- Add extra if statement that checks for an empty buffer and return the correct error in this case
- Add unit test

Result:

Less risk for users


normanmaurer:
Fix for https://github.com/cloudflare/quiche/issues/817

ghedo (approved):

PR #851, opened 2021-02-15
limit send capacity over multiple stream_send() calls

ghedo:
A connection send capacity is limited by the size of the flow and
congestion windows. However, cwnd is not reduced between stream_send()
calls, which might cause more data than necessary to be buffered.

Instead of using the current available cwnd size for each call, buffer
the value and reuse it over a single "flight".

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -884,6 +884,9 @@ pub struct Connection {
     /// Whether we send MAX_DATA frame.
     almost_full: bool,
 
+    /// Number of stream data bytes that can be buffered.
+    tx_cap: usize,
+
     /// Total number of bytes sent to the peer.
     tx_data: u64,
 
@@ -1238,6 +1241,8 @@ impl Connection {
             max_rx_data,
             max_rx_data_next: max_rx_data,
             almost_full: false,
+
+            tx_cap: 0,
 
             tx_data: 0,
             max_tx_data: 0,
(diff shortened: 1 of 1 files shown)


LPardue (approved):

LPardue (approved):

src/lib.rs
```diff
@@ -8647,6 +8654,68 @@ mod tests {
...
+        assert_eq!(pipe.server.stream_recv(12, &mut buf), Ok((6, true)));
+
+        assert_eq!(pipe.server.tx_cap, 11565);
+
+        assert_eq!(pipe.server.stream_send(0, &buf[..5000], false), Ok(5000));
+        assert_eq!(pipe.server.stream_send(4, &buf[..5000], false), Ok(5000));
+        assert_eq!(pipe.server.stream_send(8, &buf[..5000], false), Ok(1565));
+
+        // No more connection send capacity.
+        assert_eq!(pipe.server.stream_send(12, &buf[..5000], false), Ok(0));
```
junhochoi, line 8714: `stream_send()` documentation says it will return `Done` if no data was written (https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L2951). Is `Ok(0)` acceptable return value too?

ghedo: That's not been the case for a while AFAICT. In any case fix should go into a separate PR since it was the case before this change too.

junhochoi (approved):

PR #853, opened 2021-02-16
coalesce outgoing packets

ghedo:
Instead of writing one QUIC packet at a time, `send()` will now try to
coalesce multiple packets onto the same UDP datagram.

In order to support that, padding of Initial packets needs to be moved
outside the QUIC packet itself, so instead of using PADDING frames, we
simply leave some free space at the end of the datagram.

One notable change, is that Initial packets from the server are now also
padded as needed to reach the minimum 1200 length. This is to ensure
that the network path actually supports the minimum size in both
directions.

Though in practice this doesn't change much, since in most cases the
packet is pretty full already so the padding done from the server is
minimal

---

This fixes a `MUST` violation (padding of server Initial packets) in the later draft versions, which prevents interop with some implementations, and supersedes https://github.com/cloudflare/quiche/pull/802.

Also, this requires https://github.com/cloudflare/quiche/pull/852 to avoid breaking tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -309,7 +309,7 @@ const MAX_ACK_RANGES: usize = 68;
 // The highest possible stream ID allowed.
 const MAX_STREAM_ID: u64 = 1 << 60;
 
-/// The default max_datagram_size used in congestion control.
+// The default max_datagram_size used in congestion control.
 const MAX_SEND_UDP_PAYLOAD_SIZE: usize = 1200;
 
 // The default length of DATAGRAM queues.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -2060,6 +2060,68 @@ impl Connection {
...
+        }
+
+        // Pad UDP datagram if it contains a QUIC Initial packet.
+        if has_initial && left > 0 && done < MIN_CLIENT_INITIAL_LEN {
+            let pad_len = cmp::min(left, MIN_CLIENT_INITIAL_LEN - done);
+
+            // Fill padding area with null bytes, to avoid leaking information
+            // in case the application reuses the packet buffer.
+            let pad_ptr = (&mut out[done..done + pad_len]).as_mut_ptr();
+            unsafe { std::ptr::write_bytes(pad_ptr, 0, pad_len) };
```
junhochoi, line 2116:
```suggestion
            out[done..done + pad_len].fill(0);
```
Looks like it can be simplified and safe as above, rust 1.50 and later (CI use already 1.50)
https://doc.rust-lang.org/std/primitive.slice.html#method.fill

ghedo: Updated, thanks.

junhochoi (approved):

PR #856, opened 2021-02-19
send STOP_SENDING frame when stream's read-side is shutdown

ghedo:


src/stream.rs
```diff
@@ -384,6 +389,18 @@ impl StreamMap {
         }
     }
 
+    /// Adds or removes the stream ID to/from the stopped streams set with the
+    /// given error code.
+    ///
+    /// If the stream was already in the list, this does nothing.
+    pub fn mark_stopped(&mut self, stream_id: u64, reset: bool, error_code: u64) {
```
LPardue, line 396:
```suggestion
    pub fn mark_stopped(&mut self, stream_id: u64, stop: bool, error_code: u64) {
```

maybe?

LPardue (approved):

src/stream.rs
ghedo: Ah, yes. Fixed.

PR #860, opened 2021-02-23
h3: only emit Finished events for request/push streams

ghedo:
After 5d39b5b1, we started collecting GREASE and other unknown type
streams correctly. However the way these were handled meant that we were
issuing Finished events regardless of the stream's type.

This makes FInished events be emitted only for request and push streams.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -1294,7 +1294,7 @@ impl Connection {
                 };
 
                 if conn.stream_finished(s) {
-                    self.finished_streams.push_back(s);
+                    self.process_finished_stream(s);
                 }
 
                 // TODO: check if stream is completed so it can be freed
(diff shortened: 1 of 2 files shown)


LPardue:
Looking good - think you missed one case at https://github.com/cloudflare/quiche/pull/860/files#diff-05f5bdcd2bf3c962d0806f69b20955595bd549f8f584e0ed8fc43063a92ffd98R1252

ghedo:
@LPardue AFAICT that's where the finished stream are popped from the queue, but what we are interested in here is not adding streams to the queue in the first place (i.e. `push_back()`).

LPardue:
So my comment was a bit terse :). What I was trying to say is that I think the fix could possibly go further; define push_finished/pop_finished functions that incorporate the checks you added here. That would make it a bit clearer of the intent, in the various parts of mod.rs that handle this stuff.

ghedo:
Not sure how that's better? I kind of like the fact that I don't have to pass the `conn` object around TBH, as it's only needed to check if the stream is finished. We also don't need to do anything for pop, as we need to do this check only once.

LPardue (approved):

LPardue:
Makes sense, shipit!

PR #862, opened 2021-02-24
send RESET_STREAM frame when stream's write-side is shutdown

ghedo:
This also ensures that the stream's send buffer is considered "complete"
when all previously buffered data is dropped, by marking all the dropped
data as acked, so that the stream can be collected properly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3228,9 +3228,10 @@ impl Connection {
                 self.streams.mark_readable(stream_id, false);
             },
 
-            // TODO: send RESET_STREAM
             Shutdown::Write => {
-                stream.send.shutdown()?;
+                let final_size = stream.send.shutdown()?;
+
+                self.streams.mark_reset(stream_id, true, err, final_size);
 
                 // Once shutdown, the stream is guaranteed to be non-writable.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -6556,34 +6557,79 @@ mod tests {
...
+
+        // Server reads the data.
         let mut r = pipe.server.readable();
+        assert_eq!(r.next(), Some(4));
         assert_eq!(r.next(), None);
 
+        assert_eq!(pipe.server.stream_recv(4, &mut buf), Ok((15, true)));
+
+        // Stream is collected on both sides.
+        // TODO: assert_eq!(pipe.client.streams.len(), 0);
```
LPardue, line 6628: what is the TODO dependent on?

LPardue (approved):

src/lib.rs
ghedo: Ah, I forgot about that. Should be fixed now. Basically just needed to make the client go through readable streams in the test.

ghedo: Oh, wait nvm, fixing that will require some changes to how RESET_STREAM is handled, see https://github.com/cloudflare/quiche/compare/reset_stream

PR #864, opened 2021-02-26
WIP: http3_test: support regex in EXPECT_REQ_HEADERS

LPardue:


tools/http3_test/tests/httpbin_tests.rs
```diff
@@ -957,6 +957,22 @@ mod httpbin_tests {
                 if let Some(expected_headers) = &expect_req_headers() {
                     for (name, val) in expected_headers {
                         if let Some(expected_value) = val.as_str() {
+                            // Check if the header value is actually a special
+                            // "regex header value" - prefix "htt3_test_regex". If so,
+                            // test it's pattern otherwise just check the value verbatim.
+                            if let Some(pattern) =
+                                expected_value.strip_prefix("http3_test_regex")
```
junhochoi, line 964: For example, when expected value is `http3_test_regex\d+` when it will check the value is all number?
Would like to have some comment on usage, or test.

PR #865, opened 2021-03-01
introduce a possibility for manual stream flow control by an application

sxgunchenko:
This feature allows an application to slow down the reception of packets from the peer.
Without this, the application needs to buffer packets. And since it cannot stop the library
from sending MAX_DATA frames, the buffer can grow to quite large sizes in case the processing
speed is much lower than the packet reception speed.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -195,6 +195,9 @@ void quiche_config_enable_dgram(quiche_config *config, bool enabled,
                                 size_t recv_queue_len,
                                 size_t send_queue_len);
 
+// Configures whether to enable automatic flow control limits update on streams.
+void quiche_config_enable_stream_flow_control_auto_update(quiche_config *config, bool enabled);
+
 // Frees the config object.
 void quiche_config_free(quiche_config *config);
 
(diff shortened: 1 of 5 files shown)


LPardue:
Can you explain a bit more about the problem this is solving? It would seem that some of the goal is already achieved by simply not calling stream_recv() when you don't want to buffer stream data in the application and release flow control.

sxgunchenko:
Suppose we have 2 streams (with IDs 1 and 2) on a connection and the application has different paces of data processing on the streams (1 is slower).

`-> UDP pkt -> DATA(1), DATA(2) -> h3.poll() -> Event::Data(1), Event::Data(2)`

Suppose the application is not ready to process the data on stream 1, but it is ready on stream 2. For now, there is no possibility to skip the data event on the prior stream (https://github.com/cloudflare/quiche/blob/master/src/h3/mod.rs#L1281). 

So, the application has 2 options: 
1) stop reading on the entire connection - it is obviously poor behaviour
2) do read the data (which leads to automatic MAX_DATA frames sending) and buffer it on the streams the app is not ready to process - in this case, the buffer may grow to quite large sizes

With the feature of manual flow control, the application could buffer currently available data on the stream it's not ready to process and do not send MAX_DATA at the moment, but only after it has processed this data chunk.

(The second possible solution I wonder is to allow to skip data events while `poll()`ing the connection.)

LPardue:
Thanks for the explanation. There is already a PR that adopts your suggested second solution, please see https://github.com/cloudflare/quiche/pull/765. It's an API change but sounds like it will address your problem and it's ready to try out

sxgunchenko:
Yup. It seems like that PR solves the issue. Thank you!

Issue #766, opened 2020-12-01
Expose error code / message of APPLICATION_CLOSE when receiving the frame


normanmaurer:
It would be very useful to be able to get access to the error code / message of an received APPLICATION_CLOSE frame. This way a user of quiche can better understand on why a connection was closed in the first place.

For example I am using quiche in netty and currently building an h3 layer on top of it in which I would love to expose some more details about why a remote peer closed the connection

normanmaurer:
The same may also be true for `CONNECTION_CLOSE`.

Maybe it would be just good enough to expose something that can be "queried" once a connection is marked as closed ?

normanmaurer:
Any preference / ideas here ? I would be happy to work on a PR as well.

normanmaurer:
Any idea ? I just encountered the problem again that the connection close frame was received due ALPN mismatch but there is no way for me to know without enabling trace logging.

LPardue:
Seems like a reasonable idea. There's a few ways this could be done, have you got any sketches of what would address your case? Are you interested in knowing of a connection error that was **sent**?

normanmaurer:
I am mainly interested in the receiving side (at least atm)... I was thinking of maybe just expose an "getter" on the connection (as an optional)

normanmaurer:
I just wanted to ensure this is a "reasonable" way how to expose something like this before trying to come up with a PR.

LPardue:
The simplest option is probably something that is the mirror of https://docs.quic.tech/quiche/struct.Connection.html#method.close. So you can access `bool: app, err: u64, reason: [u8]` of the received CONNECTION_CLOSE frame. Independent of the exact API, it seems we'd need to store the information in the library, which could be done during processing (`src/lib.rs:process_frame()`).

normanmaurer:
yes... And the same also for `APPLICATION_CLOSE`.. @LPardue if this sounds reasonable I will work on a PR, wdyt ?

LPardue:
yep. It's worth clarifying that in QUIC transport terms, the wire frame is always CONNECTION_CLOSE but with different type codepoints and fields, so quiche models these as different frame::Frame::Types even if they are conceptually similar. For simplicity I think just one accesor, with the `bool: app` is going to be cleanest. But while you work on this consider if you think knowing the received "transport CONNECTION_CLOSE" frame's `Frame Type` field is useful (see https://tools.ietf.org/html/draft-ietf-quic-transport-34#section-19.19)

[...]

PR #879, opened 2021-03-24
Expose CONNECTION_CLOSE's Error Code and Reason Phrase via received_close_info()

kevinjohna6:
Addresses #766

Renames existing `error`, `app_error`, `app_reason` fields to have a `local_` prefix.
Adds fields for `peer_error`, etc.
Adds a `received_close_info` method to `Connection`, as well as the FFI.
Includes tests.

I am especially unfamiliar with conventions for the FFI. Are there tests for it?

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -338,6 +338,13 @@ uint64_t quiche_conn_peer_streams_left_uni(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
+// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small
+int quiche_conn_received_close_info(quiche_conn *conn,
+                                    bool *is_app,
+                                    uint64_t *error_code,
+                                    uint8_t *reason,
+                                    size_t *reason_len);
+
 // Initializes the stream's application data.
 //
 // Stream data can only be initialized once. Additional calls to this method
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -338,7 +338,7 @@ uint64_t quiche_conn_peer_streams_left_uni(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
-// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small
+// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small.
 int quiche_conn_received_close_info(quiche_conn *conn,
                                     bool *is_app,
                                     uint64_t *error_code,
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -339,6 +339,9 @@ uint64_t quiche_conn_peer_streams_left_uni(quiche_conn *conn);
 bool quiche_conn_is_closed(quiche_conn *conn);
 
 // Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small.
+//
+// Always sets reason_len to the length of reason, so an exact length
+// buffer can be malloc'd and filled with a second call.
 int quiche_conn_received_close_info(quiche_conn *conn,
                                     bool *is_app,
                                     uint64_t *error_code,
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -338,14 +338,11 @@ uint64_t quiche_conn_peer_streams_left_uni(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
-// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small.
-//
-// Always sets reason_len to the length of reason, so an exact length
-// buffer can be malloc'd and filled with a second call.
-int quiche_conn_received_close_info(quiche_conn *conn,
+// Returns true if Some, false if None
+bool quiche_conn_peer_error(quiche_conn *conn,
                                     bool *is_app,
                                     uint64_t *error_code,
-                                    uint8_t *reason,
+                                    const uint8_t **reason,
                                     size_t *reason_len);
 
 // Initializes the stream's application data.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -828,6 +828,15 @@ impl Config {
         self.dgram_send_max_queue_len = send_queue_len;
     }
 }
+/// Container for data sent and received with CONNECTION_CLOSE frames
+struct ConnectionError {
+    /// ApplicationClose or ConnectionClose (both are CONNECTION_CLOSE frames)
+    is_app: bool,
+    /// Error Code: A variable-length integer error code
+    error_code: u64,
+    /// Reason Phrase: Additional diagnostic information for the closure
+    reason_phrase: Vec<u8>,
+}
 
 /// A QUIC connection.
 pub struct Connection {
@@ -911,14 +920,15 @@ pub struct Connection {
     /// Received address verification token.
     token: Option<Vec<u8>>,
 
-    /// Error code to be sent to the peer in CONNECTION_CLOSE.
-    error: Option<u64>,
+    /// Error code and reason to be sent to the peer in a CONNECTION_CLOSE
+    /// frame. Either ConnectionClose or ApplicationClose depending on
+    /// is_app.
+    local_error: Option<ConnectionError>,
 
-    /// Error code to be sent to the peer in APPLICATION_CLOSE.
-    app_error: Option<u64>,
-
-    /// Error reason to be sent to the peer in APPLICATION_CLOSE.
-    app_reason: Vec<u8>,
+    /// Error code and reason received from the peer in a CONNECTION_CLOSE
+    /// frame. Either ConnectionClose or ApplicationClose depending on
+    /// is_app.
+    peer_error: Option<ConnectionError>,
 
     /// Received path challenge.
     challenge: Option<Vec<u8>>,
@@ -1260,10 +1270,9 @@ impl Connection {
 
             token: None,
 
-            error: None,
+            local_error: None,
 
-            app_error: None,
-            app_reason: Vec::new(),
+            peer_error: None,
 
             challenge: None,
 
@@ -1534,7 +1543,7 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        let is_closing = self.error.is_some() || self.app_error.is_some();
+        let is_closing = self.is_closing();
 
         if is_closing {
             return Err(Error::Done);
@@ -2137,7 +2146,7 @@ impl Connection {
             return Err(Error::Done);
         }
 
-        let is_closing = self.error.is_some() || self.app_error.is_some();
+        let is_closing = self.is_closing();
 
         if !is_closing {
             self.do_handshake()?;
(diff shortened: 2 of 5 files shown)


LPardue:
Thanks for the PR!

The first observation is that the reason phrase has a strict upper bound of the largest frame size allowed by the packet size. In practice the length of the reason is going to be small (and if not is at least bounded) but the receiver has no way to know what the length of the phrase may or may not be. The C API here seems a bit awkward to use if the caller doesn't care about the reason phrase and just wants the error code.  A different approach would be to split the method in two: one just returns the error code and is_app, the other returns the reason phrase (maybe the same way we handle HTTP headers using a callback; see https://github.com/cloudflare/quiche/blob/master/src/h3/ffi.rs#L122). Interested to see what others think.

The second observation is based on seeing the code. There's a bit of duplication going on, which might be the source of a bug like here https://github.com/cloudflare/quiche/pull/879/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R3861. 

I don't think there is any way for the protocol to have both an transport error and application error at the same time, so an alternative idea that might make some of the code a bit clearer is to define a struct containing error_code, is_app, and reason;  e.g.

```
struct ConnectionError {
  u64: error_code,
  bool: is_app,
  reason: Vec<u8>
}
```

then instantiate local and peer structs

```
peer_error: Option<ConnectionError>,
local_error: Option<ConnectionError>
```

This would require supporting changes to existing code but I think it would be a net improvement

kevinjohna6:
I agree that passing the reason as a pointer and length rather than filling a buffer would be better.
I am not sure what memory safety considerations need to be taken for this, so I opted for the copy.
Can we provide the FFI a malloc'd buffer rather than relying on a callback just for the single reason phrase, or is it fine just to use a callback?
The `error_code` and `is_app` should still be set even when the `reason` does not fit in the provided buffer, so I think the only other change that may be needed is more conventional return codes.

Regarding the "bug" in received_close_info, peer_app_reason is (unreasonably) used for both the application and connection close.
Quiche does not support sending a reason phrase with ConnectionClose packets, although this seems to be allowed by the latest draft.
So local_app_reason would only be used for ApplicationClose, while peer_app_reason is used for both CONNECTION_CLOSE types.
I used `peer_app_reason` just to keep symmetry, as it would be empty anyway for ConnectionClose events sent from Quiche.
This is not an issue if we use a ConnectionError like struct (and/or if we allow setting Reason Phrase with ConnectionError).

I'm not exactly sure what you are referring to as code duplication, but I agree using a struct for both would be an improvement. I avoided defining a new struct as per the discussion in the issue.

Although it was not part of the issue, I think allowing Quiche to send a reason with ConnectionClose packets may be a reasonable change as part of this PR, in combination with using a struct like ConnectionError.

I will commit some of these changes, but it is fine if we revert due to further discussion.

LPardue:
> I agree that passing the reason as a pointer and length rather than filling a buffer would be better.
I am not sure what memory safety considerations need to be taken for this, so I opted for the copy.
Can we provide the FFI a malloc'd buffer rather than relying on a callback just for the single reason phrase, or is it fine just to use a callback?
The error_code and is_app should still be set even when the reason does not fit in the provided buffer, so I think the only other change that may be needed is more conventional return codes.

Good points. Maybe a method that retrieves the error code could indicate the length of the reason phrase, Then the reason phrase getter can just accept a buffer of any length and truncate if needed? 

> I'm not exactly sure what you are referring to as code duplication, but I agree using a struct for both would be an improvement. I avoided defining a new struct as per the discussion in the issue.

Basically I meant there were 6 fields used to represent two logical objects - which makes it easy to think there is a bug if there wasn't ;) I think an internal struct makes sense, even if we chose not to expose that struct to the API.

kevinjohna6:
I did not see #783. I separated that change from this PR.

src/lib.rs
```diff
@@ -4345,11 +4388,23 @@ impl Connection {
 
             frame::Frame::PathResponse { .. } => (),
 
-            frame::Frame::ConnectionClose { .. } => {
+            frame::Frame::ConnectionClose {
+                error_code, reason, ..
```
LPardue, line 4392: If you destructure this as `reason_phrase` then it simplifies the creation of `ConnectionError` below. Some for the ApplicationClose case below.

include/quiche.h
```diff
@@ -338,6 +338,13 @@ uint64_t quiche_conn_peer_streams_left_uni(quiche_conn *conn);
 // Returns true if the connection is closed.
 bool quiche_conn_is_closed(quiche_conn *conn);
 
+// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small
```
LPardue, line 341:
```suggestion
// Returns 0 if Some, 1 if None, -1 if Some but reason_len is too small.
```

LPardue (commented):
Left some nits but otherwise looking good with this approach.

src/lib.rs
```diff
@@ -828,6 +828,15 @@ impl Config {
         self.dgram_send_max_queue_len = send_queue_len;
     }
 }
+/// Container for data sent and received with CONNECTION_CLOSE frames
+struct ConnectionError {
+    /// ApplicationClose or ConnectionClose (both are CONNECTION_CLOSE frames)
+    is_app: bool,
+    /// Error Code: A variable-length integer error code
```
ghedo, line 835:
```suggestion

    /// The error code carried by the `CONNECTION_CLOSE` frame.
```

src/lib.rs
```diff
@@ -828,6 +828,15 @@ impl Config {
         self.dgram_send_max_queue_len = send_queue_len;
     }
 }
+/// Container for data sent and received with CONNECTION_CLOSE frames
+struct ConnectionError {
+    /// ApplicationClose or ConnectionClose (both are CONNECTION_CLOSE frames)
```
ghedo, line 833:
```suggestion
    /// Whether the error came from the application or the transport layer.
```

src/lib.rs
```diff
@@ -911,14 +920,15 @@ pub struct Connection {
     /// Received address verification token.
     token: Option<Vec<u8>>,
 
-    /// Error code to be sent to the peer in CONNECTION_CLOSE.
-    error: Option<u64>,
-
-    /// Error code to be sent to the peer in APPLICATION_CLOSE.
-    app_error: Option<u64>,
+    /// Error code and reason to be sent to the peer in a CONNECTION_CLOSE
+    /// frame. Either ConnectionClose or ApplicationClose depending on
```
ghedo, line 924:
```suggestion
    /// frame.
```

src/lib.rs
```diff
@@ -911,14 +920,15 @@ pub struct Connection {
...
+    /// is_app.
```
ghedo, line 925:
```suggestion
```

src/lib.rs
```diff
@@ -911,14 +920,15 @@ pub struct Connection {
...
+    local_error: Option<ConnectionError>,
 
-    /// Error reason to be sent to the peer in APPLICATION_CLOSE.
-    app_reason: Vec<u8>,
+    /// Error code and reason received from the peer in a CONNECTION_CLOSE
+    /// frame. Either ConnectionClose or ApplicationClose depending on
```
ghedo, line 929:
```suggestion
    /// frame.
```

src/lib.rs
```diff
@@ -911,14 +920,15 @@ pub struct Connection {
...
+    /// is_app.
```
ghedo, line 930:
```suggestion
```

src/lib.rs
```diff
@@ -3828,6 +3838,33 @@ impl Connection {
         self.closed
     }
 
+    /// Returns true if the connection is closing.
+    #[inline]
+    pub fn is_closing(&self) -> bool {
```
ghedo, line 3843: This function doesn't actually do what the description say it does. It only returns true if the local endpoint set an error, not whether the error is actually sent (so the closing process is actually initiated), or whether an error is _received_.

Since we already have the `is_draining()` method that covers all of the above, we shouldn't add this, as it simply adds confusion and doesn't seem to be particularly useful anyway.

src/lib.rs
```diff
@@ -3828,6 +3838,33 @@ impl Connection {
...
+        self.local_error.is_some()
+    }
+
+    /// Returns the error code and reason phrase from a connection close or
```
ghedo, line 3847:
```suggestion
    /// Returns the error received from the peer, if any.
```

src/lib.rs
```diff
@@ -3828,6 +3838,33 @@ impl Connection {
...
+    /// application close event received from the peer, if any.
```
ghedo, line 3848:
```suggestion
```

src/lib.rs
```diff
@@ -3828,6 +3838,33 @@ impl Connection {
...
+    /// The values contained in the tuple are symmetric with the [`close()`]
+    /// method.
+    ///
+    /// Note that a `Some` return value does not necessarily imply
+    /// [`is_closed()`] or any other connection state.
+    ///
+    /// [`close()`]: struct.Connection.html#method.close
+    /// [`is_closed()`]: struct.Connection.html#method.is_closed
+    #[inline]
+    pub fn received_close_info(&self) -> Option<(bool, u64, &[u8])> {
```
ghedo, line 3859:
```suggestion
    pub fn peer_error(&self) -> Option<(bool, u64, &[u8])> {
```

ghedo, line 3859: Why not just return the `ConnectionClose` structure rather than a tuple? (that is, `ConnectionClose` would need to be a public type).

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
         );
     }
 
+    #[test]
+    fn received_close_info() {
```
ghedo, line 8872:
```suggestion
    fn peer_error() {
```

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
...
+        let mut pipe = testing::Pipe::default().unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        assert_eq!(pipe.server.close(false, 0x1234, b"hello?"), Ok(()));
+
+        assert_eq!(pipe.advance(), Ok(()));
+
+        assert_eq!(
+            pipe.client.received_close_info(),
```
ghedo, line 8881:
```suggestion
            pipe.client.peer_error(),
```

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
...
+            Some((false, 0x1234u64, &b"hello?"[..]))
+        );
+    }
+
+    #[test]
+    fn app_received_close_info() {
```
ghedo, line 8887:
```suggestion
    fn app_peer_error() {
```

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
...
+        let mut pipe = testing::Pipe::default().unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        assert_eq!(pipe.server.close(true, 0x1234, b"hello!"), Ok(()));
+
+        assert_eq!(pipe.advance(), Ok(()));
+
+        assert_eq!(
+            pipe.client.received_close_info(),
```
ghedo, line 8896:
```suggestion
            pipe.client.peer_error(),
```

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn app_received_close_info() {
+        let mut pipe = testing::Pipe::default().unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        assert_eq!(pipe.server.close(true, 0x1234, b"hello!"), Ok(()));
+
```
ghedo, line 8892:
```suggestion
```

src/lib.rs
```diff
@@ -8805,6 +8868,36 @@ mod tests {
         );
     }
 
+    #[test]
+    fn received_close_info() {
+        let mut pipe = testing::Pipe::default().unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        assert_eq!(pipe.server.close(false, 0x1234, b"hello?"), Ok(()));
+
```
ghedo, line 8877:
```suggestion
```

ghedo (changes requested):
Thanks for the PR and sorry for the delay. I added a few comments, mostly nits, but there are a few regarding the new public APIs that need to be addressed.

src/lib.rs
LPardue: I'll take some blame here. In the current code in `recv_single` and `send_single` there's `is_closing` based on has a view of whether the connection has an error or not. So I think I suggested we could make an `is_closing()` method. But I do agree with ghedo that this can be a confusing public API, so apologies for any churn. Since the check is now just a single parameter check it doesn't save much to wrap it in a method.

ghedo (approved):

ghedo:
I applied some minor changes that I missed before, and merged in https://github.com/cloudflare/quiche/commit/8736dbec4c9b0ddf95949baaeb134b2599eb4ac6. Thank you again @kevinjohna6!

ghedo:
And added ab3de258f9fde2ce3f6ccaa722853af3620a7d9f as well to fix the build that I accidentally broken in the manual merge :facepalm:

PR #880, opened 2021-03-24
Add SSL_ERROR_WANT_CERTIFICATE_VERIFY and SSL_ERROR_PENDING_SESSION h…

normanmaurer:
…andling

Motivation:

For more advanced use-cases the certificate verification or session lookup may be executed asynchronously.  In this case SSL_ERROR_WANT_CERTIFICATE_VERIFY / SSL_ERROR_PENDING_SESSION will be returned

Modifications:

Add mapping for SSL_ERROR_WANT_CERTIFICATE_VERIFY / SSL_ERROR_PENDING_SESSIO

Result:

Be able to to use asynchronous certificate verification and session lookup

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -848,6 +848,9 @@ fn map_result_ssl(ssl: &Handshake, bssl_result: c_int) -> Result<()> {
                 // SSL_ERROR_SYSCALL
                 5 => Err(Error::TlsFail),
 
+                // SSL_ERROR_PENDING_SESSION
+                11 => Err(Error::Done),
+
                 // SSL_ERROR_PENDING_CERTIFICATE
                 12 => Err(Error::Done),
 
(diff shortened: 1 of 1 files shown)


src/tls.rs
```diff
@@ -848,6 +848,9 @@ fn map_result_ssl(ssl: &Handshake, bssl_result: c_int) -> Result<()> {
                 // SSL_ERROR_SYSCALL
                 5 => Err(Error::TlsFail),
 
+                // SSL_ERROR_PENDING_SESSION
```
ghedo, line 851: AFAICT this is only relevant to legacy session ID-based resumption, which is forbidden in TLS 1.3 (and QUIC), so I don't think we need this (which is why it was left out originally IIRC).

But on the other hand it also doesn't hurt I guess.

src/tls.rs
```diff
@@ -857,6 +860,9 @@ fn map_result_ssl(ssl: &Handshake, bssl_result: c_int) -> Result<()> {
                 // SSL_ERROR_PENDING_TICKET
                 14 => Err(Error::Done),
 
+                //  SSL_ERROR_WANT_CERTIFICATE_VERIFY
```
ghedo, line 863: Nit:
```suggestion
                // SSL_ERROR_WANT_CERTIFICATE_VERIFY
```

ghedo (approved):
Thanks, lgtm. I'll apply the small nit manually when mergin.

ghedo (approved):

PR #881, opened 2021-03-24
Add getters for scid and dcid

ehaydenr:


LPardue:
I think the general idea is ok. What concerns me though is that connection IDs can be complicated to understand and they can change depending on the various phases of a connection or the role - the API should probably explain some of the caveats that a user might expect around this. 

It would also be nice to develop the tests more to assert on some actual values so that people can follow the logic. That means that we can't use `Pipe::default` so there's some more work to do but a lot of the existing test boilerplate can be reused,

src/lib.rs
```diff
@@ -3777,6 +3777,18 @@ impl Connection {
         self.handshake.lock().unwrap().peer_cert()
     }
 
+    /// Returns the source connection ID
```
ghedo, line 3780:
```suggestion
    /// Returns the source connection ID.
```

src/lib.rs
```diff
@@ -3777,6 +3777,18 @@ impl Connection {
...
+    #[inline]
+    pub fn source_id(&self) -> ConnectionId {
+        ConnectionId::from_ref(self.scid.as_ref())
+    }
+
+    /// Returns the destination connection ID
```
ghedo, line 3786:
```suggestion
    /// Returns the destination connection ID.
```

src/lib.rs
```diff
@@ -3777,6 +3777,18 @@ impl Connection {
...
+    #[inline]
+    pub fn dest_id(&self) -> ConnectionId {
```
ghedo, line 3788:
```suggestion
    pub fn destination_id(&self) -> ConnectionId {
```

src/lib.rs
```diff
@@ -8916,6 +8928,13 @@ mod tests {
 
         assert_eq!(pipe.advance(), Ok(()));
     }
+
+    #[test]
+    fn connection_ids() {
+        let pipe = testing::Pipe::default().unwrap();
+        assert!(pipe.client.source_id().len() > 0);
+        assert!(pipe.client.dest_id().len() > 0);
```
ghedo, line 8936:
```suggestion
        assert!(pipe.client.destination_id().len() > 0);
```

ghedo (dismissed):
Sorry for the delay. I added a few nits, but no need for you to do anything as they are minor and I can just apply them during merge (but note the change to one of the methods' name).

Also, it's worth noting that QUIC allows a connection to have multiple source/destination IDs, and while we don't support this right now in quiche, we will need to in the future for proper connection migration support, so this API may change as well.

ghedo:
I removed the test since it didn't seem particularly useful, and also added a note to the docs to warn that the connection IDs can change during the connection (to avoid potential confusion), and merged as https://github.com/cloudflare/quiche/commit/6fc099108e01ff051d22db482a13c18b711901c1. Thank you again @ehaydenr!

ehaydenr:
Thanks!

PR #890, opened 2021-04-04
cc: CUBIC - detecting spurious congestion events

junhochoi:
Implementing detecting spurious congestion events and restore to the previous CC state in CUBIC.

Current algorithm is as follows:
- When the congestion event happen, save the current CUBIC state before cwnd reduction.
- When the current recovery episode ends and the new lost count is smaller than RESTORE_COUNT_THREHOLD (unit of a packet), restore from the saved prior state. It means that this can be a random loss, not a congestion loss which usually come with many multiple losses.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -49,6 +49,8 @@ pub static CUBIC: CongestionControlOps = CongestionControlOps {
     on_packet_acked,
     congestion_event,
     collapse_cwnd,
+    checkpoint,
+    rollback,
 };
 
 /// CUBIC Constants.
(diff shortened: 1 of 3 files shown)


src/recovery/cubic.rs
```diff
@@ -162,6 +182,29 @@ fn on_packet_acked(
 
     if in_congestion_recovery {
         return;
+    } else {
+        // Detecting spurious congestion events.
+        // <https://tools.ietf.org/id/draft-ietf-tcpm-rfc8312bis-00.html#section-4.9>
```
ghedo, line 187: This doesn't actually explain how to detect the spurious event, where does the algorithm come from?

junhochoi: The draft doesn't specify a detailed algorithm how to detect it. This is my own.

lohith-bellad (approved):
LGTM

PR #893, opened 2021-04-07
recovery: add machinery for saving and rolling back congestion control state

ghedo:
This simply splits out the changes to save and restore CC state from https://github.com/cloudflare/quiche/pull/890 so e.g. it could also be used for https://github.com/cloudflare/quiche/pull/470.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -49,6 +49,8 @@ pub static CUBIC: CongestionControlOps = CongestionControlOps {
     on_packet_acked,
     congestion_event,
     collapse_cwnd,
+    checkpoint,
+    rollback,
 };
 
 /// CUBIC Constants.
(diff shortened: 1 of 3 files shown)


src/recovery/cubic.rs
```diff
@@ -646,3 +671,22 @@ mod tests {
         assert_eq!(r.cwnd(), cwnd_prev + r.max_datagram_size);
     }
 }
+
+fn checkpoint(r: &mut Recovery) {
```
junhochoi, line 675: I think we just need to have a stub here and let each PR has actual implementation. Same to rollback()

src/recovery/cubic.rs
```diff
@@ -75,6 +77,29 @@ pub struct State {
...
+
+    w_max: f64,
+
+    w_last_max: f64,
+
+    k: f64,
+
+    epoch_start: Option<Instant>,
+
+    lost_count: usize,
```
junhochoi, line 102: This part can be removed too.

src/recovery/cubic.rs
ghedo: Why? This is the generic part from draft-ietf-tcpm-rfc8312bis, why would each PR re-implement the same thing?

junhochoi (approved):

ghedo:
This was merged as part of https://github.com/cloudflare/quiche/pull/890, closing.

PR #898, opened 2021-04-14
network path awareness

ghedo:
Currently quiche does not hold any information about the network path of
a QUIC connection (such as the address of the peer), and the application
is responsible for maintaining such information.

This means that in case of a network migration, the application is
responsible for detecting the change and switching over to new path,
however there is currently no way for an application to actually
validate a new path, as required by the QUIC spec.

Adding an ad-hoc API to only expose path validation to applications
would likely be very cumbersome, due to the synchronization needed
between an application and quiche on the state of the current path, and
any other path being probed.

Instead, this change makes quiche be aware of the network path being
used.

The application needs to communicate the destination address of a
connection upon creationg (via `accept()` or `connect()`), as well as
the source address of received packets (via `recv()` and the new
`RecvInfo` structure).

In turn quiche will provide the application with the destination address
of generated packets (via `send()` and the new `SendInfo` structure).

Currently only the destination address of a connection is tracked, which
would allow quiche to handle responding to migrations transparently from
the application (but this will be added as a separate change).

Additional fields can later be added to `RecvInfo` and `SendInfo`, such
as the address of the local endpoint in order to be able to initiate
migrations, rather than just respond to them.

---

This change is quite large due to the API change, but it's mostly in tests and documentation, and it's mostly a no-op as well, as the `RecvInfo` structure is not used at all.

One nice side-effect is that even without migration support, this slightly simplifies server applications as they won't need to keep track of a client's IP address anymore (see e.g. the example servers and the quiche-server app).

TODO:
- [x] Expose via FFI
- [x] Update http3_test, fuzz, ...
- [ ] Possibly more

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -216,11 +216,14 @@ typedef struct Connection quiche_conn;
 // Creates a new server-side connection.
 quiche_conn *quiche_accept(const uint8_t *scid, size_t scid_len,
                            const uint8_t *odcid, size_t odcid_len,
+                           const struct sockaddr *from, size_t from_len,
                            quiche_config *config);
 
 // Creates a new client-side connection.
-quiche_conn *quiche_connect(const char *server_name, const uint8_t *scid,
-                            size_t scid_len, quiche_config *config);
+quiche_conn *quiche_connect(const char *server_name,
+                            const uint8_t *scid, size_t scid_len,
+                            const struct sockaddr *to, size_t to_len,
+                            quiche_config *config);
 
 // Writes a version negotiation packet.
 ssize_t quiche_negotiate_version(const uint8_t *scid, size_t scid_len,
@@ -239,6 +242,7 @@ bool quiche_version_is_supported(uint32_t version);
 
 quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
                                       const uint8_t *odcid, size_t odcid_len,
+                                      const struct sockaddr *peer, size_t peer_len,
                                       quiche_config *config, void *ssl,
                                       bool is_server);
 
@@ -259,11 +263,23 @@ void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char *log_title,
 // Configures the given session for resumption.
 int quiche_conn_set_session(quiche_conn *conn, const uint8_t *buf, size_t buf_len);
 
+typedef struct {
+    struct sockaddr *from;
+    socklen_t from_len;
+} quiche_recv_info;
+
 // Processes QUIC packets received from the peer.
-ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len);
+ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len,
+                         const quiche_recv_info *info);
+
+typedef struct {
+    struct sockaddr_storage to;
+    socklen_t to_len;
+} quiche_send_info;
 
 // Writes a single QUIC packet to be sent to the peer.
-ssize_t quiche_conn_send(quiche_conn *conn, uint8_t *out, size_t out_len);
+ssize_t quiche_conn_send(quiche_conn *conn, uint8_t *out, size_t out_len,
+                         quiche_send_info *out_info);
 
 // Reads contiguous data from a stream.
 ssize_t quiche_conn_stream_recv(quiche_conn *conn, uint64_t stream_id,
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -39,11 +39,37 @@ use libc::c_int;
 use libc::c_void;
 use libc::size_t;
 use libc::sockaddr;
-use libc::sockaddr_in;
-use libc::sockaddr_in6;
-use libc::sockaddr_storage;
-use libc::socklen_t;
 use libc::ssize_t;
+
+#[cfg(not(windows))]
+use libc::sockaddr_in;
+#[cfg(windows)]
+use winapi::shared::ws2def::SOCKADDR_IN as sockaddr_in;
+
+#[cfg(not(windows))]
+use libc::sockaddr_in6;
+#[cfg(windows)]
+use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH as sockaddr_in6;
+
+#[cfg(not(windows))]
+use libc::sockaddr_storage;
+#[cfg(windows)]
+use winapi::shared::ws2def::SOCKADDR_STORAGE_LH as sockaddr_storage;
+
+#[cfg(windows)]
+use libc::c_int as socklen_t;
+#[cfg(not(windows))]
+use libc::socklen_t;
+
+#[cfg(not(windows))]
+use libc::AF_INET;
+#[cfg(windows)]
+use winapi::shared::ws2def::AF_INET;
+
+#[cfg(not(windows))]
+use libc::AF_INET6;
+#[cfg(windows)]
+use winapi::shared::ws2def::AF_INET6;
 
 use crate::*;
 
@@ -1018,7 +1044,7 @@ pub extern fn quiche_conn_peer_streams_left_uni(conn: &mut Connection) -> u64 {
 fn std_addr_from_c(addr: &sockaddr, addr_len: socklen_t) -> SocketAddr {
     unsafe {
         match addr.sa_family as i32 {
-            libc::AF_INET => {
+            AF_INET => {
                 assert!(addr_len as usize == std::mem::size_of::<sockaddr_in>());
 
                 SocketAddr::V4(
@@ -1026,7 +1052,7 @@ fn std_addr_from_c(addr: &sockaddr, addr_len: socklen_t) -> SocketAddr {
                 )
             },
 
-            libc::AF_INET6 => {
+            AF_INET6 => {
                 assert!(addr_len as usize == std::mem::size_of::<sockaddr_in6>());
 
                 SocketAddr::V6(
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -29,6 +29,8 @@ use std::ptr;
 use std::slice;
 use std::sync::atomic;
 
+use std::net::SocketAddr;
+
 #[cfg(unix)]
 use std::os::unix::io::FromRawFd;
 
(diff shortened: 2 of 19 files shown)


ghedo:
Assuming this still builds, it should be ready for review now, at least as far as what it set out to do, which is not much, as the path information is not actually used for anything right, so this mostly just changes the API. The FFI integration is kind of delicate, due to the conversion between Rust <-> C address structures, so particular care should be placed there. Also need to verify that all Rust and C examples still work.

src/ffi.rs
```diff
@@ -848,12 +901,12 @@ pub extern fn quiche_stream_iter_free(iter: *mut StreamIter) {
...
-    pub lost: usize,
-    pub rtt: u64,
-    pub cwnd: usize,
-    pub delivery_rate: u64,
+    recv: usize,
+    sent: usize,
+    lost: usize,
+    rtt: u64,
+    cwnd: usize,
+    delivery_rate: u64,
```
LPardue, line 909: why is this required in this commit?

src/ffi.rs
```diff
@@ -948,3 +1001,57 @@ pub extern fn quiche_conn_peer_streams_left_bidi(conn: &mut Connection) -> u64 {
...
+
+            libc::AF_INET6 => {
+                assert!(addr_len as usize == std::mem::size_of::<sockaddr_in6>());
+
+                SocketAddr::V6(
+                    *(addr as *const _ as *const sockaddr_in6 as *const _),
+                )
+            },
+
+            _ => unimplemented!("unsupport address type"),
```
LPardue, line 1024:
```suggestion
            _ => unimplemented!("unsupported address type"),
```

src/ffi.rs
```diff
@@ -948,3 +1001,57 @@ pub extern fn quiche_conn_peer_streams_left_bidi(conn: &mut Connection) -> u64 {
...
+        }
+    }
+}
+
+fn std_addr_to_c(addr: &SocketAddr) -> (sockaddr_storage, socklen_t) {
```
LPardue, line 1029: seems to be a bit of repition here, could I guess you need the destructuring to get to the actual `addr`, so any suggestion came up with doesn't really help much.

src/lib.rs
```diff
@@ -448,6 +478,20 @@ impl std::convert::From<octets::BufferTooShortError> for Error {
     }
 }
 
+/// Ancillary information about infoming packets.
```
LPardue, line 481:
```suggestion
/// Ancillary information about incoming packets.
```

LPardue (approved):
code LGTM.

As long as quiche internals don't depend on doing anything with the socket info I'm happy. I think our unit tests that use local memory buffers enforces that.

src/ffi.rs
```diff
@@ -552,26 +573,54 @@ pub extern fn quiche_conn_set_session(
     }
 }
 
+#[repr(C)]
+pub struct RecvInfo<'a> {
+    from: &'a sockaddr,
```
junhochoi, line 578: In `SendInfo`, `sockaddr_storage` is used but here `sockaddr` is used. Any reason for that?

src/ffi.rs
ghedo: It's not, really, I just noticed this struct wasn't consistent with the others, and I thought it doesn't really warrant its own PR.

src/ffi.rs
ghedo: The main difference is that `SendInfo` owns the address value, so we need to hold the needed storage in it, while `RecvInfo` contains only a reference/pointer. In practice I don't think it makes any difference, but it seemed that using `struct sockaddr *` instead of `struct sockaddr_storage *` was better on the C side, so this is just consistent with that.

src/ffi.rs
ghedo: Yeah, it might be possible to make a macro or something, but every time I try to Rust macros I get confused, so I haven't tried that yet.

normanmaurer:
I will do a POC on top of the provided c API and report back this week.

junhochoi (approved):

ghedo:
Rebased again, and added a change to avoid double-copying the address when creating the FFI `SendInfo` structure.

junhochoi (approved):

include/quiche.h
```diff
@@ -216,11 +216,13 @@ typedef struct Connection quiche_conn;
 // Creates a new server-side connection.
 quiche_conn *quiche_accept(const uint8_t *scid, size_t scid_len,
                            const uint8_t *odcid, size_t odcid_len,
+                           struct sockaddr *from, size_t from_len,
```
normanmaurer, line 219: I think this should be:

```
const struct sockaddr *from
```

include/quiche.h
```diff
@@ -216,11 +216,13 @@ typedef struct Connection quiche_conn;
...
                            quiche_config *config);
 
 // Creates a new client-side connection.
 quiche_conn *quiche_connect(const char *server_name, const uint8_t *scid,
-                            size_t scid_len, quiche_config *config);
+                            size_t scid_len, struct sockaddr *to, size_t to_len,
```
normanmaurer, line 224: I think this should be:

```
const struct sockaddr *to
```

include/quiche.h
```diff
@@ -239,6 +241,7 @@ bool quiche_version_is_supported(uint32_t version);
 
 quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
                                       const uint8_t *odcid, size_t odcid_len,
+                                      struct sockaddr *peer, size_t peer_len,
```
normanmaurer, line 244: I think this should be:

```
const struct sockaddr *peer
```

include/quiche.h
```diff
@@ -259,11 +262,23 @@ void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char *log_title,
...
 
+typedef struct {
+    struct sockaddr *from;
+    socklen_t from_len;
+} quiche_recv_info;
+
 // Processes QUIC packets received from the peer.
-ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len);
+ssize_t quiche_conn_recv(quiche_conn *conn, uint8_t *buf, size_t buf_len,
+                         quiche_recv_info *info);
```
normanmaurer, line 272: const quiche_recv_info *info

normanmaurer (changes requested):
First round of review when trying to integrate this

normanmaurer:
As promised I did spent some time today to try to use it in netty. I happy to report back that all works as expected. So :shipit: 

While the code needs still some love on my side you may still be interested: https://github.com/netty/netty-incubator-codec-quic/pull/291

Special thanks to @ghedo for answering all my questions.

ghedo:
Thanks @normanmaurer!

normanmaurer:
@ghedo I just noticed that there may be some stuff missing for windows:

```
     [exec] error[E0432]: unresolved import `libc::sockaddr_in`
15793
     [exec]   --> src\ffi.rs:42:5
15794
     [exec]    |
15795
     [exec] 42 | use libc::sockaddr_in;
15796
     [exec]    |     ^^^^^^-----------
15797
     [exec]    |     |     |
15798
     [exec]    |     |     help: a similar name exists in the module: `sockaddr`
15799
     [exec]    |     no `sockaddr_in` in the root
15800
     [exec] 
15801
     [exec] error[E0432]: unresolved import `libc::sockaddr_in6`
15802
     [exec]   --> src\ffi.rs:43:5
15803
     [exec]    |
15804
     [exec] 43 | use libc::sockaddr_in6;
15805
     [exec]    |     ^^^^^^------------
15806
     [exec]    |     |     |
15807
     [exec]    |     |     help: a similar name exists in the module: `sockaddr`
15808
     [exec]    |     no `sockaddr_in6` in the root
15809
     [exec] 
15810
     [exec] error[E0432]: unresolved import `libc::sockaddr_storage`
15811
     [exec]   --> src\ffi.rs:44:5
15812
     [exec]    |
15813
     [exec] 44 | use libc::sockaddr_storage;
15814
     [exec]    |     ^^^^^^^^^^^^^^^^^^^^^^ no `sockaddr_storage` in the root
15815
     [exec] 
15816
     [exec] error[E0432]: unresolved import `libc::socklen_t`
15817
     [exec]   --> src\ffi.rs:45:5
15818
     [exec]    |
15819
     [exec] 45 | use libc::socklen_t;
15820
     [exec]    |     ^^^^^^---------
15821
     [exec]    |     |     |
15822
     [exec]    |     |     help: a similar name exists in the module: `socket`
15823
     [exec]    |     no `socklen_t` in the root
15824
     [exec] 
15825
     [exec] error[E0531]: cannot find unit struct, unit variant or constant `AF_INET` in crate `libc`
15826
     [exec]     --> src\ffi.rs:1021:19
15827
     [exec]      |
15828
     [exec] 1021 |             libc::AF_INET => {
15829
     [exec]      |                   ^^^^^^^ not found in `libc`
15830
     [exec]      |
15831
     [exec] help: consider importing one of these items
15832
     [exec]      |
15833
     [exec] 27   | use crate::tls::ffi::c_str::sys::c::AF_INET;
15834
     [exec]      |
15835
     [exec] 27   | use winapi::shared::ws2def::AF_INET;
15836
     [exec]      |
15837
     [exec] 
15838
     [exec] error[E0531]: cannot find unit struct, unit variant or constant `AF_INET6` in crate `libc`
15839
     [exec]     --> src\ffi.rs:1029:19
15840
     [exec]      |
15841
     [exec] 1029 |             libc::AF_INET6 => {
15842
     [exec]      |                   ^^^^^^^^ not found in `libc`
15843
     [exec]      |
15844
     [exec] help: consider importing one of these items
15845
     [exec]      |
15846
     [exec] 27   | use crate::tls::ffi::c_str::sys::c::AF_INET6;
15847
     [exec]      |
15848
     [exec] 27   | use winapi::shared::ws2def::AF_INET6;
15849
     [exec]      |
15850
     [exec] 
15851
     [exec] error: aborting due to 6 previous errors
15852
     [exec] 
15853
     [exec] Some errors have detailed explanations: E0432, E0531.
15854
     [exec] For more information about an error, try `rustc --explain E0432`.
15855
     [exec] error: could not compile `quiche`
15856
     [exec] 
15857
     [exec] To learn more, run the command again with --verbose.
```

ghedo:
I pushed a change to fix build on Windows... don't know if it actually works though TBH, as I don't have a machine to test this on. Also rebased again, after merging https://github.com/cloudflare/quiche/pull/938, so the `ffi` feature is actually enabled on all builds (including Windows).

normanmaurer:
@ghedo seems like it works ... at least the tests pass on windows 

https://github.com/netty/netty-incubator-codec-quic/pull/291/checks?check_run_id=2640610379

PR #906, opened 2021-04-19
disable the TLS extension for QUIC v1 on the server when downgrading

ghedo:
When a pre-v1 client connects to a v1-enabled server, we need to
re-enable the legacy TLS extension codepoint on the server-side due to
the downgrade.


junhochoi (approved):
🚢

LPardue (approved):

PR #909, opened 2021-04-26
buffer undecryptable 0-RTT packets

ghedo:
Currently, incoming packets that can't be decrypted because the
corresponding read keys are not available yet are simply dropped. On the
server side this can happen with 0-RTT packets, for example if an early
handshake callback configured in BoringSSL delays installation of read
keys.

This change implements buffering a limited number of 0-RTT packets
received before the corresponding read keys are installed, for a short
period, so that they can be correctly processed at a later time. The
buffer is limited to 10 packets (no particular reason for this number,
though it is also what Chrome uses), and it should be drained as soon as
keys become available.

It can also happen that the server ends up rejecting 0-RTT anyway, in
which case the buffer is cleared as soon as the handshake is completed.
In theory this could happen at an earlier time during the handshake, but
BoringSSL does not (yet?) expose a signal for this.

This could, in theory, also be used to support the case where 0-RTT
packets are received before an Initial one. However it is up to the
application to decide whether a connection should be accept()ed even if
no valid Initial packet has been received yet, as well as dropping the
connection object after a certain amount of time in case the handshake
doesn't complete, as the idle timer is not armed if no Initial packet is
received.

---

As you may have noticed there are no tests here, because we don't actually support sending 0-RTT packets yet...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -275,6 +275,8 @@ use std::str::FromStr;
 
 use std::sync::Mutex;
 
+use std::collections::VecDeque;
+
 /// The current QUIC wire version.
 pub const PROTOCOL_VERSION: u32 = PROTOCOL_VERSION_V1;
 
@@ -323,6 +325,9 @@ const MAX_DGRAM_FRAME_SIZE: u64 = 65536;
 
 // The length of the payload length field.
 const PAYLOAD_LENGTH_LEN: usize = 2;
+
+// The number of undecryptable that can be buffered.
+const MAX_UNDECRYPTABLE_PACKETS: usize = 10;
 
 /// A specialized [`Result`] type for quiche operations.
 ///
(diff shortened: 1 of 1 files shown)


LPardue (dismissed):

junhochoi (approved):

ghedo:
It turns out this didn't actually work, though I was able to create a test using the changes from https://github.com/cloudflare/quiche/pull/911, so this should work now I think. The test is based on the 0-RTT<->Initial re-ordering case (i.e. receiving 0-RTT before Initial), but I think it should also work in the early callback delay case too.

junhochoi:
> It turns out this didn't actually work, though I was able to create a test using the changes from #911, so this should work now I think. The test is based on the 0-RTT<->Initial re-ordering case (i.e. receiving 0-RTT before Initial), but I think it should also work in the early callback delay case too.

Can you add a test case where initial first and 0-rtt comes after? this is actual case I am seeing.

ghedo:
@junhochoi added another test, not sure if that's what you meant though.

junhochoi (approved):
LGTM. Thanks!

PR #913, opened 2021-04-27
qns: fix client test run

ghedo:
In 60f66cac, run_quiche_client_tests was changed to do "case $1", like
check_testcase, but the testcase was not passed to the function.

Also remove unused variable.


LPardue (approved):
:heavy_dollar_sign:

PR #914, opened 2021-04-27
support for sending 0-RTT packets on the client

### LINKED ISSUE #683: PSK / 0-RTT supported ?
Hi there,

I am currently investigate in using quiche within [netty](https://netty.io) by using the C API via JNI. Once thing I was wondering while looking at this was if PSK / 0-RTT is supported at all. I couldn't spot it in the C API. So I was wondering if it is not supported at all or maybe just not exposed yet via C. I tried to check the rust source but my rust skills are still very weak unfortunately so I may have missed it :/

ghedo:
This adds support for clients to generate 0-RTT packets. Currently only
STREAM and DATAGRAM frames are allowed, though this is somewhat stricter
than what is allowed in theory.

This requies exporting the peer's transport parameters as part of the
TLS session, so that they can be restored on resumption.

This also changes how epoch/packet type is picked when creating a new
packet. Packet types offer more granularity (e.g. Short vs. ZeroRTT), so
it's better to pick the type first and then derive the epoch, rather
than the other way around.

Closes #683.

---

It seems to work reasonably fine so far, but probably needs more interop testing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1496,7 +1496,25 @@ impl Connection {
     /// [`session()`]: struct.Connection.html#method.session
     #[inline]
     pub fn set_session(&mut self, session: &[u8]) -> Result<()> {
-        self.handshake.lock().unwrap().set_session(session)
+        let mut b = octets::Octets::with_slice(session);
+
+        let session_len = b.get_u64()? as usize;
+        let session_bytes = b.get_bytes(session_len)?;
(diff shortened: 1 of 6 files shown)


ghedo:
Rebased on top of https://github.com/cloudflare/quiche/pull/915.

junhochoi (approved):

PR #919, opened 2021-05-04
Expose C API that is needed for session resumption and 0-RTT

normanmaurer:
Motivation:

Quiche added support for session resumption and 0-RTT lately but the API was not exposed via the C api.

This was done in https://github.com/cloudflare/quiche/pull/911 and https://github.com/cloudflare/quiche/pull/914.

Modifications:

Add C API to also be able to make use of the functionality via C.

Result:

Be able to use session resumption and 0-RTT via the C api

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -314,6 +314,13 @@ void quiche_conn_on_timeout(quiche_conn *conn);
 int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
                       const uint8_t *reason, size_t reason_len);
 
+// Configures the given session for resumption.
+int quiche_conn_set_session(quiche_conn *conn, const uint8_t *buf, size_t buf_len);
+
+// Returns the serialized cryptographic session for the connection..
+void quiche_conn_get_session(quiche_conn *conn, const uint8_t **out,
+                                   size_t *out_len);
+
 // Returns a string uniquely representing the connection.
 void quiche_conn_trace_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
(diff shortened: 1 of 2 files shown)


include/quiche.h
```diff
@@ -314,6 +314,13 @@ void quiche_conn_on_timeout(quiche_conn *conn);
 int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
                       const uint8_t *reason, size_t reason_len);
 
+// Configures the given session for resumption.
+int quiche_conn_set_session(quiche_conn *conn, const uint8_t *buf, size_t buf_len);
+
+// Returns the serialized cryptographic session for the connection..
+void quiche_conn_get_session(quiche_conn *conn, const uint8_t **out,
```
ghedo, line 321: Should follow the naming of the Rust function here IMO, i.e. `quiche_conn_session()`.

include/quiche.h
```diff
@@ -314,6 +314,13 @@ void quiche_conn_on_timeout(quiche_conn *conn);
...
+                                   size_t *out_len);
```
ghedo, line 322: Nit: should align this with the `quiche_conn`.

include/quiche.h
ghedo: Nit: Also, we try to order functions here like in `lib.rs`, so both of these should be moved around (same applies `ffi.rs`).

ghedo (approved):
Looks good overall, just added a couple of minor comments, though I can fix these myself during merge.

ghedo:
Merged in 0a08fd78134dc35351ce773f143015ed12f53ae1, thank you @normanmaurer!

include/quiche.h
normanmaurer: > Should follow the naming of the Rust function here IMO, i.e. quiche_conn_session().
No strong preference here... I just thought as we have a `set` version adding `get` makes it more clear what this is all about.

PR #920, opened 2021-05-06
Validate that a valid version is specified when create a Config

normanmaurer:
Motivation:

We should validate that the version is valid that was used to create the Config object and if not fail.

Modifications:

- Add validation for the version
- Add some tests

Result:

Version is validated during config creation

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -499,6 +499,15 @@ pub struct Config {
     max_send_udp_payload_size: usize,
 }
 
+// See https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-15
+fn is_reserved_version(version: u32) -> bool {
+    version == 0x00000000 ||
+        version
+            .to_le_bytes()
+            .iter()
+            .all(|b| (b & 0b0000_1111) == 0b0000_1010)
+}
+
 impl Config {
     /// Creates a config object with the given version.
     ///
@@ -509,6 +518,9 @@ impl Config {
     /// # Ok::<(), quiche::Error>(())
     /// ```
     pub fn new(version: u32) -> Result<Config> {
+        if !is_reserved_version(version) && !version_is_supported(version) {
+            return Err(Error::UnknownVersion);
+        }
         let tls_ctx = Mutex::new(tls::Context::new()?);
 
         Ok(Config {
@@ -5594,6 +5606,20 @@ mod tests {
 
         let mut pipe = testing::Pipe::with_client_config(&mut config).unwrap();
         assert_eq!(pipe.handshake(), Err(Error::UnknownVersion));
+    }
+
+    #[test]
+    fn config_version_reserved() {
+        Config::new(0xbabababa).unwrap();
+        Config::new(0x00000000).unwrap();
+    }
+
+    #[test]
+    fn config_version_invalid() {
+        assert_eq!(
+            Config::new(0xb1bababa).err().unwrap(),
+            Error::UnknownVersion
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -5585,6 +5594,17 @@ mod tests {
...
 
+    #[test]
+    fn config_version_reserved() {
+        Config::new(0xbabababa).unwrap();
+        Config::new(0x00000000).unwrap();
+    }
+
+    #[test]
+    fn config_version_invalid() {
+        assert_eq!(Config::new(0xb1bababa).err().unwrap(), Error::UnknownVersion);
```
normanmaurer, line 5605: used `err().unwrap()` as the `Config` doesn't declare `#[derive(Debug)]`. Not sure if there is a reason for this. If not we could do so and use `unwrap_err()`

normanmaurer:
The build failure is odd as I didn't change anything related to these files... @ghedo any idea ?

ghedo:
@normanmaurer most of the failures are because of the new Rust release, see https://github.com/cloudflare/quiche/pull/921, though the formatting lint failure is because of the changes in this PR https://github.com/cloudflare/quiche/pull/920/checks?check_run_id=2520952393 (need to run `cargo +nightly fmt` with latest Rust+rustfmt nightly to fix).

normanmaurer:
Will fix tomorrow

src/lib.rs
```diff
@@ -499,6 +499,15 @@ pub struct Config {
     max_send_udp_payload_size: usize,
 }
 
+// See https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-15
+fn is_reserved_version(version: u32) -> bool {
+    version == 0x00000000 ||
```
ghedo, line 504: The spec is a bit terse, but I think what it's saying is that `0x00000000` is reserved only for use within [version negotiation packets](https://quicwg.org/base-drafts/rfc9000.html#name-version-negotiation-packet), so I don't think we should allow users to configure this version.

src/lib.rs
```diff
@@ -509,6 +518,9 @@ impl Config {
     /// # Ok::<(), quiche::Error>(())
     /// ```
     pub fn new(version: u32) -> Result<Config> {
+        if !is_reserved_version(version) && !version_is_supported(version) {
+            return Err(Error::UnknownVersion);
+        }
```
ghedo, line 523: Nit:
```suggestion
        }

```

src/lib.rs
```diff
@@ -499,6 +499,15 @@ pub struct Config {
     max_send_udp_payload_size: usize,
 }
 
+// See https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-15
+fn is_reserved_version(version: u32) -> bool {
+    version == 0x00000000 ||
+        version
+            .to_le_bytes()
+            .iter()
+            .all(|b| (b & 0b0000_1111) == 0b0000_1010)
```
ghedo, line 508: I _think_ this whole thing could be simplified to just:
```rust
    version & 0xfafafafa == version
```
?

The `0xfafafafa` can then also be declared as a named const, so we don't hardcode it, e.g.
```rust
const RESERVED_VERSION_MASK: u32 = 0xfafafafa
```
or something like that.

src/lib.rs
```diff
@@ -499,6 +499,15 @@ pub struct Config {
     max_send_udp_payload_size: usize,
 }
 
+// See https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-15
```
ghedo, line 502:
```suggestion
// See https://quicwg.org/base-drafts/rfc9000.html#section-15
```
:D

ghedo (changes requested):
Thanks @normanmaurer, added a few comments.

normanmaurer:
@ghedo addressed... thanks!

ghedo (approved):

ghedo:
Merged, thanks @normanmaurer!

PR #926, opened 2021-05-11
do not try coalesce Initial and Short

junhochoi:
For Initial packets we need to make the UDP datagram at least 1200
bytes following the transport spec. However when the last packet
in the datagram is Short and the datagram is not fully filled, we try
to do null padding at the end which causes a decryption failure on
the server because Short packet doesn't have an explicit length.

This fix is try to avoid coalesce Initial and Short packet in the
same datagram. It will cause 2 different datagrams so may be
a little inefficient but it only happens when we send Initial and
a small Short packet (mostly, only in the beginning).


junhochoi:
> It seems like this just kind of defeats the purpose of coalescing packets in the first place, particularly on the server which shouldn't be affected by this anyway.
> 
> I think we should instead fallback to padding using frames in this specific case, or maybe just not coalesce 1-RTT packets on the client side, though I don't know whether there is some other edge-case where this would be required by the server, so the padding fallback is probably the safer option.

I added a check to limit this behavior to the client. PADDING looks like a little tricky (we need to add before Short), I'll take a look.

ghedo:
@junhochoi one option could be to pass a "minimum packet length" when an initial packet has been previous created from `send()` into `send_single()` which can then decide whether to add padding to the `Short` packet or not after it generated all the frames.

PR #929, opened 2021-05-11
h3: make quiche_h3_recv_dgram signature consistent

LPardue:
Seems we missed the header file...


junhochoi (approved):
I think we need to add datagram example in C to catch this type of error in CI.

PR #930, opened 2021-05-11
Add PADDING frame to fill up the datagram with Initial packet

junhochoi:
For Initial packets we need to make the UDP datagram at least 1200
bytes following the transport spec. However when the last packet
in the datagram is Short and the datagram is not fully filled, we try
to do null padding at the end which causes a decryption failure on
the server because Short packet doesn't have an explicit length.

To avoid this issue, when Short packet is added in the same
datagram with Initial packet, add a PADDING frame to Short packet
to fill up the datagram size. If Short packet is not added,
we still add zero padding at the end of packet.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2286,14 +2286,15 @@ impl Connection {
 
         // Generate coalesced packets.
         while left > 0 {
-            let (ty, written) =
-                match self.send_single(&mut out[done..done + left]) {
-                    Ok(v) => v,
+            let (ty, written) = match self
+                .send_single(&mut out[done..done + left], has_initial)
+            {
+                Ok(v) => v,
 
-                    Err(Error::BufferTooShort) | Err(Error::Done) => break,
+                Err(Error::BufferTooShort) | Err(Error::Done) => break,
 
-                    Err(e) => return Err(e),
-                };
+                Err(e) => return Err(e),
+            };
 
             done += written;
             left -= written;
@@ -2334,7 +2335,9 @@ impl Connection {
         Ok(done)
     }
 
-    fn send_single(&mut self, out: &mut [u8]) -> Result<(packet::Type, usize)> {
+    fn send_single(
+        &mut self, out: &mut [u8], has_initial: bool,
+    ) -> Result<(packet::Type, usize)> {
         let now = time::Instant::now();
 
         if out.is_empty() {
@@ -2974,6 +2977,19 @@ impl Connection {
             // app_limited to false.
             self.recovery.update_app_limited(false);
             return Err(Error::Done);
+        }
+
+        // When coalescing a 1-RTT packet, we can't add padding in the UDP
+        // datagram, so use PADDING frames instead.
+        //
+        // This is only needed if an Initial packet has already been written to
+        // the UDP datagram, as Initial always requires padding.
+        if has_initial && pkt_type == packet::Type::Short && left >= 1 {
+            let frame = frame::Frame::Padding { len: left };
+
+            if push_frame_to_pkt!(b, frames, frame, left) {
+                in_flight = true;
+            }
         }
 
         // Pad payload so that it's always at least 4 bytes.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2964,6 +2976,17 @@ impl Connection {
             }
         }
 
+        // Create PADDING frame for the rest of the packet.
+        // PADDING frame can't be a single frame in the packet,
+        // so only add it when there is a frame already.
+        if add_padding && left >= 1 && !frames.is_empty() {
```
ghedo, line 2982: Should probably move this a few lines down after the existing `frames.is_empty()` check, so we don't have to duplicate it.

src/lib.rs
```diff
@@ -2286,14 +2287,23 @@ impl Connection {
...
+            // When there is Initial packet already and try to add
+            // Short packet in the same datagram, add PADDING frame.
+            if has_initial {
+                if let Ok(packet::Type::Short) = self.write_pkt_type() {
+                    add_padding = true;
+                }
+            }
+
+            let (ty, written) = match self
+                .send_single(&mut out[done..done + left], add_padding)
```
ghedo, line 2299: It might be better to simply pass the `has_initial` value to `send_single()` so that we can unify the logic in one place inside it, rather than split in two parts, something like:
```rust
if has_initial && pkt_type == packet::Type::Short && left > 1 ...
```
this also means we wouldn't have to duplicate the `write_pkt_type()` call in `send()`, which seems somewhat "cleaner".

ghedo (changes requested):
Added a couple of comments, but looks good overall:

Should also add a test though. The following seems to work:
```rust
    #[test]
    fn coalesce_padding_short() {
        let mut buf = [0; 65535];

        let mut pipe = testing::Pipe::default().unwrap();

        // Client sends first flight.
        let len = pipe.client.send(&mut buf).unwrap();
        assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
        assert_eq!(pipe.server.recv(&mut buf[..len]), Ok(len));

        // Server sends first flight.
        let len = pipe.server.send(&mut buf).unwrap();
        assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
        assert_eq!(pipe.client.recv(&mut buf[..len]), Ok(len));

        let len = pipe.server.send(&mut buf).unwrap();
        assert_eq!(pipe.client.recv(&mut buf[..len]), Ok(len));

        // Client sends stream data.
        assert_eq!(pipe.client.is_established(), true);
        assert_eq!(pipe.client.stream_send(4, b"hello", true), Ok(5));

        // Client sends second flight.
        let len = pipe.client.send(&mut buf).unwrap();
        assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
        assert_eq!(pipe.server.recv(&mut buf[..len]), Ok(len));

        // None of the sent packets should have been dropped.
        assert_eq!(pipe.client.sent_count, pipe.server.recv_count);
        assert_eq!(pipe.server.sent_count, pipe.client.recv_count);
    }
```

src/lib.rs
junhochoi: Moved.

src/lib.rs
junhochoi: Updated.

junhochoi:
I added the test you suggested and updated based on your feedback. Thanks!

junhochoi:
> Added a couple of comments, but looks good overall:
> 
> Should also add a test though. The following seems to work:
> 
> ```rust
>     #[test]
>     fn coalesce_padding_short() {
>         let mut buf = [0; 65535];
> 
>         let mut pipe = testing::Pipe::default().unwrap();
> 
>         // Client sends first flight.
>         let len = pipe.client.send(&mut buf).unwrap();
>         assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
>         assert_eq!(pipe.server.recv(&mut buf[..len]), Ok(len));
> 
>         // Server sends first flight.
>         let len = pipe.server.send(&mut buf).unwrap();
>         assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
>         assert_eq!(pipe.client.recv(&mut buf[..len]), Ok(len));
> 
>         let len = pipe.server.send(&mut buf).unwrap();
>         assert_eq!(pipe.client.recv(&mut buf[..len]), Ok(len));
> 
>         // Client sends stream data.
>         assert_eq!(pipe.client.is_established(), true);
>         assert_eq!(pipe.client.stream_send(4, b"hello", true), Ok(5));
> 
>         // Client sends second flight.
>         let len = pipe.client.send(&mut buf).unwrap();
>         assert_eq!(len, MIN_CLIENT_INITIAL_LEN);
>         assert_eq!(pipe.server.recv(&mut buf[..len]), Ok(len));
> 
>         // None of the sent packets should have been dropped.
>         assert_eq!(pipe.client.sent_count, pipe.server.recv_count);
>         assert_eq!(pipe.server.sent_count, pipe.client.recv_count);
>     }
> ```

Added this test and updated based on your feedback. Thanks!

ghedo (approved):

ghedo:
Thanks, I changed the comment in the code to add more details, but lgtm otherwise, so merged.

PR #933, opened 2021-05-12
Expose connection source / destination id via c API

normanmaurer:
Motivation:

We should expose functions to access the source and destination ids via the c API as well

Modifications:

Expose functions in the C API

Result:

Be able to access the ids of a connection

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -320,6 +320,12 @@ int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
 // Returns a string uniquely representing the connection.
 void quiche_conn_trace_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
+// Returns the source connection id.
+void quiche_conn_source_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
+
+// Returns the destination connection id.
+void quiche_conn_destination_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
+
 // Returns the negotiated ALPN protocol.
 void quiche_conn_application_proto(quiche_conn *conn, const uint8_t **out,
                                    size_t *out_len);
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -323,7 +323,7 @@ void quiche_conn_trace_id(quiche_conn *conn, const uint8_t **out, size_t *out_le
 // Returns the source connection ID.
 void quiche_conn_source_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
-// Returns the destination connection id.
+// Returns the destination connection ID.
 void quiche_conn_destination_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
 // Returns the negotiated ALPN protocol.
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -320,7 +320,7 @@ int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
 // Returns a string uniquely representing the connection.
 void quiche_conn_trace_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
-// Returns the source connection id.
+// Returns the source connection ID.
 void quiche_conn_source_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
 // Returns the destination connection id.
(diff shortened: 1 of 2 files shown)


normanmaurer:
@ghedo the test-failure is not related to the change: https://github.com/cloudflare/quiche/pull/933/checks?check_run_id=2569169423

src/ffi.rs
```diff
@@ -766,6 +766,26 @@ pub extern fn quiche_conn_trace_id(
     *out_len = trace_id.len();
 }
 
+#[no_mangle]
+pub extern fn quiche_conn_source_id(
+    conn: &mut Connection, out: &mut *const u8, out_len: &mut size_t,
+) {
+    let id = conn.source_id().to_vec();
```
ghedo, line 773: This value is dropped once the function returns, so I think this would end up returning a pointer to freed memory to the application.

Instead of doing `to_vec()` (which allocates a new value withing this function), I think `as_ref()` would be better, so we would use a pointer to the actual value held by the connection object (which currently has the same lifetime as the connection itself), rather than a newly allocated but short-lived value.

But probably worth confirming with valgrind in an application that uses these new functions.

src/ffi.rs
```diff
@@ -766,6 +766,26 @@ pub extern fn quiche_conn_trace_id(
...
+
+    *out = id.as_ptr();
+    *out_len = id.len();
+}
+
+#[no_mangle]
+pub extern fn quiche_conn_destination_id(
+    conn: &mut Connection, out: &mut *const u8, out_len: &mut size_t,
+) {
+    let id = conn.destination_id().to_vec();
```
ghedo, line 783: Same here.

ghedo (changes requested):

src/ffi.rs
normanmaurer: Sounds about right ... let me change it

normanmaurer:
@ghedo ok this not worked … will investigate tomorrow

normanmaurer:
@ghedo I think this should be ready to go now. PTAL!

src/ffi.rs
```diff
@@ -766,6 +766,27 @@ pub extern fn quiche_conn_trace_id(
     *out_len = trace_id.len();
 }
 
+#[no_mangle]
+pub extern fn quiche_conn_source_id(
+    conn: &mut Connection, out: &mut *const u8, out_len: &mut size_t,
+) {
+    let conn_id = conn.source_id();
+    let id = conn_id.as_ref();
+    *out = id.as_ptr();
```
junhochoi, line 775: Since the maximum size is already known (max cid len is 20 bytes in quic-transport spec). I think it's simpler for caller to provide 20 bytes buffer and this function copy the cid value into the provided buffer and return the actual cid len. Also who is responsible to free this pointer?

normanmaurer: I just thought it's a good idea to keep it the same as other functions in terms of arguments ... Also noone need to free anything as we just grab the memory which is wrapped in the id that should have the same lifetime as the connection

normanmaurer: @ghedo thoughts ?

include/quiche.h
```diff
@@ -320,6 +320,12 @@ int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
 // Returns a string uniquely representing the connection.
 void quiche_conn_trace_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
 
+// Returns the source connection id.
```
LPardue, line 323:
```suggestion
// Returns the source connection ID.
```

include/quiche.h
```diff
@@ -320,6 +320,12 @@ int quiche_conn_close(quiche_conn *conn, bool app, uint64_t err,
...
+void quiche_conn_source_id(quiche_conn *conn, const uint8_t **out, size_t *out_len);
+
+// Returns the destination connection id.
```
LPardue, line 326:
```suggestion
// Returns the destination connection ID.
```

LPardue (commented):
nits :D

normanmaurer:
@LPardue thanks... fixed :)

ghedo (approved):

normanmaurer:
@ghedo <3 Thanks a lot!

src/ffi.rs
ghedo: I think if possible we should try to follow what the Rust API does, so the current approach looks fine to me, but we can change later if we see copying is a better option.

PR #935, opened 2021-05-17
Draft: Implement BBR

### LINKED ISSUE #514: Support BBR congestion control
I use configurable congestion control to implement BBR congestion control. It is doable with some caveats.

1. Need an accurate pacing implementation. I tried to implement a pacing module on the C part using libev/libuv timer but the accuracy of the timer is not good enough. I also tried to use `set_socket_opt` to leverage kernel pacing but it can not support multiple pacing rates on same socket. I read your [blog](https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/). Is there any update on the pacing implementation?
2. I implemented the BBR congestion control by porting the BBR implementation in [lsquic](https://github.com/litespeedtech/lsquic) into the Rust part. However, their data structure is different from QUICHE's. It involves a lot of structs translation and duplicate. I notice that some parts of the BBR code such as delivery rate estimation(#375) and windowed min-max filter over (#398) is ported. Are you actively working on the BBR implementation?

rafibaum:
I've recently been working on a version of BBR for quiche for a research project. It's broadly based on information found in the [draft BBR RFC](https://datatracker.ietf.org/doc/html/draft-cardwell-iccrg-bbr-congestion-control), the [Linux kernel implementation](https://github.com/torvalds/linux/blob/master/net/ipv4/tcp_bbr.c), and [Google's implementation in their (similarly named) QUIC library](https://source.chromium.org/chromium/chromium/src/+/master:net/third_party/quiche/src/quic/core/congestion_control/bbr_sender.cc).

This version of BBR implements the core BBR algorithm, the packet conservation loss handling mechanism, and the long-term sampling functionality which so far has only been implemented in the Linux kernel implementation and isn't detailed in the RFC or the QUIC implementation. I've had to make a number of adaptations to the algorithm to get it to work in quiche and there's still some features missing:

- No ack aggregation handling.
- Packet conservation adapted since we're not using the TCP state machine anymore, adapted from how Google handles this.
- Under constant packet loss (1%) bitrate seems to collapse. May be a delivery rate estimation issue but needs investigating.

Also more general caveats, there's no unit testing and comments are sparse. I'm submitting this draft now to see if you'd be receptive to getting BBR merged in and what changes you'd want to see here before starting code review. As part of my research, I've done a bit of evaluation and I've seen typical BBR behaviour where the throughput is higher than what CUBIC usually gets you but it incurs higher packet loss rates.

Depends on #770 
Closes #514

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -61,8 +61,11 @@ static void debug_log(const char *line, void *argp) {
 static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
     static uint8_t out[MAX_DATAGRAM_SIZE];
 
+    quiche_send_info send_info;
+
     while (1) {
-        ssize_t written = quiche_conn_send(conn_io->conn, out, sizeof(out));
+        ssize_t written = quiche_conn_send(conn_io->conn, out, sizeof(out),
+                                           &send_info);
 
         if (written == QUICHE_ERR_DONE) {
             fprintf(stderr, "done writing\n");
@@ -74,7 +77,10 @@ static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
             return;
         }
 
-        ssize_t sent = send(conn_io->sock, out, written, 0);
+        ssize_t sent = sendto(conn_io->sock, out, written, 0,
+                              (struct sockaddr *) &send_info.to,
+                              send_info.to_len);
+
         if (sent != written) {
             perror("failed to send");
             return;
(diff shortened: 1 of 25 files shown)


normanmaurer:
@rafibaum I am not a quiche developer and also no lawyer so it would be good to "confirm" my assumption, but if you did base some of your work on the code that you found in the linux kernel it might be problematic to pull the code in if the origin was licensed under the GPL.

rafibaum:
The BBR module is, depending on which Git repository you look at, either dual licensed under BSD (2-clause) and GPL or under Apache 2.0, either of which is compatible. The QUIC implementation is also Apache 2.0.

normanmaurer:
@rafibaum Perfect ... nevermind then

rafibaum:
@normanmaurer no worries! Licenses are often disregarded so I'm glad you checked. I'll add attribution for both codebases later on.

PR #954, opened 2021-06-01
h3: don't use UTF-8 strings to represent headers

ghedo:
There is not requirement for HTTP/3 header names and values to be valid
UTF-8, so don't require it.


src/h3/mod.rs
```diff
@@ -146,8 +146,8 @@
 //!
 //!             if method.value() == "GET" && path.value() == "/" {
 //!                 let resp = vec![
-//!                     quiche::h3::Header::new(":status", &200.to_string()),
-//!                     quiche::h3::Header::new("server", "quiche"),
+//!                     quiche::h3::Header::new(b":status", &200.to_string()),
```
junhochoi, line 149:
```suggestion
//!                     quiche::h3::Header::new(b":status", &200.to_string().as_bytes()),
```

junhochoi (approved):

PR #960, opened 2021-06-04
properly handle incoming RESET_STREAM frames

ghedo:
When a RESET_STREAM frame is received, unless the stream was already
readable for other reason, the application wouldn't know about it.

For example, if one peer sent STREAM frames without setting the fin
flag, then sent a RESET_STREAM frame completing the stream (even without
an error), the other peer would correctly close the receive-side of the
stream, but not notify the application.

This would prevent, for example, issuing HTTP/3 events to the application.

This change causes the stream to be marked as readable when a
RESET_STREAM frame is received, as if a STREAM frame with fin=true was
received. In addition, a `StreamReset` error will be returned by the
`stream_recv()` method when called by the application.

This also changes the behaviour in case a RESET_STREAM frame is
received, but either the application hasn't read the data yet, or the
data hasn't been received at all (i.e. there is a gap between the
current known stream offset, and the final size reported in the
RESET_STREAM frame).

In this case we now drop any already buffered data (so it won't be
returned to the application), and any missing data will also be dropped
on receipt. The application will still be notified by making the stream
readable.

For HTTP/3, a new "Reset" event is added, which propagates the error
code from the RESET_STREAM frame up to the application, without causing
the whole connection to be closed when `poll()` is called.

### DIFF
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -584,7 +584,7 @@ pub enum Event {
 
     /// Stream was reset.
     ///
-    /// The associated data represents the error code received by the peer.
+    /// The associated data represents the error code sent by the peer.
     Reset(u64),
 
     /// DATAGRAM was received.
@@ -4551,7 +4551,7 @@ mod tests {
         assert_eq!(s.poll_server(), Ok((stream, Event::Reset(42))));
         assert_eq!(s.poll_server(), Err(Error::Done));
 
-        // Sending RESET_STREAM again shouldn't trigger another Finished event.
+        // Sending RESET_STREAM again shouldn't trigger another Reset event.
         assert_eq!(
             s.pipe.send_pkt_to_server(pkt_type, &frames, &mut buf),
             Ok(39)
--- a/src/h3/mod.rs
+++ b/src/h3/mod.rs
@@ -164,6 +164,10 @@
 //!             // Peer terminated stream, handle it.
 //!         },
 //!
+//!         Ok((stream_id, quiche::h3::Event::Reset(err))) => {
+//!             // Peer reset the stream, handle it.
+//!         },
+//!
 //!         Ok((_flow_id, quiche::h3::Event::Datagram)) => (),
 //!
 //!         Ok((goaway_id, quiche::h3::Event::GoAway)) => {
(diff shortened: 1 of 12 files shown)


src/h3/mod.rs
```diff
@@ -574,6 +582,11 @@ pub enum Event {
     /// Stream was closed,
     Finished,
 
+    /// Stream was reset.
+    ///
+    /// The associated data represents the error code received by the peer.
```
LPardue, line 587: This reads a bit weird, can we say "sent by the peer", or maybe just "The associated data represents the error code"

ghedo: Oh, right, "sent by the peer" makes a lot more sense than "received by the peer".
```suggestion
    /// The associated data represents the error code sent by the peer.
```

LPardue:
It's not clear to me if the new code will also make `conn.stream_finished()` true. The reason that's important is because  `h3::process_control_stream()` checks stream condition internally and will attempt to convert quiche stream closure of control streams into an explicit `Error::ClosedCriticalStream`. So either resets already included in the `stream_finished()` checks, or we'll want to update `process_control_stream()` to detect resets, send a CONNECTION_CLOSE, and translate the error up to the application.

src/h3/mod.rs
```diff
@@ -4485,6 +4503,183 @@ mod tests {
...
+        assert_eq!(
+            s.pipe.send_pkt_to_server(pkt_type, &frames, &mut buf),
+            Ok(39)
+        );
+
+        // Server issues Reset event for the stream.
+        assert_eq!(s.poll_server(), Ok((stream, Event::Reset(42))));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Sending RESET_STREAM again shouldn't trigger another Finished event.
```
LPardue, line 4554:
```suggestion
        // Sending RESET_STREAM again shouldn't trigger another Reset event.
```

ghedo:
@LPardue it should already do that in master because `poll()` always tries to read from control streams, so it doesn't matter if they become readable or not after being reset.

LPardue:
yeah you're right and I was able to manually confirm this by tsting this PR locally and noodling with the `h3::tests::reset_stream()` test.

LPardue (dismissed):
one nit other LGTM! :ship:

PR #962, opened 2021-06-11
Add transport byte counters.

lohith-bellad:
1. Sent bytes.
2. Received bytes.
3. Lost bytes.
4. Retransmitted bytes.
5. Retransmitted packets -> packets with transmitted frames.
6. MTU -> currently using max datagram size configured

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -418,7 +418,7 @@ typedef struct {
     // The number of QUIC packets that were lost.
     size_t lost;
 
-    // The number of QUIC packets with retranmitted data.
+    // The number of sent QUIC packets with retranmitted data.
     size_t retrans;
 
     // The estimated round-trip time of the connection (in nanoseconds).
@@ -433,7 +433,7 @@ typedef struct {
     // The number of recevied bytes.
     uint64_t recv_bytes;
 
-    // The number of stream bytes lost.
+    // The number of bytes lost.
     uint64_t lost_bytes;
 
     // The number of stream bytes retransmitted.
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -418,11 +418,29 @@ typedef struct {
     // The number of QUIC packets that were lost.
     size_t lost;
 
+    // The number of QUIC packets with retranmitted data.
+    size_t retrans;
+
     // The estimated round-trip time of the connection (in nanoseconds).
     uint64_t rtt;
 
     // The size of the connection's congestion window in bytes.
     size_t cwnd;
+
+    // The number of sent bytes.
+    uint64_t sent_bytes;
+
+    // The number of recevied bytes.
+    uint64_t recv_bytes;
+
+    // The number of stream bytes lost.
+    uint64_t lost_bytes;
+
+    // The number of stream bytes retransmitted.
+    uint64_t stream_retrans_bytes;
+
+    // The current PMTU for the connection.
+    size_t pmtu;
 
     // The estimated data delivery rate in bytes/s.
     uint64_t delivery_rate;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5061,7 +5061,7 @@ pub struct Stats {
     /// The number of QUIC packets that were lost.
     pub lost: usize,
 
-    /// The number of QUIC packets with retransmitted data.
+    /// The number of sent QUIC packets with retransmitted data.
     pub retrans: usize,
 
     /// The estimated round-trip time of the connection.
@@ -5076,7 +5076,7 @@ pub struct Stats {
     /// The number of received bytes.
     pub recv_bytes: u64,
 
-    /// The number of stream bytes lost.
+    /// The number of bytes lost.
     pub lost_bytes: u64,
 
     /// The number of stream bytes retranmitted.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -980,6 +980,9 @@ pub struct Connection {
     /// Total number of sent packets.
     sent_count: usize,
 
+    /// Total number of packets sent with data retransmitted.
+    retrans_count: usize,
+
     /// Total number of bytes received from the peer.
     rx_data: u64,
 
@@ -1005,6 +1008,16 @@ pub struct Connection {
     /// Total number of bytes the server can send before the peer's address
     /// is verified.
     max_send_bytes: usize,
+
+    /// Total number of bytes retransmitted over the connection.
+    /// This counts only STREAM and CRYPTO data.
+    stream_retrans_bytes: u64,
+
+    /// Total number of bytes sent over the connection.
+    sent_bytes: u64,
+
+    /// Total number of bytes recevied over the connection.
+    recv_bytes: u64,
 
     /// Streams map, indexed by stream ID.
     streams: stream::StreamMap,
@@ -1359,6 +1372,9 @@ impl Connection {
 
             recv_count: 0,
             sent_count: 0,
+            retrans_count: 0,
+            sent_bytes: 0,
+            recv_bytes: 0,
 
             rx_data: 0,
             max_rx_data,
@@ -1369,6 +1385,8 @@ impl Connection {
 
             tx_data: 0,
             max_tx_data: 0,
+
+            stream_retrans_bytes: 0,
 
             max_send_bytes: 0,
 
@@ -2238,6 +2256,8 @@ impl Connection {
 
         let read = b.off() + aead_tag_len;
 
+        self.recv_bytes += read as u64;
+
         // An Handshake packet has been received from the client and has been
         // successfully processed, so we can drop the initial state and consider
         // the client's address to be verified.
@@ -2440,6 +2460,8 @@ impl Connection {
 
         let epoch = pkt_type.to_epoch()?;
 
+        let stream_retrans_bytes = self.stream_retrans_bytes;
+
         // Process lost frames.
         for lost in self.recovery.lost[epoch].drain(..) {
             match lost {
@@ -2448,6 +2470,8 @@ impl Connection {
                         .crypto_stream
                         .send
                         .retransmit(offset, length);
+
+                    self.stream_retrans_bytes += length as u64;
                 },
 
                 frame::Frame::StreamHeader {
@@ -2482,6 +2506,8 @@ impl Connection {
                             incremental,
                         );
                     }
+
+                    self.stream_retrans_bytes += length as u64;
                 },
 
                 frame::Frame::ACK { .. } => {
@@ -2514,6 +2540,10 @@ impl Connection {
 
                 _ => (),
             }
+        }
+
+        if stream_retrans_bytes > self.stream_retrans_bytes {
+            self.retrans_count += 1;
         }
 
         let mut left = b.cap();
@@ -3224,6 +3254,7 @@ impl Connection {
         self.pkt_num_spaces[epoch].next_pkt_num += 1;
 
         self.sent_count += 1;
+        self.sent_bytes += written as u64;
 
         if self.dgram_send_queue.byte_size() > self.recovery.cwnd_available() {
             self.recovery.update_app_limited(false);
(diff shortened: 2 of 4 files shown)


junhochoi (commented):
* tx_data and rx_data doesn't include packet header and non stream frames, causing underreporting
* Datagram is counted?

lohith-bellad:
> * tx_data and rx_data doesn't include packet header and non stream frames, causing underreporting
> * Datagram is counted?

made changes

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
     /// is verified.
     max_send_bytes: usize,
 
+    /// Total nummber of bytes retransmitted over the connection.
+    retrans_data: u64,
+
+    /// Total bytes sent counter over the connection.
+    bytes_sent: u64,
+
+    /// Total bytes recevied over the connection.
```
ghedo, line 1018:
```suggestion
    /// Total number of bytes received over the connection.
```

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
     /// is verified.
     max_send_bytes: usize,
 
+    /// Total nummber of bytes retransmitted over the connection.
+    retrans_data: u64,
+
+    /// Total bytes sent counter over the connection.
```
ghedo, line 1015:
```suggestion
    /// Total number of bytes sent over the connection.
```

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
     /// is verified.
     max_send_bytes: usize,
 
+    /// Total nummber of bytes retransmitted over the connection.
```
ghedo, line 1012:
```suggestion
    /// Total number of bytes retransmitted over the connection.
```

src/lib.rs
```diff
@@ -2448,6 +2470,7 @@ impl Connection {
                         .crypto_stream
                         .send
                         .retransmit(offset, length);
+                    self.retrans_data += length as u64;
```
ghedo, line 2473:
```suggestion

                    self.retrans_data += length as u64;
```

src/lib.rs
```diff
@@ -5024,12 +5060,30 @@ pub struct Stats {
...
+    /// The number of received bytes.
+    pub recv_bytes: u64,
+
+    /// The number of bytes lost.
+    pub lost_bytes: u64,
+
+    /// The number of bytes retranmitted.
+    pub retrans_bytes: u64,
+
+    /// pMTU for the QUIC connection.
```
ghedo, line 5084:
```suggestion
    /// The current PMTU for the connection.
```

src/lib.rs
```diff
@@ -5024,12 +5060,30 @@ pub struct Stats {
...
+    /// The number of QUIC packets with retransmitted data.
+    pub retrans: usize,
+
     /// The estimated round-trip time of the connection.
     pub rtt: time::Duration,
 
     /// The size of the connection's congestion window in bytes.
     pub cwnd: usize,
 
+    /// The number of sent bytes.
```
ghedo, line 5072: It looks like in other cases (e.g. `recv_bytes` and `lost_bytes`) we count the total number of bytes in a packet (including the header), while `sent_bytes` only counts the payload. Is this on purpose? If so it needs to be documented I think.

src/lib.rs
```diff
@@ -5024,12 +5060,30 @@ pub struct Stats {
...
+    pub sent_bytes: u64,
+
+    /// The number of received bytes.
+    pub recv_bytes: u64,
+
+    /// The number of bytes lost.
+    pub lost_bytes: u64,
+
+    /// The number of bytes retranmitted.
```
ghedo, line 5081: `retrans_bytes` also only counts STREAM and CRYPTO data AFAICT, so should specify in the description I think.

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
     /// is verified.
     max_send_bytes: usize,
 
+    /// Total nummber of bytes retransmitted over the connection.
+    retrans_data: u64,
```
ghedo, line 1013: Maybe call this `retrans_bytes` to be consistent with other fields?

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
...
+
+    /// Total bytes sent counter over the connection.
+    bytes_sent: u64,
```
ghedo, line 1016: Rename to `sent_bytes` for consistency.

src/lib.rs
```diff
@@ -1006,6 +1009,15 @@ pub struct Connection {
...
+
+    /// Total bytes recevied over the connection.
+    bytes_recv: u64,
```
ghedo, line 1019: Rename to `recv_bytes` for consistency.

ghedo (changes requested):
Added a few comments, mostly just style/nits.

ghedo:
Also, you need to update the structure in `src/ffi.rs` and `include/quiche.h` as well.

junhochoi:
@lohith-bellad I think we still have a discrepancy issue - packet length vs data length. `retrans_data` and `lost_bytes`  is counting a stream data. Since a packet can contain retransmitted frame and new frame (or other frames needed for the protocol), it's hard to tell this is a retransmit packet as TCP does.

I suggest we have a following counters for clearer definition:

- `bytes_packet_recv` - counting all incoming bytes including packet header and frames across all epochs, at recv()
- `bytes_packet_send` - counting all outgoing bytes including packet header and frames across all epochs, at send()
- `bytes_stream_recv` - counting all incoming stream payload, at stream_recv()
- `bytes_stream_send` - counting all outgoing stream payload, at stream_send()

Above are including all retransmits. These are only for streams:

- `bytes_stream_lost` - counting all outgoing stream payload detected as lost
- `bytes_stream_retransmit` - counting all outgoing stream payload retransmitted

We still don't have a way to count retransmitted frames (e.g. MAX_DATA), but this can be a good start.

ghedo:
@junhochoi why do we need `bytes_stream_recv` / `bytes_stream_send`? It also seems that an application could just count that on its own if it needs it.

junhochoi:
> @junhochoi why do we need `bytes_stream_recv` / `bytes_stream_send`? It also seems that an application could just count that on its own if it needs it.

Same thing can apply to bytes_recv/send? Also when http3 is used app doesn't know all quic payload, such as a size of qpack-decoded header frame.

My point is we need to see packet and stream separately because "bytes_lost" metric is a stream level one. Probably we don't need to add `bytes_stream_recv/send` if not needed right now, but just want to be clear on the name where the metric came from.

LPardue:
agree with Junho here. In future I'd also like to see some way to retrieve buffered but not-sent bytes, acessible via both Stats and directly in the API.

lohith-bellad:
@junhochoi lost_bytes and retrans_bytes are at stream level. Hard to get these for connection level. Thought getting stream level would be simpler approximation.

Would it make sense to add stream word in above two counters to be more clearer ?

yes retrans_count as the comment says will be incremented only if packet contains retransmitted data, good approximation. Yes as TCP we cannot say this entire packet contains retransmitted data.

We can have `bytes_stream_recv` and `bytes_stream_sent` taken up later ?

junhochoi:
> Would it make sense to add stream word in above two counters to be more clearer ?
I think so.

> We can have `bytes_stream_recv` and `bytes_stream_sent` taken up later ?

Sure. we can add them later if needed

include/quiche.h
```diff
@@ -418,12 +418,30 @@ typedef struct {
...
     // The size of the connection's congestion window in bytes.
     size_t cwnd;
 
+    // The number of sent bytes.
+    uint64_t sent_bytes;
+
+    // The number of recevied bytes.
+    uint64_t recv_bytes;
+
+    // The number of bytes lost.
```
junhochoi, line 436:
```suggestion
    // The number of stream bytes lost.
```

include/quiche.h
```diff
@@ -418,12 +418,30 @@ typedef struct {
...
+    uint64_t stream_lost_bytes;
+
+    // The number of bytes retransmitted.
```
junhochoi, line 439:
```suggestion
    // The number of stream bytes retransmitted.
```

src/lib.rs
```diff
@@ -5024,12 +5062,30 @@ pub struct Stats {
...
     /// The size of the connection's congestion window in bytes.
     pub cwnd: usize,
 
+    /// The number of sent bytes.
+    pub sent_bytes: u64,
+
+    /// The number of received bytes.
+    pub recv_bytes: u64,
+
+    /// The number of bytes lost.
```
junhochoi, line 5080:
```suggestion
    /// The number of stream bytes lost.
```

src/lib.rs
```diff
@@ -5024,12 +5062,30 @@ pub struct Stats {
...
+    pub stream_lost_bytes: u64,
+
+    /// The number of bytes retranmitted.
```
junhochoi, line 5083:
```suggestion
    /// The number of stream bytes retranmitted.
```

junhochoi (approved):
some nits, but LGTM

lohith-bellad:
> Added a few comments, mostly just style/nits.

Taken care

LPardue (dismissed):

junhochoi (dismissed):

src/lib.rs
```diff
@@ -4266,8 +4298,14 @@ impl Connection {
             recv: self.recv_count,
             sent: self.sent_count,
             lost: self.recovery.lost_count,
+            retrans: self.retrans_count,
             cwnd: self.recovery.cwnd(),
             rtt: self.recovery.rtt(),
+            sent_bytes: self.sent_bytes,
+            stream_lost_bytes: self.recovery.bytes_lost,
```
ghedo, line 4305: AFAICT `self.recovery.bytes_lost` tracks total packet bytes, including packet headers and other frames (but excluding non-ack-eliciting packets), so either the `stream_` prefix is wrong here, or it needs to get the data somewhere else.

ghedo (changes requested):

src/lib.rs
```diff
@@ -1975,6 +1993,7 @@ impl Connection {
                     let pkt = (b.buf()[..pkt_len]).to_vec();
 
                     self.undecryptable_pkts.push_back((pkt, *info));
+                    self.recv_bytes += pkt_len as u64;
```
ghedo, line 1996: Oh, sorry, I missed this before, but this will cause the bytes of buffered packets to be counted twice, so this should probably be removed. Was it added on purpose?

ghedo (changes requested):
Added one more comment that I missed before, then it should be good to go.

Though it's probably worth noting that the way "retrans_count" is tracked means that it's not necessarily very accurate, as data from multiple stream frames could be retransmitted as a single frame, or split into multiple frames, or not retransmitted at all, e.g. if the stream is closed in the meantime, but for now it's fine, we can improve on it later on.

src/lib.rs
```diff
@@ -5024,12 +5061,30 @@ pub struct Stats {
     /// The number of QUIC packets that were lost.
     pub lost: usize,
 
+    /// The number of QUIC packets with retransmitted data.
```
ghedo, line 5064:
```suggestion
    /// The number of sent QUIC packets with retransmitted data.
```

src/lib.rs
```diff
@@ -5024,12 +5061,30 @@ pub struct Stats {
...
     /// The size of the connection's congestion window in bytes.
     pub cwnd: usize,
 
+    /// The number of sent bytes.
+    pub sent_bytes: u64,
+
+    /// The number of received bytes.
+    pub recv_bytes: u64,
+
+    /// The number of stream bytes lost.
```
ghedo, line 5079:
```suggestion
    /// The number of bytes lost.
```

ghedo (approved):
Added some final nits, will squash during merge.

ghedo (dismissed):

include/quiche.h
```diff
@@ -418,12 +418,30 @@ typedef struct {
     // The number of QUIC packets that were lost.
     size_t lost;
 
+    // The number of QUIC packets with retranmitted data.
```
ghedo, line 421:
```suggestion
    // The number of sent QUIC packets with retranmitted data.
```

include/quiche.h
```diff
@@ -418,12 +418,30 @@ typedef struct {
...
     // The size of the connection's congestion window in bytes.
     size_t cwnd;
 
+    // The number of sent bytes.
+    uint64_t sent_bytes;
+
+    // The number of recevied bytes.
+    uint64_t recv_bytes;
+
+    // The number of stream bytes lost.
```
ghedo, line 436:
```suggestion
    // The number of bytes lost.
```

PR #965, opened 2021-06-15
Report bytes received and bytes sent in quiche::Stats

nox:
This is a breaking change.


nox:
I am not exactly sure `rx_data` and `tx_data` are the proper fields to use for that.

nox:
Oh I didn't see #962 before, closing this.

PR #966, opened 2021-06-15
qlog: add means to provide external event time

LPardue:
Previously, the `add_event()` method populated the qlog event_time
field by calling `std::time::Instant::elapsed()`, which causes an in
direct call for `std::time::Instand::now()`. This incurs additional
system calls that are wasteful, especially if events are logged in
rapid succession or the time is available elsewhere.

This change introduces the `add_event_with_instant()` method that
allows callers to provide their own Instant that they wish to use
when logging an event.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2024,14 +2024,17 @@ impl Connection {
                 Some(&hdr.dcid),
             );
 
-            q.add_event(qlog::event::Event::packet_received(
-                hdr.ty.to_qlog(),
-                qlog_pkt_hdr,
-                Some(Vec::new()),
-                None,
-                None,
-                None,
-            ))
+            q.add_event_with_instant(
+                qlog::event::Event::packet_received(
+                    hdr.ty.to_qlog(),
+                    qlog_pkt_hdr,
+                    Some(Vec::new()),
+                    None,
+                    None,
+                    None,
+                ),
+                now,
+            )
             .ok();
         });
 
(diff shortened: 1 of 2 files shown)


LPardue:
The PR initially failed on the multiarch build but it seems completely unrelated and a rerun cleared it.

```
failures:

---- h3::tests::reset_finished_at_client stdout ----
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(40)`,
 right: `Ok(39)`', src/h3/mod.rs:4666:9
```

junhochoi:
> The PR initially failed on the multiarch build but it seems completely unrelated and a rerun cleared it.
> 
> ```
> failures:
> 
> ---- h3::tests::reset_finished_at_client stdout ----
> thread 'main' panicked at 'assertion failed: `(left == right)`
>   left: `Ok(40)`,
>  right: `Ok(39)`', src/h3/mod.rs:4666:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Opened #968 for fixing those tests.

junhochoi:
How heavy is Instant::now() ? In Linux it runs on VDSO (other platform can be different) so it should be cheap most of times. Also reusing timestamp value can make the event timestamp looks all same within a single API call (e.g. recv()), not sure this can make a visible difference.

ghedo (approved):

PR #967, opened 2021-06-15
qlog: add support for importance levels

LPardue:
Previously, qlog data_moved events were generated any time stream data
was read from or written to quiche. Since these events can be costly to
log, this change makes them optional when qlog is enabled.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -255,6 +255,12 @@ bool quiche_conn_set_keylog_path(quiche_conn *conn, const char *path);
 // Enables keylog to the specified file descriptor. Unix only.
 void quiche_conn_set_keylog_fd(quiche_conn *conn, int fd);
 
+enum quiche_qlog_level {
+    QUICHE_QLOG_CORE = 0,
+    QUICHE_QLOG_BASE = 1,
+    QUICHE_QLOG_EXTRA = 2,
+};
+
 // Enables qlog to the specified file path. Returns true on success.
 bool quiche_conn_set_qlog_path(quiche_conn *conn, const char *path,
                           const char *log_title, const char *log_desc);
@@ -262,6 +268,9 @@ bool quiche_conn_set_qlog_path(quiche_conn *conn, const char *path,
 // Enables qlog to the specified file descriptor. Unix only.
 void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char *log_title,
                              const char *log_desc);
+
+void quiche_conn_set_qlog_level(quiche_conn *conn,
+                                enum quiche_qlog_level log_level);
 
 // Configures the given session for resumption.
 int quiche_conn_set_session(quiche_conn *conn, const uint8_t *buf, size_t buf_len);
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -588,6 +588,14 @@ pub extern fn quiche_conn_set_qlog_fd(
 }
 
 #[no_mangle]
+#[cfg(feature = "qlog")]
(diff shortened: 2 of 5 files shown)


junhochoi (approved):

LPardue:
Discussed some offline. Making some of the qlog event logging optional is a nice feature to have but we'd prefer not to break the existing API. 

Will have a think trhough some other options, but we'll probably still disable `data_moved` events by default.

cc: @rmarx you might be interested in knowing how implementations are trying to tailor qlog creation.

rmarx:
Thanks for the heads-up.

I can definitely imagine that `data_moved` is something you don't want to log all the time indeed :) 

My original ideas for qlog interfaces would be that you can set a "log level" that more or less corresponds to the `importance` indicators in the qlog drafts (so Core, Base and Extra, see https://www.ietf.org/archive/id/draft-ietf-quic-qlog-main-schema-00.html#name-event-importance-indicators). I do see that `data_moved` is currently marked as Base, so maybe that should be Extra instead :) (though I don't like correlating verbosity with importance conceptually). 

You can of course define custom levels with a custom mapping to the rough Importance indicators on a per-implementation level as well (say crucial, core, normal, debug (data_moved might be here), verbose). 

A feature that was once suggested by Facebook and is tracked at https://github.com/quicwg/qlog/issues/72 is to include the events that you're logging in the qlog explicitly, so it's clear whether the event was not logged by configuration or if it just didn't occur in the trace. That type of thing would allow a very fine-grained interface where you could indicate on a per-event-type basis what you want to log or not, but I'm sure that's not something all implementations might want.

Either way, I personally don't like the `bool log_data_moved` approach you took in this PR and would prefer to see it updated with a `QLOG_LEVEL level` type of deal at the least ;)

LPardue:
Ahh, I think I confused core and base.

So yeah I think we can easily implement broad logging levels here. And the qlog crate can probably do a better job of documenting the events is does support at different levels. 

What I'd like on the top is further fine tuning that would allow apps to selectively disable specific events in a given log level. Something like rust log filters but a bit different https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html

rmarx:
Yes, the "selectively disable specific events in a given log level" is what I was trying to explain near the end of my comment with the link to https://github.com/quicwg/qlog/issues/72, as that would allow indicating in qlog which events were enabled/disabled for a given run. 

I've also wondered if we can't use better names than Core and Base... ideal bikeshedding topic

LPardue:
I've updated the PR to support importance levels.

Internally, the qlog crate gets some refactoring. When adding the `importance` to events, dependent code gets quite boilerplatey and repetitive. So I introduced some `from` trait impl to make mapping between event types, categories and importance a bit easier to manage with support from compiler. The change in `tools/qlog/src/event.rs` looks huge because of that but it is a simple change really.

Externally, the qlog streamer object needs to be constructed with a logging level. Quiche will default this to `Base` level, which is what it was implicitly doing already. The new quiche API `set_qlog_level` can be used to switch this to `Core` level to reduce the events (practically today, just the data_moved ones).

LPardue:
Actually, if we're going to touch this code, we might as well bump the qlog version we are targetting. The reason being that the serialized format changes mean that some of this work would be immediately deprecated when we bump.

With some prototype code, I found we can't mix & match the serialization style with the old scheme, so we'll need to do that too. It's going to end up an even larger changeset but due to the interdependencies it makes sense to do it all combined together.

LPardue:
overtaken by #1013

PR #970, opened 2021-06-16
stream_send(...) should return Err(Error::Done) if no bytes are send

normanmaurer:
Motivation:

The documention of stream_send(...) states that Err(Error::Done) is returned if no bytes were send. This is not always the case and Ok(0) might be returned.

Modifications:

Return Err(Error::Done) if tx_cap is 0.

Result:

Return value matches the documentation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3448,6 +3448,9 @@ impl Connection {
         // Truncate the input buffer based on the connection's send capacity if
         // necessary.
         let cap = self.tx_cap;
+        if cap == 0 {
+            return Err(Error::Done)
+        }
 
         let (buf, fin) = if cap < buf.len() {
             (&buf[..cap], false)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8080,7 +8080,7 @@ mod tests {
             Ok(15)
         );
         assert_eq!(pipe.advance(), Ok(()));
-        assert_eq!(pipe.client.stream_send(8, b"a", false), Ok(0));
+        assert_eq!(pipe.client.stream_send(8, b"a", false), Err(Error::Done));
         assert_eq!(pipe.advance(), Ok(()));
 
         let mut r = pipe.server.readable();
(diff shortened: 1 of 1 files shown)


LPardue (commented):
I think this change breaks how 0-length FIN works currently.

normanmaurer:
@LPardue i think you are right and we should only return Done if no fin was send ... correct ?

LPardue:
Yeah, in the first check of cap is 0, also check that  input fin is false before returning Done. We should probably also update the method description to state that 0-length FIN is supported even if capacity is zero.

normanmaurer:
@LPardue will fix formatting error and update docs

src/lib.rs
```diff
@@ -10447,7 +10450,7 @@ mod tests {
         assert_eq!(pipe.server.stream_send(8, &buf[..5000], false), Ok(2000));
 
         // No more connection send capacity.
-        assert_eq!(pipe.server.stream_send(12, &buf[..5000], false), Ok(0));
+        assert_eq!(pipe.server.stream_send(12, &buf[..5000], false), Err(Error::Done));
```
LPardue, line 10453:
```suggestion
        assert_eq!(
            pipe.server.stream_send(12, &buf[..5000], false),
            Err(Error::Done)
        );
```

normanmaurer:
@LPardue docs updated... PTAL

normanmaurer:
@LPardue @ghedo test failure is not related to this change

src/lib.rs
```diff
@@ -3390,6 +3390,10 @@ impl Connection {
     /// On success the number of bytes written is returned, or [`Done`] if no
     /// data was written (e.g. because the stream has no capacity).
     ///
+    /// This methods also allows to send 0-length FIN `STREAM_FRAME` frames,
+    /// which means it is possible to use it with 0-length bytes while the
+    /// FIN flag is set. In this case it will return [`Ok(0)`].
```
LPardue, line 3395: Maybe something like

```suggestion
    /// Applications can provide a 0-length buffer with the fin flag set to 
    /// true. This will lead to a 0-length FIN STREAM frame being sent at the 
    /// latest offset. This is the only case where [`Ok(0)`] is returned.
```

normanmaurer:
@LPardue sounds good...

LPardue (approved):
LGTM

LPardue:
thanks @normanmaurer !

PR #976, opened 2021-06-19
recovery: fix overflow in PRR limit calculation

omarkilani:
This is a follow on fix to the one in https://github.com/cloudflare/quiche/commit/fd2c866c80bad1d2aea0f2e34d13d6bb8b0824a6

We had issues with Cloudflare not delivering objects properly over QUIC (see CF ticket #2183294).

I went through a whole bunch of debugging, repro, etc. Then decided I could run nginx + quiche and see if I could hit the same issue.

Turns out I could:

```
2021/06/19 15:09:56 [debug] 126004#0: *12 quic read handler
thread '<unnamed>' panicked at 'attempt to subtract with overflow', src/recovery/prr.rs:75:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
2021/06/19 15:09:56 [notice] 126003#0: signal 17 (SIGCHLD) received from 126004
2021/06/19 15:09:56 [alert] 126003#0: worker process 126004 exited on signal 6 (core dumped)
```

My test runner is available here:

https://gist.github.com/omarkilani/1a10dadcdb2d24739ab4fe96c29b7a09

`thing` is available here:

https://github.com/omarkilani/thing

Without this patch, it would fail on the second run because...:

```
on_packet_acked[1]: delivered_data = 1200, pipe = 1200, ssthresh = 482400, max_datagram_size = 1200, self.prr_delivered = 2231251
on_packet_acked[1]: delivered_data = 1200, pipe = 0, ssthresh = 482400, max_datagram_size = 1200, self.prr_delivered = 2232451
on_packet_acked[1]: delivered_data = 1200, pipe = 1039602, ssthresh = 482400, max_datagram_size = 1200, self.prr_delivered = 2233651
on_packet_acked[2]: self.prr_delivered = 2234851, ssthresh = 482400, self.recoverfs = 1015600, self.prr_out = 2261253
```

Which:

```
X = ((prr_delivered * ssthresh + recoverfs - 1) / recoverfs) = 1061533
Y = prr_out = 2261253
X - Y = -1199720 (subtract with overflow)
```

With the patch the test succeeds for unlimited runs. 😃

PR #977, opened 2021-06-20
recovery: fix more PRR calculation issues

junhochoi:
In addition to #976, fix additional divide by zero case.
Also test cases for all overflow added, including #973.


ghedo:
FTR, I cherry-picked the commit from https://github.com/cloudflare/quiche/pull/976 so that @omarkilani gets properly credited for the fix (thanks!).

ghedo (approved):

PR #979, opened 2021-06-20
ci: add libm to nightly C example

junhochoi:
CI is failing when building C examples on nightly:

```
/usr/bin/ld: /home/runner/work/quiche/quiche/examples/build/debug/libquiche.a(std-363511a59ca6fedb.std.2k3kczsj-cgu.0.rcgu.o): undefined reference to symbol 'fmaf@@GLIBC_2.2.5'
```

stable is ok without it, so I think it's nightly issue.


junhochoi:
It's a band-aid so need to find a solution. since we use [libm](https://github.com/rust-lang/libm), this should not happen.

ghedo:
Mrged as part of https://github.com/cloudflare/quiche/pull/977.

PR #981, opened 2021-06-21
nginx: add command: http3_cc_algorithm

lilongyan-huawei:
users can select congestion control algorithm

### DIFF
--- a/extras/nginx/nginx-1.16.patch
+++ b/extras/nginx/nginx-1.16.patch
@@ -29,8 +29,8 @@ Subject: [PATCH] Initial QUIC and HTTP/3 implementation using quiche
  src/http/v3/ngx_http_v3.c               | 2228 +++++++++++++++++++++++
  src/http/v3/ngx_http_v3.h               |   77 +
  src/http/v3/ngx_http_v3_filter_module.c |   68 +
- src/http/v3/ngx_http_v3_module.c        |  286 +++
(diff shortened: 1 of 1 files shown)


extras/nginx/nginx-1.16.patch
```diff
@@ -4035,6 +4035,13 @@ index 000000000..30955c4cd
 +      offsetof(ngx_http_v3_srv_conf_t, idle_timeout),
 +      NULL },
 +
++    { ngx_string("http3_cc_algorithm"),
++     NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_TAKE1,
++     ngx_conf_set_num_slot,
```
ghedo, line 4040: Instead of using the CC algorithm IDs directly, it's probably better to add the `set_cc_algorithm_name()` function to the FFI API (https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L894), and then use a string for this configuration value.

ghedo (changes requested):
Added a comment, but thanks for the patch @lilongyan-huawei.

ghedo:
Also, please rebase your branch on latest `master`.

ghedo:
Superseded by https://github.com/cloudflare/quiche/commit/e434e42173f73bb533f6a39cd853eac752731fd9.

PR #991, opened 2021-06-30
qlog: align time units to qlog draft 02 style

LPardue:
The qlog crate is primarily based on the qlog draft 01 specification.
This allowed configuration of time units used for logging, either in
milliseconds or microseconds. The qlog crate only supports relative time
logging and defaults to millisecond units in integer form.

The unfortunate side effect of millisecond units is that events occurring
close together get aliased into the same millisecond. Tools like qvis
can then misrepresent things.

In qlog draft 02, the time unit handling changed. It is no longer
configurable, instead it is always milliseconds, represented as floating
point. Previously, an event at relative time 100085 microseconds, would
be logged as "100". With this change it is logged as "100.085".

This change updates the time unit handling to align with draft 02. It
removes time unit configurability and logs event times as floating
point.

### DIFF
--- a/tools/qlog/src/lib.rs
+++ b/tools/qlog/src/lib.rs
@@ -93,7 +93,6 @@
 //!     Some("Example qlog trace description".to_string()),
 //!     Some(qlog::Configuration {
 //!         time_offset: Some("0".to_string()),
-//!         time_units: Some(qlog::TimeUnits::Ms),
 //!         original_uris: None,
 //!     }),
 //!     None,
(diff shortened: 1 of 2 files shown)


tools/qlog/src/lib.rs
```diff
@@ -734,7 +706,14 @@ impl QlogStreamer {
 
         let maybe_terminate = if contains_frames { "" } else { "]" };
 
-        let ev_time = serde_json::to_string(&EventField::RelativeTime(rel)).ok();
+        let dur = if cfg!(test) {
+            std::time::Duration::from_secs(0)
+        } else {
+            now.duration_since(self.start_time)
+        };
+        let rel_time = dur.as_micros() as f64 / 1000.0;
```
junhochoi, line 714:
```suggestion
        let rel_time = dur.as_secs_f64() / 1000.0;
```

junhochoi (approved):

Issue #983, opened 2021-06-23
almost_full tends to true [theoretically]


Sparika:
The following is only theoretical as I've not checked if it happens or not. But looking at the code and running some simplified computations, I have the impression that the almost_full function may tends to true once the received offset is sufficiently large compared to the initial max_data.

Here, `max_data_next` is increased by the `len` of the emited packets.
https://github.com/cloudflare/quiche/blob/f86de8f18e930f823e00b43b08eee47dc52aa635/src/stream.rs#L872

and `almost_full` is true if `max_data_next/2 > max_data - len` (here, `len` is the total length received).
https://github.com/cloudflare/quiche/blob/f86de8f18e930f823e00b43b08eee47dc52aa635/src/stream.rs#L951

I believe `max_data_next` increases constantly but `max_data - (total)_len` is heuristically bounded by the regular increase of `max_data`. Supposing that this bound is the initial value of `max_data`, then once `max_data_next > 2*initial_max_data` the stream tends to be always declared almost full.

If this is indeed true, the practical implication would be that MAX_STREAM_DATA frames would be send at a high frequency once this threshold is passed. Is it the behavior intended?

Sparika:
I can confirm I am observing the behavior I've described in practice. The setup is a server sending a large amount of data to a client. 

The client is configured with the following
```
        config.set_initial_max_data(1_000_000);
        config.set_initial_max_stream_data_uni(1_000_000);
        config.set_initial_max_stream_data_bidi_local(1_000_000);
        config.set_initial_max_stream_data_bidi_remote(1_000_000);
        config.set_initial_max_streams_bidi(100);
        config.set_initial_max_streams_uni(100);
```

```
[... All traces before that point ar "OK", i.e. not "ALMOST FULL"]
RecvBuf { data: [], off: 332053, len: 332053, max_data: 1000000, max_data_next: 1332053, fin_off: None, drain: false } OK
RecvBuf { data: [], off: 332100, len: 332100, max_data: 1000000, max_data_next: 1332100, fin_off: None, drain: false } OK
RecvBuf { data: [], off: 333403, len: 333403, max_data: 1000000, max_data_next: 1333403, fin_off: None, drain: false } ALMOST FULL
RecvBuf { data: [], off: 333450, len: 333450, max_data: 1333403, max_data_next: 1333450, fin_off: None, drain: false } OK
[... 326 OK traces]
[...]

PR #992, opened 2021-06-30
Changes control flow credits increases to a fixed amount of half init…

Sparika:
…ial max_data

See issue #983 

Previous implementation of update_max_data() and related code has an issue that may result in
max_data increments becoming smaller over time and increasing the amount of MAX_STREAM_DATA
frames sent.

Another potential issue would be to have the difference between max_data and max_data_next
increasing over time.

The implemented solution sets a `max_data_incr` that is equal to half the initial `max_data`.
`max_data_incr` is used to determine when the stream is almost full, i.e. `len > max_data -
max_data_incr` and to set the new `max_data`, i.e. `max_data = max_data + max_data_incr`. The
control flow credits always increase by half the initial `max_data` and the unspent control
flow credits (`max_data - len`) is never larger than the initial `max_data`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -989,9 +989,8 @@ pub struct Connection {
     /// Local flow control limit for the connection.
     max_rx_data: u64,
 
-    /// Updated local flow control limit for the connection. This is used to
-    /// trigger sending MAX_DATA frames after a certain threshold.
-    max_rx_data_next: u64,
+    /// The maximum offset increment that is used for the connection.
+    max_rx_data_incr: u64,
 
     /// Whether we send MAX_DATA frame.
     almost_full: bool,
@@ -1378,7 +1377,7 @@ impl Connection {
 
             rx_data: 0,
             max_rx_data,
-            max_rx_data_next: max_rx_data,
+            max_rx_data_incr: max_rx_data /2,
             almost_full: false,
 
             tx_cap: 0,
@@ -2752,16 +2751,17 @@ impl Connection {
             }
 
             // Create MAX_DATA frame as needed.
-            if self.almost_full && self.max_rx_data < self.max_rx_data_next {
+            if self.almost_full {
+                let max_rx_data_next = self.max_rx_data + self.max_rx_data_incr;
                 let frame = frame::Frame::MaxData {
-                    max: self.max_rx_data_next,
+                    max: max_rx_data_next,
                 };
 
                 if push_frame_to_pkt!(b, frames, frame, left) {
                     self.almost_full = false;
 
                     // Commits the new max_rx_data limit.
-                    self.max_rx_data = self.max_rx_data_next;
+                    self.max_rx_data = max_rx_data_next;
 
                     ack_eliciting = true;
                     in_flight = true;
@@ -3350,8 +3350,6 @@ impl Connection {
                 return Err(e);
             },
         };
-
-        self.max_rx_data_next = self.max_rx_data_next.saturating_add(read as u64);
 
         let readable = stream.is_readable();
 
@@ -5001,11 +4999,11 @@ impl Connection {
 
     /// Returns true if the connection-level flow control needs to be updated.
     ///
-    /// This happens when the new max data limit is at least double the amount
-    /// of data that can be received before blocking.
+    /// This happens when the current offset for the connection is more than half of the
+    /// previously allocated control flow credit since last MAX_DATA frame.
+    /// max_rx_data - max_rx_data_incr < rx_data < max_rx_data < max_rx_data + max_rx_data_incr
     fn should_update_max_data(&self) -> bool {
-        self.max_rx_data_next != self.max_rx_data &&
-            self.max_rx_data_next / 2 > self.max_rx_data - self.rx_data
+        self.rx_data > self.max_rx_data - self.max_rx_data_incr
     }
 
     /// Returns the idle timeout value.
@@ -6884,10 +6882,10 @@ mod tests {
             iter.next(),
             Some(&frame::Frame::MaxStreamData {
                 stream_id: 4,
-                max: 30
+                max: 22
             })
         );
-        assert_eq!(iter.next(), Some(&frame::Frame::MaxData { max: 46 }));
+        assert_eq!(iter.next(), Some(&frame::Frame::MaxData { max: 45 }));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


ghedo (changes requested):
Thank you for the PR, @Sparika!

The change looks good to me, however, please have a look at the failed CI builds. The code needs to be reformatted (you can run `cargo +nightly fmt` after installing latest Rust nightly), and it also looks like there are a couple of tests failing.


In addition, we should also implement a similar change for the `MAX_DATA` frame too (that is, for connection-level flow control, see https://github.com/cloudflare/quiche/blob/master/src/lib.rs#L2754), as it works in a similar way, and is affected by the same problem.

Sparika:
`max_data_incr` is the value by which `max_data` will be incremented regularly, i.e. each time `len > max_data - max_data_incr`. Because the increment is triggered each time len is greater than `max_data - max_data_incr` and the new limit is set to `max_data = max_data + max_data_incr`, the available control flow credit is always less than `2x max_data_incr`. More explicitly: `max_data - max_data_incr < len < max_data < max_data + max_data_incr`. By setting `max_data_incr = max_data/2` using the initial value, the code ensures that the available control flow credit is always less than `2x ([initial]max_data/2) == [initial]max_data`.  

Supposing a symmetrical connection with a throughput equal to the initial max_data. Then after 1/2 RTT, the connection would have consumed `max_data/2`, triggering the increase `max_data = max_data + max_data/2` from the receiver. The MAX_DATA frame would arrive 1/2 RTT later at the sender, just in time for refreshing the limit for the sender. I've not run test or graphs for that part but that would surely be interesting.

Regarding dynamically changing the control flow limit, my understanding is that the limit ensures the receiver does not over-commit on memory, and so I believe that whatever the BDP or other network properties, the network should not influence the control flow limit. Otherwise that would allow the network to force the receiver into committing more memory than intended. As I understand, this serves as a way to communicate the buffer limit to the sender while having a throughput otherwise unlimited (except for congestion control). 

Feel free to correct if I'm wrong in anything. ;)

LPardue:
> Regarding dynamically changing the control flow limit, my understanding is that the limit ensures the receiver does not over-commit on memory, and so I believe that whatever the BDP or other network properties, the network should not influence the control flow limit

That kind of depends. Autotuning flow control based on network perf is a common practice. To address the concern that you raise, what's needed is some receiver-local implementation control of the window scaling maximum. Right now quiche is quite basic - an application states the initial limit and we use that as the upper bound forever more. 

Autotuning can indeed help implementations be _more_ efficient or conservative with memory because they can afford to set lower initial limits for "light/slow" connections and allow the "heavy/faster" connections to make use of the freed up resource.

ghedo:
Let's be clear here though. The point of this PR is to fix a problem with the existing and currently used flow control update mechanism, the point is _not_ to completely change it. That is out of topic for this PR.

Sparika:
I did the same quick and dirty test I've done in #983, and this time I've plotted the `max_data` and `len` of the stream. The first graph shows what's happening with the master branch, the second shows this PR. In the first graph, the linear increase in `max_data` is because the value is continuously increasing while on the second graphs, the stairs pattern shows that the value is updated quite less frequently.

image

src/lib.rs
```diff
@@ -6980,9 +6979,11 @@ mod tests {
 
         pipe.server.stream_recv(4, &mut buf).unwrap();
 
+        assert_eq!(pipe.server.streams.get(4).unwrap().recv.max_data_next(), 22);
+
         let frames = [frame::Frame::Stream {
```
Sparika, line 6984: @ghedo Hi Alessandro, 

I've a question regarding some of the unit tests such as this one that sends frames to the server. In these tests, a frame is sent to the server, but it has no effect on the unit tests since `pipe.server.stream_recv` is not called afterwards.

I believe that the server receives the frame but do nothing with it. In this example:
In the master branch, the first frame adds 7 bytes to the `max_data_next value` and so the MAX_STREAM_DATA frame that is received has a max value of 22 (15+7). The one byte from the second frame does not affect the `max_data_next` value.
In my PR, I have to send 9 bytes in the first frame to cause the server sending a MAX_STREAM_DATA frame (`len > max_data - max_data_incr == 9 > 15-7)`. Sending 7 bytes in the first frame and 2 in the second does not cause MAX_STREAM_DATA frame. 

I've left the frames in this test and in the test on line 6864 because I'm not sure of the intent here. Just wanted to raise the point. ;)

junhochoi:
> Supposing a symmetrical connection with a throughput equal to the initial max_data. Then after 1/2 RTT, the connection would have consumed `max_data/2`, triggering the increase `max_data = max_data + max_data/2` from the receiver. The MAX_DATA frame would arrive 1/2 RTT later at the sender, just in time for refreshing the limit for the sender. I've not run test or graphs for that part but that would surely be interesting.

Let's see the following example. I have the following dev environment:

- MacOS (as a client) -> Linux VM (Debian Buster) (as a server)
- server: running quiche-server at linux vm, from master branch.
- client: run two different version, your PR and master right before your PR (5afe2882)
- set linux qdisc as follows: this is to simulate 1000Mbit bandwidth/25ms rtt/0% packet loss. Note that this link's BDP is *3125000* (1000Mbit * 25ms)
```tc qdisc add dev enp2s1 root netem delay 25ms 0ms rate 1000Mbit limit 2116 reorder 0 loss 0```
- Note that client's max_data and max_data as default will be 10000000 and 1000000. (see `tools/apps/arc/args.rs`)

Now, run the client with the following option, downloading 100Mbyte file.

- with default option (--max-stream-data 1000000)
- with `--max-stream-data 4000000` (4000000 is just a higher number than the BDP above).

Here is what I get (total download time as seconds):

- this PR + default: *6.031s*
- this PR + max-stream-data 4000000: 2.276s
- master branch + default: 3.666s
- master branch + max-stream-data 4000000: 2.304s

As I expected, current PR seems to start to throttle where current BDP > initial max_stream_data (max_data is not considered here because current default is pretty high - 10M - no impact on the current test)

On the other hand, we can see the problem of current master branch here but its behavior can be a separate discussion.

junhochoi:
@Sparika I understand sending less frames can save the number of packets going back to the sender. (note that # of saved frames > # of saved packets because we need to send ACK and other frames to the sender) but from my perspective it should be balanced with the performance as well.

Sparika:
My latest commit implemented changes requested by @ghedo. Is there anything I should change to get the PR accepted?

ghedo:
Thanks for the contribution. The fixed-increment flow-control scheme changed by this PR was replaced by the window-based flow-control implementation in commit `e79ea93bf`. The old fields no longer exist and current code derives credit updates from the receive window and consumed data. Closing this PR as superseded.

PR #995, opened 2021-07-05
stream: in-order incoming data fast path

ghedo:
When processing incoming stream data, a whole bunch of logic is
dedicated to handling overlaps with data already received. This involves
looping over all buffered data.

However, when the incoming range buffer is past the buffered data range,
there is clearly no overlap, so the looping logic can be skipped.


LPardue (approved):
The diff makes this change look much bigger than it really is. LGTM

PR #996, opened 2021-07-05
stream: use FxHash for stream maps

ghedo:
Stream IDs are just numbers, so using the default HashMap hasher is kind
of pointless and unnecessarily expensive. FxHash is supposedly faster
for this kind of input.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ log = { version = "0.4", features = ["std"] }
 libc = "0.2"
 libm = "0.2"
 ring = "0.16"
+rustc-hash = "1"
 lazy_static = "1"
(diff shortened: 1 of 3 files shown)


LPardue:
maybe even no hash would work....  

https://crates.io/crates/nohash

junhochoi:
Do you have a number how much this PR can reduce CPU resource or speed up?

ghedo:
@LPardue yeah, good idea, we can probably just implement our own `StreamIdHasher` as well, so we don't need to add an external dependency. Closing this for now.

@junhochoi haven't measured this one yet, but the default hasher shows up in flamegraphs a bit all over the place. It's not a huge amount, but if it's easy enough to reduce then I think we should do it (if anything it would reduce noise in flamegraphs).

PR #999, opened 2021-07-06
qlog: convert numbers to actual numbers

LPardue:
In qlog-01, many of the fields that carry numbers were basically defined
as string in order to accommodate perceived problems with JSON and 64-bit
numbers. QUIC is rife with potentially 64-bit numbers but in practice,
many of these fields will contain values that are perfectly legal for
the duration of a connection. In later versions of qlog, many of the
fields have been changed to actual number types.

This change updates many qlog fields and API parameters to be closer the
qlog specifications today by using numbers e.g. stream IDs are now u64
rather than String . There general several benefits to this, we no
longer need to create owned Strings to pass around, we don't waste
effort creating strings that are immediately serialized again by the
streaming serializer, we save space in the serialized logs because
numbers are no longer surrounded with pointless quotes, and we save
space in the buffered serializer.

This change also builds on the recent update to qlog time units.
Several fields in the metrics_updated event are now represented as f32
which encodes milliseconds with decimal precision.

While working through the commit, I was affected by the overloading of
String for qlog numbers and bytes see
https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#name-bytes.
In order to capture all the cases where we explicitly want to log bytes,
I added a simple type definition, `Bytes`, for disambiguation. In
future, we might substitute this for a more complex type that can
satisfy the qlog requirements.

### DIFF
--- a/tools/qlog/src/lib.rs
+++ b/tools/qlog/src/lib.rs
@@ -92,7 +92,7 @@
 //!     Some("Example qlog trace".to_string()),
 //!     Some("Example qlog trace description".to_string()),
 //!     Some(qlog::Configuration {
-//!         time_offset: Some("0".to_string()),
+//!         time_offset: Some(0.0),
 //!         original_uris: None,
 //!     }),
 //!     None,
@@ -122,7 +122,7 @@
 //! #     Some("Example qlog trace".to_string()),
 //! #     Some("Example qlog trace description".to_string()),
 //! #     Some(qlog::Configuration {
-//! #         time_offset: Some("0".to_string()),
+//! #         time_offset: Some(0.0),
 //! #         original_uris: None,
 //! #     }),
 //! #     None
@@ -140,8 +140,7 @@
 //!     Some(&dcid),
 //! );
 //!
-//! let frames =
-//!     vec![qlog::QuicFrame::crypto("0".to_string(), "1000".to_string())];
+//! let frames = vec![qlog::QuicFrame::crypto(0, 0)];
 //!
 //! let event = qlog::event::Event::packet_sent_min(
 //!     qlog::PacketType::Initial,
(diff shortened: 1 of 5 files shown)


tools/qlog/src/lib.rs
```diff
@@ -1194,8 +1195,8 @@ pub enum EventData {
     ServerListening {
         ip_v4: Option<String>,
         ip_v6: Option<String>,
-        port_v4: u64,
-        port_v6: u64,
+        port_v4: u32,
```
junhochoi, line 1198: Can we use `u16` for port numbers? (same for L1213/1214)

tools/qlog/src/lib.rs
```diff
@@ -1349,26 +1350,26 @@ pub enum EventData {
...
+        timer_granularity: Option<u16>,
+        initial_rtt: Option<f32>,
 
-        max_datagram_size: Option<u64>,
+        max_datagram_size: Option<u32>,
         initial_congestion_window: Option<u64>,
-        minimum_congestion_window: Option<u64>,
-        loss_reduction_factor: Option<u64>,
-        persistent_congestion_threshold: Option<u64>,
+        minimum_congestion_window: Option<u32>,
```
junhochoi, line 1360: Since initial_congestion_window is u64, I think we also use u64 here for consistency.

tools/qlog/src/lib.rs
LPardue: these are taken straight from https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-server_listening

tools/qlog/src/lib.rs
LPardue: Taken straight from https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-parameters_set-2

tools/qlog/src/lib.rs
junhochoi: ah ok. I will look qlog definitions. lgtm

junhochoi (approved):

PR #1000, opened 2021-07-06
crypto: use BoringCrypto for AEAD operations

ghedo:
This replaces the use of ring::aead with the BoringCrypto AEAD API.

While the ring AEAD API is nicer and less verbose, BoringCrypto's
`EVP_AEAD_CTX_seal_scatter()` allows passing plaintext data in two
separate buffers ("in" and "extra_in"), which will enable later
optimizations to avoid copying plaintext stream data in the output
packet buffer before encryption.

Other crypto operations like HKDF and QUIC header protection are still
implemented using ring, though this might change in the future (if
anything, because building and using 2 different crypto libraries at the
same time is kind of stupid).

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -24,8 +24,13 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::mem::MaybeUninit;
+
 use ring::aead;
 use ring::hkdf;
+
+use libc::c_int;
+use libc::c_void;
 
 use crate::Error;
 use crate::Result;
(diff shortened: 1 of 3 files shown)


src/crypto.rs
```diff
@@ -241,30 +272,47 @@ impl Seal {
...
+        let rc = unsafe {
+            EVP_AEAD_CTX_seal_scatter(
+                &self.ctx,                  // ctx
+                buf.as_mut_ptr(),           // out
+                buf[in_len..].as_mut_ptr(), // out_tag
+                &mut out_tag_len,           // out_tag_len
+                tag_len + extra_in_len,     // max_out_tag_len
+                nonce[..].as_ptr(),         // nonce
+                nonce.len(),                // nonce_len
+                buf.as_ptr(),               // in
```
LPardue, line 304:
```suggestion
                buf.as_ptr(),               // inp
```

to be pedantic

LPardue (approved):

PR #1001, opened 2021-07-06
recovery: only log qlog metrics when fields change

LPardue:
Previously, when the qlog feature was enabled, quiche would log all
recovery metrics (that it cares to log) any time there was a recovery
event. This is legal but increases duplication in the logs.

With this change, we make the event emission smarter by only generating
them if at least one of the event data fields has changed. In testing,
the file size is reduced by up to 25%, the qvis congestion view
continues to work just fine, and the sequence diagram has improved
readability,

builds on https://github.com/cloudflare/quiche/pull/999 so needs to come after it

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -146,6 +146,9 @@ pub struct Recovery {
 
     // RFC6937 PRR.
     prr: prr::PRR,
+
+    #[cfg(feature = "qlog")]
+    qlog_metrics: QlogMetrics,
 }
 
 impl Recovery {
@@ -230,6 +233,9 @@ impl Recovery {
             last_packet_scheduled_time: None,
 
             prr: prr::PRR::default(),
+
+            #[cfg(feature = "qlog")]
+            qlog_metrics: QlogMetrics::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/recovery/mod.rs
```diff
@@ -855,22 +861,18 @@ impl Recovery {
...
+            min_rtt: self.min_rtt,
+            smoothed_rtt: self.rtt(),
+            latest_rtt: self.latest_rtt,
+            rttvar: self.rttvar,
+            cwnd: self.cwnd() as u64,
+            bytes_in_flight: self.bytes_in_flight as u64,
+            ssthresh: self.ssthresh as u64,
+        };
+
+        self.qlog_metrics.maybe_update(qlog_metrics)
```
ghedo, line 875: Why not do something like:
```rust
if self.qlog_metrics != qlog_metrics {
    self.qlog_metrics = qlog_metrics;

    return Some(qlog::event::Event::metrics_updated(
        ...
    ));
}

None
```
instead of comparing each field manually?

src/recovery/mod.rs
```diff
@@ -1071,6 +1073,111 @@ fn sub_abs(lhs: Duration, rhs: Duration) -> Duration {
     }
 }
 
+// We don't need to log all qlog metrics every time there is a recovery event.
+// Instead, we can log only the MetricsUpdated event data fields that we care
+// about, only when they change. To support this, the QLogMetrics structure
+// keeps a running picture of the fields.
+#[derive(Default)]
+#[cfg(feature = "qlog")]
+struct QlogMetrics {
```
ghedo, line 1082: Could maybe just use the `qlog::event::Event` type rather than having a new struct?

ghedo (changes requested):

src/recovery/mod.rs
LPardue: We only want create an event with the changed fields in it. The common case is that bytes in flight changes often while the other field don't, so comparing the objects will probably return true and then we'd have to go through each one to figure out which to put in the new event that gets emitted.

src/recovery/mod.rs
LPardue: I tried something similar to your suggestion. The ugly thing is that the event data is `pub enum EventData` of variant `MetricsUpdated`, which made it harder to access and manipulate the actual data after the object was created. Furthermore, since qlog uses Option for every field, it makes comparison and update code a bit more onerous, especially where some fields use floats and equivalence comparison is more cumbersome.

junhochoi (dismissed):
I was wondering if there is a case when nothing changed but need to be reported (e.g. duplicated packet of ack), but it looks good to me.

LPardue:
I don't know if that's need for quiche-generated qlog itself - we know that packets arrive and that the absence of a metrics updated event means that nothing changed. 

I guess if it turns out useful to has something specific, we could make the behaviour a bit more configurable (i.e. an option to force flush an event).. But for the time being this seems ok.

src/recovery/mod.rs
ghedo: Oh, I missed that this would create partial events as well. There's probably a way to clean this up a bit and make it less repetitive, but we can look at it in a future change.

PR #1002, opened 2021-07-07
avoid calling is_in_early_data() unnecessarily

ghedo:
Since servers cannot send 0-RTT packets, `is_in_early_data()` will be
called unnecessarily. While it's not a huge cost, avoiding it is pretty
simple and as a side-effect, it also reduces noise in flamegraphs.


LPardue (approved):
nice shortcut ;)

PR #1006, opened 2021-07-08
Adaptive reordering fixes

ghedo:
Some additional fixes to adaptive reordering thresholds.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -215,11 +215,12 @@ fn on_packet_acked(
     if r.congestion_recovery_start_time.is_some() {
         let new_lost = r.lost_count - r.cubic_state.prior.lost_count;
 
-        if r.congestion_window < r.cubic_state.prior.congestion_window &&
-            new_lost < RESTORE_COUNT_THRESHOLD
-        {
-            rollback(r);
-            return;
+        if new_lost < RESTORE_COUNT_THRESHOLD {
+            let did_rollback = rollback(r);
+
+            if did_rollback {
+                return;
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/recovery/cubic.rs
```diff
@@ -213,13 +213,12 @@ fn on_packet_acked(
...
-        if let Some(prior) = &r.cubic_state.prior {
-            let new_lost = r.lost_count - prior.lost_count;
+        let new_lost = r.lost_count - r.cubic_state.prior.lost_count;
 
-            if r.congestion_window < prior.congestion_window &&
-                new_lost < RESTORE_COUNT_THRESHOLD
-            {
-                rollback(r);
+        if new_lost < RESTORE_COUNT_THRESHOLD {
+            let did_rollback = maybe_rollback(r);
```
junhochoi, line 219: What's the reason behind this change? Here we can safely assume checkpoint() is already called (it's already in recovery episode), calling rollback() doesn't need additional check.

ghedo: See the second commit's message. This is just to make the rollback avoidance if prior cwnd is smaller more generic, so each called to `rollback()` doesn't need to check this (i.e. we need this for adaptive reordering too).

ghedo: So basically this moves the ``if r.congestion_window < prior.congestion_window` into `rollback()` directly, but here we need to check if the rollback actually happened (or at least that's what the previous code did), hence `maybe_rollback()`.

junhochoi: I see. If so, why not adding `bool` type to `rollback()` and merge with `maybe_rollback()`?

ghedo: No real reason. I just thought this might be slightly smaller, but your way is fine to me too. Updated commit to do that.

src/recovery/cubic.rs
```diff
@@ -213,13 +213,12 @@ fn on_packet_acked(
...
 
-            if r.congestion_window < prior.congestion_window &&
-                new_lost < RESTORE_COUNT_THRESHOLD
-            {
-                rollback(r);
+        if new_lost < RESTORE_COUNT_THRESHOLD {
+            let did_rollback = rollback(r);
+
+            if did_rollback {
                 return;
```
junhochoi, line 222:
```suggestion
            if rollback(r) {
                return;
```

junhochoi (approved):

src/recovery/cubic.rs
ghedo: So, I did that originally, but then clippy complained that the `if`s could be merged together, though I found:
```rust
if new_lost < RESTORE_COUNT_THRESHOLD && rollback() {
```
kind of hard to read (as the `rollback()` call looks a bit hidden, and would be easy to pass over when reading), so I did it this way to silence clippy.

junhochoi: no problem then. 👍

PR #1012, opened 2021-07-15
add DATAGRAM APIs to work with Vec<u8> directly

ghedo:
These can be used by applications to avoid copying DATAGRAM data, and
instead provide or receive an owned Vec<u8> directly.

This also prevents an intermediate copy when writing HTTP/3 DATAGRAMs,
as well as when receiving DATAGRAM frames.

### DIFF
--- a/src/dgram.rs
+++ b/src/dgram.rs
@@ -46,13 +46,14 @@ impl DatagramQueue {
         }
     }
 
-    pub fn push(&mut self, data: &[u8]) -> Result<()> {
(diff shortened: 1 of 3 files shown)


LPardue:
blast from the past! 

I think we started with vec and debated back and forth before picking slices and copies. We should have just supported both to start with :smile:

LPardue (approved):

junhochoi (approved):

PR #1013, opened 2021-07-15
qlog: update to qlog 02, support event importance levels

LPardue:
This change updates the qlog crate to support the qlog 02 schemas.

The major change compared to qlog 01 is that the EventFields design has
been entirely removed. The loose concept of an event has been replaced
by a concrete Event type. Previously, the qlog crate defined a loosely
coupled data model and provided a bunch of helper methods in the `event`
module to help users craft well-formed, semantically correct objects.
With qlog 02, the relationships become much tighter. The `event` module
helpers no longer added much value, and introduced a layer of
indirection from the actual qlog data types. For these reasons, in this
change the `event` module has been deleted and client code is expected
to construct event object directly. For example, previously client code
might have used the `qlog::event::Event::packet_sent()` method to
construct an object that would serialize to a qlog `packet_sent` event.
With this change, clients need to instead build an instance of
`qlog::EventData::PacketSent` and then create a `qlog::Event` for
serialization. The crate docs examples have all been updated.

In additional, many qlog event types have been updated to match their
qlog 02 schema definitions. Event fields may have been renamed, retyped,
moved or deleted. Check the specification out for full details.

As part of this work, support for event importance levels has also been
added. Each event now has a level - Core, Base, or Extra. The field
value is not included in the serialized qlogs. Instead it is used to
determine what events are actually written by the streaming serializer.
The `qlog::QlogStreamer` object must now be constructed with a
`qlog::ImportanceLogLevel`, which controls the events that get
serialized.

The quiche library and apps have been updated to support the qlog crate
changes. By default, quiche will log at Base importance, which includes
events at Core and Base level. This shouldn't change the default logging
behaviour if apps enable a writer with `set_qlog()`. A new method,
`set_qlog_level()` has been added to change the logging level. For
instance, lowering the level to Core will avoid `data_moved` events
being written out, which reduces qlog size at the cost of losing some
fidelity of quiche stream operations.


LPardue:
This is a boulder to review, sorry. But once the event fields came out, everything explodes and needs fixing. You probably want to start from `tools/qlog/src/lib.rs` and work your way up.

Basically, I've ditched all the helper methods and tried to make things more compiler-friendly. Users of the qlog crate can focus on making type-specific `EventData` objects and sticking them into qlog so that it does most of the heavy lifting. From any type of EventData, we can reverse engineer the qlog type, category and importance. That's what all the blocks of `From` trait impls are about. 

The changes to quiche itself are pretty minimal, we just have to replace those old helper functions with explicit object creation. Building the structs directly means it becomes a bit easier to read that code.

tools/qlog/src/lib.rs
```diff
@@ -920,20 +923,258 @@ pub enum EventType {
...
-    Data(EventData),
+    #[serde(skip)]
+    ty: EventType,
 }
 
-#[derive(Serialize, Clone)]
+impl Event {
+    /// Returns a new `Event` object with the provided data at the default time.
+    pub fn new(data: EventData) -> Self {
+        Self::with_time(0.0, data)
```
ghedo, line 1051: It looks like in most cases the `time` field is set to `0`, so what is its purpose?

src/lib.rs
```diff
@@ -5572,38 +5627,51 @@ impl TransportParams {
     /// Creates a qlog event for connection transport parameters and TLS fields
     #[cfg(feature = "qlog")]
     pub fn to_qlog(
-        &self, owner: qlog::TransportOwner, version: u32, alpn: &[u8],
-        cipher: Option<crypto::Algorithm>,
-    ) -> qlog::event::Event {
-        let ocid = qlog::HexSlice::maybe_string(
+        &self, owner: qlog::TransportOwner, cipher: Option<crypto::Algorithm>,
+    ) -> qlog::EventData {
```
ghedo, line 5631: Why not return `qlog::Event` directly? Would save having to call `qlog::Event::new` from the caller.

src/recovery/mod.rs
```diff
@@ -914,7 +914,7 @@ impl Recovery {
     }
 
     #[cfg(feature = "qlog")]
-    pub fn maybe_qlog(&mut self) -> Option<qlog::event::Event> {
+    pub fn maybe_qlog(&mut self) -> Option<qlog::EventData> {
```
ghedo, line 917: Same, why not return `qlog::Event` directly? Would save having to call `qlog::Event::new` from the caller.

ghedo (changes requested):
Haven't really looked at all the changes yet, but added a couple of comments before I forget.

tools/qlog/src/lib.rs
LPardue: Agree it looks a bit weird. The `time` field is super important for qlog. I started with it as mandatory param but realised the quiche code handling of events meant that the time value was always 0 to start and then always updated later. On reflection, I think we can probably tweak quiche's behaviour to just construct the Event with the time on a single pass. I'll go back and tweak things in this PR.

LPardue: Please take a look at the new commit I just pushed. I've refactored things a bit to make the code for building Events less convoluted as a user.

src/lib.rs
LPardue: In the new commit, I stuck with returning EventData here so that the call site can drive the population of the event time field, which depends on access to some of the connection's parameters. Attempting to construct the event in to_qlog would require passing more parameters, which didn't seem much clearer to me.

src/recovery/mod.rs
LPardue: same answer :)

junhochoi:
> A new method, set_qlog_level() has been added to change the logging level.

Can you update quiche-apps (probably a new command line option?) to control the logging level?
I prefer if it can be controlled via environment variable (e.g. QLOGLEVEL?) but it's a different request.

LPardue:
> Can you update quiche-apps (probably a new command line option?) to control the logging level?

I'd like to do that in a follow up PR

tools/qlog/src/lib.rs
ghedo: So, it seems like the `add_event_with_instant()` calls were replaced with `add_event()`, but then all the relative time calculation code is duplicated in quiche itself, why?

LPardue: The main benefit is that it avoids having to pass a mutable object to be updated inside the qlog crate. 

A secondary benefit is that it gives flexibility to create events at arbitrary timepoints. Which could help us log events (server listening, etc) that occur before before the QUIC connection proper.

src/lib.rs
```diff
@@ -1290,21 +1299,75 @@ macro_rules! push_frame_to_pkt {
...
+            if let Some($qlog_streamer_ref) = &mut $qlog.streamer {
                 $body
             }
         }
     }};
 }
 
+/// Executes the provided body if the qlog feature is enabled, quiche has been
+/// configured with a log writer, the event's importance is within the
+/// confgured level, and the qlog start time is set.
```
ghedo, line 1319: Nit:
```suggestion
/// confgured level.
```

tools/qlog/src/lib.rs
```diff
@@ -52,23 +52,8 @@
...
-//! for qlog analysis tools. For information is available at
-//! https://tools.ietf.org/html/draft-marx-qlog-main-schema-01#section-3.3.4
-//!
-//! In order to make using qlog a bit easier, this crate expects a qlog Event to
-//! consist of the following EventFields in the following order:
-//! [`EventField::RelativeTime`], [`EventField::Category`],
-//! [`EventField::Event`] and [`EventField::Data`]. A set of methods are
-//! provided to assist in creating a Trace and appending events to it in this
-//! format.
+//! A [`Trace`] contains metadata,such as the [`VantagePoint`] of capture and
```
ghedo, line 55:
```suggestion
//! A [`Trace`] contains metadata such as the [`VantagePoint`] of capture and
```

tools/qlog/src/lib.rs
```diff
@@ -101,16 +86,13 @@
...
-//! valid events to a trace. However, there is nothing preventing you from
-//! creating the events manually.
-//!
-//! The following example demonstrates how to log a QUIC packet
-//! containing a single Crypto frame. It uses the [`QuicFrame::crypto()`],
-//! [`packet_sent_min()`] and [`push_event()`] methods to create and log a
-//! PacketSent event and its EventData.
+//! The following example demonstrates how to log a qlog QUIC `packet_sent`
+//! event containing a single Crypto frame. It constructs the necessary elements
+//! of the [`Event`], the then appends it to the trace with [`push_event()`].
```
ghedo, line 93: I think this comment needs some reworking? Besides "the then", it later mentions `QuicFrame::crypto()` which was removed.

ghedo (dismissed):
Some final nits, otherwise lgtm :tada:

tools/qlog/src/lib.rs
```diff
@@ -101,16 +86,13 @@
...
+//! [`QuicFrame`] objects can be constructed manually, or using a helper method.
+//! In this case we use [`QuicFrame::crypto()`].
```
LPardue, line 95:
```suggestion
//! of the [`Event`], then appends it to the trace with [`push_event()`].
```

tools/qlog/src/lib.rs
LPardue: good spot. I fixed "the then" and just removed the part about frame construction because it should be obvious how to instantiate a struct.

ghedo (approved):

PR #1014, opened 2021-07-23
Fix typo in tests

jgalenson:
This patch was written by prb@google.com and davidben@google.com.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6547,7 +6547,7 @@ mod tests {
             .load_priv_key_from_pem_file("examples/cert.key")
             .unwrap();
         config
-            .set_application_protos(b"\x06proto1\06proto2")
+            .set_application_protos(b"\x06proto1\x06proto2")
             .unwrap();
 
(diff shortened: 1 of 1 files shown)


junhochoi (approved):

davidben:
You all may, after merging this, want to look into `set_application_protos`. It seems to, itself, try to parse the list, but doesn't notice syntax errors. After https://boringssl-review.googlesource.com/c/boringssl/+/46804, BoringSSL will do this for you, but your own parsers ought to also be correct.

Beyond that, taking the wire format is not a very friendly API. BoringSSL doing it comes from OpenSSL and is probably in part because "complex" types are annoying to represent in C. But I don't see why the Rust API doesn't take a `Vec<Vec<u8>>` or so.

ghedo:
Merged, thanks!

@davidben quiche exposes a C API that wraps the Rust one, so it made sense at the time to keep them consistent, even if that meant going with the less friendly way. Though we might reconsider this later on.

PR #1016, opened 2021-07-26
write DATAGRAM frame data directly into the output packet

ghedo:
Similarly to how STREAM and CRYPTO frames are written to the output
packet buffer, write the DATAGRAM frame header and data into the output
packet buffer directly.

This doesn't really make much of a difference in terms of performance,
but it enables a later change to use the crypto scatter seal API to
avoid unnecessary copies of data.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -155,6 +155,10 @@ pub enum Frame {
     Datagram {
         data: Vec<u8>,
     },
+
+    DatagramHeader {
+        length: usize,
+    },
 }
 
 impl Frame {
@@ -517,16 +521,12 @@ impl Frame {
             },
 
             Frame::Datagram { data } => {
-                let mut ty: u8 = 0x30;
+                encode_dgram_header(data.len() as u64, b)?;
 
-                // Always encode length
-                ty |= 0x01;
-
-                b.put_varint(u64::from(ty))?;
-
-                b.put_varint(data.len() as u64)?;
                 b.put_bytes(data.as_ref())?;
             },
+
+            Frame::DatagramHeader { .. } => (),
         }
 
         Ok(before - b.cap())
@@ -723,8 +723,14 @@ impl Frame {
 
             Frame::Datagram { data } => {
                 1 + // frame type
-                octets::varint_len(data.len() as u64) + // length
+                2 + // length, always encode as 2-byte varint
                 data.len() // data
+            },
+
+            Frame::DatagramHeader { length } => {
+                1 + // frame type
+                2 + // length, always encode as 2-byte varint
+                *length // data
             },
         }
     }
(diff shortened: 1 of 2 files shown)


src/frame.rs
```diff
@@ -1130,6 +1137,20 @@ pub fn encode_stream_header(
     Ok(())
 }
 
+pub fn encode_dgram_header(length: u64, b: &mut octets::OctetsMut) -> Result<()> {
+    let mut ty: u8 = 0x30;
+
+    // Always encode length
+    ty |= 0x01;
+
+    b.put_varint(u64::from(ty))?;
```
LPardue, line 1146: The hope with the old code was that we might go back one day and find a smart way to decide if we will include the datagram length or not.

The changes in this PR seem to be taking the decision to include length further, which is OK. I wonder if we should just use 0x31 type directly.

ghedo: This is similar to how `STREAM` frames are encoded as well. The original idea was to set the flags separately for clarity, so e.g. we can have a comment to describe what's going on.

LPardue: oh yeah, in that case its fine

src/lib.rs
```diff
@@ -2943,11 +2943,39 @@ impl Connection {
...
+                    let hdr_off = b.off();
+                    let hdr_len = 1 + // frame type
+                        2; // length, always encode as 2-byte varint
+
+                    if (hdr_len + len) <= left {
                         // Front of the queue fits this packet, send it
                         match self.dgram_send_queue.pop() {
                             Some(data) => {
-                                let frame = frame::Frame::Datagram { data };
+                                let (mut dgram_hdr, mut dgram_payload) =
```
LPardue, line 2954: I found this hard to understand first of all. The comments in the STREAM frame writing also apply here, which helped my understanding. It might be overkill to copy all of the commentary but spelling out that you start by writing the payload and then jump back to write the header would really help

src/lib.rs
```diff
@@ -2943,11 +2943,39 @@ impl Connection {
         {
             if let Some(max_dgram_payload) = self.dgram_max_writable_len() {
                 while let Some(len) = self.dgram_send_queue.peek_front_len() {
-                    if (len + frame::MAX_DGRAM_OVERHEAD) <= left {
```
LPardue, line 2946: you remove use of frame::MAX_DGRAM_OVERHEAD here but it is still used at L2940, is that intentional?

src/lib.rs
ghedo: I copied some of the comments, might as well copy all of them :D

src/lib.rs
ghedo: Yeah, we need _some_ value there, and we don't know what the front DATAGRAM's size is. We can probably rework this code a bit more to remove the constant from there too, but it seems better to leave that for a future change to avoid making this too big.

src/lib.rs
LPardue: thanks!

src/lib.rs
LPardue: WFM

LPardue (approved):

PR #1020, opened 2021-07-29
fix needless_borrow clippy warnings

ghedo:
There's a new Rust stable release, so it's time again to fix all the new clippy warnings.


LPardue (approved):
Clippy always has our back

junhochoi (approved):

PR #1024, opened 2021-08-16
h3: return InternalError in a case of stream read logic failure

LPardue:
Applications read H3 streams via the `poll()` method. The quiche H3
stream struct provides indirect access to a backing buffer that is used
to read data from the QUIC stream, depending on the context of the H3
stream (control, request, etc.). These methods provide some guardrails
against inconsistent state.

This change extends the current behavior by returning a
quiche::h3:InternalError if the buffer reading code detects a logical
mismatch.


junhochoi (approved):

LPardue:
overtaken by events

PR #1025, opened 2021-08-18
expose local_error to applications

### LINKED ISSUE #1023: Determining why a connection is closed / expose local_error
On a quic server, while garbage collecting connections, I would like to determine why the connection closed.
Most of this was made possible by exposing `peer_error`, but I am not sure how to differentiate between a client timeout or a `CONNECTION_CLOSE` sent by the server, without maintaining state external to quiche.

As a possible solution, I have dealt with this by exposing `local_error` (fc17d0a), and checking both peer_error and local_error to determine the reason for close.

kevinjohna6:
Closes #1023.

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -388,6 +388,14 @@ bool quiche_conn_peer_error(quiche_conn *conn,
                             const uint8_t **reason,
                             size_t *reason_len);
 
+// Returns true if a connection error was queued or sent, and updates the provided
+// parameters accordingly.
+bool quiche_conn_local_error(quiche_conn *conn,
+                            bool *is_app,
+                            uint64_t *error_code,
+                            const uint8_t **reason,
+                            size_t *reason_len);
+
 // Initializes the stream's application data.
 //
 // Stream data can only be initialized once. Additional calls to this method
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4458,6 +4458,21 @@ impl Connection {
         self.peer_error.as_ref()
     }
 
+    /// Returns the error [`close()`] was called with, or internally
+    /// created quiche errors, if any.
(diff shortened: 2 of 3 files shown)


include/quiche.h
```diff
@@ -388,6 +388,14 @@ bool quiche_conn_peer_error(quiche_conn *conn,
                             const uint8_t **reason,
                             size_t *reason_len);
 
+// Returns true if a connection error was queued, and updates the provided
```
LPardue, line 391:
```suggestion
// Returns true if a connection error was queued or sent, and updates the provided
```

src/lib.rs
```diff
@@ -4458,6 +4458,20 @@ impl Connection {
         self.peer_error.as_ref()
     }
 
+    /// Returns the error [`close()`] was called with, if any.
+    ///
+    /// Note that a `Some` return value does not necessarily imply
```
LPardue, line 4463: We should also note that quiche can internally close connections, so a value might be returned even if the user/app didn't call `close()` explicitly.

src/lib.rs
```diff
@@ -10644,6 +10658,23 @@ mod tests {
         );
     }
 
+    #[test]
+    fn local_error() {
+        let mut pipe = testing::Pipe::default().unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        assert_eq!(pipe.server.close(true, 0x1234, b"hello!"), Ok(()));
```
LPardue, line 10666: Please add a check before `close()` is called too.

LPardue (changes requested):

LPardue (approved):

ghedo:
Merged, thanks @kevinjohna6!

PR #1027, opened 2021-08-24
h3: add support for additional HEADERS

LPardue:
HTTP/3 request and response message exchanges allow multiple HEADERS
frames to be sent, which allows for trailer sections or non-final
responses. Previously, quiche has only supported generating a single
HEADERS frame per request or response, and didn't support receiving
them.

This change adds support for sending and receiving multiple HEADERS
frames on a request stream. It enforces the sequencing rules defined in
https://datatracker.ietf.org/doc/html/draft-ietf-quic-http-34#section-4.1.
Sending additional HEADERS is done using the new
`send_additional_headers()` method. On the receive side, rather than add
a new Event enum variant for these additional uses of HEADERS, this
change reuses Event::Headers. Previosuly, this event included a
`has_body` boolean, which has been renamed to `more_frames` to better
align with how HTTP/3 streams work.

### DIFF
--- a/nginx/nginx-1.16.patch
+++ b/nginx/nginx-1.16.patch
@@ -2569,7 +2569,7 @@ index 000000000..d7c4fbf1d
 +        return;
 +    }
 +
-+    stream->in_closed = !quiche_h3_event_headers_has_body(ev);
++    stream->in_closed = !quiche_h3_event_headers_has_more_frames(ev);
 +
 +    ngx_http_v3_run_request(stream->request);
 +}
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -2631,7 +2631,7 @@ impl Connection {
                     return Err(Error::FrameUnexpected);
                 }
 
-                // Servers reject too many HEADERS frames
+                // Servers reject too many HEADERS frames.
                 if let Some(s) = self.streams.get_mut(&stream_id) {
                     if self.is_server && s.headers_received_count() == 2 {
                         conn.close(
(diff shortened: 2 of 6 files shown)


LPardue:
Oh wow. The build failure is in `data_event_rearm`, there's a part of the test that says `// Send more data, then HEADERS, then more data.` - that isn't a valid sequence, so I guess this change has revealed a latent bug. Can we just remove that part of the test, or does it invalidate the rearm code/test?

LPardue:
@ghedo since there is no way to send DATA frames after headers, I tried addressing the broken test by sending more data on a different stream and basically interleaving operations to see if the DATA event rearms properly.

A different approach might be, if we add an event for PRIORITY_UPDATE, to interleave that with DATA events. If we like the sound of that I can replace the `fix data_event_rearm test` commit with an alternative. Or we change it later, or whatever.

nginx/nginx-1.16.patch
```diff
@@ -50,7 +50,7 @@ index 2c7af1040..abf920bae 100644
 @@ -25,6 +25,10 @@ if [ $USE_OPENSSL = YES ]; then
      . auto/lib/openssl/conf
  fi
- 
+
```
junhochoi, line 53: looks like most of nginx diff is not intended other than new API use?

LPardue: bah, editor stripping whitespace

quiche/src/h3/mod.rs
```diff
@@ -2568,6 +2631,20 @@ impl Connection {
                     return Err(Error::FrameUnexpected);
                 }
 
+                // Servers reject too many HEADERS frames
```
ghedo, line 2634:
```suggestion
                // Servers reject too many HEADERS frames.
```

ghedo (approved):

PR #1033, opened 2021-09-02
recovery: hystart++ draft 03

junhochoi:
- Implement hystart++ draft 03 for cubic and reno
- A new mode Conservative Slow Start (CSS) introduced instead of
  Limited Slow Start (LSS). The difference is CSS is still under
  slow start (cwnd < ssthresh)
- Add hystart.reset() hook to be called in collapse_cwnd() CC
  hook to start in a clean state.
- Update hystart test in cubic and test 2 mode transitions:
  SS -> CSS -> SS and SS -> CSS -> CA.
- For various rtt, simply use Duration.
- Bump minimum rust version to 1.53 due to the use of Duration::MAX.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -230,31 +230,36 @@ fn on_packet_acked(
         r.bytes_acked_sl += packet.size;
 
         if r.bytes_acked_sl >= r.max_datagram_size {
-            r.congestion_window += r.max_datagram_size;
+            if r.hystart.in_css(epoch) {
+                r.congestion_window +=
+                    r.hystart.css_cwnd_inc(r.max_datagram_size);
+            } else {
+                r.congestion_window += r.max_datagram_size;
+            }
+
             r.bytes_acked_sl -= r.max_datagram_size;
         }
 
-        if r.hystart.enabled() &&
-            epoch == packet::EPOCH_APPLICATION &&
-            r.hystart.try_enter_lss(
-                packet,
-                r.latest_rtt,
-                r.congestion_window,
-                now,
-                r.max_datagram_size,
-            )
-        {
+        if r.hystart.on_packet_acked(
+            epoch,
+            packet,
+            r.latest_rtt,
+            r.congestion_window,
+            now,
+            r.max_datagram_size,
+        ) {
+            // Exit to congestion avoidance if CSS ends.
             r.ssthresh = r.congestion_window;
         }
     } else {
         // Congestion avoidance.
         let ca_start_time;
 
-        // In LSS, use lss_start_time instead of congestion_recovery_start_time.
-        if r.hystart.in_lss(epoch) {
-            ca_start_time = r.hystart.lss_start_time().unwrap();
+        // In CSS, use css_start_time instead of congestion_recovery_start_time.
+        if r.hystart.in_css(epoch) {
+            ca_start_time = r.hystart.css_start_time().unwrap();
 
-            // Reset w_max and k when LSS started.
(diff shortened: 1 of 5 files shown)


src/recovery/hystart.rs
```diff
@@ -80,6 +88,26 @@ impl Hystart {
...
+            current_round_min_rtt: Duration::MAX,
+
+            css_baseline_min_rtt: Duration::MAX,
+
+            ..Default::default()
+        }
+    }
+
+    pub fn reset(&mut self) {
+        *self = Self {
```
ghedo, line 102: Can probably just do:
```rust
*self = Self::new(self.enabled);
```
to avoid duplicating.

src/recovery/hystart.rs
```diff
@@ -88,93 +116,107 @@ impl Hystart {
...
+                    self.css_start_time = None;
+                    self.css_round_count = 0;
                 }
             }
         }
 
-        self.lss_start_time.is_some()
+        // Check if we reached the end of the round.
+        if let Some(end_pkt_num) = self.window_end {
+            if packet.pkt_num >= end_pkt_num {
```
ghedo, line 192: What if a packet later than `window_end` is acked before `window_end` (e.g. due to re-ordering)? Is it ok to end the round anyway? The draft doesn't seem to talk about this.

ghedo (changes requested):
lgtm, just added a couple of minor comments.

src/recovery/hystart.rs
junhochoi: thanks. fixed

src/recovery/hystart.rs
junhochoi: The algorithm in the draft is based on TCP, so the definition of the round is different in QUIC. When reordering is enough to trigger a congestion event we will end SS/CSS any way. If this is tolerable within reordering threshold, I don't have too much concern about it.

src/recovery/cubic.rs
```diff
@@ -230,31 +230,35 @@ fn on_packet_acked(
         r.bytes_acked_sl += packet.size;
 
         if r.bytes_acked_sl >= r.max_datagram_size {
-            r.congestion_window += r.max_datagram_size;
+            r.congestion_window += if r.hystart.in_css(epoch) {
```
LPardue, line 233: nit: I find this code construction a bit hard to grok. Could maybe split the conditional into out

src/recovery/hystart.rs
```diff
@@ -44,32 +44,40 @@ const MIN_RTT_THRESH: Duration = Duration::from_millis(4);
```
LPardue, line 44: I can't add a nit comment at the right [line](https://github.com/cloudflare/quiche/pull/1033/files#diff-591f0b10723101e64a47f8f97c5c655656c4c39e621a2c4dc0232348f4174c35R31)

but can we change the document reference to https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-hystartplusplus-03 please? This got really confusing  when reviewing because now the tcpm and balasubramanian docs are both -03

LPardue:
lgtm modulo nits

src/recovery/cubic.rs
junhochoi: updated.

src/recovery/hystart.rs
junhochoi: Updated URL. Thanks!

LPardue (approved):

ghedo (approved):

PR #1034, opened 2021-09-02
build: drop support for building with Android NDK < 19

ghedo:
It looks like newer versions of BoringSSL need a more recent version of
the toolchain (https://github.com/cloudflare/quiche/pull/1028), and
being able to update BoringSSL is more important than supporting ancient
Android versions.


LPardue (dismissed):

README.md
```diff
@@ -290,78 +290,45 @@ the BoringSSL directory with the ``QUICHE_BSSL_PATH`` environment variable:
...
 ```bash
  $ export ANDROID_NDK_HOME=/usr/local/share/android-ndk
 ```
 
-- Install the Rust toolchain for Android architectures needed:
+Then the Rust toolchain for the Android architectures needed can be installed as
+follows:
 
 ```bash
  $ rustup target add aarch64-linux-android arm-linux-androideabi armv7-linux-androideabi i686-linux-android x86_64-linux-android
```
junhochoi, line 308:
```suggestion
 $ rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
```

`arm-linux-androideabi` is not supported by cargo-ndk.

junhochoi (approved):

PR #1037, opened 2021-09-07
[WIP] Implement stateless reset

### LINKED ISSUE #850: Stateless Reset Feature
Is the stateless reset already implemented (draft section 10.4)? I seem to be able to find code related to the tokens but I am not sure where and how it is processed. If it is implemented could someone give me a nudge to where the tokens are generated and where they are processed.
If it is not implemented yet, is there a schedule for the feature in the future?

su225:
Fixes #850

Implements issuing and detection of stateless reset tokens and takes appropriate actions. Like the rest of the implementation here, currently it is assumed that only one stateless reset token per connection. (In future, if multiple connection IDs are supported then it would be a hash table lookup from given ID to the token).

Tasks
- [X] Implement stateless reset token detection if provided
- [ ] Unit tests for all the above items

Finally refine and make the implementation secure :) 

Signed-off-by: su225 <suchithjn22@gmail.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1815,6 +1815,16 @@ impl Connection {
             return Err(Error::Done);
         }
 
+        // Check if the peer has sent the stateless reset token. If yes, then
+        // move the connection to the draining state immediately.
+        // https://www.rfc-editor.org/rfc/rfc9000.html#section-10.3.1-5
+        if let Some(ref token) = self.local_transport_params.stateless_reset_token {
+            if buf.ends_with(token) {
+                self.draining_timer = Some(now + (self.recovery.pto() * 3));
+                return Err(Error::Done);
+            }
+        }
+
         let mut b = octets::OctetsMut::with_slice(buf);
 
         let mut hdr =
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -433,6 +433,9 @@ pub enum Error {
 
     /// Error in congestion control.
     CongestionControl,
+
+    /// Error when the peer sends the stateless reset token
+    StatelessReset
 }
 
 impl Error {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1815,6 +1815,16 @@ impl Connection {
             return Err(Error::Done);
         }
 
+        // Check if the peer has sent the stateless reset token. If yes, then
+        // move the connection to the draining state immediately.
+        // https://www.rfc-editor.org/rfc/rfc9000.html#section-10.3.1-5
+        if let Some(ref token) = self.local_transport_params.stateless_reset_token {
```
ghedo, line 1821: From https://www.rfc-editor.org/rfc/rfc9000.html#section-10.3.1-2:
>  However, the comparison MUST be performed when the first packet in an incoming datagram either cannot be associated with a connection or cannot be decrypted.

so this should happen way later, and only after we failed to decrypt the packet, and also probably if e.g. decoding the header fails.

It might be easier to just move this to the `Err` case in `recv()`, and first check if `done == 0` (which means this is the first packet in the datagram).

ghedo (changes requested):
Thanks for the PR! I know this is still WIP, but added a comment.

Also, before you go ahead and implement the remaining tasks, it would be better to split this across multiple PRs, so e.g. one PR for detecting resets + tests (like this one), one for issuing tokens, etc...

src/lib.rs
su225: Thanks for the review. I'll fix that.

su225:
@ghedo - Sure. In this PR, I'll just limit it to detecting stateless reset tokens and taking appropriate actions.

src/lib.rs
su225: By the way, it also has the following - https://www.rfc-editor.org/rfc/rfc9000.html#section-10.3.1-5
> If the last 16 bytes of the datagram are identical in value to a stateless reset token, the endpoint MUST enter the draining period and not send any further packets on this connection.

I interpreted it as - if the last 16 bytes are equal to the stateless reset token **no matter whether the packet could be processed or not**, then draining period should be entered.

PR #1039, opened 2021-09-09
h3: store all received SETTINGS parameters and expose them

LPardue:
SETTINGS frames can contain parameters with unknown or reserved identifiers.
quiche has no direct usage for these parameters and previously discarded them.
With this change we store all received parameters in the order that they were
sent, and add a new `peer_settings_raw()` to access them,

### DIFF
--- a/examples/http3-client.c
+++ b/examples/http3-client.c
@@ -98,6 +98,14 @@ static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
     ev_timer_again(loop, &conn_io->timer);
 }
 
+static int for_each_setting(uint64_t identifier, uint64_t value,
+                           void *argp) {
+    fprintf(stderr, "got HTTP/3 SETTING: %lu=%lu\n",
+            identifier, value);
+
+    return 0;
+}
+
 static int for_each_header(uint8_t *name, size_t name_len,
                            uint8_t *value, size_t value_len,
                            void *argp) {
@@ -109,6 +117,7 @@ static int for_each_header(uint8_t *name, size_t name_len,
 
 static void recv_cb(EV_P_ ev_io *w, int revents) {
     static bool req_sent = false;
+    static bool settings_received = false;
 
     struct conn_io *conn_io = w->data;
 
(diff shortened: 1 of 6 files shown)


punkeel (approved):

examples/http3-client.c
```diff
@@ -244,6 +253,16 @@ static void recv_cb(EV_P_ ev_io *w, int revents) {
...
             }
 
+            if (!settings_received) {
+                int rc = quiche_h3_for_each_setting(conn_io->http3,
+                                                    for_each_setting,
+                                                    NULL);
+
+                if (rc == 0) {
+                    settings_received = true;
+                }
```
junhochoi, line 263: since `quiche_h3_for_each_setting()` always return 0, I think there is no need to check the return value.
```suggestion
                quiche_h3_for_each_setting(conn_io->http3,
                                                    for_each_setting,
                                                    NULL);

                settings_received = true;
```

junhochoi (approved):

src/h3/mod.rs
```diff
@@ -604,11 +604,22 @@ pub enum Event {
     GoAway,
 }
 
+/// An HTTP/3 setting.
+#[derive(Clone, Debug, PartialEq)]
+pub struct RawSetting {
```
ghedo, line 609: Why not just use `(u64, u64)` tuples? Would save us from having to expose another type.

src/h3/mod.rs
```diff
@@ -1410,6 +1423,13 @@ impl Connection {
         Ok(())
     }
 
+    /// Gets the raw settings from peer including unknown and reserved types.
+    ///
+    /// The order of settings is the same as received in the SETTINGS frame.
+    pub fn peer_settings_raw(&self) -> Option<Vec<RawSetting>> {
+        self.peer_settings.raw.clone()
```
ghedo, line 1430: Could just return a slice here, rather than `Vec`, so the application can decide if it wants to clone or not.

src/h3/ffi.rs
```diff
@@ -84,6 +84,31 @@ pub extern fn quiche_h3_conn_new_with_transport(
     }
 }
 
+#[no_mangle]
+pub extern fn quiche_h3_for_each_setting(
+    conn: &h3::Connection,
+    cb: extern fn(identifier: u64, value: u64, argp: *mut c_void) -> c_int,
+    argp: *mut c_void,
+) -> c_int {
+    let rc = match conn.peer_settings_raw() {
```
ghedo, line 93: Don't really need to assign the value to a variable and then return the variable, can just leave the `match` without the final `;` to have the value returned automatically.

ghedo (changes requested):
Added some comments, but looks ok overall.

LPardue:
applied suggestions, cheers

ghedo (approved):

PR #987, opened 2021-06-29
apps: Linux GSO support for quiche-server

junhochoi:
When GSO can be enabled (on a supported linux platform),
now quiche-server sends multiple packets using sendmsg()
with GSO. For non-supported platforms should work same as before.

PR #1040, opened 2021-09-10
lib: window-based flow control

junhochoi:
This PR reimplments #529 which tried to implement a QUIC flow
control specified in https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit?usp=sharing

It's basically window-based approach where window size is determined
by current receiving speed.

Since lot of things changed from #529, I tried to minimize the
changes in the existing code. Most of the logic is in src/flowcontrol.rs
and it will replace max_data and max_data_next logic.

### DIFF
--- a/src/flowcontrol.rs
+++ b/src/flowcontrol.rs
@@ -1,0 +1,220 @@
+// Copyright (C) 2021, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::time::Duration;
+use std::time::Instant;
+
+// When autotuning the receiver window, decide how much
+// we increase the window.
+const WINDOW_INCREASE_FACTOR: u64 = 2;
+
+// When autotuning the receiver window, check if the last
+// update is within RTT * this constant.
+const WINDOW_TRIGGER_FACTOR: u32 = 2;
+
+#[derive(Default, Debug)]
+pub struct FlowControl {
+    /// Total consumed bytes by the receiver.
+    consumed: u64,
+
+    /// Flow control limit.
+    max_data: u64,
+
+    /// The receive window. This value is used for updating
+    /// flow control limit.
+    window: u64,
+
+    /// The maximum receive window.
+    max_window: u64,
+
+    /// Last update time of max_data for autotuning the window.
+    last_update: Option<Instant>,
+}
+
+impl FlowControl {
+    pub fn new(max_data: u64, window: u64, max_window: u64) -> Self {
+        Self {
+            max_data,
+
+            window,
+
+            max_window,
+
+            ..Default::default()
+        }
+    }
+
+    /// Returns the current window size.
+    pub fn window(&self) -> u64 {
+        self.window
+    }
+
+    /// Returns the current flow limit.
+    pub fn max_data(&self) -> u64 {
+        self.max_data
+    }
+
+    /// Update consumed bytes.
+    pub fn add_consumed(&mut self, consumed: u64) {
+        self.consumed += consumed;
+    }
+
+    /// Returns true if the flow control needs to update max_data.
+    ///
+    /// This happens when the available window is smaller than the half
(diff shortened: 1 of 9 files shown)


junhochoi:
This PR mostly improves client download performance.

- Network: 1Gbit link, 20ms latency
- Server: running quiche-server (master + #987 branch)
- Client:
master branch: quiche-client downloading 1G file: 23.397s
```
[2021-09-10T04:26:49.594639000Z INFO  quiche_apps::common] 1/1 response(s) received in 23.397971791s, closing...
```
this branch: quiche-client download 1G file: 10.062s
```
[2021-09-10T04:30:40.581162000Z INFO  quiche_apps::common] 1/1 response(s) received in 10.062495518s, closing...
```

For comparison, h2load with quic patch on the same client:
```
$ h2load --npn-list h3-29 -n 1 https://buster:8443/1000m.dat
finished in 9.53s, 0.10 req/s, 104.88MB/s
```

src/lib.rs
```diff
@@ -359,6 +359,16 @@ const MAX_UNDECRYPTABLE_PACKETS: usize = 10;
 
 const RESERVED_VERSION_MASK: u32 = 0xfafafafa;
 
+// The default size of the receiver connection flow control window.
+const DEFAULT_CONNECTION_WINDOW: u64 = 48 * 1024;
+
+// The maximum size of the receiver connection flow control window.
+const MAX_CONNECTION_WINDOW: u64 = 24 * 1024 * 1024;
+
+// How much larger the connection flow control window need to be larger than
```
LPardue, line 368:
```suggestion
// How much larger the connection flow control window needs to be than
```

src/stream.rs
```diff
@@ -49,6 +52,12 @@ const SEND_BUFFER_SIZE: usize = 5;
 #[cfg(not(test))]
 const SEND_BUFFER_SIZE: usize = 4096;
 
+// The default size of the receiver stream flow control window.
+const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
```
LPardue, line 56: Does this default override the configuration value set by the application in e.g. `set_initial_max_stream_data_bidi_local()` and `set_initial_max_stream_data_bidi_remote()`?

Maybe instead we need to take into account those values. And that would mean we need separate flow control behaviour for bidi streams initiated locally, bidi streams initiated remotely, and uni stream initiated remotely.

src/lib.rs
```diff
@@ -359,6 +359,16 @@ const MAX_UNDECRYPTABLE_PACKETS: usize = 10;
 
 const RESERVED_VERSION_MASK: u32 = 0xfafafafa;
 
+// The default size of the receiver connection flow control window.
+const DEFAULT_CONNECTION_WINDOW: u64 = 48 * 1024;
```
LPardue, line 363: Should the default really be based on `set_initial_max_data()`?

src/flowcontrol.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /// Update consumed bytes.
+    pub fn add_consumed(&mut self, consumed: u64) {
+        self.consumed += consumed;
+    }
+
+    /// Returns true if the flow control needs to update max_data.
+    ///
+    /// This happens when the available window is smaller than the half
```
LPardue, line 83:
```suggestion
    /// This happens when the available window is smaller than half
```

src/flowcontrol.rs
```diff
@@ -0,0 +1,214 @@
...
+    ) {
+        if let Some(last_update) = self.last_update {
+            if now - last_update < rtt * WINDOW_TRIGGER_FACTOR {
+                self.window =
+                    cmp::min(self.window * WINDOW_INCREASE_FACTOR, max_window);
+            }
+        }
+    }
+
+    /// Make sure the lower bound of the window is same to
```
LPardue, line 116:
```suggestion
    /// Make sure the lower bound of the window is same as
```

src/flowcontrol.rs
```diff
@@ -0,0 +1,214 @@
...
+        assert_eq!(fc.should_update_max_data(), true);
+
+        let max_data_next = fc.max_data_next();
+        assert_eq!(max_data_next, consumed + w);
+
+        fc.update_max_data(Instant::now());
+        assert_eq!(fc.max_data(), max_data_next);
+
+        // Window size should be doubled.
+        fc.autotune_window(Instant::now(), Duration::from_millis(100), 100);
```
LPardue, line 189: Can we use fixed instants rather than `now()` just in case some timing issues pop up in CI?

src/lib.rs
```diff
@@ -359,6 +359,16 @@ const MAX_UNDECRYPTABLE_PACKETS: usize = 10;
...
+
+// The maximum size of the receiver connection flow control window.
+const MAX_CONNECTION_WINDOW: u64 = 24 * 1024 * 1024;
```
LPardue, line 366: We might want to make the maximum configurable by the application. But for now this is ok.

src/flowcontrol.rs
junhochoi: thanks. fixed

src/stream.rs
junhochoi: DEFAULT_{CONNECTION,STREAM}_WINDOW is a different value from initial stream/connection max_data. in the flowcontrol.rs there is `window` variable is limited by those default values and (so, max_data will be consumed + window). Since those are hardcoded, it's a good idea to make it configurable, but don't be confused with max_data transport parameter which defines initial limit only and nothing to do with window size.

src/lib.rs
junhochoi: See my comment on DEFAULT_STREAM_WINDOW above.

src/stream.rs
junhochoi: I added `config.set_max_connection_window()` and `config.set_max_stream_window()` API to set the max window size of the connection and the stream.

src/lib.rs
junhochoi: the max window (connection, stream) is now configurable.

src/stream.rs
ghedo: I'm still confused. In the code we have `cmp::min(max_data, DEFAULT_STREAM_WINDOW)`, which means tha if the application sets a lower initial stream data value (i.e. `max_data`) than the "default" value and a lower max window value, we will just ignore the application-set values and simply use `DEFAULT_STREAM_WINDOW` as the initial window size. And later, when autotune is done, the new window will just be `max_window`.

So it's still not very clear to me what the point of this constant is. Why not just use `max_data` without doing the `cmp::min()`?

ghedo: And while we are at it, why even do autotuning at all, and not just use `max_window` from the start?

ghedo: Actually, I just realized I was probably confusing min with max :facepalm: ...

junhochoi: `cmp::min()` is used for flow control window initialization when the max_data (whether it's connection or stream) is too small (e.g. tests) in such case, having a small window will make sense and it's helpful for passing most of existing tests without a large modification (I only changed 2 tests). In practive it doesn't have much meaning

junhochoi: > And while we are at it, why even do autotuning at all, and not just use `max_window` from the start?

Interesting. I think the reason to have a flow control window is to properly manage receiver buffer size. If it's too high each connection/stream might need more memory even for a small BDP connection and MAX_DATA update might not be delivered in time.

I see some of implementation use a very high max_data in the beginning (e.g. h2load + quic's initial max_data is  2^30-1 and window size is 2^26-1, but it's a test tool). In practical use I believe it's better to tune the window size dynamically.

ghedo (approved):

PR #1043, opened 2021-09-13
http3_test: remove unused fields

ghedo:
They make builds with latest Rust nightly sad.


LPardue:
This structure is basically a reverse engineer of the HTTP bin JSON, so its sad if we forget the available fields we could use for future testing. Can we just mark them unused rather than remove them outright?

ghedo:
Well, we can always add the fields back, but silencing works for me too, updated.

LPardue:
I think the challenge is remembering what HTTP bin supports, to know that we can use it to test :smile: . So allowing this dead code works for me!

LPardue (approved):

PR #1044, opened 2021-09-15
add method to check if stream can hold more data

ghedo:
- [ ] Need to figure out how to test.

Regarding the `DATA` case, I think in the future we could change `send_body()` to not check the capacity and not slice the input buffer at all (so the frame header's length field will just be the full input buffer length), and then simply expect the application to do the right thing and write the remaining bytes once the stream becomes writable (applications _should_ be doing this already anyway to avoid dropping data, but they don't _have to_ do it), this way the capacity check would fall to the transport layer. The `send_headers()` case would still require "atomic" writes though.

Another idea would be to have a `stream_send_atomic()` method that either writes all data, or none of it, but that would require copying h3 headers/data into a single buffer, unless this is implemented in multiple steps ("start atomic send" -> "write data" -> "write data" -> ... -> "end atomic send"), but that API would be much more complicated than what this PR does.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3848,6 +3848,50 @@ impl Connection {
         stream.is_readable()
     }
 
+    /// Returns true if the stream has enough send capacity.
+    ///
+    /// When `len` more bytes can be buffered into the given stream's send
+    /// buffer, `true` will be returned, `false` otherwise.
+    ///
+    /// In the latter case, if the additional data can't be buffered due to
+    /// flow control limits, the peer will also be notified.
+    ///
+    /// If the specified stream doesn't exist (including when it has already
+    /// been completed and closed), the [`InvalidStreamState`] error will be
+    /// returned.
+    ///
+    /// In addition, if the peer has signalled that it doesn't want to receive
+    /// any more data from this stream by sending the `STOP_SENDING` frame, the
+    /// [`StreamStopped`] error will be returned.
+    ///
+    /// [`InvalidStreamState`]: enum.Error.html#variant.InvalidStreamState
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3848,6 +3848,50 @@ impl Connection {
...
+    /// In addition, if the peer has signalled that it doesn't want to receive
+    /// any more data from this stream by sending the `STOP_SENDING` frame, the
+    /// [`StreamStopped`] error will be returned.
+    ///
+    /// [`InvalidStreamState`]: enum.Error.html#variant.InvalidStreamState
+    /// [`StreamStopped`]: enum.Error.html#variant.StreamStopped
+    #[inline]
+    pub fn stream_writable(
+        &mut self, stream_id: u64, len: usize,
+    ) -> Result<bool> {
```
LPardue, line 3872: It might be nice to return which window is too small. Thinking purely for audit purpose. Rather than complicate error type, could substitute bool for an enum like {Ok, StreamBlocked, ConnectionBlocked}.

LPardue:
I think this PR is a good compromise for how the h3 layer operates today - later changes will need some rethinks. For instance, send_headers() already (without this) is going to have problems when we come around to needing to emit QPACK encoder stream instructions too.

src/lib.rs
ghedo: That seems like a fairly niche use case, so I'd rather keep this API simple. But we could add `max_tx_data` / `max_rx_data` to `Stats` so that an application can figure this out on its own if it wants to (we already have cwnd there as well).

LPardue: Yeah that's cool

junhochoi (approved):

PR #1046, opened 2021-09-16
avoid counting reset streams' unsent data towards flow control limit

ghedo:
Currently the send-side flow control allowance is enforced and reduced
whenever stream data is buffered in the corresponding stream's send
buffer (i.e. when the application calls `stream_send()`). This is to
limit the amount of buffered data to what the connection can actually
take and avoid over-buffering.

However whenever a stream is reset (e.g. `STOP_SENDING` is received, or
`stream_shutdown()` is called), the stream's final size will include all
data that was previously buffered, even if not all of it was actually
sent on the wire yet.

To aovid that, we can make the final size be the offset of the first
unsent byte, and then "claw back" the remaining data from the flow
control allowance when a stream is reset.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2329,10 +2329,7 @@ impl Connection {
         }
 
         // Update send capacity.
-        self.tx_cap = cmp::min(
-            self.recovery.cwnd_available() as u64,
-            self.max_tx_data - self.tx_data,
-        ) as usize;
+        self.update_tx_cap();
 
         self.recv_count += 1;
 
@@ -3803,7 +3800,14 @@ impl Connection {
             },
 
             Shutdown::Write => {
-                let final_size = stream.send.shutdown()?;
+                let (final_size, unsent) = stream.send.shutdown()?;
+
+                // Claw back some flow control allowance from data that was
+                // buffered but not actually sent before the stream was reset.
+                self.tx_data = self.tx_data.saturating_sub(unsent);
+
+                // Update send capacity.
+                self.update_tx_cap();
 
                 self.streams.mark_reset(stream_id, true, err, final_size);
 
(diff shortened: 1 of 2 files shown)


LPardue (approved):
Changes look good. It would be nice to have a test or two to check the RESET_STREAM final size and the internal tx_send and tx_cap values (before and after STOP_SENDING)  when we have unsent stream data.

ghedo:
@LPardue I added tests for both STOP_SENDING and stream_shutdown() cases that try to send data up to flow control limit, then reset with unsent data, and then verify that more data can be written (rather than checking `tx_cap` directly, but the idea is the same). I also checked that they fail with current master.

src/lib.rs
```diff
@@ -8323,6 +8421,74 @@ mod tests {
...
+            Err(Error::Done)
+        );
+        assert_eq!(
+            pipe.server.stream_send(8, b"hello", false),
+            Err(Error::Done)
+        );
+
+        // Server shuts down stream.
+        assert_eq!(pipe.server.stream_shutdown(4, Shutdown::Write, 42), Ok(()));
+        assert_eq!(pipe.advance(), Ok(()));
```
LPardue, line 8477: Does this unintentionally cause the client to ack or MAX_DATA?

LPardue: In other words, are we sure the following send is allowed because we reset, and not because the client just granted more capacity

ghedo: `stream_shutdown()` drops buffers before anything is sent, and previously we only sent 5 bytes (less than half of the fc window), so there shouldn't be any MAX_DATA. But FWIW, while making the tests I was running with the `tx frm ...` logs enabled and I didn't see any MAX_DATA (the server does seem to send DATA_BLOCKED anyway though, but that should be harmless I think, and our code just ignores it anyway).

ghedo: I added `ssert_eq!(pipe.client.should_update_max_data(), false);` just to make sure. The value shouldn't change during `advance()` since there is no data being sent.

LPardue (approved):
:shipit:

PR #1055, opened 2021-09-28
cubic: no rollback during slow start

junhochoi:
While the stored cwnd/ssthresh is in slow start state, don't try
rollback. This is not to delay to exit slow start when there is a loss.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -385,6 +385,11 @@ fn checkpoint(r: &mut Recovery) {
 }
 
 fn rollback(r: &mut Recovery) -> bool {
+    // Don't go back to slow start.
+    if r.cubic_state.prior.congestion_window < r.cubic_state.prior.ssthresh {
+        return false;
+    }
+
     if r.congestion_window >= r.cubic_state.prior.congestion_window {
         return false;
     }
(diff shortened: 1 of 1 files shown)


LPardue (dismissed):

src/recovery/cubic.rs
```diff
@@ -385,6 +385,11 @@ fn checkpoint(r: &mut Recovery) {
 }
 
 fn rollback(r: &mut Recovery) -> bool {
+    // don't go back to slow start
```
ghedo, line 388:
```suggestion
    // Don't go back to slow start.
```

src/recovery/cubic.rs
```diff
@@ -928,7 +933,39 @@ mod tests {
             now + rtt + Duration::from_millis(5),
         );
 
-        // cwnd is restored to the previous one.
+        // This is from slow start, no rollback
```
ghedo, line 936:
```suggestion
        // This is from slow start, no rollback.
```

src/recovery/cubic.rs
```diff
@@ -928,7 +933,39 @@ mod tests {
...
+        assert_eq!(r.cwnd(), cur_cwnd);
+
+        let now = now + rtt;
+
+        // Trigger another congestion event
```
ghedo, line 941:
```suggestion
        // Trigger another congestion event.
```

src/recovery/cubic.rs
```diff
@@ -928,7 +933,39 @@ mod tests {
...
+        let rtt = Duration::from_millis(100);
+
+        let acked = vec![Acked {
+            pkt_num: 0,
+            // To exit from recovery
+            time_sent: now + rtt,
+            size: r.max_datagram_size,
+        }];
+
+        // Ack more than cwnd bytes with rtt=100ms
```
ghedo, line 958:
```suggestion
        // Ack more than cwnd bytes with rtt=100ms.
```

src/recovery/cubic.rs
```diff
@@ -928,7 +933,39 @@ mod tests {
...
+        r.update_rtt(rtt, Duration::from_millis(0), now);
+
+        // Trigger detecting sprurious congestion event
```
ghedo, line 961:
```suggestion
        // Trigger detecting sprurious congestion event.
```

ghedo (dismissed):
Added some nits but lgtm.

ghedo (dismissed):

junhochoi:
> Actually, I'm confused. The commit message says "don't rollback during slow start", but what the code does is avoiding rollback _to_ slow start. That is, the commit message seems to imply that the _current_ cwnd / ssthresh are used to determine whether to rollback or not, but the code uses the prior ones.
> 
> I think either the commit message or the code need to be updated to match each other.

Thanks. Current code is correct - I updated the commit message to match current behavior.

ghedo (approved):

PR #1056, opened 2021-09-28
cubic: proportional rollback threshold

junhochoi:
Currently a fixed 10 packets threshold is used for detecting a small
loss. When cwnd is very high (more than 1000 packets, for example),
this makes the detection less sensitive. New approach is to make a
rollback threshold relative to cwnd (currently 20%) to make it work
with various size of cwnd in real time.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -61,9 +61,12 @@ const BETA_CUBIC: f64 = 0.7;
 
 const C: f64 = 0.4;
 
-/// The packet count threshold to restore to the prior state if the
-/// lost packet count since the last checkpoint is less than the threshold.
-const RESTORE_COUNT_THRESHOLD: usize = 10;
+/// Threshold for rolling back state, as percentage of lost packets relative to
+/// cwnd.
+const ROLLBACK_THRESHOLD_PERCENT: usize = 20;
+
+/// Minimum threshold for rolling back state, as number of packets.
+const MIN_ROLLBACK_THRESHOLD: usize = 2;
 
 /// Default value of alpha_aimd in the beginning of congestion avoidance.
 const ALPHA_AIMD: f64 = 3.0 * (1.0 - BETA_CUBIC) / (1.0 + BETA_CUBIC);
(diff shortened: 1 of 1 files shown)


src/recovery/cubic.rs
```diff
@@ -61,9 +61,13 @@ const BETA_CUBIC: f64 = 0.7;
 
 const C: f64 = 0.4;
 
-/// The packet count threshold to restore to the prior state if the
-/// lost packet count since the last checkpoint is less than the threshold.
-const RESTORE_COUNT_THRESHOLD: usize = 10;
+/// The packet count threshold (%) in relative to cwnd, to roll back to
+/// the prior state if the lost packet count since the last checkpoint
+/// is less than the threshold.
```
LPardue, line 66: I find this sentence hard to parse but have no suggestion on improvements.

src/recovery/cubic.rs
```diff
@@ -210,12 +214,16 @@ fn on_packet_acked(
     // <https://tools.ietf.org/id/draft-ietf-tcpm-rfc8312bis-00.html#section-4.9>
     //
     // When the recovery episode ends with recovering
-    // a few packets (less than RESTORE_COUNT_THRESHOLD), it's considered
-    // as spurious and restore to the previous state.
+    // a few packets (less than cwnd / mss * ROLLBACK_THRESHOLD_PERCENT(%)), it's
+    // considered as spurious and restore to the previous state.
     if r.congestion_recovery_start_time.is_some() {
         let new_lost = r.lost_count - r.cubic_state.prior.lost_count;
+        let rollback_threshold = (r.congestion_window / r.max_datagram_size) *
```
LPardue, line 221: Is `rollback_threshold` a hypothetical value based on maximally packed packets?

junhochoi: Right. Initially I was thinking to use bytes_in_flight but sometimes it can be very low and didn't work well.

src/recovery/cubic.rs
```diff
@@ -61,9 +61,13 @@ const BETA_CUBIC: f64 = 0.7;
 
 const C: f64 = 0.4;
 
-/// The packet count threshold to restore to the prior state if the
-/// lost packet count since the last checkpoint is less than the threshold.
-const RESTORE_COUNT_THRESHOLD: usize = 10;
+/// The packet count threshold (%) in relative to cwnd, to roll back to
```
ghedo, line 64:
```suggestion
/// Threshold for rolling back state, as percentage of lost packets relative to cwnd.
```

src/recovery/cubic.rs
```diff
@@ -61,9 +61,13 @@ const BETA_CUBIC: f64 = 0.7;
...
+/// the prior state if the lost packet count since the last checkpoint
+/// is less than the threshold.
+const ROLLBACK_THRESHOLD_PERCENT: usize = 20;
+
+/// Minimum count of rollback threshold (packets)
```
ghedo, line 69:
```suggestion
/// Minimum threshold for rolling back state, in number of packets.
```

src/recovery/cubic.rs
```diff
@@ -61,9 +61,13 @@ const BETA_CUBIC: f64 = 0.7;
 
 const C: f64 = 0.4;
 
-/// The packet count threshold to restore to the prior state if the
-/// lost packet count since the last checkpoint is less than the threshold.
-const RESTORE_COUNT_THRESHOLD: usize = 10;
+/// The packet count threshold (%) in relative to cwnd, to roll back to
+/// the prior state if the lost packet count since the last checkpoint
```
ghedo, line 65:
```suggestion
```

src/recovery/cubic.rs
ghedo, line 66:
```suggestion
```

ghedo (dismissed):
LGTM, added some comments.

src/recovery/cubic.rs
ghedo: I added a suggestion. I don't think we need to repeat the whole thing, since the mechanism is described in the code where the constant is used.

ghedo (approved):

PR #1058, opened 2021-10-01
Add sys/socket.h to include list

Lukasa:
The quiche.h header files use struct sockaddr extensively, meaning that
they rely on having sys/socket.h types defined. This include is missing
in the quiche.h header file, sadly.

This patch adds that include, so that the quiche.h header file stands
alone.


ghedo (approved):

ghedo:
Merged, thanks @Lukasa!

PR #1065, opened 2021-10-18
apps: print header lists as strings not bytes

LPardue:
When quiche switched headers to bytes, some of our debug printing
started displaying the headers as bytes, which isn't helpful.

This change restores the header list display back to strings.


junhochoi (dismissed):

LPardue:
rebased after the repo layout changes, no other changes applied

ghedo (approved):

PR #1069, opened 2021-10-23
cc: debug info

junhochoi:
For congestion control specific debug log, add debug_string()
hook to the congestion control operations.

### DIFF
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -53,6 +53,7 @@ pub static CUBIC: CongestionControlOps = CongestionControlOps {
     checkpoint,
     rollback,
     has_custom_pacing,
+    debug_write,
 };
 
 /// CUBIC Constants.
@@ -434,6 +435,14 @@ fn rollback(r: &mut Recovery) -> bool {
 
 fn has_custom_pacing() -> bool {
     false
+}
+
+fn debug_write(r: &Recovery, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+    write!(
+        f,
+        "cubic={{ k={} w_max={} max_bytes_in_flight={} }} ",
+        r.cubic_state.k, r.cubic_state.w_max, r.cubic_state.max_bytes_in_flight,
+    )
 }
 
 #[cfg(test)]
--- a/src/recovery/cubic.rs
+++ b/src/recovery/cubic.rs
@@ -53,7 +53,7 @@ pub static CUBIC: CongestionControlOps = CongestionControlOps {
     checkpoint,
     rollback,
     has_custom_pacing,
-    debug_write,
+    debug_fmt,
 };
 
 /// CUBIC Constants.
@@ -437,7 +437,7 @@ fn has_custom_pacing() -> bool {
     false
 }
 
-fn debug_write(r: &Recovery, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+fn debug_fmt(r: &Recovery, f: &mut std::fmt::Formatter) -> std::fmt::Result {
     write!(
         f,
         "cubic={{ k={} w_max={} max_bytes_in_flight={} }} ",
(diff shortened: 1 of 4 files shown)


LPardue (dismissed):

src/recovery/mod.rs
```diff
@@ -1041,6 +1043,9 @@ impl std::fmt::Debug for Recovery {
             write!(f, "hystart={:?} ", self.hystart)?;
         }
 
+        // CC-specific debug info
+        write!(f, "{} ", (self.cc_ops.debug_string)(self))?;
```
ghedo, line 1047: Instead of returning a string, it seems like it would be better to pass the `std::fmt::Formatter` directly to the callback. This way the cb can call `write!()` directly, without having to allocate a temporary string that would then get copied into the formatter and discarded anyway.

ghedo (changes requested):

src/recovery/mod.rs
junhochoi: Updated as recommended. Thanks!

src/recovery/cubic.rs
```diff
@@ -53,6 +53,7 @@ pub static CUBIC: CongestionControlOps = CongestionControlOps {
     checkpoint,
     rollback,
     has_custom_pacing,
+    debug_write,
```
ghedo, line 56:
```suggestion
    debug_fmt,
```

src/recovery/cubic.rs
```diff
@@ -436,6 +437,14 @@ fn has_custom_pacing() -> bool {
     false
 }
 
+fn debug_write(r: &Recovery, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
ghedo, line 440:
```suggestion
fn debug_fmt(r: &Recovery, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```

src/recovery/mod.rs
```diff
@@ -993,6 +993,9 @@ pub struct CongestionControlOps {
     pub rollback: fn(r: &mut Recovery) -> bool,
 
     pub has_custom_pacing: fn() -> bool,
+
+    pub debug_write:
```
ghedo, line 997:
```suggestion
    pub debug_fmt:
```

src/recovery/mod.rs
```diff
@@ -1050,6 +1053,9 @@ impl std::fmt::Debug for Recovery {
             write!(f, "hystart={:?} ", self.hystart)?;
         }
 
+        // CC-specific debug info
+        (self.cc_ops.debug_write)(self, f)?;
```
ghedo, line 1057:
```suggestion
        (self.cc_ops.debug_fmt)(self, f)?;
```

src/recovery/reno.rs
```diff
@@ -47,6 +47,7 @@ pub static RENO: CongestionControlOps = CongestionControlOps {
     checkpoint,
     rollback,
     has_custom_pacing,
+    debug_write,
```
ghedo, line 50:
```suggestion
    debug_fmt,
```

src/recovery/reno.rs
```diff
@@ -149,6 +150,10 @@ fn has_custom_pacing() -> bool {
     false
 }
 
+fn debug_write(_r: &Recovery, _f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
ghedo, line 153:
```suggestion
fn debug_fmt(_r: &Recovery, _f: &mut std::fmt::Formatter) -> std::fmt::Result {
```

ghedo (changes requested):
One more nitpick, I think `debug_fmt` would be a better name for the callback, given that the `Debug` trait method is called `fmt` (I'll apply this change myself, since it's small enough).

ghedo (approved):

PR #1079, opened 2021-11-05
qlog event nits

LPardue:
Noticed during some interop testing, some of our event definition fields were a bit wrong compared to the spec. I've fixed them up here as individual commits but we can squash before merging as there's nothing really complicated going on.

### DIFF
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -856,7 +856,7 @@ impl Frame {
                 stream_id: *stream_id,
                 offset: data.off() as u64,
                 length: data.len() as u64,
-                fin: data.fin(),
+                fin: data.fin().then(|| true),
                 raw: None,
             },
 
@@ -869,7 +869,7 @@ impl Frame {
                 stream_id: *stream_id,
                 offset: *offset,
                 length: *length as u64,
-                fin: *fin,
+                fin: fin.then(|| true),
                 raw: None,
             },
 
--- a/src/frame.rs
+++ b/src/frame.rs
@@ -33,6 +33,8 @@ use crate::ranges;
 use crate::stream;
 
 #[cfg(feature = "qlog")]
+use qlog::events::quic::AckedRanges;
+#[cfg(feature = "qlog")]
 use qlog::events::quic::ErrorSpace;
 #[cfg(feature = "qlog")]
 use qlog::events::quic::QuicFrame;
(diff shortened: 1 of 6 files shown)


src/frame.rs
```diff
@@ -869,7 +874,7 @@ impl Frame {
                 stream_id: *stream_id,
                 offset: *offset,
                 length: *length as u64,
-                fin: *fin,
+                fin: if *fin { Some(true) } else { None },
```
LPardue, line 877: This is probably the most valuable change, saving some extra bytes for nearly every stream event that is logged

src/frame.rs
```diff
@@ -856,7 +861,7 @@ impl Frame {
                 stream_id: *stream_id,
                 offset: data.off() as u64,
                 length: data.len() as u64,
-                fin: data.fin(),
+                fin: if data.fin() { Some(true) } else { None },
```
ghedo, line 864:
```suggestion
                fin: data.fin().then(|| true),
```

src/frame.rs
ghedo, line 877:
```suggestion
                fin: fin.then(|| true),
```

tools/qlog/src/events/quic.rs
```diff
@@ -200,10 +201,19 @@ pub enum StreamState {
...
+    Transport,
+    #[serde(alias = "application_error")]
+    Application,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+#[serde(untagged)]
+pub enum ErrorCode {
+    TransportError(TransportError),
+    ApplicationError(ApplicationError),
```
ghedo, line 215: It seems a bit weird that the `quic` module references an `h3` type. Maybe `ErrorCode` and `ErrorSpace` should go in `mod.rs` instead?

ghedo (changes requested):
Small nits, but lgtm. Keeping separate commits is better I think, but need to squash the nits into the corresponding commits.

tools/qlog/src/events/quic.rs
LPardue: good spot! The current code models the qlog datamodel and `ApplicationError` is defined in the H3 spec, I agree it feels wrong but I'd say that it is probably a bug in the spec. To maintain consitency we could leave it as is now, and fixup when qlog changes?

see https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#name-applicationerror and https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-connectioncloseframe

LPardue: made https://github.com/quicwg/qlog/issues/183

src/frame.rs
LPardue: applied in the relevant commit, thanks

src/frame.rs
LPardue: applied in the relevant commit, thanks

tools/qlog/src/events/quic.rs
LPardue: I've dropped this commit from the PR because there isn't an easy resolution right now, we can fix it seperately pending the outcome

ghedo (approved):

PR #1083, opened 2021-11-12
cubic: remove direct dependency on libm crate

ghedo:


junhochoi:
This means we will need `-lm` for C applications now on?

ghedo:
@junhochoi we already do, don't we?
https://github.com/cloudflare/quiche/blob/master/examples/Makefile#L20
https://github.com/cloudflare/quiche/blob/master/extras/nginx/nginx-1.16.patch#L121

junhochoi:
Right but it's a workaround (#979 and #1042) for a recent linux build and actually macos is ok without `-lm`. This looks like rust + linux platform issue (hopefully fixed in later version?)

Or I'm ok if we remove libm this time and `-lm` by default for C apps -- then we need to update docs too.

ghedo:
Ok, thanks. It seems better to just keep things as they are now.

PR #1084, opened 2021-11-12
examples: fix MacOS X C examples and add to CI

junhochoi:
Fix a build error (not caught in linux) and add C example build
to macOS X CI.

### DIFF
--- a/examples/http3-client.c
+++ b/examples/http3-client.c
@@ -100,7 +100,7 @@ static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
 
 static int for_each_setting(uint64_t identifier, uint64_t value,
                            void *argp) {
-    fprintf(stderr, "got HTTP/3 SETTING: %lu=%lu\n",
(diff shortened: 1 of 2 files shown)


ghedo (approved):

examples/http3-client.c
```diff
@@ -100,7 +100,7 @@ static void flush_egress(struct ev_loop *loop, struct conn_io *conn_io) {
 
 static int for_each_setting(uint64_t identifier, uint64_t value,
                            void *argp) {
-    fprintf(stderr, "got HTTP/3 SETTING: %lu=%lu\n",
+    fprintf(stderr, "got HTTP/3 SETTING: %" PRId64 "=%" PRId64 "\n",
```
ghedo, line 103: These values are unsigned, so I think we need `PRIu64` here?
```suggestion
    fprintf(stderr, "got HTTP/3 SETTING: %" PRIu64 "=%" PRIu64 "\n",
```

ghedo (changes requested):

examples/http3-client.c
junhochoi: Fixed. Thanks

ghedo (approved):

PR #1085, opened 2021-11-15
Resolve "unknown specifier socklen_t" and "uses undefined struct 'timespec'" build errors on Windows.

krashanoff:
I noticed that while compiling this library into a Qt project with the MSVC toolchain that `socklen_t` and `struct timespec` are undefined. They are defined appropriately in `ws2tcpip.h` and `time.h` on Windows systems. This pull request adds the appropriate headers to the `WIN32` includes of `quiche.h`.

* Includes ws2tcpip.h to provide socklen_t on Windows machines.
* Includes time.h to provide timespec struct on Windows machines.

Resolves #1048

PR #1090, opened 2021-11-18
Include missing header

normanmaurer:
Motivation:

As the quiche_recv_info and quiche_send_info structs use timespec we need to also include the time header.

Modifications:

Include the missing header

Result:

quiche.h can be used on windows / linux / macOS without the need of the user to include the time header

### DIFF
--- a/include/quiche.h
+++ b/include/quiche.h
@@ -37,8 +37,10 @@ extern "C" {
 
 #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
 #include <winsock2.h>
+#include <time.h>
 #else
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


ghedo:
This is similar to https://github.com/cloudflare/quiche/pull/1085. Leaving Linux aside, that also includes `ws2tcpip.h` on Windows, do we need that?

normanmaurer:
@ghedo yes `ws2tcpip.h` is needed as well.

ghedo:
Ok, I rebased this on top of the other PR and merged as https://github.com/cloudflare/quiche/commit/3ab88cad6e67bb08307d02e19adb01f54e9dd334, thanks @normanmaurer!

PR #1091, opened 2021-11-18
prefer arrays for small fixed-size fields

ehaydenr:
Path Challenge/Response data is exactly 8 bytes. Stateless Reset Token is exactly 16 bytes. If we know the sizes in advance, there's no reason for the dynamic allocation.


LPardue (approved):

ghedo:
Merged, thanks @ehaydenr!

PR #1094, opened 2021-11-22
[WIP] software packet pacing

junhochoi:
This implements software-based pacing for quiche-server.

### DIFF
--- a/tools/apps/src/bin/quiche-server.rs
+++ b/tools/apps/src/bin/quiche-server.rs
@@ -37,6 +37,9 @@ use std::rc::Rc;
 
 use std::cell::RefCell;
 
+use std::time::Duration;
+use std::time::Instant;
+
 use ring::rand::*;
 
 use quiche_apps::args::*;
@@ -149,9 +152,12 @@ fn main() {
         //
         // TODO: use event loop that properly supports timers
         let timeout = match continue_write {
-            true => Some(std::time::Duration::from_secs(0)),
+            true => Some(Duration::from_secs(0)),
 
-            false => clients.values().filter_map(|c| c.conn.timeout()).min(),
+            false => clients
+                .values()
+                .filter_map(|c| c.conn.timeout().min(c.pacing_delay))
+                .min(),
         };
 
         poll.poll(&mut events, timeout).unwrap();
@@ -345,6 +351,7 @@ fn main() {
                     siduck_conn: None,
                     app_proto_selected: false,
                     bytes_sent: 0,
+                    pacing_delay: None,
                 };
 
                 clients.insert(scid.clone(), client);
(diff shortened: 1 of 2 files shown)


francoismichel:
Hello,

Thank you for the work, I love using quiche and find the implem very concise and clean ! 

I have a few questions regarding this PR and how quiche works regarding pacing. Nothing critical, I am nitpicking a bit because I would like to use quiche for running experiments.

- I see that you ignore pacing if the pacing delay is lower than 1ms. Does it mean that we here completely ignore pacing if the rate is lower than 1 pkt every ms (with 1350B packets this means ~10Mb/s which seems rather low to me ?) Is there a reason for doing so ? It looks to me that it might give odd results with pacing-based CC algos like BBR. 
- How to to cope with both ACK delays and packet pacing ? In quiche, when ACK frames are generated, the ACK delay is the time elapsed from the last received pkt time to the moment the ACK frame is created. The `conn.send()` method returns the instant the packet should be sent, so sending the packet at the correct timestamp would give inaccuracies when estimating the RTT especially for low bitrates (we do not take the pacing delay into the ACK delay). It seems in this PR that we send the pkt directly *and then* wait for the amount of time determined by the pacing delay. Is that to avoid this ACK delay inaccuracy problem ? It seems the best way to avoid that problem would be to have `conn.send()` return only packets to be sent out directly, but it would prevent us from using pacing at the UDP socket level (but I guess this trade-off between throughput and ACK delay accuracy could be tuned through config options)

What is the priority you give to merging this PR ?

junhochoi:
Thanks for your comment. This PR is still early and buggy and your points are all valid.
- Initial idea of ignoring <1ms is to merge some packets for GSO in the future, but the logic/rate can be changed.
- Packet should be sent at the given time, not later as it does today. this need to be fixed but hard to do in the current event loop.

If you are looking for pacing for now, see #1075 (has a linux kernel dependency but it should work)

> Hello,
> 
> Thank you for the work, I love using quiche and find the implem very concise and clean !
> 
> I have a few questions regarding this PR and how quiche works regarding pacing. Nothing critical, I am nitpicking a bit because I would like to use quiche for running experiments.
> 
> * I see that you ignore pacing if the pacing delay is lower than 1ms. Does it mean that we here completely ignore pacing if the rate is lower than 1 pkt every ms (with 1350B packets this means ~10Mb/s which seems rather low to me ?) Is there a reason for doing so ? It looks to me that it might give odd results with pacing-based CC algos like BBR.
> * How to to cope with both ACK delays and packet pacing ? In quiche, when ACK frames are generated, the ACK delay is the time elapsed from the last received pkt time to the moment the ACK frame is created. The `conn.send()` method returns the instant the packet should be sent, so sending the packet at the correct timestamp would give inaccuracies when estimating the RTT especially for low bitrates (we do not take the pacing delay into the ACK delay). It seems in this PR that we send the pkt directly _and then_ wait for the amount of time determined by the pacing delay. Is that to avoid this ACK delay inaccuracy problem ? It seems the best way to avoid that problem would be to have `conn.send()` return only packets to be sent out directly, but it would prevent us from using pacing at the UDP socket level (but I guess this trade-off between throughput and ACK delay accuracy could be tuned through config options)
> 
> What is the priority you give to merging this PR ?

francoismichel:
Okay thank you for your answer !

junhochoi:
Closing this one since this PR is not a right fix and need focus on #1138

PR #1100, opened 2021-11-29
convert repo to cargo workspace

ghedo:
This should make it easier to add new crates to the repo (e.g. no need to add CI jobs for each one), and just generally makes more sense than stuffing everything that is not quiche itself under `tools/`.

Pretty much all the changes are related to build scripts and related things. All the actual code is just moved around.


LPardue:
hard to review, if CI passes I hope that covers all bases :D

LPardue (approved):

ghedo:
So, I noticed that the doc tests were not being run, however this appears to have been a problem for a while (unrelated to this PR), but I fixed it by adding an additional `cargo test --doc` step because apparently this is a known cargo problem (https://github.com/rust-lang/cargo/issues/6669), that is you can't run examples and doc tests in the same command.

PR #1111, opened 2021-12-07
add C-API quiche_conn_set_token

lostk1ng:
As client may cache token for address validation for future connections, we may need to set token to the very first Initial packet, so let's expose this API.

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -282,6 +282,9 @@ void quiche_conn_set_qlog_fd(quiche_conn *conn, int fd, const char *log_title,
 // Configures the given session for resumption.
 int quiche_conn_set_session(quiche_conn *conn, const uint8_t *buf, size_t buf_len);
 
+// Configures the given token for Initial Packet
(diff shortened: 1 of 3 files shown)


lostk1ng:
@ghedo Can you review this?

ghedo:
Sorry for the delay. This looks ok, though we don't really expose a way for clients to get the tokens yet, so not sure how useful this would be in practice.

lostk1ng:
@ghedo We should wait for [NEW_TOKEN frame to be supported](https://github.com/cloudflare/quiche/issues/1112) and expose API for client to get NEW_TOKEN. Then we can use quiche_conn_set_token API to set token to Initial packet.

msoxzw:
I propose another design like [neqo](https://github.com/mozilla/neqo/blob/058313b69c04672285a4d42743f93beca53abef0/neqo-transport/src/connection/mod.rs#L581
), which merges resumption token and address validation token into one token, thereby benefiting existing applications without modifying codes.

PR #1121, opened 2021-12-28
Handle Key Update Request - RFC 9001/6

### LINKED ISSUE #1115: Connections stop working with AES-ciphers because of missing Key Update
If I am not missing something, it seems *quiche* does not implement "Key Update".

https://datatracker.ietf.org/doc/html/rfc9001#section-6

---

If a peer initiates a key update setting the Key Phase bit, I am not able to see any logic handling this.

https://github.com/cloudflare/quiche/blob/master/quiche/src/packet.rs#L274

The Key Phase bit in the header is never actually used.

---

Please note that for `AEAD_AES_128_GCM` and `AEAD_AES_256_GCM`, the confidentiality limit is 2^23 encrypted packets (https://datatracker.ietf.org/doc/html/rfc9001#section-6.6), so, in that case, after about `8388608` packets an endpoint might not be able to trasmit anymore.

Am I missing something?

BiagioFesta:
Logic for handling peer-initiated key update requests, as described in [RFC-9001#6](https://datatracker.ietf.org/doc/html/rfc9001#section-6).

In short, this includes the logic for handling key-update requested by peers.

It does not include logic for initiating key-update (based on AEAD confidentiality value). It might be proposed in a dedicated PR. However, most of the logic should be already included in this PR.

- [x] [Responding to a Key Update](https://datatracker.ietf.org/doc/html/rfc9001#section-6.2);
- [x] [Timing of Receive Key Generation](https://datatracker.ietf.org/doc/html/rfc9001#section-6.3);
- [x] [Sending with Updated Keys](https://datatracker.ietf.org/doc/html/rfc9001#section-6.4);
- [x] [Receiving with Different Keys](https://datatracker.ietf.org/doc/html/rfc9001#section-6.5);
  - [ ] [`SHOULD`] timeout retain old read keys (see `TODO` in the code).

This should fix #1115.

Therefore, QUIC-quiche implementation should work without any disruption even with other QUIC implementations using AEAD ciphers.

*Any feedback/suggestions are of course more than welcome.*

### DIFF
--- a/quiche/src/crypto.rs
+++ b/quiche/src/crypto.rs
@@ -131,45 +131,38 @@ impl Algorithm {
 pub struct Open {
     alg: Algorithm,
 
-    ctx: EVP_AEAD_CTX,
+    secret: Vec<u8>,
 
-    hp_key: aead::quic::HeaderProtectionKey,
+    header: HeaderProtectionKey,
 
-    nonce: Vec<u8>,
+    packet: PacketKey,
 }
 
 impl Open {
     pub fn new(
-        alg: Algorithm, key: &[u8], iv: &[u8], hp_key: &[u8],
+        alg: Algorithm, key: &[u8], iv: &[u8], hp_key: &[u8], secret: &[u8],
     ) -> Result<Open> {
         Ok(Open {
             alg,
 
-            ctx: make_aead_ctx(alg, key)?,
+            secret: Vec::from(secret),
 
-            hp_key: aead::quic::HeaderProtectionKey::new(
-                alg.get_ring_hp(),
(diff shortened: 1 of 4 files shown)


LPardue:
Thank you @BiagioFesta. We still have some folks on holidays so review will be a bit delayed.

BiagioFesta:
CI failures seem to not be related with the CR itself.

Any updates on the review?

ghedo:
Sorry for the (huge) delay on this. I made some modifications (mostly just clean-up and refactoring, but also implemented the discard timer), and created #1456. Closing this now.

PR #1131, opened 2022-01-19
stream: fix reception of data overlapping several chunks

### LINKED ISSUE #1130: Receiver loses data under specific reception patterns
While running some experiments with quiche endpoints at both ends, I experience some “connection deadlock” when the receiver looses quite a lot of packets. After investigation, it seems like the implementation of `RecvBuf::write()` does not properly handle some specific reception patterns, such as the one provided here in snippet (that you can insert in the `mod tests` of `stream.rs`).
```rust
#[test]
fn overlapping_end_twice_read() {
    let mut recv = RecvBuf::new(std::u64::MAX, DEFAULT_STREAM_WINDOW);
    assert_eq!(recv.len, 0);

    let mut buf = [0; 32];

    let first = RangeBuf::from(b"he", 0, false);
    let second = RangeBuf::from(b"ow", 4, false);
    let third = RangeBuf::from(b"rl", 7, false);
    let fourth = RangeBuf::from(b"helloworld", 0, true);

    assert!(recv.write(third).is_ok());
    assert_eq!(recv.len, 9);
    assert_eq!(recv.off, 0);
    assert_eq!(recv.data.len(), 1);

    assert!(recv.write(second).is_ok());
    assert_eq!(recv.len, 9);
    assert_eq!(recv.off, 0);
    assert_eq!(recv.data.len(), 2);

    assert!(recv.write(first).is_ok());
    assert_eq!(recv.len, 9);
    assert_eq!(recv.off, 0);
    assert_eq!(recv.data.len(), 3);

    assert!(recv.write(fourth).is_ok());
    assert_eq!(recv.len, 10);
    assert_eq!(recv.off, 0);
    assert_eq!(recv.data.len(), 6);

    let (len, fin) = recv.emit(&mut buf).unwrap();
    assert_eq!(len, 10);
    assert_eq!(fin, true);
    assert_eq!(&buf[..len], b"helloworld");
    assert_eq!(recv.len, 10);
    assert_eq!(recv.off, 10);

    assert_eq!(recv.emit(&mut buf), Err(Error::Done));
}
```
When incoming data overlaps several already received data chunks, [the condition covering buffer's end overlap](https://github.com/cloudflare/quiche/blob/a9ee599a7bf9f4365d4368c1b0d8e7b92bf2424f/quiche/src/stream.rs#L839) is `true` several times in a single `for` iteration. In such case, when reassigning to `tmp_buf` when it is `Some(lost_buf)`, `lost_buf` will be lost. While the sender sees its packet containing the overlapping STREAM frame acknowledged, the receiver is still missing some data chunks that it involuntary lost and the transfer cannot complete properly.

qdeconinck:
A sender might sometimes redimension its STREAM frames, making the
receiver get some out-of-order small chunks of data. At some point,
a sender might send a larger chunk overlapping several already
received one. In that case, the condition covering the case the new
buffer's end overlaps existing buffer can be entered several times
in the for loop. However, the `tmp_buf` variable might not be
`None` at assignment time, causing data loss.

Instead of relying on a one-slot `Option<RangeBuf>`, rely on a
small-sized `VecDeque<RangeBuf>`.

Fixes #1130.

### DIFF
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -803,11 +803,10 @@ impl RecvBuf {
             }
         }
 
-        let mut tmp_buf = Some(buf);
+        let mut tmp_bufs = VecDeque::with_capacity(2);
+        tmp_bufs.push_back(buf);
 
-        while let Some(mut buf) = tmp_buf {
-            tmp_buf = None;
-
+        'tmp: while let Some(mut buf) = tmp_bufs.pop_front() {
             // Discard incoming data below current stream offset. Bytes up to
             // `self.off` have already been received so we should not buffer
             // them again. This is also important to make sure `ready()` doesn't
@@ -827,7 +826,7 @@ impl RecvBuf {
                 for b in &self.data {
                     // New buffer is fully contained in existing buffer.
                     if buf.off() >= b.off() && buf.max_off() <= b.max_off() {
-                        return Ok(());
+                        continue 'tmp;
                     }
 
                     // New buffer's start overlaps existing buffer.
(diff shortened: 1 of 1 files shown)


ghedo (approved):

ghedo:
Merged in https://github.com/cloudflare/quiche/commit/74a3b0473f8cd3eecfc1f6060c09711f67f07a6c, thank you @qdeconinck!

PR #1139, opened 2022-01-27
h3: initial support for PRIORITY_UPDATE frame

LPardue:
Introduce support for the two PRIORITY_UPDATE frames defined in
  https://datatracker.ietf.org/doc/draft-ietf-httpbis-priority/12/.
  
  This change only includes basic frame handling and verification.
  Acting on frame contents, such as generating a new event in `poll()` is
  left as a TODO.


LPardue:
May as well just do things under a single PR #1155. Closing this one.

PR #1145, opened 2022-01-31
Restore FFI functions to return TLS session and peer cert

ghedo:
There's a slight change to the Rust API as well, but I think this is better anyway (see commit message).


ghedo:
Did a quick test with the `http3-client.c` example, and this seems to work fine, and without valgrind complaining about it.

LPardue (approved):
LGTM, thanks

junhochoi (approved):

PR #1149, opened 2022-02-01
Use a simpler hasher for stream collections

LPardue:
This is a follow up to https://github.com/cloudflare/quiche/pull/996, which implements the alternative suggestion of using a locally-defined "no-op" hasher.

### DIFF
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -58,6 +58,39 @@ const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
 /// The maximum size of the receiver stream flow control window.
 pub const MAX_STREAM_WINDOW: u64 = 16 * 1024 * 1024;
 
+/// A simple no-op hasher for Stream IDs.
+///
+/// The QUIC protocol and quiche library guarantees stream ID uniqueness, so
+/// we can save effort by avoiding using a more complicated algorithm.
+#[derive(Default)]
+pub struct StreamIdHasher {
+    id: u64,
+}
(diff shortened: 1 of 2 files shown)


quiche/src/stream.rs
```diff
@@ -58,18 +58,49 @@ const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
...
+
+    #[inline]
+    fn write(&mut self, _: &[u8]) {
+        // We need a default write() for the trait but stream IDs will always
+        // be a u64 so we just delegate to write_u64.
+        unimplemented!()
+    }
+}
+
+pub type BuildStreamIdHasher = std::hash::BuildHasherDefault<StreamIdHasher>;
```
ghedo, line 89: We should also define a `StreamIdHashMap` or whatever, so we don't need to repeat `BuildStreamIdHasher` all over the place, and so we can also enforce the fact that the key is a `u64` so this doesn't get accidentally misused.

Something like:
```suggestion
type BuildStreamIdHasher = std::hash::BuildHasherDefault<StreamIdHasher>;

pub type StreamIdHashMap<V> = HashMap<u64, V, BuildStreamIdHasher>;
```
and then update all the places where the stream `HashMap` is used.

LPardue: Ah I tried something like that but got caught out by the variants all over the place. In my haste to push, I didn't think to make it genric. Good Idea, I'll do that.

ghedo (approved):

PR #1156, opened 2022-02-08
ffi: expose transport-related HTTP/3 errors

LPardue:
An HTTP/3 operation might fail due to a transport-related reason. This
usually happens if a stream write can't occur because the peer stopped
it. This should be expected behaviour and not a critical error for
applications. In Rust, the quiche::h3::Error::TransportError
encapsulates the underlying QUIC error, so applications can match only
quiche::Error and swallow certain types. In contrast, the FFI interface
collapses all HTTP/3 transport-related errors into a single value,
preventing FFI-based applications from being able to also swallow.

This change exposes the encapsulated QUIC transport errors to the FFI by
applying a simple numerical shift to the QUIC error codes.

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -631,56 +631,56 @@ enum quiche_h3_error {
     QUICHE_H3_ERR_VERSION_FALLBACK = -20,
 
     // QUIC transport had no more work to do.
-    QUICHE_H3_TRANSPORT_ERR_DONE = -1001,
+    QUICHE_H3_TRANSPORT_ERR_DONE = QUICHE_ERR_DONE - 1000,
 
     // The provided buffer is too short for QUIC transport.
-    QUICHE_H3_TRANSPORT_ERR_BUFFER_TOO_SHORT = -1002,
+    QUICHE_H3_TRANSPORT_ERR_BUFFER_TOO_SHORT = QUICHE_ERR_BUFFER_TOO_SHORT - 1000,
 
     // The provided packet cannot be parsed by QUIC transport because its
     // version is unknown.
-    QUICHE_H3_TRANSPORT_ERR_UNKNOWN_VERSION = -1003,
+    QUICHE_H3_TRANSPORT_ERR_UNKNOWN_VERSION = QUICHE_ERR_UNKNOWN_VERSION - 1000,
 
     // The provided packet cannot be parsed by QUIC transport because it
     // contains an invalid frame.
-    QUICHE_H3_TRANSPORT_ERR_INVALID_FRAME = -1004,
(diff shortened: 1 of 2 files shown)


junhochoi (commented):
This looks breaking change, not sure how many people already uses QUICHE_H3_ERR_TRANSPORT_ERROR.

LPardue:
What would break though? These codes are mainly just useful as a lookup when people encounter an error from quiche.

quiche/include/quiche.h
```diff
@@ -630,6 +629,58 @@ enum quiche_h3_error {
     // The requested operation cannot be served over HTTP/3. Peer should retry
     // over HTTP/1.1.
     QUICHE_H3_ERR_VERSION_FALLBACK = -20,
+
+    // QUIC transport had no more work to do.
+    QUICHE_H3_TRANSPORT_ERR_DONE = -1001,
```
ghedo, line 634: Instead of hard-coding the numbers, maybe do `QUICHE_ERR_DONE - 1000` or something like that?

It could even be a macro to avoid potential typos, e.g.

```c
enum quiche_h3_error {
    ...

    #define QUICHE_H3_TRANSPORT_ERR_DEF(err) \
        QUICHE_H3_TRANSPORT_ERR_##err = QUICHE_ERR_##err - 1000

    QUICHE_H3_TRANSPORT_ERR_DEF(DONE),

    QUICHE_H3_TRANSPORT_ERR_DEF(BUFFER_TOO_SHORT),

    ...
};
```

ghedo (changes requested):

quiche/include/quiche.h
LPardue: On one hand the macro is neat. On the other hand, it seems to push the requirement on applications to do the name expansion themselves i.e. when comparing or switching on a returned error code they still need to type out the name in full. 

So I've gone with the simpler suggestion of avoiding hardcoding. This will mean that if people print the error code value out, they can't easily grep for it in the header. However, the calculation is trivial enough that it shouldn't be too hard to figure out.

quiche/include/quiche.h
```diff
@@ -630,6 +629,58 @@ enum quiche_h3_error {
     // The requested operation cannot be served over HTTP/3. Peer should retry
     // over HTTP/1.1.
     QUICHE_H3_ERR_VERSION_FALLBACK = -20,
+
+    // QUIC transport had no more work to do.
```
ghedo, line 633: Nit: it just occurred to me that we don't really need to duplicate the decriptions here, could just be `// See QUICHE_ERR_...` or something like that, so it's easier to keep everything in sync in case we change something in the future.

ghedo (changes requested):

quiche/include/quiche.h
LPardue: good idea!

ghedo (approved):

PR #1157, opened 2022-02-11
h3: change priority type to byte array and use sfv

LPardue:
h3: change priority type and add optional parser

The `quiche::h3::send_response_with_priority()` method allows
applications to specify a stream priority that is used by quiche's
stream prioritization.

Previously, the priority was passed as a String value that was expected
to contain Extensible Priority parameters in a serialized Structured
Fields Dictionary format; see
https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority-12#section-4.
Internally, the String was manually parsed to extract the "u" and "i"
parameters.

With this change, we convert the priority parameter from a String to a
new `quiche::h3::Priority` type that holds only the parameters we use
inside quiche. Applications are now required to handle Extensible
Priorities field values themselves. A `TryFrom<&[u8]>` trait
implementation can be used by applications for parsing the serialized
format NB: in the C API, the equivalent is the
`parse_extensible_priority()` method. This relies on the sfv crate,
which can be included as an optional dependency for quiche (and thus
controlled via features). sfv offers a Structured Fields compliant
parser, if it fails to read the expected Extensible Priority format, an
error is returned. If the format is ok but values are outside the
allowed range, they get clamped. Using sfv already caught a few bugs in
the old approach and I've added test cases for those.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.53 as build
+FROM rust:1.54 as build
 
 WORKDIR /build
 
--- a/apps/Cargo.toml
+++ b/apps/Cargo.toml
@@ -19,6 +19,7 @@ boringssl-boring-crate = ["quiche/boringssl-boring-crate"]
 sfv = ["quiche/sfv"]
 
 default = ["qlog", "sfv"]
+
 [dependencies]
 docopt = "1"
 env_logger = "0.6"
--- a/apps/Cargo.toml
+++ b/apps/Cargo.toml
@@ -15,8 +15,10 @@ qlog = ["quiche/qlog"]
 # Use BoringSSL provided by the boring crate.
 boringssl-boring-crate = ["quiche/boringssl-boring-crate"]
 
-default = ["qlog"]
+# Enable sfv support.
+sfv = ["quiche/sfv"]
 
+default = ["qlog", "sfv"]
 [dependencies]
 docopt = "1"
 env_logger = "0.6"
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -798,7 +798,7 @@ ssize_t quiche_h3_send_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
 ssize_t quiche_h3_recv_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
                             uint64_t stream_id, uint8_t *out, size_t out_len);
 
-// Try to parse Extensible Priority field value.
+// Try to parse an Extensible Priority field value.
 int quiche_h3_parse_extensible_priority(uint8_t *priority,
                                             size_t priority_len,
                                             quiche_h3_priority *parsed);
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -768,6 +768,7 @@ typedef struct {
     size_t value_len;
 } quiche_h3_header;
 
+// Extensible Priorities parameters.
 typedef struct {
     uint8_t urgency;
     bool incremental;
@@ -800,8 +801,8 @@ ssize_t quiche_h3_recv_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
 
 // Try to parse an Extensible Priority field value.
 int quiche_h3_parse_extensible_priority(uint8_t *priority,
-                                            size_t priority_len,
-                                            quiche_h3_priority *parsed);
+                                        size_t priority_len,
+                                        quiche_h3_priority *parsed);
 
 // Returns whether the peer enabled HTTP/3 DATAGRAM frame support.
 bool quiche_h3_dgram_enabled_by_peer(quiche_h3_conn *conn,
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -768,6 +768,11 @@ typedef struct {
     size_t value_len;
 } quiche_h3_header;
 
+typedef struct {
+    uint8_t urgency;
+    bool incremental;
+} quiche_h3_priority;
+
 // Sends an HTTP/3 request.
 int64_t quiche_h3_send_request(quiche_h3_conn *conn, quiche_conn *quic_conn,
                                quiche_h3_header *headers, size_t headers_len,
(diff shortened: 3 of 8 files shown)


LPardue:
note that the other priority related PRs will need to be updated to reflect this change from String to bytes

LPardue:
Dockerfile bump required because of a new dependency failing with 1.53 (8 months old version)

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.53 as build
+FROM rust:1.58.1 as build
```
junhochoi, line 1: I think this is to match with the minimum version mentioned [here](https://github.com/cloudflare/quiche/blob/master/README.md?plain=1#L249)
If minimum version is changed we need to bump README too.

quiche/src/h3/mod.rs
```diff
@@ -2127,6 +2137,64 @@ fn grease_value() -> u64 {
     31 * n + 33
 }
 
+/// Try to parse Extensible Priority field value.
+///
+/// The field value is expected to be a Structured Fields Dictionary; see
+/// https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority-12.
```
ghedo, line 2143: Also here.

quiche/src/h3/mod.rs
```diff
@@ -2127,6 +2137,64 @@ fn grease_value() -> u64 {
...
+/// If the `u` or `i` fields are contained with correct types, a constructed
+/// Priority object is returned. Note that urgency values outside of valid range
+/// (0 through 7) are clamped to 7.
+///
+/// If the `u` or `i` fields are contained with the wrong types,
+/// Error::Done is returned.
+///
+/// Omitted parameters will yield default values.
+#[cfg(feature = "sfv")]
+pub fn parse_extensible_priority(priority: &[u8]) -> Result<Priority> {
```
ghedo, line 2154: This could probably be implemented as a [`TryFrom<&[u8]>`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) trait on the `Priority` type instead of a function on its own.

ghedo (changes requested):
Ideally, I think quiche shouldn't really provide the priority format parsing functionality at all, as it's somewhat orthogonal to HTTP/3 and is more of a general HTTP feature.

But we were already providing this functionality before and if anything, this change makes the quiche API less coupled to the priority format, so it's a good improvement, though in the future we might want to think about moving this somewhere else instead.

Anyway, added some comments.

Dockerfile
LPardue: It would be nice to actually support formal MSRV now too

junhochoi: I tried some rust version and currently the practical minimum version is `1.54.0` because of https://github.com/paupino/rust-decimal#minimum-rust-compiler-version (one of the indirect dependency)

Ideally `rust-version` need to be used but its minimum is `1.56.0` https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

junhochoi: `rust-decimal` is added in the dependency list because of newly added `sfv`.

quiche/include/quiche.h
```diff
@@ -768,6 +768,11 @@ typedef struct {
     size_t value_len;
 } quiche_h3_header;
 
+typedef struct {
```
ghedo, line 771: Need to add the description from the Rust struct here.

quiche/include/quiche.h
```diff
@@ -793,6 +798,11 @@ ssize_t quiche_h3_send_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
 ssize_t quiche_h3_recv_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
                             uint64_t stream_id, uint8_t *out, size_t out_len);
 
+// Try to parse Extensible Priority field value.
+int quiche_h3_parse_extensible_priority(uint8_t *priority,
+                                            size_t priority_len,
+                                            quiche_h3_priority *parsed);
```
ghedo, line 804: Nit: indentation :stuck_out_tongue:

quiche/include/quiche.h
```diff
@@ -793,6 +798,11 @@ ssize_t quiche_h3_send_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
 ssize_t quiche_h3_recv_body(quiche_h3_conn *conn, quiche_conn *quic_conn,
                             uint64_t stream_id, uint8_t *out, size_t out_len);
 
+// Try to parse Extensible Priority field value.
```
ghedo, line 801:
```suggestion
// Try to parse an Extensible Priority field value.
```

apps/Cargo.toml
```diff
@@ -15,8 +15,10 @@ qlog = ["quiche/qlog"]
 # Use BoringSSL provided by the boring crate.
 boringssl-boring-crate = ["quiche/boringssl-boring-crate"]
 
-default = ["qlog"]
+# Enable sfv support.
+sfv = ["quiche/sfv"]
 
+default = ["qlog", "sfv"]
```
ghedo, line 21:
```suggestion
default = ["qlog", "sfv"]

```

quiche/src/h3/mod.rs
```diff
@@ -603,6 +614,98 @@ pub enum Event {
...
+    pub fn new(urgency: u8, incremental: bool) -> Self {
+        Priority {
+            urgency,
+            incremental,
+        }
+    }
+}
+
+#[cfg(feature = "sfv")]
+impl TryFrom<&[u8]> for Priority {
```
ghedo, line 645:
```suggestion
#[cfg_attr(docsrs, doc(cfg(feature = "sfv")))]
impl TryFrom<&[u8]> for Priority {
```

quiche/src/h3/mod.rs
```diff
@@ -603,6 +614,98 @@ pub enum Event {
     GoAway,
 }
 
+/// Extensible Priorities parameters.
```
ghedo, line 617: Maybe expand this a bit to say that the priority object can be generated by parsing an extensible priority value with the `TryFrom` thing?

ghedo (changes requested):
Added a few comments, but looks good to me overall.

quiche/include/quiche.h
LPardue: done

quiche/src/h3/mod.rs
LPardue: PTAL

ghedo (approved):

PR #1161, opened 2022-02-18
Make qlog packet event logging atomic

LPardue:
When we switched from JSON to JSON-SEQ, we modified the qlog serializer but didn't change it's behaviour wrt QUIC Packet events (rx/tx). The old behaviour TL;DR was:

1. qlog streamer user passes streamer a packet event
2. Streamer writes out packet header information. Then it locks itself into frame writing mode.
2. While locked, no other events cand be written.
3. qlog streamer user passes 0 or more frames.
4. qlog streamer user finalizes frames write. Streamer unlocks itself.

There are a few downsides to that approach:

* Locking prevents other events being written. For example, while processing frames, any additional events would be ignored by the serializer. While callers could buffer events themselves and log them later, it adds complication and we decided never to do that inside quiche.
* Related to the above. We had several places where processing might shortcircuit or terminate. We ended up peppering quiche with calls to finalize frame writing.
* It's possible to produce an incomplete log line if the connection is destroyed before.
* It's less efficient to write a partial JSON object out.

The changes in the PR make packet event logging atomic. The qlog streamer is simplified by removing all locking and frame-specific logic. As a consequence, the quiche library requires updates in the packet send() and receive() function. Rather than serialize individual frames as they are added or read from a packet, we buffer the qlog representation of a frame, and then later pass that in a single call to the streamer.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2213,7 +2213,7 @@ impl Connection {
         );
 
         #[cfg(feature = "qlog")]
-        let mut qlog_frames = self.qlog.streamer.as_ref().map(|_| vec![]);
+        let mut qlog_frames = vec![];
 
         let mut payload = packet::decrypt_pkt(
             &mut b,
@@ -2278,9 +2278,7 @@ impl Connection {
             let frame = frame::Frame::from_bytes(&mut payload, hdr.ty)?;
 
             qlog_with_type!(QLOG_PACKET_RX, self.qlog, _q, {
-                if let Some(frames) = &mut qlog_frames {
-                    frames.push(frame.to_qlog());
-                }
+                qlog_frames.push(frame.to_qlog());
             });
 
             if frame.ack_eliciting() {
@@ -2313,7 +2311,7 @@ impl Connection {
             let ev_data =
                 EventData::PacketReceived(qlog::events::quic::PacketReceived {
                     header: qlog_pkt_hdr,
-                    frames: Some(qlog_frames.unwrap_or_default()),
+                    frames: Some(qlog_frames),
                     is_coalesced: None,
                     retry_token: None,
                     stateless_reset_token: None,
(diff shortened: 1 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -2143,37 +2143,8 @@ impl Connection {
...
-                    stateless_reset_token: None,
-                    supported_versions: None,
-                    raw: Some(qlog_raw_info),
-                    datagram_id: None,
-                });
-
-            q.add_event_data_with_instant(ev_data, now).ok();
-        });
+        #[cfg(feature = "qlog")]
+        let mut qlog_frames = self.qlog.streamer.as_ref().map(|_| vec![]);
```
ghedo, line 2147: I don't think there's much point in making this an Option, as `qlog_with_type!()` will already skip the push calls if there is no qlog streamer, and just calling `Vec::new()`/`vec![]` unconditionally shouldn't really do much harm.

quiche/src/lib.rs
```diff
@@ -3352,6 +3356,19 @@ impl Connection {
             pn
         );
 
+        #[cfg(feature = "qlog")]
+        let mut qlog_frames = self.qlog.streamer.as_ref().map(|_| vec![]);
```
ghedo, line 3360: Same as above.

ghedo (changes requested):

quiche/src/lib.rs
```diff
@@ -3418,6 +3420,17 @@ impl Connection {
             pn
         );
 
+        #[cfg(feature = "qlog")]
+        let mut qlog_frames = vec![];
```
ghedo, line 3424:
```suggestion
        let mut qlog_frames = Vec::with_capacity(frames.len());
```

ghedo (changes requested):
Added a nit, but looks good otherwise.

quiche/src/lib.rs
LPardue: applied, thanks!

ghedo (approved):

PR #1162, opened 2022-02-22
cc: add send_quantum API

junhochoi:
For higher performance sending multiple packets is a common
practice, such as sendmmsg() or linux sendmsg() GSO. To
help with that, send_quantum is added and it can be modified
in the congestion controller. It's a hint for the application,
it may or may not use the value for controlling the batch
size of packets.

For default value, initcwnd is used(RFC9002 section 7.7).

Also, conn.send_quantum() and C API (quiche_conn_send_quantum())
is provided for the application.

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -43,12 +43,12 @@ use quiche_apps::args::*;
 
 use quiche_apps::common::*;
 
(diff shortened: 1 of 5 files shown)


ghedo (approved):
I expanded the documentation a bit, and also slightly rearranged the orider of the functions. LGTM otherwise.

PR #1164, opened 2022-02-23
expose boringssl ctx

ehaydenr:
Allow clients to configure boring SslContext ahead of time. This is useful when you'd like to configure things that quiche's API doesn't expose (e.g. SNI callback).

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -610,19 +610,19 @@ impl Config {
     /// # Ok::<(), quiche::Error>(())
     /// ```
     pub fn new(version: u32) -> Result<Config> {
-        Self::_new_for_tls_ctx(version, tls::Context::new()?)
+        Self::with_tls_ctx(version, tls::Context::new()?)
     }
 
     /// Creates a config object with the given version and SslContext. This is
-    /// useful for applications which wish to manually configure SslContext
-    #[cfg(feature = "boring")]
-    pub fn new_for_tls_ctx(
+    /// useful for applications that wish to manually configure SslContext
+    #[cfg(feature = "boring-sys")]
+    pub fn with_boring_ssl_ctx(
         version: u32, tls_ctx: boring::ssl::SslContext,
     ) -> Result<Config> {
-        Self::_new_for_tls_ctx(version, tls::Context::from_boring(tls_ctx))
+        Self::with_tls_ctx(version, tls::Context::from_boring(tls_ctx))
     }
 
-    fn _new_for_tls_ctx(version: u32, tls_ctx: tls::Context) -> Result<Config> {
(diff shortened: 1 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -610,12 +610,23 @@ impl Config {
     /// # Ok::<(), quiche::Error>(())
     /// ```
     pub fn new(version: u32) -> Result<Config> {
+        Self::_new_for_tls_ctx(version, tls::Context::new()?)
+    }
+
+    /// Creates a config object with the given version and SslContext. This is
+    /// useful for applications which wish to manually configure SslContext
```
LPardue, line 617:
```suggestion
    /// useful for applications that wish to manually configure SslContext
```

quiche/src/lib.rs
```diff
@@ -610,12 +610,23 @@ impl Config {
...
+    #[cfg(feature = "boring")]
+    pub fn new_for_tls_ctx(
```
LPardue, line 619: consider perhaps instead `with_boring_ssl_ctx`

quiche/src/lib.rs
```diff
@@ -610,12 +610,23 @@ impl Config {
...
+        version: u32, tls_ctx: boring::ssl::SslContext,
+    ) -> Result<Config> {
+        Self::_new_for_tls_ctx(version, tls::Context::from_boring(tls_ctx))
+    }
+
+    fn _new_for_tls_ctx(version: u32, tls_ctx: tls::Context) -> Result<Config> {
```
LPardue, line 625: as above, then this could become `with_tls_ctx`

LPardue:
It might be useful to update one pair of examples to with this API in order to get CI to exercise it.

quiche/src/lib.rs
ehaydenr: fixed

ehaydenr:
Added a test that should exercise it:
```
cargo test --features boring-sys
```

Is that sufficient or did you have something else in mind?

LPardue:
Tests are good, thanks!

I was thinking maybe we could change quiche-client / quiche-server to actually switch over to this feature by default. Curious to hear what @ghedo thinks though, so perhaps don't rush to do that.

junhochoi:
CI fauilre is a different issue: https://github.com/cloudflare/quiche/pull/1167

ghedo:
Merged as https://github.com/cloudflare/quiche/commit/3204f89beed236f0ea458380ff0e196cbb4bf78e, thanks @ehaydenr!

Note that I slightly tweaked the documentation, and also changed the code to use `into_ptr()` rather than `as_ptr()`+`mem::forget()`.

Since the `boring-sys` feature is not the name of an optional dependency anymore, in the future we will probably want to rename it to something else as well, e.g. `boringssl-boring-crate`, to be more consistent with the existing `boringssl-vendored` feature.

PR #1167, opened 2022-02-24
ci: use windows-2019 for windows build

junhochoi:
Today I see `windows-latest` is updated to `windows-2022` and the
build start to fail on 32bit build. Until we figure it out, pin the version
to 2019.


junhochoi:
It seems VS2022 is fully 64bit https://devblogs.microsoft.com/dotnet/msbuild-and-64-bit-visual-studio-2022/

ghedo (approved):

PR #1170, opened 2022-02-25
lib: send STREAM_DATA_BLOCKED frames less often

LPardue:
NB: this reduces the frequency of blocked frames at the same offset from "every send()" to once. If we want a middle ground, like every N packets, that is feasible too.

When applications try to send stream data with `stream_send()`, the
number of bytes actually written might be less than supplied due to
insufficient stream or connection capacity. In these cases, quiche
internally marks the stream as blocked and, when `conn.send()` is next
called the mark causes a STREAM_DATA_BLOCKED frame to be emitted in a
new packet.

Previously, every call to `stream_send()` would cause the stream to be
marked and therefore generate a STREAM_DATA_BLOCKED frame. In cases,
where `conn.send()` is called frequently, the blocked frame might be the
only frame in the packet, which is wasteful.

This change reduces the frequency of STREAM_DATA_BLOCKED frames by
avoiding sending them when the offset of sent stream data is unchanged.
When a stream is first blocked during `stream_send()`, a frame is
emitted and the blocked offset is recorded. Subsequent blocked frames
are only emitted if the application is able to write more data and get
blocked at a new offset.

### DIFF
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -1244,7 +1244,7 @@ impl SendBuf {
         self.blocked_at = blocked_at;
     }
 
-    /// The last offset the stream blocked at, if any.
+    /// The last offset the stream was blocked at, if any.
     pub fn blocked_at(&self) -> Option<u64> {
         self.blocked_at
     }
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -1239,7 +1239,7 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
-    /// Updates the last offset the stream was blocked_at, if any.
+    /// Updates the last offset the stream was blocked at, if any.
     pub fn update_blocked_at(&mut self, blocked_at: Option<u64>) {
         self.blocked_at = blocked_at;
     }
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -1239,12 +1239,12 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
-    /// Updates the last offset we were blocked_at, if any.
+    /// Updates the last offset the stream was blocked_at, if any.
     pub fn update_blocked_at(&mut self, blocked_at: Option<u64>) {
         self.blocked_at = blocked_at;
     }
 
-    /// The last offset we were blocked at, if any.
+    /// The last offset the stream blocked at, if any.
     pub fn blocked_at(&self) -> Option<u64> {
         self.blocked_at
     }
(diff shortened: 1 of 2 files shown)


quiche/src/stream.rs
```diff
@@ -1236,6 +1239,16 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
+    /// Updates the last offset we were blocked_at, if any.
+    pub fn update_blocked_at(&mut self, blocked_at: Option<u64>) {
+        self.blocked_at = blocked_at;
+    }
+
+    /// The last offset we were blocked at, if any.
```
ghedo, line 1247:
```suggestion
    /// The last offset the stream blocked at, if any.
```

quiche/src/stream.rs
```diff
@@ -1236,6 +1239,16 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
+    /// Updates the last offset we were blocked_at, if any.
```
ghedo, line 1242:
```suggestion
    /// Updates the last offset the stream was blocked_at, if any.
```

ghedo (changes requested):
Small nits, but lgtm.

quiche/src/stream.rs
```diff
@@ -1236,6 +1239,16 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
+    /// Updates the last offset the stream was blocked_at, if any.
+    pub fn update_blocked_at(&mut self, blocked_at: Option<u64>) {
+        self.blocked_at = blocked_at;
+    }
+
+    /// The last offset the stream blocked at, if any.
```
ghedo, line 1247: Ooops, my bad
```suggestion
    /// The last offset the stream was blocked at, if any.
```

quiche/src/stream.rs
```diff
@@ -1236,6 +1239,16 @@ impl SendBuf {
         self.max_data = cmp::max(self.max_data, max_data);
     }
 
+    /// Updates the last offset the stream was blocked_at, if any.
```
ghedo, line 1242:
```suggestion
    /// Updates the last offset the stream was blocked at, if any.
```

quiche/src/stream.rs
```diff
@@ -1082,6 +1082,9 @@ pub struct SendBuf {
     /// The maximum offset we are allowed to send to the peer.
     max_data: u64,
 
+    /// The last offset we were blocked at, if any.
```
ghedo, line 1085:
```suggestion
    /// The last offset the stream was blocked at, if any.
```

ghedo (approved):

PR #1175, opened 2022-02-28
pacing: get_packet_send_time() is no longer Option<>

junhochoi:
Since `recovery.last_packet_scheduled_time` always have a
default value (current timestamp), Having it `Option<Instant>`
type doesn't make sense much. No changes on public API.


junhochoi:
@lohith-bellad PTAL

ghedo (approved):

PR #1180, opened 2022-03-04
cc: BBR congestion control module

junhochoi:
Implementing `bbr` congestion control algorithm module. 

To try this module, you'll need `quiche-server` supports packet pacing only with `SO_TXTIME` feature in recent linux kernel (#1138). Also you'll need to configure `fq` in `tc qdisc` on the server for pacing to work properly.

Currently this PR is rebased on #1238 implementing GSO-friendly pacer. You can test with quiche-server using GSO in #1138 

TODOs
- [x] fix bug/issues
- [x] add more unit tests

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -203,6 +203,7 @@ void quiche_config_set_disable_active_migration(quiche_config *config, bool v);
 enum quiche_cc_algorithm {
     QUICHE_CC_RENO = 0,
     QUICHE_CC_CUBIC = 1,
+    QUICHE_CC_BBR = 2,
 };
 
 // Sets the congestion control algorithm used.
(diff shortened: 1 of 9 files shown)


ghedo:
Haven't reviewed the whole thing yet, but one comment I have is that I don't think BBR should be the default algorithm due to the requirement for pacing, which could lead applications to be kinda broken as we have no way to enforce that they pace, and its other other side-effects (e.g. fairness to other algorithms, increase in spurious retransmissions, ...).

CUBIC seems like a saner default (in line with what Linux does).

We can change the default in the apps once the pacing PR is merged, but even then we might need to limit it to Linux.

junhochoi:
> CUBIC seems like a saner default (in line with what Linux does).

Sure - I removed the commit which makes `bbr` default.

ghedo (approved):

PR #1193, opened 2022-03-11
Connection ID handling

qdeconinck:
The main features that this commit introduces are the following.

Monitoring of QUIC events

This commit introduces a new public Connection method called
`poll()`, acting similarly to the `poll()` method in the h3 module.
This enables the application to handle various newly introduced
events, currently only specific to Connection IDs.

Support for Connection ID changes

The connection is no more bound to a single SCID-DCID pair and can
now advertise and use other Connection IDs through
NEW_CONNECTION_ID and RETIRE_CONNECTION_ID frames. From an API
viewpoint, a host can indicate in its QUIC config the maximum
number of active connection IDs it can support
(`set_active_connection_id_limit()`). Then, it can actively provide
Connection IDs to its peer (`new_source_cid()`) and retire
peer-provided Connection IDs (`retire_destination_cid()`). Passive
operations (like receiving a new Destination Connection ID or
retiring a source Connection ID) are monitored through the `poll()`
method. Connection IDs are handled by a dedicated structure,
`ConnectionIdentifiers`, located in the newly introduced `cid`
module.

Retro-compatibility notes

The provided features were designed to minimise breaking changes to
the application. Still, taking advantages of all the features
brought by this commit requires the application to listen to events
returned by `poll()`. By default, the generation of these events is
disabled to avoid allocating memory for events that will never be
read. An application wanting to handle such events needs to
explicitly request them by calling the `enable_events()` method of
the `Config` structure.

### DIFF
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -48,6 +48,7 @@ pub struct CommonArgs {
     pub dgrams_enabled: bool,
     pub dgram_count: u64,
     pub dgram_data: String,
+    pub max_active_cids: u64,
 }
 
 /// Creates a new `CommonArgs` structure using the provided [`Docopt`].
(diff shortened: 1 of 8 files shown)


qdeconinck:
I just updated the PR with two main changes.

* The internal structure to store CIDs is now a bounded, non-empty `VecDeque` instead of a `HashMap`.
* Breaking changes (i.e., different return type for `source_id()` and `destination_id()`) were removed.

qdeconinck:
Closing this as #1223 includes the main changes of this PR.

PR #1195, opened 2022-03-17
cc: hystart++ draft 04

junhochoi:
Update based on [draft-ietf-tcpm-hystartplusplus-04](https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-hystartplusplus-04).
Major change is removing `LOW_CWND`.

### DIFF
--- a/quiche/src/recovery/hystart.rs
+++ b/quiche/src/recovery/hystart.rs
@@ -28,7 +28,7 @@
 //!
 //! This implementation is based on the following I-D:
 //!
-//! <https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-hystartplusplus-03>
+//! <https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-hystartplusplus-04>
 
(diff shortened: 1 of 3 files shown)


quiche/src/recovery/hystart.rs
```diff
@@ -143,11 +141,10 @@ impl Hystart {
 
         // Slow Start.
         if self.css_start_time().is_none() {
-            if cwnd >= (LOW_CWND * max_datagram_size) &&
-                self.rtt_sample_count >= N_RTT_SAMPLE
+            if self.rtt_sample_count >= N_RTT_SAMPLE &&
+                self.current_round_min_rtt != Duration::MAX &&
+                self.last_round_min_rtt != Duration::MAX
             {
-                self.rtt_sample_count = 0;
```
junhochoi, line 149: This part is removed because it was not in the spec and rtt_sample_count will be reset when a new round starts.

ghedo (approved):

PR #1201, opened 2022-03-19
split octets module into its own crate

ghedo:
Some builds fail because the crate hasn't been published to crates.io yet.


LPardue (dismissed):
LGTM thanks!

junhochoi (dismissed):

ghedo:
The crate is published, and the builds are green now.

PR #1200, opened 2022-03-19
Use mio v0.8 in examples

sile:
This PR updates the version of `mio` used in the examples from v0.6 to v0.8 (latest minor version).

After this change, I've confirmed that the `http3-server` and` http3-client` examples work as before:

```console
$ cd quiche/
$ mkdir examples/root/
$ echo 'Hello World!' > examples/root/hello

// Server
$ cargo run --example http3-server

// Client (another shell)
$ cargo run --example http3-client https://127.0.0.1:4433/hello
Hello World!
```

Closes https://github.com/cloudflare/quiche/issues/1196

PR #1202, opened 2022-03-20
Use mio v0.8 in apps and tools

sile:
This is a follow-up PR of #1200 and updates the mio version to the latest one (v0.8).

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -61,22 +61,19 @@ fn main() {
     let args = ServerArgs::with_docopt(&docopt);
 
     // Setup the event loop.
-    let poll = mio::Poll::new().unwrap();
+    let mut poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
 
     // Create the UDP listening socket, and register it with the event loop.
     let socket = net::UdpSocket::bind(args.listen).unwrap();
+    socket.set_nonblocking(true).unwrap();
 
     info!("listening on {:}", socket.local_addr().unwrap());
 
-    let socket = mio::net::UdpSocket::from_socket(socket).unwrap();
-    poll.register(
-        &socket,
-        mio::Token(0),
-        mio::Ready::readable(),
-        mio::PollOpt::edge(),
-    )
-    .unwrap();
+    let mut socket = mio::net::UdpSocket::from_std(socket);
+    poll.registry()
+        .register(&mut socket, mio::Token(0), mio::Interest::READABLE)
+        .unwrap();
 
     // Create the configuration for the QUIC connections.
     let mut config = quiche::Config::new(quiche::PROTOCOL_VERSION).unwrap();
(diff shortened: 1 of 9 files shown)


apps/src/bin/quiche-server.rs
```diff
@@ -61,22 +61,19 @@ fn main() {
     let args = ServerArgs::with_docopt(&docopt);
 
     // Setup the event loop.
-    let poll = mio::Poll::new().unwrap();
+    let mut poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
 
     // Create the UDP listening socket, and register it with the event loop.
     let socket = net::UdpSocket::bind(args.listen).unwrap();
+    socket.set_nonblocking(true).unwrap();
```
junhochoi, line 69: Do we need this?

apps/src/bin/quiche-server.rs
```diff
@@ -61,22 +61,19 @@ fn main() {
...
 
-    let socket = mio::net::UdpSocket::from_socket(socket).unwrap();
-    poll.register(
-        &socket,
-        mio::Token(0),
-        mio::Ready::readable(),
-        mio::PollOpt::edge(),
-    )
-    .unwrap();
+    let mut socket = mio::net::UdpSocket::from_std(socket);
```
junhochoi, line 73: I think we can merge L68 (calling `bind()`) and this line into calling `mio::net::UdpSocket::bind()` once.

junhochoi (changes requested):

apps/src/bin/quiche-server.rs
sile: Without this line, the socket will blocks on the recv_from call below (as far as I confirmed in my local env): https://github.com/cloudflare/quiche/blob/0107fb90694d0144bcf76a125c67abb84e2fe204/apps/src/bin/quiche-server.rs#L170

Please also see the comment in the mio repo: https://github.com/tokio-rs/mio/blob/b2bbc1029849c939031fb396475d1a19da9786a6/tests/udp_socket.rs#L102-L105

apps/src/bin/quiche-server.rs
sile: Sounds good, thanks!
The result of `mio::net::UdpSocket::bind()` seems a non-blocking socket by default, so we can eliminate the `set_nonblocking` call too.

sile (commented):
@junhochoi Thank you for your review. I will reflect your commets soon.

apps/src/bin/quiche-server.rs
sile: Fixed in https://github.com/cloudflare/quiche/pull/1202/commits/935e8d6ac79340a4172da931cd90f66ec963ea67
Note that I also updated the UDP socket creation code merged in the previous PR (#1200).

junhochoi (approved):
LGTM

ghedo:
This is merged now, thanks @sile!

sile:
🎉

Issue #828, opened 2021-01-28
Support for "server's preferred address"


tbu-:
Does the quiche client support connecting to the server's preferred address as stated in [Section 9.6 of draft34](https://tools.ietf.org/html/draft-ietf-quic-transport-34#section-9.6)?

The code says things like
https://github.com/cloudflare/quiche/blob/fcaedfa486b402d331025629cb6e8895c3885394/src/lib.rs#L4598
so I'm going to assume "no".

You can consider this a feature request for supporting the "server's preferred address". :)

Issue #840, opened 2021-02-04
Feature request: Connection migration


tbu-:
According to #698, connection migration isn't supported.

I think supporting connection migration would need an API change, as quiche would need to be made aware of IP addresses, in order to do address validation etc.

Would PRs going in such a direction be accepted? I guess a design decision would be warranted first.

LPardue:
Connection migration is a nice feature of QUIC and it would be great for quiche to support it but for now it is still a TBD. Adding it requires some upfront design to make sure the feature aligns with the API paradigms, and really needs some upfront thinking about how to test and validate it. If you have ideas or PoC then we could consider them but you given the potential complexity I can't so how hard or easy that is.

tbu-:
Fixed by #1223.

PR #1208, opened 2022-03-27
add support for configuring and communicating preferred address

kckeiks:
Hello,
I have a wip PR for review for issue #828. This change allows users to configure the server's preferred address via Config. After the handshake, the preferred addresses can be accessed via the Connection.

- [X] implement encoding and decoding of params (`PreferredAddressParams`)
- [X] allow users to configure preferred address params via`Config`
- [X] allow clients to access the server's preferred address after the handshake
- [X] add/update unit tests 
- [x] update `TransportParams.to_qlog`
- [X] ~allow clients to update the `Connection`'s peer address with the preferred address~ #1223 

I will add docs once I get some feedback about these changes.

### DIFF
--- a/qlog/src/lib.rs
+++ b/qlog/src/lib.rs
@@ -627,6 +627,24 @@ pub struct Token {
     pub details: Option<String>,
 }
 
+impl Token {
+    pub fn new(
+        ty: Option<TokenType>,
+        length: Option<u32>,
+        stateless_reset_token: Option<Vec<u8>>,
+        details: Option<String>
+    ) -> Self {
+        Token {
+            ty,
+            length,
+            data: HexSlice::maybe_string(
+                stateless_reset_token.as_ref(),
+            ),
+            details,
+        }
+    }
+}
+
 pub struct HexSlice<'a>(&'a [u8]);
 
 impl<'a> HexSlice<'a> {
--- a/qlog/src/lib.rs
+++ b/qlog/src/lib.rs
@@ -629,17 +629,13 @@ pub struct Token {
 
 impl Token {
     pub fn new(
-        ty: Option<TokenType>,
-        length: Option<u32>,
-        stateless_reset_token: Option<Vec<u8>>,
-        details: Option<String>
+        ty: Option<TokenType>, length: Option<u32>,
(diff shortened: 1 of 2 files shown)


kckeiks:
>allow clients to update the Connection's peer address with the preferred address

I'm not %100 sure of all the tasks to complete this part. I believe we should update `Connection.peer_addr` (add a setter). I need to look at it more closely but would be great to get some guidance.

ghedo:
@kckeiks Thanks for the PR!

> I believe we should update Connection.peer_addr

Yes, but note that before sending packets to a new address the client needs to validate the path first, by sending a `PATH_CHALLENGE` frame (see https://datatracker.ietf.org/doc/html/rfc9000#section-9.6.2 ) and quiche currently doesn 't currently do this on its own since we don't support migration yet (it does support creating the frame, but some code will need to be added to verify the challenge response).

kckeiks:
> @kckeiks Thanks for the PR!
> 
> > I believe we should update Connection.peer_addr
> 
> Yes, but note that before sending packets to a new address the client needs to validate the path first, by sending a `PATH_CHALLENGE` frame (see https://datatracker.ietf.org/doc/html/rfc9000#section-9.6.2 ) and quiche currently doesn 't currently do this on its own since we don't support migration yet (it does support creating the frame, but some code will need to be added to verify the challenge response).

Thanks for the info! I see that this enhancement has an [issue](https://github.com/cloudflare/quiche/issues/840) and adding this feature is still TBD. Maybe I can think of some ideas and offer them for consideration.

kckeiks:
> #840 was now added by #1223, does that mean that this pull request is now unblocked?

Yes, it is. I will update this PR shortly today.

junhochoi:
I rerun the failed test. It looks like flaky

LPardue:
The qlog specs recently changed to extract StatelessResetToken into its own type. The changes on https://github.com/cloudflare/quiche/pull/1294 should help avoid this PR needing to make some changes of its own.

PSUdaemon:
@kckeiks Can you rebase this on the changes from #1294 please?

PR #1213, opened 2022-04-09
apps: fix rounding down max_send_burst

junhochoi:
Currently when send_quantum() is smaller than the maximum
buffer size, it can make a small leftover because it's not
rounded down to the maximum datagram size. Now it takes the
mininum of max buffer size or send_quantum() first
and rounds down by the maximum datagram size.

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -456,10 +456,9 @@ fn main() {
         // packets to be sent.
         continue_write = false;
         for client in clients.values_mut() {
-            let max_send_burst = client
-                .conn
-                .send_quantum()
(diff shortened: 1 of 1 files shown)


junhochoi:
This fix will reduce a lot of variance in quiche-server performance test.

e.g. 100Mbps/50ms network, downloading 100M bytes -> 100k -> 100M bytes, 50 times

quiche-server version | MEDIAN of Total Time (s) | STDEVP of Total Time (s)
-- | -- | --
1213 | 25.36 | 1.13
master | 25.85 | 9.59

Average time doesn't change much but the variance reduced a lot.

ghedo (approved):

PR #1215, opened 2022-04-13
recovery: update delivery_rate on all packets

junhochoi:
Currently `delivery_rate.on_packet_sent()` is called
only packets with in-flight data, which makes non
in-flight packets not updated with some metadata such
as `delivered`, etc. They have zero payload, but
need to be included in the rate estimation and
app_limited updates. Otherwise there can be unsync
between `Rate.last_sent_packet` and `Rate.latest_acked`
(`last_sent_packet` doesn't include non-ack-eliciting
packets but `latest_acked` does), which can mess up
`Rate.end_of_app_limited` tracking.

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -267,6 +267,9 @@ impl Recovery {
         self.largest_sent_pkt[epoch] =
             cmp::max(self.largest_sent_pkt[epoch], pkt_num);
 
+        self.delivery_rate
+            .on_packet_sent(&mut pkt, self.bytes_in_flight, now);
+
         if in_flight {
             if ack_eliciting {
                 self.time_of_last_sent_ack_eliciting_pkt[epoch] = Some(now);
@@ -276,12 +279,6 @@ impl Recovery {
 
             self.update_app_limited(
                 (self.bytes_in_flight + sent_bytes) < self.congestion_window,
-            );
-
-            self.delivery_rate.on_packet_sent(
-                &mut pkt,
-                self.bytes_in_flight,
-                now,
             );
 
             self.on_packet_sent_cc(sent_bytes, now);


ghedo:
The delivery rate draft says `SendPacket()` should be called "upon transmitting or retransmitting a data packet", and "data packet" seems to imply an in-flight packet, so I wonder if to avoid the de-sync problem we should instead call `UpdateRateSample()` only for in-flight packets too instead? Not quite sure what the consequences of either solution are though.

junhochoi:
> The delivery rate draft says `SendPacket()` should be called "upon transmitting or retransmitting a data packet", and "data packet" seems to imply an in-flight packet, so I wonder if to avoid the de-sync problem we should instead call `UpdateRateSample()` only for in-flight packets too instead? Not quite sure what the consequences of either solution are though.

The difference is that TCP doesn't increase its sequence number when there is no data but QUIC increases packet number in every new packet. so 'largest_acked' need to include all packets.

Note that we use a packet number to detect [end-of-app-limited](https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/delivery_rate.rs#L127). In the spec it uses `delivered` but then we need to figure out [`pipe`](https://www.ietf.org/archive/id/draft-cheng-iccrg-delivery-rate-estimation-02.html#name-detecting-application-limit).

Our behavior is coming from Chromium using [packet number approach](https://source.chromium.org/chromium/chromium/src/+/main:net/third_party/quiche/src/quiche/quic/core/congestion_control/bandwidth_sampler.cc;l=190;bpv=1;bpt=1?q=BandwidthSampler::OnPacketSent&ss=chromium%2Fchromium%2Fsrc) and its behavior seems matching to this PR (doesn't look whether it's ack-eliciting or not), so I think it's ok.

This bug makes an problem when tested in #1180 - when it goes to ProbeRTT and last few packets before ProbeRTT has no data, it can accidentally updates btlbw (and eventually cwnd) with very low rate.

ghedo (approved):

PR #1223, opened 2022-05-04
Connection migration support

qdeconinck:
This PR provides provides connection migration support to quiche. It is composed of 3 incremental parts (the current 3 commits), starting from CID handling support, then path validation and finally connection migration. See the commit messages for further details about the changes introduced by each of them.

Note that while most of the changes are additive, there are a few breaking API changes introduced by the path validation support. Namely,

- `RecvInfo` now contains the `to` field, indicating the local host address on which the packet was received.
- `SendInfo` now contains the `from` field, indicating from which local host address the packet should be sent.
- Both `connect` and `accept` now takes an owned `ConnectionId` for the source CID and an additional `SocketAddr` parameter corresponding to the (initial) local host address that is used by the connection.

### DIFF
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -48,6 +48,8 @@ pub struct CommonArgs {
     pub dgrams_enabled: bool,
     pub dgram_count: u64,
     pub dgram_data: String,
+    pub max_active_cids: u64,
+    pub enable_active_migration: bool,
 }
 
 /// Creates a new `CommonArgs` structure using the provided [`Docopt`].
@@ -68,6 +70,8 @@ pub struct CommonArgs {
 /// --dgram-proto PROTO         DATAGRAM application protocol.
 /// --dgram-count COUNT         Number of DATAGRAMs to send.
 /// --dgram-data DATA           DATAGRAM data to send.
+/// --max-active-cids NUM       Maximum number of active Connection IDs.
+/// --enable-active-migration   Enable active connection migration.
 ///
 /// [`Docopt`]: https://docs.rs/docopt/1.1.0/docopt/
 impl Args for CommonArgs {
@@ -143,6 +147,11 @@ impl Args for CommonArgs {
 
         let disable_hystart = args.get_bool("--disable-hystart");
 
+        let max_active_cids = args.get_str("--max-active-cids");
+        let max_active_cids = max_active_cids.parse::<u64>().unwrap();
+
+        let enable_active_migration = args.get_bool("--enable-active-migration");
+
         CommonArgs {
             alpns,
             max_data,
@@ -160,6 +169,8 @@ impl Args for CommonArgs {
             dgrams_enabled,
             dgram_count,
             dgram_data,
+            max_active_cids,
+            enable_active_migration,
         }
     }
 }
@@ -183,6 +194,8 @@ impl Default for CommonArgs {
             dgrams_enabled: false,
             dgram_count: 0,
             dgram_data: "quack".to_string(),
+            max_active_cids: 2,
+            enable_active_migration: false,
         }
     }
 }
@@ -216,6 +229,9 @@ Options:
   --no-grease              Don't send GREASE.
   --cc-algorithm NAME      Specify which congestion control algorithm to use [default: cubic].
   --disable-hystart        Disable HyStart++.
+  --max-active-cids NUM    The maximum number of active Connection IDs we can support [default: 2].
+  --enable-active-migration   Enable active connection migration.
+  --perform-migration      Perform connection migration on another source port.
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --session-file PATH      File used to cache a TLS session for resumption.
@@ -237,6 +253,7 @@ pub struct ClientArgs {
     pub connect_to: Option<String>,
     pub session_file: Option<String>,
     pub source_port: u16,
+    pub perform_migration: bool,
 }
 
 impl Args for ClientArgs {
@@ -303,6 +320,8 @@ impl Args for ClientArgs {
         let source_port = args.get_str("--source-port");
         let source_port = source_port.parse::<u16>().unwrap();
 
+        let perform_migration = args.get_bool("--perform-migration");
+
         ClientArgs {
             version,
             dump_response_path,
@@ -316,6 +335,7 @@ impl Args for ClientArgs {
             connect_to,
             session_file,
             source_port,
+            perform_migration,
         }
     }
 }
@@ -335,6 +355,7 @@ impl Default for ClientArgs {
             connect_to: None,
             session_file: None,
             source_port: 0,
+            perform_migration: false,
         }
     }
 }
@@ -367,6 +388,8 @@ Options:
   --dgram-data DATA           Data to send for certain types of DATAGRAM application protocol [default: brrr].
   --cc-algorithm NAME         Specify which congestion control algorithm to use [default: cubic].
   --disable-hystart           Disable HyStart++.
+  --max-active-cids NUM       The maximum number of active Connection IDs we can support [default: 2].
+  --enable-active-migration   Enable active connection migration.
   -h --help                   Show this screen.
 ";
 
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -90,7 +90,8 @@ fn main() {
     config.set_initial_max_stream_data_uni(conn_args.max_stream_data);
     config.set_initial_max_streams_bidi(conn_args.max_streams_bidi);
     config.set_initial_max_streams_uni(conn_args.max_streams_uni);
-    config.set_disable_active_migration(true);
+    config.set_disable_active_migration(!conn_args.enable_active_migration);
+    config.set_active_connection_id_limit(conn_args.max_active_cids);
 
     config.set_max_connection_window(conn_args.max_window);
     config.set_max_stream_window(conn_args.max_stream_window);
@@ -133,11 +134,15 @@ fn main() {
     let conn_id_seed =
         ring::hmac::Key::generate(ring::hmac::HMAC_SHA256, &rng).unwrap();
 
+    let mut next_client_id = 0;
+    let mut clients_ids = ClientIdMap::new();
     let mut clients = ClientMap::new();
 
     let mut pkt_count = 0;
 
     let mut continue_write = false;
+
+    let local_addr = socket.local_addr().unwrap();
 
     loop {
         // Find the shorter timeout from all the active connections.
@@ -216,8 +221,8 @@ fn main() {
 
             // Lookup a connection based on the packet's connection ID. If there
             // is no connection matching, create a new one.
-            let client = if !clients.contains_key(&hdr.dcid) &&
-                !clients.contains_key(&conn_id)
+            let client = if !clients_ids.contains_key(&hdr.dcid) &&
+                !clients_ids.contains_key(&conn_id)
             {
                 if hdr.ty != quiche::Type::Initial {
                     error!("Packet is not Initial");
@@ -307,9 +312,14 @@ fn main() {
                 debug!("New connection: dcid={:?} scid={:?}", hdr.dcid, scid);
 
                 #[allow(unused_mut)]
-                let mut conn =
-                    quiche::accept(&scid, odcid.as_ref(), from, &mut config)
-                        .unwrap();
+                let mut conn = quiche::accept(
+                    &scid,
+                    odcid.as_ref(),
+                    local_addr,
+                    from,
+                    &mut config,
+                )
+                .unwrap();
 
                 if let Some(keylog) = &mut keylog {
                     if let Ok(keylog) = keylog.try_clone() {
@@ -332,9 +342,12 @@ fn main() {
                     }
                 }
 
+                let client_id = next_client_id;
+
                 let client = Client {
                     conn,
                     http_conn: None,
+                    client_id,
                     partial_requests: HashMap::new(),
                     partial_responses: HashMap::new(),
                     siduck_conn: None,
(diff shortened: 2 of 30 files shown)


normanmaurer:
@qdeconinck can you check the test-failure ?

qdeconinck:
The integration indicates a failure with `tests::reset_stream_data_recvd`with the target `quiche_multiarch (armv7-unknown-linux-gnueabihf)`. However, I'm not able to reproduce it.
```
qdeconinck@server:~/git/quiche$ cross test --target=armv7-unknown-linux-gnueabihf --verbose --all-targets --features=ffi,qlog
running 20 tests
test tests::get_bytes ... ok
test tests::get_bytes_mut ... ok
test tests::get_u ... ok
test tests::get_u_mut ... ok
test tests::get_varint ... ok
test tests::get_varint_mut ... ok
test tests::peek_bytes ... ok
test tests::peek_bytes_mut ... ok
test tests::peek_u ... ok
test tests::peek_u_mut ... ok
test tests::put_bytes ... ok
test tests::put_u ... ok
test tests::put_varint ... ok
test tests::slice ... ok
test tests::slice_last ... ok
test tests::slice_last_mut ... ok
test tests::slice_mut ... ok
test tests::split ... ok
test tests::split_at ... ok
test tests::varint_too_large - should panic ... ok

test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/deps/qlog-8e3d1a131b8c1e3a`

running 7 tests
test events::quic::tests::packet_header ... ok
test streamer::tests::serialization_states ... ok
test tests::packet_sent_event_no_frames ... ok
test tests::packet_sent_event_some_frames ... ok
test tests::trace_no_events ... ok
test tests::trace_seq_no_events ... ok
test tests::trace_single_transport_event ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/deps/quiche-7f38ad5740b58ee6`

running 388 tests
[...]
test result: ok. 388 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 82.75s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/client-4fdb7e1e89d4d433`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/http3_client-5b3b46fc488749cb`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/http3_server-4fca75b077c783ca`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/qpack_decode-f95ee9ee3a677544`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/qpack_encode-41eade3245482513`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/examples/server-cd750cae0dd058b8`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/deps/quiche_apps-53201971cdab96a0`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/deps/quiche_client-f5857d3a376ae482`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/linux-runner armv7 /target/armv7-unknown-linux-gnueabihf/debug/deps/quiche_server-807fbaebbba3012c`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

qdeconinck:
After investigation, it seems that the failing test was due to the slowness of the `quiche_multiarch` containers, sometimes increasing the value of the ACK delay above 63 and thus increasing by one byte the size of the packet. After adapting the test, all the checks should now pass.

apps/src/args.rs
```diff
@@ -68,6 +70,8 @@ pub struct CommonArgs {
 /// --dgram-proto PROTO         DATAGRAM application protocol.
 /// --dgram-count COUNT         Number of DATAGRAMs to send.
 /// --dgram-data DATA           DATAGRAM data to send.
+/// --max-active-cids NUM       Maximum number of active Connection IDs.
+/// --migrate                   Enable active connection migration.
```
LPardue, line 74: I'm not a huge fan of this flag name since it can be overloaded. How about `--enable-active-migration`?

apps/src/args.rs
```diff
@@ -216,6 +229,9 @@ Options:
   --no-grease              Don't send GREASE.
   --cc-algorithm NAME      Specify which congestion control algorithm to use [default: cubic].
   --disable-hystart        Disable HyStart++.
+  --max-active-cids NUM    The maximum number of active Connection IDs we can support [default: 2].
+  --migrate                Enable active connection migration.
```
LPardue, line 233: Same comment here regarding the name.

apps/src/client.rs
```diff
@@ -227,50 +251,57 @@ pub fn connect(
...
-            };
+            for event in &events {
+                let token: usize = event.token().into();
+                let socket = &sockets[token];
+                let (len, from) = match socket.recv_from(&mut buf) {
+                    Ok(v) => v,
+
+                    Err(e) => {
+                        // There are no more UDP packets to read, so end the read
+                        // loop.
```
LPardue, line 262: Is this comment true? Isn't there a possibility that there are other packets to read from other sockets?

apps/src/client.rs
```diff
@@ -227,50 +251,57 @@ pub fn connect(
...
+                        if e.kind() == std::io::ErrorKind::WouldBlock {
+                            trace!("recv() would block");
```
LPardue, line 264: we probably want to add detail to this line to say which socket would block

apps/src/client.rs
```diff
@@ -227,50 +251,57 @@ pub fn connect(
...
+
+                        return Err(ClientError::Other(format!(
+                            "recv() failed: {:?}",
+                            e
+                        )));
+                    },
+                };
 
-            trace!("got {} bytes", len);
+                trace!("got {} bytes", len);
```
LPardue, line 275: from experience, it helps to be as descriptive in these lines as possible. So like above let's consider if there is any way for the log line to discriminate _exactly what_ received the bytes.

apps/src/client.rs
```diff
@@ -378,38 +410,113 @@ pub fn connect(
...
-                Ok(v) => v,
-
-                Err(quiche::Error::Done) => {
-                    trace!("done writing");
-                    break;
+        // Handle path events.
+        while let Ok(qe) = conn.poll_path() {
+            match qe {
+                quiche::PathEvent::New(..) => {
+                    panic!("Should not see new path");
```
LPardue, line 417: I appreciate this is an semantic error but it would be great to avoid the explicit panic. This makes me wonder, does the library event return such an event?

Perhaps a clearer thing to do would be to use `unreachable!()` ?

apps/src/client.rs
```diff
@@ -78,14 +78,24 @@ pub fn connect(
         std::net::SocketAddr::V6(_) => "[::]:0",
     };
 
+    let mut sockets =
```
LPardue, line 81: If we are only going to use at most 2 sockets in the migration, I think it would be better to make that explicit (i.e. just name the objects) rather than have a potentially unbounded vector that requires magic indexes, what do you think?

apps/src/client.rs
```diff
@@ -446,3 +553,16 @@ pub fn connect(
 
     Ok(())
 }
+
+/// Generate a new pair of Source Connection ID and reset token.
+fn generate_cid_and_reset_token<T: SecureRandom>(
```
LPardue, line 558: There are two of these and they seem identical?

If so, please put common code in apps/src/common.rs

apps/src/common.rs
```diff
@@ -89,6 +89,8 @@ pub struct Client {
 
     pub http_conn: Option<Box<dyn HttpConn>>,
 
+    pub id: u64,
```
LPardue, line 92: Is there a more descriptive term than just `id` that we could use?

quiche/include/quiche.h
```diff
@@ -520,6 +621,12 @@ typedef struct {
 
     // DATAGRAM frame extension parameter, if any.
     ssize_t peer_max_datagram_frame_size;
+
+    // The stats of the connection's paths.
+    quiche_path_stats paths[8];
```
LPardue, line 626: where does `8` come from?

apps/src/args.rs
qdeconinck: Naming is complicated :) but yes, `--enable-active-migration` actually better describes what this option does. Will update that and the following occurrences.

apps/src/bin/quiche-server.rs
```diff
@@ -449,6 +470,71 @@ fn main() {
...
+            while let Ok(qe) = client.conn.poll_path() {
+                match qe {
+                    PathEvent::New(local_addr, peer_addr) => {
+                        info!("Seen new path ({}, {})", local_addr, peer_addr);
+                        // Directly probe the new path.
+                        client
+                            .conn
+                            .probe_path(local_addr, peer_addr)
+                            .expect("cannot probe");
+                    },
```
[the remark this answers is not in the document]
qdeconinck: Updated that, also integrated the `trace_id` of the client's connection for each received event.

apps/src/client.rs
qdeconinck: Good catch! Indeed, we should instead iterate over the events and loop on a given socket until there is nothing more to read on that one. Will update the code.

apps/src/client.rs
qdeconinck: Indeed. I added in the log the local address on which the socket is bound.

apps/src/client.rs
qdeconinck: Did the same as above.

apps/src/client.rs
qdeconinck: On a client-side connection, it should not. I replaced that to `unreachable!()`.

apps/src/client.rs
qdeconinck: I have no objection doing explicit naming instead of the current `sockets` solution. My original motivation to do so was to handle the variable number of sockets with slightly less code. I now have `socket` and `migrate_socket`, the second being an `Option` depending whether the application wants to perform migration or not.

apps/src/client.rs
qdeconinck: Good point, moved the duplicate codes into common.rs

apps/src/common.rs
qdeconinck: Maybe `client_id`? It is used as an intermediate identifier between Connection IDs and actual client's connections.

quiche/include/quiche.h
qdeconinck: It is actually a magic number to ensure the structure can hold up to 8 path's statistics. I could instead put a simple pointer along with a length, but it is unclear to me whether the caller should allocate the required space or the `ffi` should instead do this.

ghedo:
I pushed a commit that does a bit of refactoring and clean-up. For the most part there should be no functional change, however I did rename or otherwise change some of the new public APIs as follows:

* Changed back `accept()` and `connect()` to take a `&ConnectionId` instead of `ConnectionId<'static>

* Removed `Config::enable_events()`. Rationale being that an application is probably going to be broken without it as there's no way to actually disable migration in general.

* Renamed `path_send()` and `path_send_quantum()` to `send_on_path()` and * `send_quantum_on_path()`.

* Renamed `poll_path()` to `path_event_next()` and changed it to return `Option<>` instead of `Result<>`. In retrospect calling the HTTP/3 method "poll()" wasn't such a good idea due to clash and potential confusion with the standard `Future::poll()` method, so we might have to change it in the future.

* Renamed `retired_scids()` to `retired_scid_next()` and changed it to return a single `ConnectionId` wrapped in `Option<>`. This way we don't have to pre-allocate a whole vector of CIDs and we can also remove the custom iterator type.

* Renamed `sendable_paths()` to `paths_iter()`.

* Renamed `spare_scid_spots()` to `source_cids_left()`.

* Removed FFI API for now as it was incomplete and there were some issues. We'll add it back at a later time.

(I might have missed something in the list above)

quiche/src/lib.rs
```diff
@@ -5548,12 +6816,20 @@ fn drop_pkt_on_err(
     Error::Done
 }
 
+fn addr_tuple_to_string(from: SocketAddr, to: SocketAddr) -> String {
```
ehaydenr, line 6819: I think this avoids the intermediate allocation:
```
use std::net::SocketAddr;

fn main() {
    println!("Hello {}", AddrTupleFmt("1.2.3.4:80".parse().unwrap(), "4.5.6.7:80".parse().unwrap()));
}

struct AddrTupleFmt(SocketAddr, SocketAddr);

impl std::fmt::Display for AddrTupleFmt {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        let AddrTupleFmt(from, to) = &self;
        if from.ip().is_unspecified() || to.ip().is_unspecified() {
            Ok(())
        } else {
            f.write_fmt(format_args!("({} -> {})", from, to))
        }
    }
}
```

ghedo: Good idea, thanks.

qdeconinck:
The proposed refactoring looks good to me.

ghedo (approved):

ghedo:
Merged :tada:, thanks @qdeconinck for all your work and @ehaydenr for the review!

Issue #1217, opened 2022-04-15
quiche creates aliased mutable references to Connection; undefined behavior


withoutboats:
Looking at the API of quiche, I noted the use of `Pin` and was curious. Looking at the issues, I see #1210, which points out that `Connection` implements `Unpin`, and therefore is not protected by `Pin`. This issue is completely correct, and the API is unsound for that reason.

However, there is a deeper soundness problem in quiche. I think that any use of quiche will contain undefined behavior right now.

Every connection method takes `&mut self`; meanwhile, another pointer to the `Connection` is stored in the SSL context with `SSL_set_ex_data`, and revived to an `&mut Connection` in various callbacks passed to BoringSSL. These two mutable references to the same `Connection` are aliasing: a method on Connection calls into boring SSL which calls one of these callbacks, putting both mutable references on the stack at the same time. This is undefined behavior.

At minimum, this can possibly cause miscompilation as the optimizer might exploit the fact that the mutable reference to the connection should not alias with any other reference, when it actually does. More seriously, if these SSL callbacks can be called from a different thread, this will cause unpredictable data races (I don't know enough about BoringSSL to know if this is possible).

The connection should be stored in an `UnsafeCell` and all mutable accesses to it should never alias. This can possibly be done with carefully written unsafe code if the BoringSSL callbacks will never be used from another thread. It can also be done with some sort of synchronization primitive like RefCell or Mutex or RwLock at the expense of runtime overhead.

I would recommend then wrapping the Connection type in an opaque type, as discussed back in #45. The outer connection would contain a `Box<UnsafeCell<ConnectionInner>>` and be careful to never hold a reference to ConnectionInner while calling into BoringSSL that may call one of the QUIC callbacks.

PR #1227, opened 2022-05-13
only store &mut Connection in SSL ex_data when necessary

### LINKED ISSUE #1210: Connection implements Unpin but needs to be pinned
`quiche::connect` returns a `Pin<Box<Connection>>`. According to https://github.com/cloudflare/quiche/pull/249#issue-519321868 this is because `Connection` contains pointers to its internal contents.

However `Connection` automatically implements `Unpin`, so you can move it out of the pin via `Pin::into_inner`, causing unsafety in safe code.

`Connection` should have a `PhantomPinned` member, and methods on it taking `&mut self` should be changed into taking `self: Pin<&mut Self>`.

ghedo:
We currently need to access the quiche connection state during the TLS
handshake inside BoringSSL's various callbacks, and to do this, when a
Connection structure in created, we store a pointer to it in the SSL
state using the "ex data" API.

This however has the disadvantage that the pointer to the Connection
structure needs to be stable for the whole duration of the connection,
which means that Connection is wrapped in `Pin<Box<...>>` to avoid
applications accidentally moving the structure (thus invalidating the
pointer stored in ex_data).

But we don't actually need the ex data pointer during the whole
connection, in fact, we only need it when calling specific BoringSSL
APIs (e.g. `SSL_do_handshake()`) that in turn call the callbacks
registered by quiche, which means that we can actually set the ex data
pointer just before calling those APIs, and clear it immediately after.

This removes the needs to pin the Connection structure, as it can't be
moved while we hold a mutable reference it (due to being inside one of
quiche's methods), and simplifies the API a bit (as applications don't
need the `Pin<Box<...>>` wrapper anymore).

Fixes #1210.

### DIFF
--- a/quiche/src/tls.rs
+++ b/quiche/src/tls.rs
@@ -371,12 +371,8 @@ impl Handshake {
         unsafe { SSL_get_error(self.as_ptr(), ret_code) }
     }
 
-    pub fn init(
-        &mut self, conn: *const Connection, is_server: bool,
-    ) -> Result<()> {
+    pub fn init(&mut self, is_server: bool) -> Result<()> {
         self.set_state(is_server);
-
-        self.set_ex_data(*QUICHE_EX_DATA_INDEX, conn)?;
 
         self.set_min_proto_version(TLS1_3_VERSION);
         self.set_max_proto_version(TLS1_3_VERSION);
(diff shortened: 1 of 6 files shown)


LPardue (approved):
neat

withoutboats:
This also resolves #1217. Looks great!

quiche/src/tls.rs
```diff
@@ -558,13 +554,21 @@ impl Handshake {
         map_result_ssl(self, rc)
     }
 
-    pub fn do_handshake(&mut self) -> Result<()> {
+    pub fn do_handshake(&mut self, conn_ptr: *mut Connection) -> Result<()> {
```
withoutboats, line 557: You might be better off passing `&mut Connection` here and the other method rather than a raw pointer, to guarantee that connection is borrowed until the handshake is done.

ghedo: Due to how this method is called in `lib.rs` (`self.handshake.do_handshake(...)`), doing `self.handshake.do_handshake(self)` is considered double mutable reference, so using `&mut Connection` directly doesn't seem possible without some larger refactoring.

ghedo: Yeah, I was working on a separate change to address the issue you reported specifically, see https://github.com/cloudflare/quiche/pull/1231.

PR #1231, opened 2022-05-18
avoid storing Connection pointer in TLS ex data

### LINKED ISSUE #1217: quiche creates aliased mutable references to Connection; undefined behavior

ghedo:
Follow-up to 3c7a597a3765995a5434eb08f838caba65112a6f.

Instead of storing a pointer to the whole `Connection` structure in the
TLS ex data, this change introduces a new `ExData` structure that
provides references to specific `Connection` fields needed in the TLS
callbacks.

This way we avoid having aliased mutable references to a `Connection`
when the BoringSSL callbacks are being executed (which is undefined
behaviour).

This needs to make `QUICHE_EX_DATA_INDEX` private again, since the
`ExData` structure is not public and can't be used by applications
(though nothing should be using this right now).

Fixes #1217.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4860,13 +4860,27 @@ impl Connection {
     ///
     /// If the connection is already established, it does nothing.
     fn do_handshake(&mut self) -> Result<()> {
-        let conn_ptr = self as &mut Connection as *mut Connection;
+        let mut ex_data = tls::ExData {
+            application_protos: &self.application_protos,
 
-        if self.is_established() {
-            return self.handshake.process_post_handshake(conn_ptr);
+            pkt_num_spaces: &mut self.pkt_num_spaces,
+
+            session: &mut self.session,
+
+            local_error: &mut self.local_error,
+
+            keylog: self.keylog.as_mut(),
+
+            trace_id: &self.trace_id,
+
+            is_server: self.is_server,
+        };
+
+        if self.handshake_completed {
+            return self.handshake.process_post_handshake(&mut ex_data);
         }
 
-        match self.handshake.do_handshake(conn_ptr) {
+        match self.handshake.do_handshake(&mut ex_data) {
             Ok(_) => (),
 
             Err(Error::Done) => {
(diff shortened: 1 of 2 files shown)


withoutboats:
This definitely solves the issue in #1217. Another approach would have been to permanently store these fields in a struct that's a field of Connection; then you wouldn't have double indirection when you use them from BoringSSL. Both approaches seem fine to me.

ghedo:
Yeah, I thought about grouping these fields together, but some of them are kind of unrelated to each other other than the fact they are accessed in the callbacks, so it seemed weird to re-arrange the whole `Connection` for this.

LPardue (approved):

PR #1236, opened 2022-05-23
dgram: don't allocate full queue upfront

ghedo:
Instead of allocating the max queue size upfront, allocate a smaller
size and allow the queue to grow to the maximum limit.

### DIFF
--- a/quiche/src/dgram.rs
+++ b/quiche/src/dgram.rs
@@ -29,6 +29,8 @@ use crate::Result;
 
 use std::collections::VecDeque;
 
+const DGRAM_QUEUE_DEFAULT_LEN: usize = 256;
+
 /// Keeps track of DATAGRAM frames.
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


LPardue (approved):

quiche/src/dgram.rs
```diff
@@ -40,7 +42,9 @@ pub struct DatagramQueue {
 impl DatagramQueue {
     pub fn new(queue_max_len: usize) -> Self {
         DatagramQueue {
-            queue: VecDeque::with_capacity(queue_max_len),
+            queue: VecDeque::with_capacity(
```
inikulin, line 45: Seems to be a breaking change and would need a docs update? Also, will `queue_max_len` be useful in any way now?

ghedo: This only changes how much of the queue is allocated upfront. The max len limit still applies.

PR #1237, opened 2022-05-23
Add a new API for setting application protocols from a slice of byte strings

jyn514:
This avoids requiring clients to do tedious and error-prone length calculations manually.
As a bonus, it avoids some unnecessary copies in quiche itself.

- Relax the lifetime constraints on `Octets`
- Rename `set_application_protos` to `set_application_protos_wire_format`
- Add a new `set_application_protos` function that takes a slice of byte strings
- Change `set_alpn` to take a slice of slices instead of a slice of `Vec`
- Changes apps to use the new API

### DIFF
--- a/quiche/examples/client.rs
+++ b/quiche/examples/client.rs
@@ -81,9 +81,13 @@ fn main() {
     config.verify_peer(false);
 
     config
-        .set_application_protos(
-            b"\x0ahq-interop\x05hq-29\x05hq-28\x05hq-27\x08http/0.9",
-        )
+        .set_application_protos(&[
+            b"hq-interop",
+            b"hq-29",
+            b"hq-28",
+            b"hq-27",
+            b"http/0.9",
+        ])
         .unwrap();
 
     config.set_max_idle_timeout(5000);
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.56 as build
+FROM rust:1.57 as build
 
 WORKDIR /build
 
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ is disabled by default), by passing ``--features ffi`` to ``cargo``.
 Building
 --------
 
-quiche requires Rust 1.56 or later to build. The latest stable Rust release can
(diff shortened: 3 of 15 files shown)


quiche/examples/client.rs
```diff
@@ -82,7 +82,7 @@ fn main() {
 
     config
         .set_application_protos(
-            b"\x0ahq-interop\x05hq-29\x05hq-28\x05hq-27\x08http/0.9",
+            &[b"hq-interop", b"5hq-29", b"hq-28", b"hq-27", b"http/0.9"],
```
jyn514, line 85: this `5` is a bug, will fix

LPardue:
I think we can delete `alpns::length_prefixed()` function with this change.

LPardue:
FWIW the current build errors in Stable appear to be due to a stale version of `octets` being pulled in via crates.io

```
    Packaging quiche v0.14.0 (/home/runner/work/quiche/quiche/quiche)
   Archiving COPYING
   Archiving Cargo.lock
    Updating crates.io index
note: package `octets v0.1.0` added to the packaged Cargo.lock file, was originally sourced from `/home/runner/work/quiche/quiche/octets`

<snip>

error[E0499]: cannot borrow `b` as mutable more than once at a time
Error:    --> src/lib.rs:860:31
    |
860 |         while let Ok(proto) = b.get_bytes_with_u8_length() {
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `b` was mutably borrowed here in the previous iteration of the loop

error[E0597]: `proto` does not live long enough
Error:    --> src/lib.rs:861:30
    |
861 |             protos_list.push(proto.buf());
    |             -----------------^^^^^^^^^^^-
    |             |                |
    |             |                borrowed value does not live long enough
    |             borrow later used here
862 |         }
    |         - `proto` dropped here while still borrowed

Some errors have detailed explanations: E0499, E0597.
```

not sure if there are ways to circumvent that failure, it's a bit ugly if we can't test packaging quiche with crate updates in the same changeset

ghedo (approved):

ghedo (approved):

ghedo:
Merged, thanks @jyn514!

PR #1238, opened 2022-05-26
recovery: add GSO-friendly pacer

junhochoi:
Currently we update the timestamp for next packet
(`last_packet_scheduled_time`) for each packet based on
the current pacing rate, which can confuse the
rtt measurement because the recorded timestamp
and actual time sent can be different, especially when
the application sends in burst using send_quantum
and batches a group packets using GSO. In such case
rtt measured can be smaller than expected, or even 0
when the packet is not sent in a specified time.

Pacer provides the timestamp for next packet to be sent
based on the current send_quantum, pacing rate
and last updated time. It is reset when a valid packet is
received (including ACK and non-ACK packets) to
start a new burst.

Pacer is GSO-friendly, which means it will provides a same
timestamp for same sized (except last one) packets in a
GSO group.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2462,6 +2462,10 @@ impl Connection {
 
         self.ack_eliciting_sent = false;
 
+        // Reset pacer and start a new burst when a valid
+        // packet is received.
+        self.recovery.pacer.reset(now);
+
         Ok(read)
     }
 
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -145,9 +145,7 @@ pub struct Recovery {
     hystart: hystart::Hystart,
 
     // Pacing.
-    pacing_rate: u64,
-
-    last_packet_scheduled_time: Instant,
+    pub pacer: pacer::Pacer,
 
     // RFC6937 PRR.
     prr: prr::PRR,
@@ -162,6 +160,9 @@ pub struct Recovery {
 
 impl Recovery {
     pub fn new(config: &Config) -> Self {
+        let initial_congestion_window =
+            config.max_send_udp_payload_size * INITIAL_WINDOW_PACKETS;
+
         Recovery {
             loss_detection_timer: None,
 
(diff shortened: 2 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -1897,6 +1897,11 @@ impl Connection {
             }
         }
 
+        // Reset pacer and start a new burst if packet received.
+        if done > 0 {
+            self.recovery.pacer.reset();
```
ghedo, line 1902: This will be called even when an invalid packet is called (e.g. random Internet junk), is this ok? Maybe just move this at the end of `recv_single()`? This way we can also avoid the `Instant::now()` call inside, and just pass the `now` we already have.

quiche/src/recovery/mod.rs
```diff
@@ -321,37 +322,33 @@ impl Recovery {
 
     pub fn set_pacing_rate(&mut self, rate: u64) {
         if rate != 0 {
+            self.pacer.update(self.send_quantum, rate);
+
             self.pacing_rate = rate;
```
ghedo, line 327: Do we still need this field? Looks like it is only used in `schedule_next_packet()` to check `pacing_rate == 0`, and we could just use `pacer.pace` instead (e.g. adding a `rate()` method to `Pacer`).

quiche/src/recovery/mod.rs
```diff
@@ -321,37 +322,33 @@ impl Recovery {
...
         &mut self, epoch: packet::Epoch, now: Instant, packet_size: usize,
     ) {
         // Don't pace in any of these cases:
         //   * Packet epoch is not EPOCH_APPLICATION.
-        //   * Packet contains only ACK frames.
-        //   * The start of the connection.
-        if epoch != packet::EPOCH_APPLICATION ||
-            packet_size == 0 ||
-            self.bytes_sent <= self.congestion_window ||
+        //   * Packet contains has no data.
```
ghedo, line 340: It looks like you removed the `packet_size == 0` check, so this comment is not true anymore?

But also, typo :)

```suggestion
        //   * Packet contains no data.
```

ghedo (changes requested):

quiche/src/lib.rs
junhochoi: Good idea. Moved into recv_single().

quiche/src/recovery/mod.rs
junhochoi: `Recovery.pacing_rate` is removed and replaced by `pacer.rate()`.

ghedo (approved):
LGTM. I just tweaked the pacer module documentation a bit.

PR #1244, opened 2022-06-13
fix stream limit enforcement

ehaydenr:
RFC9000 2.1: A stream ID that is used out of order results in all streams of that type with lower-numbered stream IDs also being opened.


ghedo (approved):

ghedo:
Merged, thanks @ehaydenr!

PR #1247, opened 2022-06-16
feat: add capability to choose source port from client

RushilMehra:


lohith-bellad (approved):
LGTM

LPardue (dismissed):

apps/src/args.rs
```diff
@@ -298,6 +300,13 @@ impl Args for ClientArgs {
             None
         };
 
+        let source_port = args.get_str("--source-port");
+        let source_port = if !source_port.is_empty() {
+            source_port.parse::<u16>().unwrap()
+        } else {
+            0
+        };
```
junhochoi, line 308: Since it has a default value, you don't need to call `is_empty()`.

```suggestion
        let source_port = source_port.parse::<u16>().unwrap();
```

RushilMehra: Good point, thanks

junhochoi (approved):

PR #1252, opened 2022-06-17
Add support for `--trust-origin-ca-pem`

jyn514:
This allows testing local servers, not just production servers with a public CA certificate.

### DIFF
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -232,6 +232,7 @@ Options:
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --session-file PATH      File used to cache a TLS session for resumption.
   --source-port PORT       Source port to use when connecting to the server [default: 0].
+  --trust-origin-ca-pem <file>  Path to the pem file of the origin's CA, if not publicly trusted.
   -h --help                Show this screen.
 ";
 
@@ -250,6 +251,7 @@ pub struct ClientArgs {
     pub session_file: Option<String>,
     pub source_port: u16,
     pub perform_migration: bool,
+    pub trust_origin_ca_pem: Option<String>,
 }
 
 impl Args for ClientArgs {
@@ -318,6 +320,13 @@ impl Args for ClientArgs {
 
         let perform_migration = args.get_bool("--perform-migration");
 
+        let trust_origin_ca_pem = args.get_str("--trust-origin-ca-pem");
+        let trust_origin_ca_pem = if !trust_origin_ca_pem.is_empty() {
+            Some(trust_origin_ca_pem.to_string())
+        } else {
+            None
+        };
+
         ClientArgs {
             version,
             dump_response_path,
@@ -332,6 +341,7 @@ impl Args for ClientArgs {
             session_file,
             source_port,
             perform_migration,
+            trust_origin_ca_pem,
         }
     }
 }
@@ -352,6 +362,7 @@ impl Default for ClientArgs {
             session_file: None,
             source_port: 0,
             perform_migration: false,
+            trust_origin_ca_pem: None,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


quiche/src/tls.rs
```diff
@@ -1022,7 +1022,10 @@ extern fn new_session(ssl: *mut SSL, session: *mut SSL_SESSION) -> c_int {
 fn map_result(bssl_result: c_int) -> Result<()> {
     match bssl_result {
         1 => Ok(()),
-        _ => Err(Error::TlsFail),
+        _ => {
+            log_ssl_error();
```
jyn514, line 1026: when testing this, I used a .crt instead of .pem by accident, and was very confused why I got TlsFail. This makes it a little easier to debug.

junhochoi (approved):

apps/src/args.rs
```diff
@@ -219,6 +219,7 @@ Options:
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --session-file PATH      File used to cache a TLS session for resumption.
+  --trust-origin-ca-pem <file>  Path to the pem file of the origin's CA, if not publicly trusted.
```
ghedo, line 222:
```suggestion
  --trust-ca <file>  Path to a CA trust file in PEM format.
```

apps/src/args.rs
```diff
@@ -235,6 +236,7 @@ pub struct ClientArgs {
     pub method: String,
     pub connect_to: Option<String>,
     pub session_file: Option<String>,
+    pub trust_origin_ca_pem: Option<String>,
```
ghedo, line 239:
```suggestion
    pub trust_ca: Option<String>,
```

apps/src/args.rs
```diff
@@ -298,6 +300,13 @@ impl Args for ClientArgs {
             None
         };
 
+        let trust_origin_ca_pem = args.get_str("--trust-origin-ca-pem");
```
ghedo, line 303:
```suggestion
        let trust_ca = args.get_str("--trust-ca");
```

ghedo: ...and so forth :D

apps/src/client.rs
```diff
@@ -89,7 +89,18 @@ pub fn connect(
...
+        config
+            .load_verify_locations_from_file(trust_origin_ca_pem)
+            .map_err(|e| {
+                ClientError::Other(format!(
+                    "error loading origin CA file : {}",
+                    e
+                ))
+            })?;
+    } else {
+        config.verify_peer(!args.no_verify);
```
ghedo, line 102: Why put this in the `else` block? I could see `--no-verify` being used together with the new option for testing. Say for example I connect to a server with the new option enabled, but it still doesn't work, I might try to append `--no-verify` to quickly check if that works, but with the current implementation that would also require removing the other option which can be annoying.

ghedo (changes requested):
LGTM overall, but added some comments.

ghedo:
Also needs rebase plz.

apps/src/args.rs
jyn514: Hmm, can it be `--trust-ca-pem` at least? It wasn't clear to me at first that only PEM files are accepted.

Or maybe it's possible to accept either .pem or .crt files? is that the difference between `load_verify_locations_from_file` and `use_certificate_chain_file`?

ghedo: Well, that's kind of what the help text is for ("...in PEM format.").

Also, no, `use_certificate_chain_file` is for configuring the full certificate chain that is presented to the peer, while `load_verify_locations_from_file` is for configuring the CA certificates used to _verify_ the peer's certificate.

jyn514:
I'm not planning to keep working on this.

PR #1253, opened 2022-06-17
MSVC generator hack should be only applied to MSVC

msoxzw:
Related to https://github.com/cloudflare/boring/pull/77


junhochoi:
If we do this, I think it's good to add some CI test for windows non-msvc (mingw?) toolchain.

msoxzw:
Not really. Contrary to https://github.com/cloudflare/boring/pull/77, MinGW toolchain could find BoringSSL library, but subsequently, its linker "ld" fails to link.

ghedo (approved):

PR #1256, opened 2022-06-21
tls: allow retrieving the full certificate chain exposed by the peer

mbiggio:
In validating a peer's certificate chain, in case intermediate CAs are
present, it is necessary to retrieve the full chain exposed by the peer.
Only the root CA must be trusted present in the certificate store the caller

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -5566,6 +5566,17 @@ impl Connection {
         self.handshake.peer_cert()
     }
 
+    /// Returns the peer's certificate chain (if any) as a vector of DER-encoded
+    /// buffers.
+    ///
+    /// The certificate at index 0 is the peer's leaf certificate, the other
+    /// certificates (if any) are the chain certificate authorities used to
+    /// sign the leaf certificate.
+    #[inline]
+    pub fn peer_cert_chain(&self) -> Option<Vec<&[u8]>> {
+        self.handshake.peer_cert_chain()
+    }
+
     /// Returns the serialized cryptographic session for the connection.
     ///
     /// This can be used by a client to cache a connection's session, and resume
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -4592,10 +4592,12 @@ impl Connection {
...
     }
 
-    /// Returns the peer's leaf certificate (if any) as a DER-encoded buffer.
+    /// Returns the certificate at index `i` (if any) in the certificate
+    /// chain exposed by the peer as a DER-encoded buffer. The peer's leaf
+    /// certificate is the one at index 0.
     #[inline]
-    pub fn peer_cert(&self) -> Option<&[u8]> {
-        self.handshake.peer_cert()
+    pub fn peer_cert(&self, i: i32) -> Option<&[u8]> {
```
ghedo, line 4599: I think it would be better to add a whole new public `peer_cert_chain()` method that returns the full chain in one go, rather than change the existing API. It could return a `Option<Vec<&[u8]>>` or something like that.

ghedo (changes requested):
Thanks for the PR, added a comment.

quiche/src/lib.rs
mbiggio: Thanks for reviewing. Yes, I completely agree, I changed the PR accordingly.

quiche/src/lib.rs
```diff
@@ -4598,6 +4598,14 @@ impl Connection {
         self.handshake.peer_cert()
     }
 
+    /// Returns the peer's certificate chain (if any) as a vector of DER-encoded buffers.
```
ghedo, line 4601:
```suggestion
    /// Returns the peer's certificate chain (if any) as a vector of DER-encoded buffers.
    ///
```

ghedo (changes requested):

quiche/src/lib.rs
```diff
@@ -9438,6 +9447,29 @@ mod tests {
...
+    fn peer_cert_chain() {
+        let mut config = Config::new(PROTOCOL_VERSION).unwrap();
+        config
+            .load_cert_chain_from_pem_file("examples/cert-big.crt")
+            .unwrap();
+        config
+            .load_priv_key_from_pem_file("examples/cert.key")
+            .unwrap();
+        config
+            .set_application_protos(b"\x06proto1\x06proto2")
```
ghedo, line 9460:
```suggestion
            .set_application_protos(&[b"proto1", b"proto2"])
```

ghedo (changes requested):

quiche/src/tls.rs
```diff
@@ -645,6 +645,39 @@ impl Handshake {
...
+            let num = sk_num(chain);
+            if num <= 0 {
+                return None;
+            }
+
+            let mut cert_chain = vec![];
+            for i in 0..num {
+                let buffer =
+                    map_result_ptr(sk_value(chain, i) as *const CRYPTO_BUFFER)
+                    .ok()?;
```
ghedo, line 662:
```suggestion
                        .ok()?;
```

ghedo (approved):

ghedo:
Merged, thanks @mbiggio!

PR #1258, opened 2022-06-21
process_post_handshake short-circuit

ehaydenr:
in a download test I noticed 3.29% of CPU time spent in do_handshake. After this change it is no longer showing up in perf output. I'm not familiar enough with BoringSSL QUIC APIs to say for certain this is correct usage.


ghedo (approved):

alexj0l:
Hi @ehaydenr . May I ask how did you perfed the quiche? Is there any tool for that? I would like to test the Throughput

PR #1260, opened 2022-06-22
ci: disable mingw32 test

junhochoi:
Looks like mingw32 test executable crashes outside the
tests occasionally. Since windows tests are covered by
msvc builds, mingw32 is switched to build only, not
running test to prevent CI flakiness.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -165,7 +165,15 @@ jobs:
         with:
           args: install nasm
 
+      - name: Run cargo build
+        if: matrix.target == 'x86_64-pc-windows-gnu'
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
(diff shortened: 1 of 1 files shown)


.github/workflows/stable.yml
```diff
@@ -165,7 +165,14 @@ jobs:
         with:
           args: install nasm
 
+      - name: Run cargo build
```
ghedo, line 168: This will cause all the crates to be built, and then get re-built immediately after for the tests step. Can we limit this to just `matrix.target == 'x86_64-pc-windows-gnu'`?

ghedo (changes requested):

.github/workflows/stable.yml
junhochoi: Sure. Done.

ghedo (approved):

PR #1264, opened 2022-06-27
Change generic struct names to use quiche prefix in FFI

tbu-:
Otherwise, this clashes with projects that define their own `struct
Connection` or `struct Config`.


ghedo (approved):

ghedo:
Merged, thanks @tbu-!

PR #1265, opened 2022-06-27
quiche-client: add --send-priority-update param

LPardue:
quiche-client has supported two methods of sending extensible priorities:

  1) via a header, controlled by the `-H` parameter. This applies to all
     requests, if more than one URL (or the `-n` parameter) was
     provided.
  2) via a custom querystring format such as `?u=1&i=1`. This format is
     understood by quiche-server, facilitating experimentation with
     priorities by supplying multiple URLs with individual priorities.

With this change, we enhance the second method with the optional
`--send-priority-update` parameter. When it is supplied, quiche-client
will use the querystring as input to construct an individual
PRIORITY_UPDATE frame for the request.

quiche-server has been updated to reflect this change. It now attempts
to take PRIORITY_UPDATE frames when building a response, and sets the
response priority accordingly.


apps/src/common.rs
```diff
@@ -1488,9 +1524,43 @@ impl HttpConn for Http3Conn {
...
+                        match quiche::h3::Priority::try_from(priority.as_slice())
+                        {
+                            Ok(v) => v,
+                            Err(_) => quiche::h3::Priority::default(),
+                        };
+
+                    #[cfg(not(feature = "sfv"))]
+                    let priority = quiche::h3::Priority::default();
+
+                    info!(
```
junhochoi, line 1559: `debug!()` is better?

apps/Cargo.toml
```diff
@@ -26,6 +26,7 @@ env_logger = "0.6"
 mio = { version = "0.8", features = ["net", "os-poll"] }
 url = "1"
 log = "0.4"
+octets = { version = "0.2", path = "../octets" }
```
junhochoi, line 29: Do we need `version` here?

LPardue: I just copied what quiche/Cargo.toml does

apps/src/common.rs
LPardue: This is on the server, where receiving the priority event trigger, and receiving request headers are logged at info level. Having the server state what priority it is applying at info level seems helpful to me.

apps/Cargo.toml
LPardue: ah, yes!

junhochoi (approved):

PR #1269, opened 2022-07-01
apps: add limited support for HTTP/3 settings tuning

LPardue:
With this change, add support for three new quiche app CLI options that
control HTTP/3 settings parameter values.

When these options are not provided, the applications will omit sending
the HTTP/3 settings, which by RFC 9114 means that the peer will use the
default values.

When the `--max-field-section-size` option is provided, the setting is
sent with that value.

quiche does not support dynamic QPACK. If the
`--qpack-max-table-capacity` or `--qpack-blocked-streams` options are
provided, their values are clamped to 0 before being sent in a setting.


ghedo (approved):

DieLokoMotive:
11

PR #1275, opened 2022-07-05
recovery: binary search sent packets

ehaydenr:


ehaydenr:
This feels like more of a bandaid. I've submitted some other changes that have reduced the queue size (e.g. #1312 )

PR #1280, opened 2022-07-13
h3: add recv_dgram_vec

ehaydenr:


ehaydenr:
Never mind - now that `octets` is a public crate, the caller can do this themselves

PR #1291, opened 2022-08-03
h3: remove dgram API in order to add support for RFC version

LPardue:
h3: add HTTP/3 datagram RFC version
    
    With this change, we add support for advertising the HTTP/3 datagram SETTING
    value (0x33) defined in draft 11 and RFC-to-be when an h3 Connection is
    configured to have datagrams enabled.
    
    The format of the the HTTP/3 datagrams are pretty much structurally identical,
    so there is not much meaningful difference between versions. Applications that
    rely on HTTP/3 datagrams, such as to use CONNECT-UDP, would apply their own
    semantics to these datagrams and can manage that themselves. Therefore, this
    change doesn't provide fine grain control over versions.

h3: remove HTTP/3 datagram API
    
    An HTTP/3 datagram is a series of fields carried in a QUIC DATAGRAM frame
    payload. In draft 00, draft 11 and what will be the RFC there is a single field,
    which is a variable-length integer.
    
    Previously, the h3 module provided support functions that were a thin shim over
    quiche's transport-layer datagram methods, which helped with the variable-length
    encoding of the first field in the HTTP/3 datagram. However, in the meantime we
    spun out the Octets crate, which allows applications to easily handle varints
    themselves. Furthermore, changes in specifications like CONNECT-UDP mean that
    applications are probably required to do additional processing of varint fields
    that might exist after the first field. At this stage, the h3 datagram API
    provides minimal utility and in the worst case is confusing for applications
    that would use it for different purposes.
    
    This change removes quiche::h3::send_dgram() and quiche::h3::recv_dgram()
    functions. Instead applications are now responsible for serialization /
    deserialization of HTTP/3 datagrams, with the recommendation being that they use
    the Octets crate. This is demonstrated by related changes to the apps.
    
    Since the Octets crate does not provide a C API itself, helper function have
    been added to quiche's C API to support varint handling.

### DIFF
--- a/apps/src/common.rs
+++ b/apps/src/common.rs
@@ -313,6 +313,26 @@ pub fn priority_from_query_string(url: &url::Url) -> Option<Priority> {
     }
 }
 
+fn send_h3_dgram(
+    conn: &mut quiche::Connection, flow_id: u64, dgram_content: &[u8],
+) -> quiche::Result<()> {
+    info!(
+        "sending HTTP/3 DATAGRAM on flow_id={} with data {:?}",
+        flow_id, dgram_content
+    );
+
+    let len = octets::varint_len(flow_id) + dgram_content.len();
+    let mut d = vec![0; len];
+    let mut b = octets::OctetsMut::with_slice(&mut d);
+
+    b.put_varint(flow_id)
+        .map_err(|_| quiche::Error::BufferTooShort)?;
+    b.put_bytes(dgram_content)
+        .map_err(|_| quiche::Error::BufferTooShort)?;
+
+    conn.dgram_send(&d)
+}
+
 pub trait HttpConn {
     fn send_requests(
         &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
@@ -1177,17 +1197,8 @@ impl HttpConn for Http3Conn {
             let mut dgrams_done = 0;
 
             for _ in ds.dgrams_sent..ds.dgram_count {
-                info!(
-                    "sending HTTP/3 DATAGRAM on flow_id={} with data {:?}",
-                    ds.flow_id,
-                    ds.dgram_content.as_bytes()
-                );
-
-                match self.h3_conn.send_dgram(
-                    conn,
-                    0,
-                    ds.dgram_content.as_bytes(),
-                ) {
+                match send_h3_dgram(conn, ds.flow_id, ds.dgram_content.as_bytes())
+                {
                     Ok(v) => v,
 
                     Err(e) => {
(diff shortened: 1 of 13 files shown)


quiche/src/h3/frame.rs
```diff
@@ -520,6 +525,14 @@ fn parse_settings_frame(
                 qpack_blocked_streams = Some(value);
             },
 
+            SETTINGS_H3_DATAGRAM_00 => {
```
ehaydenr, line 528: since the bodies are the same, consider collapsing this into:
```
SETTINGS_H3_DATAGRAM_00 | SETTINGS_H3_DATAGRAM =>
```

quiche/include/quiche.h
```diff
@@ -462,7 +462,6 @@ bool quiche_stream_iter_next(quiche_stream_iter *iter, uint64_t *stream_id);
 
 // Frees the given stream iterator object.
 void quiche_stream_iter_free(quiche_stream_iter *iter);
-
 typedef struct {
```
ghedo, line 465: Nit:
```suggestion

typedef struct {
```

ghedo (changes requested):
LGTM other than minor nitpick.

quiche/src/h3/frame.rs
DieLokoMotive: SETTINGS_H3_DATAGRAM_00 | SETTINGS_H3_DATAGRAM =>

DieLokoMotive: 11

DieLokoMotive: > since the bodies are the same, consider collapsing this into:
> ```
> SETTINGS_H3_DATAGRAM_00 | SETTINGS_H3_DATAGRAM =>
> ```

11

apps/src/common.rs
```diff
@@ -313,6 +313,26 @@ pub fn priority_from_query_string(url: &url::Url) -> Option<Priority> {
...
+fn send_h3_dgram(
+    conn: &mut quiche::Connection, flow_id: u64, dgram_content: &[u8],
+) -> quiche::Result<()> {
+    info!(
+        "sending HTTP/3 DATAGRAM on flow_id={} with data {:?}",
+        flow_id, dgram_content
+    );
+
+    let len = octets::varint_len(flow_id) + dgram_content.len();
+    let mut d = vec![0; len];
```
vkrasnov, line 325: I know this is similar to current implementation, but reallocating the entire datagram and copying it is extremely wasteful

vkrasnov: The API should expose a method that allows to pass a generic that implements AsRef<[u8]>, and then those can be queued with the flow_id and the actual copying only happens when they are sent on the wire. At the very list a method that accepts a Vec<u8> and avoids reallocating

ghedo (approved):

PR #1298, opened 2022-08-12
quic: configurable drain timeout

ehaydenr:
When the quiche-calculated RTT is high, the draining_timer can become very large. Make the draining timer more configurable and make sure the drain timeout is never larger than the idle timeout.


ehaydenr:
will implement in the application instead by dropping after `is_draining` and before `is_closed`

PR #1307, opened 2022-09-01
h3: support SETTINGS_ENABLE_CONNECT_PROTOCOL

ehaydenr:


quiche/src/h3/mod.rs
```diff
@@ -552,6 +554,13 @@ impl Config {
     pub fn set_qpack_blocked_streams(&mut self, v: u64) {
         self.qpack_blocked_streams = Some(v);
     }
+
+    /// Sets or omits the `SETTINGS_ENABLE_CONNECT_PROTOCOL` setting.
+    ///
+    /// The default value is `0`.
+    pub fn set_connect_protocol_enabled(&mut self, v: u64) {
```
LPardue, line 561: this would probably be better modelled as a bool such as `enable_extended_connect(enabled: bool)`

LPardue: modelling the setting internally as an uint is fine, but there's little reason for us to let users of quiche set random values when the RFC doesn't permit that

ehaydenr: thanks, addressed in ced7972025b1c0993642df69fa09520713d320bc

quiche/src/h3/mod.rs
```diff
@@ -1280,6 +1296,17 @@ impl Connection {
...
     }
 
+    /// Returns whether the peer enabled HTTP/3 CONNECT protocol support.
+    ///
+    /// Support is signalled by the peer's SETTINGS, so this method always
+    /// returns false until they have been processed using the [`poll()`]
+    /// method.
+    ///
+    /// [`poll()`]: struct.Connection.html#method.poll
+    pub fn connect_protocol_enabled_by_peer(&self) -> bool {
```
LPardue, line 1306: for consistency with `enable_extended_connect`, how about 
```suggestion
    pub fn extended_connect_enabled_by_peer(&self) -> bool {
```

quiche/src/h3/mod.rs
```diff
@@ -1280,6 +1296,17 @@ impl Connection {
             conn.dgram_max_writable_len().is_some()
     }
 
+    /// Returns whether the peer enabled HTTP/3 CONNECT protocol support.
```
LPardue, line 1299: for consistency with enable_extended_connect, how about
```suggestion
    /// Returns whether the peer enabled extended CONNECT support.
```

LPardue (commented):
This is looking good. For completeness, we probable want to add equivalent capability to the C FFI, meaning adding `quiche_h3_config_enable_extended_connect` to https://github.com/cloudflare/quiche/blob/master/quiche/src/h3/ffi.rs#L72 and https://github.com/cloudflare/quiche/blob/master/quiche/include/quiche.h#L771

LPardue (approved):

junhochoi (approved):

PR #1310, opened 2022-09-05
Multipath support with non-zero length Connection IDs

qdeconinck:
This commits introduces the following features.

- Leave to the application to choose the multipath extensions or not
- Build on the `PathEvent`s to let the application decide which paths to use
- Provide default reasonable behavior into `quiche` while letting the
  application provide optimised behavior

== Requesting the Multipath feature

The application can request multipath through the `Config` structure using a
specific API,`set_multipath()`.
```rust
config.set_multipath(true);
```
The boolean value determines whether the underlying connection negotiates the
multipath extension. Once the handshake succeeded, the application can check
whether the connection has the multipath feature enabled using the
`is_multipath_enabled()` API on the `Connection` structure.

== Path management

The API requires the application to specify on which paths/4-tuples it wants
to send non-probing packets. Paths must first be validated before using them.
This is automatically done for servers, and client must use `probe_path()`.
Once the path is validated, the application decides whether it wants active
usage through the `set_active()` API. It provides a single API entrypoint to
request its usage or not. For active path usage, it can use the following.

```rust
if let Some(PathEvent::Validated(local, peer)) = conn.path_event_next() {
    conn.set_active(local, peer, true).unwrap();
}
```
Then, the path will then be considered to use non-probing packets.

On the other hand, if for some reason the application wants to temporarily
stop sending non-probing frames on a given path, it can do the following.

```rust
conn.set_active(local, peer, false).unwrap();
```
Note that in such state, quiche still replies to PATH_CHALLENGEs observed on
that path.

Finally, the multipath design allows a QUIC endpoint to close/abandon a
given path along with an error code and error message, without altering the
connection's operations as long as another path is available.

```rust
conn.abandon_path(local, peer, 42, "Some error message".into()).unwrap();
```

-- Retrocompatibility note

- The `Closed` variant of `PathEvent` is now a 4-tuple that, in addition to
the local and peer `SocketAddr`, also contains an `u64` error code and a
`Vec<u8>` reason message.
- There is a new variant of `PathEvent`: `PeerPathStatus` reports to the
application that the peer advertised some status for a given 4-tuple.
- There are two new `Error` variants: `UnavailablePath` and
`MultiPathViolation`.

-- Note

Currently this API is only available when multipath feature is enabled over
the session (i.e., `conn.is_multipath_enabled()` returns `true`). If the
extension is not enabled, `set_active()` and `abandon_path()` return an
`Error::InvalidState`. Actually, this API might sound "double usage" along
with the `migrate()` API (as there is no real "connection migration" with
multipath). Should we just keep the `set_active()` or similarly named API
and include the `migrate()` functionality in `set_active()`? Actually, an
client application without the multipath feature could just migrate using
`set_active(local, peer, true)`, setting the previous path in unused mode
under the hood.

== Scheduling sent packets

Similarly to the connection migration PR, there are two ways to control how
`quiche` schedules packets on paths.

The first consists in letting `quiche` handles this by itself. The
application simply uses the `send()` method. In the current master code,
`quiche` automatically handles path validation processing thanks to the
internal `get_send_path_id()` `Connection` method. The multipath patch
extends this method and iterates over all active paths following the lowest
latency path having its congestion window open heuristic (a reasonable
default in multipath protocols).

```rust
loop {
    let (write, send_info) = match conn.send(&mut out) {
            Ok(v) => v,

            Err(quiche::Error::Done) => break,

            Err(e) => {
                conn.close(false, 0x1, b"fail").ok();
                break;
            },
        };
        // write to socket bound to `send_info.from`
}
```

The second option is to let the application choose on which path it wants
to send the next packet. The application can iterate over the available paths
and their corresponding statistics using `path_stats()` and schedules packets
using `send_on_path()` method. This can be useful when the use case requires
some specific scheduling strategies. See `apps/src/client.rs` for an example
of such application-driven scheduling.

### DIFF
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -57,6 +57,7 @@ pub struct CommonArgs {
     pub qpack_blocked_streams: Option<u64>,
     pub initial_cwnd_packets: u64,
     pub multipath: bool,
+    pub multipath_old: bool,
 }
 
 /// Creates a new `CommonArgs` structure using the provided [`Docopt`].
@@ -196,6 +197,7 @@ impl Args for CommonArgs {
             .unwrap();
 
         let multipath = args.get_bool("--multipath");
+        let multipath_old = args.get_bool("--multipath-old");
 
         CommonArgs {
             alpns,
@@ -221,6 +223,7 @@ impl Args for CommonArgs {
             qpack_blocked_streams,
             initial_cwnd_packets,
             multipath,
+            multipath_old,
         }
     }
 }
@@ -251,6 +254,7 @@ impl Default for CommonArgs {
             qpack_blocked_streams: None,
             initial_cwnd_packets: 10,
             multipath: false,
+            multipath_old: false,
         }
     }
 }
@@ -289,6 +293,7 @@ Options:
   --enable-active-migration   Enable active connection migration.
   --perform-migration      Perform connection migration on another source port.
   --multipath              Enable multipath support.
+  --multipath-old          Enable (old v4) multipath support.
   -A --address ADDR ...    Specify addresses to be used instead of the unspecified address. Non-routable addresses will lead to connectivity issues.
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
@@ -484,6 +489,7 @@ Options:
   --disable-pacing            Disable pacing (linux only).
   --initial-cwnd-packets PACKETS      The initial congestion window size in terms of packet count [default: 10].
   --multipath                 Enable multipath support.
+  --multipath-old             Enable (old v4) multipath support.
   -h --help                   Show this screen.
 ";
 
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -24,6 +24,8 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::net::SocketAddr;
+
 use super::common::alpns;
 
 pub trait Args {
@@ -54,6 +56,7 @@ pub struct CommonArgs {
     pub qpack_max_table_capacity: Option<u64>,
     pub qpack_blocked_streams: Option<u64>,
     pub initial_cwnd_packets: u64,
+    pub multipath: bool,
 }
 
 /// Creates a new `CommonArgs` structure using the provided [`Docopt`].
@@ -80,6 +83,7 @@ pub struct CommonArgs {
 /// --qpack-max-table-capacity BYTES  Max capacity of dynamic QPACK decoding.
 /// --qpack-blocked-streams STREAMS  Limit of blocked streams while decoding.
 /// --initial-cwnd-packets      Size of initial congestion window, in packets.
+/// --multipath                 Enable multipath support.
 ///
 /// [`Docopt`]: https://docs.rs/docopt/1.1.0/docopt/
 impl Args for CommonArgs {
@@ -191,6 +195,8 @@ impl Args for CommonArgs {
             .parse::<u64>()
             .unwrap();
 
+        let multipath = args.get_bool("--multipath");
+
         CommonArgs {
             alpns,
             max_data,
@@ -214,6 +220,7 @@ impl Args for CommonArgs {
             qpack_max_table_capacity,
             qpack_blocked_streams,
             initial_cwnd_packets,
+            multipath,
         }
     }
 }
@@ -243,6 +250,7 @@ impl Default for CommonArgs {
             qpack_max_table_capacity: None,
             qpack_blocked_streams: None,
             initial_cwnd_packets: 10,
+            multipath: false,
         }
     }
 }
@@ -280,6 +288,8 @@ Options:
   --max-active-cids NUM    The maximum number of active Connection IDs we can support [default: 2].
   --enable-active-migration   Enable active connection migration.
   --perform-migration      Perform connection migration on another source port.
+  --multipath              Enable multipath support.
+  -A --address ADDR ...    Additional client addresses to use.
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --send-priority-update   Send HTTP/3 priority updates if the query string params 'u' or 'i' are present in URLs
@@ -309,6 +319,7 @@ pub struct ClientArgs {
     pub source_port: u16,
     pub perform_migration: bool,
     pub send_priority_update: bool,
+    pub addrs: Vec<SocketAddr>,
 }
 
 impl Args for ClientArgs {
@@ -386,6 +397,12 @@ impl Args for ClientArgs {
 
         let send_priority_update = args.get_bool("--send-priority-update");
 
+        let addrs = args
+            .get_vec("--address")
+            .into_iter()
+            .filter_map(|a| a.parse().ok())
+            .collect();
+
         ClientArgs {
             version,
             dump_response_path,
@@ -402,6 +419,7 @@ impl Args for ClientArgs {
             source_port,
             perform_migration,
             send_priority_update,
+            addrs,
         }
     }
 }
@@ -424,6 +442,7 @@ impl Default for ClientArgs {
             source_port: 0,
             perform_migration: false,
             send_priority_update: false,
+            addrs: vec![],
         }
     }
 }
@@ -464,6 +483,7 @@ Options:
   --disable-gso               Disable GSO (linux only).
   --disable-pacing            Disable pacing (linux only).
   --initial-cwnd-packets PACKETS      The initial congestion window size in terms of packet count [default: 10].
+  --multipath                 Enable multipath support.
   -h --help                   Show this screen.
 ";
 
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -289,7 +289,7 @@ Options:
   --enable-active-migration   Enable active connection migration.
   --perform-migration      Perform connection migration on another source port.
   --multipath              Enable multipath support.
-  -A --address ADDR ...    Additional client addresses to use.
+  -A --address ADDR ...    Specify addresses to be used instead of the unspecified address. Non-routable addresses will lead to connectivity issues.
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --send-priority-update   Send HTTP/3 priority updates if the query string params 'u' or 'i' are present in URLs
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -25,6 +25,7 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use std::net::SocketAddr;
+use std::str::FromStr;
 
 use super::common::alpns;
 
@@ -295,6 +296,8 @@ Options:
   --multipath              Enable multipath support.
   --multipath-old          Enable (old v4) multipath support.
   -A --address ADDR ...    Specify addresses to be used instead of the unspecified address. Non-routable addresses will lead to connectivity issues.
+  -R --rm-addr TIMEADDR ...   Specify addresses to stop using after the provided time (format time,addr).
+  -S --status TIMEADDRSTAT ...   Specify status to advertise to the peer after the provided time (format time,addr,status).
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --send-priority-update   Send HTTP/3 priority updates if the query string params 'u' or 'i' are present in URLs
@@ -325,6 +328,8 @@ pub struct ClientArgs {
     pub perform_migration: bool,
     pub send_priority_update: bool,
     pub addrs: Vec<SocketAddr>,
+    pub rm_addrs: Vec<(std::time::Duration, SocketAddr)>,
+    pub status: Vec<(std::time::Duration, SocketAddr, u64)>,
 }
 
 impl Args for ClientArgs {
@@ -408,6 +413,50 @@ impl Args for ClientArgs {
             .filter_map(|a| a.parse().ok())
             .collect();
 
+        let rm_addrs = args
+            .get_vec("--rm-addr")
+            .into_iter()
+            .filter_map(|ta| {
+                let s = ta.split(',').collect::<Vec<_>>();
+                if s.len() != 2 {
+                    return None;
+                }
+                let secs = match s[0].parse::<u64>() {
+                    Ok(s) => s,
+                    Err(_) => return None,
+                };
+                let addr = match SocketAddr::from_str(s[1]) {
+                    Ok(a) => a,
+                    Err(_) => return None,
+                };
+                Some((std::time::Duration::from_secs(secs), addr))
+            })
+            .collect();
+
+        let status = args
+            .get_vec("--status")
+            .into_iter()
+            .filter_map(|ta| {
+                let s = ta.split(',').collect::<Vec<_>>();
+                if s.len() != 3 {
+                    return None;
+                }
+                let secs = match s[0].parse::<u64>() {
+                    Ok(s) => s,
+                    Err(_) => return None,
+                };
+                let addr = match SocketAddr::from_str(s[1]) {
+                    Ok(a) => a,
+                    Err(_) => return None,
+                };
+                let status = match s[2].parse::<u64>() {
+                    Ok(s) => s,
+                    Err(_) => return None,
+                };
+                Some((std::time::Duration::from_secs(secs), addr, status))
+            })
+            .collect();
+
         ClientArgs {
             version,
             dump_response_path,
@@ -425,6 +474,8 @@ impl Args for ClientArgs {
             perform_migration,
             send_priority_update,
             addrs,
+            rm_addrs,
+            status,
         }
     }
 }
@@ -448,6 +499,8 @@ impl Default for ClientArgs {
             perform_migration: false,
             send_priority_update: false,
             addrs: vec![],
+            rm_addrs: vec![],
+            status: vec![],
         }
     }
 }
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -58,7 +58,6 @@ pub struct CommonArgs {
     pub qpack_blocked_streams: Option<u64>,
     pub initial_cwnd_packets: u64,
     pub multipath: bool,
-    pub multipath_old: bool,
 }
 
 /// Creates a new `CommonArgs` structure using the provided [`Docopt`].
@@ -198,7 +197,6 @@ impl Args for CommonArgs {
             .unwrap();
 
         let multipath = args.get_bool("--multipath");
-        let multipath_old = args.get_bool("--multipath-old");
 
         CommonArgs {
             alpns,
@@ -224,7 +222,6 @@ impl Args for CommonArgs {
             qpack_blocked_streams,
             initial_cwnd_packets,
             multipath,
-            multipath_old,
         }
     }
 }
@@ -255,7 +252,6 @@ impl Default for CommonArgs {
             qpack_blocked_streams: None,
             initial_cwnd_packets: 10,
             multipath: false,
-            multipath_old: false,
         }
     }
 }
@@ -294,7 +290,6 @@ Options:
   --enable-active-migration   Enable active connection migration.
   --perform-migration      Perform connection migration on another source port.
   --multipath              Enable multipath support.
-  --multipath-old          Enable (old v4) multipath support.
   -A --address ADDR ...    Specify addresses to be used instead of the unspecified address. Non-routable addresses will lead to connectivity issues.
   -R --rm-addr TIMEADDR ...   Specify addresses to stop using after the provided time (format time,addr).
   -S --status TIMEADDRSTAT ...   Specify availability status to advertise to the peer after the provided time (format time,addr,available).
@@ -543,7 +538,6 @@ Options:
   --disable-pacing            Disable pacing (linux only).
   --initial-cwnd-packets PACKETS      The initial congestion window size in terms of packet count [default: 10].
   --multipath                 Enable multipath support.
-  --multipath-old             Enable (old v4) multipath support.
   -h --help                   Show this screen.
 ";
 
--- a/apps/src/args.rs
+++ b/apps/src/args.rs
@@ -297,7 +297,7 @@ Options:
   --multipath-old          Enable (old v4) multipath support.
   -A --address ADDR ...    Specify addresses to be used instead of the unspecified address. Non-routable addresses will lead to connectivity issues.
   -R --rm-addr TIMEADDR ...   Specify addresses to stop using after the provided time (format time,addr).
-  -S --status TIMEADDRSTAT ...   Specify status to advertise to the peer after the provided time (format time,addr,status).
+  -S --status TIMEADDRSTAT ...   Specify availability status to advertise to the peer after the provided time (format time,addr,available).
   -H --header HEADER ...   Add a request header.
   -n --requests REQUESTS   Send the given number of identical requests [default: 1].
   --send-priority-update   Send HTTP/3 priority updates if the query string params 'u' or 'i' are present in URLs
@@ -329,7 +329,7 @@ pub struct ClientArgs {
     pub send_priority_update: bool,
     pub addrs: Vec<SocketAddr>,
     pub rm_addrs: Vec<(std::time::Duration, SocketAddr)>,
-    pub status: Vec<(std::time::Duration, SocketAddr, u64)>,
+    pub status: Vec<(std::time::Duration, SocketAddr, bool)>,
 }
 
 impl Args for ClientArgs {
@@ -450,7 +450,8 @@ impl Args for ClientArgs {
                     Err(_) => return None,
                 };
                 let status = match s[2].parse::<u64>() {
-                    Ok(s) => s,
+                    Ok(0) => false,
+                    Ok(_) => true,
                     Err(_) => return None,
                 };
                 Some((std::time::Duration::from_secs(secs), addr, status))
--- a/apps/src/client.rs
+++ b/apps/src/client.rs
@@ -27,6 +27,7 @@
 use crate::args::*;
 use crate::common::*;
 
+use std::collections::HashMap;
 use std::net::ToSocketAddrs;
 
 use std::io::prelude::*;
@@ -36,6 +37,8 @@ use std::rc::Rc;
 use std::cell::RefCell;
 
 use ring::rand::*;
+
+use slab::Slab;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
 
@@ -52,6 +55,9 @@ pub fn connect(
 ) -> Result<(), ClientError> {
     let mut buf = [0; 65535];
     let mut out = [0; MAX_DATAGRAM_SIZE];
+
+    let mut sockets = Slab::with_capacity(std::cmp::max(args.addrs.len(), 1));
+    let mut src_addrs = HashMap::new();
 
     let output_sink =
         Rc::new(RefCell::new(output_sink)) as Rc<RefCell<dyn FnMut(_)>>;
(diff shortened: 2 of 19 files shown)


qlog/src/events/quic.rs
```diff
@@ -494,6 +497,33 @@ pub enum QuicFrame {
...
+
+        ect1: Option<u64>,
+
+        ect0: Option<u64>,
+
+        ce: Option<u64>,
+    },
+
+    PathAbandon {
+        identifier_type: u8,
```
LPardue, line 514: in the absence of a qlog spec, I'd expect `identifier_type` to be u64. That makes them consistent with varint type in the frame definition.

qlog/src/events/quic.rs
```diff
@@ -494,6 +497,33 @@ pub enum QuicFrame {
...
+        path_identifier: Option<u64>,
+        error_code: u64,
+        reason: Option<String>,
+    },
+
+    PathStatus {
+        identifier_type: u8,
```
LPardue, line 521: in the absence of a qlog spec, I'd expect `identifier_type` to be u64. That makes it consistent with varint type in the frame definition.

hendrikcech:
Thanks for your work on implementing MPQUIC! I played around with this code and have the suspicion that some packets are not sent on the path that they should be sent on.

My setup: I created a mininet topology to test multipath support with the quiche-server and quiche-client applications.

```
                Path 1
         /--- s1 --- s2 ---/
     10.0.1.1           10.0.1.2
 Client h1          Server h2
     10.0.2.1           10.0.2.2 
         \--- s3 --- s4 ---/
                Path 2
```

I added `space_id` and `path_id` to the qlog `PacketHeader` that is used in the `send_single` and `recv_single` functions. From my understanding, `space_id` refers to the packet number space of the sent/received packet while `path_id` refers to the network path that the packet will be sent over / was received from. Since we have one packet number space per path, these two values should be equal.

My tests showed that, on the server, `path_id` always equals `space_id` (values are either 0 or 1). On the client, `space_id` and `path_id` are also equal in all `transport:packet_sent` events.

This is however not true for `transport:packet_received` events on the client: here, `space_id` != `path_id` in about 3% of cases (e.g., 1300 of 40000 received packets). I was not yet able to confirm if packets are actually sent on a different network path than `space_id` indicates in those cases (Wireshark fails to decode the QUIC packets).

The quiche logs hint that packets, that are assigned to the packet number space of one path, are sometimes sent on the other path:
```
[...]
[2022-11-16T14:19:49.964226504Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 path ID 0 now see SCID with seq num 1
[2022-11-16T14:19:49.974246915Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 path ID 0 now see SCID with seq num 0
[2022-11-16T14:19:49.985896148Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 path ID 0 now see SCID with seq num 1
[...]
```

Is this behavior expected or is there actually something going wrong?

Attachments:
- [0001-Add-path_id-and-space_id-to-qlog.patch.txt](https://github.com/cloudflare/quiche/files/10024211/0001-Add-path_id-and-space_id-to-qlog.patch.txt): the qlog code additions to log `path_id` and `space_id`. I used the latest commit (`d6772280`) from this PR.
- [logs.zip](https://github.com/cloudflare/quiche/files/10024256/logs.zip): qlog and quiche logs of a multipath transmission

qdeconinck:
@hendrikcech Nice to see you are experimenting with the code :)

It seems that `quiche` behaves correctly, but the server code does not send the packet on the right path. From the client logs,
```
[2022-11-16T14:19:43.721770238Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx pkt Short dcid=18f4cf5ad9df2fb7f8cb280b8ea06fafa2e45d3c key_phase=false len=1329 pn=0 src:10.0.1.2:4433 dst:10.0.2.1:8002
[2022-11-16T14:19:43.721878453Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 path ID 1 now see SCID with seq num 1
[2022-11-16T14:19:43.721925418Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx frm PATH_RESPONSE data=[02, e6, 51, c4, 58, 32, 39, 45]
[2022-11-16T14:19:43.721993077Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx frm PATH_CHALLENGE data=[26, 6b, ab, b8, 54, 64, 49, 0b]
[2022-11-16T14:19:43.722450331Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx frm PADDING len=1294
[2022-11-16T14:19:43.722732114Z TRACE quiche_apps::client] 10.0.2.1:8002: processed 1350 bytes
[2022-11-16T14:19:43.722774926Z TRACE quiche_apps::client] 10.0.2.1:8002: got 49 bytes
[2022-11-16T14:19:43.722840885Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx pkt Short dcid=14d656e811af235b787dc1ba7bc22a72524c0d80 key_phase=false len=28 pn=1 src:10.0.1.2:4433 dst:10.0.2.1:8002
[2022-11-16T14:19:43.722927964Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 peer reused CID 14d656e811af235b787dc1ba7bc22a72524c0d80 from path 0 on path 1
[2022-11-16T14:19:43.722971351Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 path ID 1 now see SCID with seq num 0
[2022-11-16T14:19:43.723004292Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx frm DATA_BLOCKED limit=10000000
[2022-11-16T14:19:43.723037715Z TRACE quiche] 14d656e811af235b787dc1ba7bc22a72524c0d80 rx frm STREAM id=7 off=0 len=1 fin=false
```

But at server side:
```
[2022-11-16T14:19:43.668765632Z INFO  quiche_server] 327765e95cde0abdb523c6de9e6b2d292c0ced18 Seen new path (10.0.1.2:4433, 10.0.2.1:8002)
[2022-11-16T14:19:43.668816125Z TRACE quiche_server] recv() would block
[2022-11-16T14:19:43.669135491Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx pkt Short dcid=18f4cf5ad9df2fb7f8cb280b8ea06fafa2e45d3c key_phase=false len=1312 pn=0 src:10.0.1.2:4433 dst:10.0.2.1:8002
[2022-11-16T14:19:43.669183980Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx frm PATH_RESPONSE data=[02, e6, 51, c4, 58, 32, 39, 45]
[2022-11-16T14:19:43.669228104Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx frm PATH_CHALLENGE data=[26, 6b, ab, b8, 54, 64, 49, 0b]
[2022-11-16T14:19:43.669267438Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx frm PADDING len=1294
[2022-11-16T14:19:43.669409946Z TRACE quiche::recovery] 327765e95cde0abdb523c6de9e6b2d292c0ced18 timer=1.023455331s latest_rtt=0ns srtt=None min_rtt=0ns rttvar=166.5ms loss_time=[None, None, None] loss_probes=[0, 0, 0] cwnd=13500 ssthresh=18446744073709551615 bytes_in_flight=1350 app_limited=true congestion_recovery_start_time=None Rate { delivered: 0, delivered_time: Instant { tv_sec: 228741, tv_nsec: 941993262 }, first_sent_time: Instant { tv_sec: 228741, tv_nsec: 941993262 }, end_of_app_limited: SpacedPktNum(0, 0), last_sent_packet: SpacedPktNum(1, 0), largest_acked: SpacedPktNum(0, 0), rate_sample: RateSample { delivery_rate: 0, is_app_limited: false, interval: 0ns, delivered: 0, prior_delivered: 0, prior_time: None, send_elapsed: 0ns, ack_elapsed: 0ns, rtt: 0ns } } pacer=Pacer { enabled: true, capacity: 13500, used: 0, rate: 0, last_update: Instant { tv_sec: 228741, tv_nsec: 941993262 }, next_time: Instant { tv_sec: 228741, tv_nsec: 941993262 }, max_datagram_size: 1350, last_packet_size: None, iv: 0ns } hystart=window_end=Some(SpacedPktNum(1, 0)) last_round_min_rtt=18446744073709551615.999999999s current_round_min_rtt=18446744073709551615.999999999s css_baseline_min_rtt=18446744073709551615.999999999s rtt_sample_count=0 css_start_time=None css_round_count=0 cubic={ k=0 w_max=0 } 
[2022-11-16T14:19:43.669725202Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx pkt Short dcid=14d656e811af235b787dc1ba7bc22a72524c0d80 key_phase=false len=11 pn=1 src:10.0.1.2:4433 dst:10.0.1.1:8001
[2022-11-16T14:19:43.669773886Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx frm DATA_BLOCKED limit=10000000
[2022-11-16T14:19:43.669803595Z TRACE quiche] 327765e95cde0abdb523c6de9e6b2d292c0ced18 tx frm STREAM id=7 off=0 len=1 fin=false
[2022-11-16T14:19:43.669936399Z TRACE quiche::recovery] 327765e95cde0abdb523c6de9e6b2d292c0ced18 timer=218.846885ms latest_rtt=104.273399ms srtt=Some(104.518737ms) min_rtt=104.273399ms rttvar=22.416993ms loss_time=[None, None, None] loss_probes=[0, 0, 0] cwnd=13500 ssthresh=18446744073709551615 bytes_in_flight=49 app_limited=true congestion_recovery_start_time=None Rate { delivered: 2194, delivered_time: Instant { tv_sec: 228741, tv_nsec: 936736141 }, first_sent_time: Instant { tv_sec: 228741, tv_nsec: 936736141 }, end_of_app_limited: SpacedPktNum(0, 1), last_sent_packet: SpacedPktNum(0, 1), largest_acked: SpacedPktNum(0, 1), rate_sample: RateSample { delivery_rate: 4958, is_app_limited: true, interval: 104.273399ms, delivered: 517, prior_delivered: 1677, prior_time: Some(Instant { tv_sec: 228741, tv_nsec: 832462742 }), send_elapsed: 0ns, ack_elapsed: 104.273399ms, rtt: 104.273399ms } } pacer=Pacer { enabled: true, capacity: 13500, used: 0, rate: 161454, last_update: Instant { tv_sec: 228741, tv_nsec: 936736141 }, next_time: Instant { tv_sec: 228741, tv_nsec: 936736141 }, max_datagram_size: 1350, last_packet_size: Some(0), iv: 0ns } hystart=window_end=Some(SpacedPktNum(0, 0)) last_round_min_rtt=18446744073709551615.999999999s current_round_min_rtt=18446744073709551615.999999999s css_baseline_min_rtt=18446744073709551615.999999999s rtt_sample_count=0 css_start_time=None css_round_count=0 cubic={ k=0 w_max=0 } 
[2022-11-16T14:19:43.670211822Z TRACE quiche_server] 327765e95cde0abdb523c6de9e6b2d292c0ced18 written 1399 bytes
```

So `quiche` at the server side indeed indicates the second packet should be sent on the original path `src:10.0.1.2:4433 dst:10.0.1.1:8001`, but for some reason the client side notices that the "`STREAM`" packet is received on the same path than the `PATH_CHALLENGE` one.

It actually seems that `quiche-server` batches the `conn.send()` into a large buffer and calls `send_to` once the buffer is full (or if there is no additional packet to be sent), but forgets to check whether packets should be sent on different 4-tuples (only the 4-tuple of the first packet in the batch is considered). I can quickly look at fixing this behaviour of `quiche-server` in a commit.

hendrikcech:
@qdeconinck Thanks for taking a look! I can confirm that these changes to quiche-server resolve the problem.

apps/src/args.rs
```diff
@@ -273,6 +281,8 @@ Options:
   --max-active-cids NUM    The maximum number of active Connection IDs we can support [default: 2].
   --enable-active-migration   Enable active connection migration.
   --perform-migration      Perform connection migration on another source port.
+  --multipath              Enable multipath support.
+  -A --address ADDR ...    Additional client addresses to use.
```
LPardue, line 285: This description is kind of confusing. 

Prior to this change, the client would look at the Server IP and pick an "any" socket using an IP family that matched the server's i.e.  0.0.0.0:0 or [::]:0 depending on v4 or v6 respectively

With this change, there aren't _additional_ address that are used on top of the old behaviour, but instead they are just the addresses that would be used. This opens a few new failure scenarios that could catch users out. For example, if a server only returns an A record and the user provided a v6 client address, then the connection would fail. 


Among the most confusing type of failure is where packets go to a black hole and the connection fails after a timeout.

We might want to tweak how the socket code handling for this option works, and make the description a bit clearer about what it does and how it might fail.

apps/src/client.rs
```diff
@@ -70,32 +76,46 @@ pub fn connect(
...
-    let mut socket =
-        mio::net::UdpSocket::bind(bind_addr.parse().unwrap()).unwrap();
-    poll.registry()
-        .register(&mut socket, mio::Token(0), mio::Interest::READABLE)
-        .unwrap();
-
-    let migrate_socket = if args.perform_migration {
-        let mut socket =
-            mio::net::UdpSocket::bind(bind_addr.parse().unwrap()).unwrap();
+    let mut addrs = Vec::new();
```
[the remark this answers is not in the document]
LPardue: Also, per my [comment about the args](https://github.com/cloudflare/quiche/pull/1310/files#r1084275282), it might help to think of the possible failure scenarios with user provided addresses and add some sanity checking to report problems rather then it just letting it fail as a timeout etc

apps/src/client.rs
```diff
@@ -601,3 +617,27 @@ pub fn connect(
...
+    rng: &T,
+) -> (quiche::ConnectionId<'static>, u128) {
+    let mut scid = [0; quiche::MAX_CONN_ID_LEN];
+    rng.fill(&mut scid).unwrap();
+    let scid = scid.to_vec().into();
+    let mut reset_token = [0; 16];
+    rng.fill(&mut reset_token).unwrap();
+    let reset_token = u128::from_be_bytes(reset_token);
+    (scid, reset_token)
+}
```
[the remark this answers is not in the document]
qdeconinck: Oh, this seems to be code I forgot to clean up, well spotted!

apps/src/args.rs
qdeconinck: You're right, the current description does not suit the actual behavior of this option. As a first step, the initial description could be rewritten as "Specify source addresses to be used instead of the unspecified address" (or "instead of letting the OS decides",...).

For the family address mismatch that may arise, I can indeed rework the code to remove all the specified addresses that do not match the family of the server one. In the case none is remaining, the code can hence fallback to the original behavior, i.e., use "0.0.0.0:0" or "[::]:0".

There still remains the issue that non-routable addresses may be provided, e.g., `fe80::/16` or other. I'm not sure we can do much here, so maybe update the description as "Specify source addresses to be used instead of the unspecified address, non-routable addresses will lead to connectivity issues" (maybe a bit long, but not sure how to make it shorter without loosing information).

toshiiw:
I made a simple client/server example and tested with this multipath patch. I noticed there might be a flaw in the ACKMP sending logic.

My code is based on the one in quiche/apps. The client opens 2 paths against the server and writes data on a single QUIC stream. It turns out that the server sometimes stops sending ACKMPs which causes a client-side idle timeout and subsequent termination of the connection. This problem can be mitigated if the send() call in the server event loop is replaced with a send_on_path() call that iterates over all available paths, just like the client example (in the multipath branch) does.

The send_single() function sends an ACKMP only if there is an unacked packet receive don the same path as specified in the function argument, while the send() function just selects a "best" path and no other paths are tried even if the selected path doesn't yield a packet data.
I think the problem lies here.

I noticed the client is retransmitting STREAM packets before a timeout but it doesn't seem to help. I haven't checked on which path those retransmits are happening.

apps/src/client.rs
```diff
@@ -485,80 +469,75 @@ pub fn connect(
             scid_sent = true;
         }
 
-        if args.perform_migration &&
+        if conn_args.multipath &&
```
FrancoLiberali, line 472: When the number of addresses to use is greater than conn.available_dcids() some addresses are ignored without giving any information. I think a warning could be added to show that this happens and it can be solved with the --max-active-cids parameter

qdeconinck:
@toshiiw This is strange, as the `get_send_path_id()` method should address this point. Could you indicate on which commit you are based, and provide logs describing the behavior? Feel free to contact me offline if preferred.

apps/src/client.rs
qdeconinck: It might be indeed interesting to add a warning message indicating so, good point!

apps/src/client.rs
```diff
@@ -116,6 +101,7 @@ pub fn connect(
     config.set_initial_max_streams_uni(conn_args.max_streams_uni);
     config.set_disable_active_migration(!conn_args.enable_active_migration);
     config.set_active_connection_id_limit(conn_args.max_active_cids);
```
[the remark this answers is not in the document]
qdeconinck: I'm not sure if we want the client to include such "magic" without proper documentation, but I can wait for other opinions. Also, the comparison should be made against `addrs.len()` and not `args.addrs.len()`, as some provided addresses might be of different families than the contacted server address.

toshiiw:
I tested with the following version.
Attached are sqlog files with packet namespace ids.
Data is sent from a client to a server (unidirectional).

```
commit d67722801b043a9b82c04fa70bf1e3240492ee23 (HEAD -> multipath)
Author: Quentin De Coninck <quentin_d@apple.com>
Date:   Mon Oct 10 17:29:15 2022 +0200
```
[sqlog.tar.gz](https://github.com/cloudflare/quiche/files/10671935/sqlog.tar.gz)

I also tested with the newest code on the multipath branch but still saw premature shutdowns.

quiche/src/lib.rs
```diff
@@ -6984,6 +7547,63 @@ impl Connection {
             }
         }
 
+        let mut consider_standby = false;
+        let dgrams_to_emit = self.dgram_max_writable_len().is_some();
```
toshiiw, line 7551: IIUC this always returns true even if self.dgram_send_queue is empty.

quiche/src/lib.rs
```diff
@@ -6984,6 +7547,63 @@ impl Connection {
...
+                    break;
+                }
+                consider_standby = true;
+            }
+        }
+
+        // When using multiple packet number spaces, let's force ACK_MP sending
+        // on their corresponding paths.
+        if self.is_multipath_enabled() {
+            if let Some(pid) =
```
toshiiw, line 7585: ...so, this code isn't executed.

qdeconinck:
@toshiiw Oooh good catch indeed! If you configure your connection to enable datagrams on the connection, I can indeed reproduce the issue! I will push a fix with the adapted multipath test now; replacing `let dgrams_to_emit = self.dgram_max_writable_len().is_some();` with `let dgrams_to_emit = self.dgram_send_queue.has_pending();` indeed solves the issue.

quiche/src/path.rs
```diff
@@ -57,15 +57,83 @@ pub enum PathState {
...
+            PathRequest::Abandon(e, r) => PathState::Closing(e, r),
+        }
+    }
+}
+
+/// The status of a path, advertised through the PATH_STATUS frame.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PathStatus {
+    /// The host should stop sending non-probing packets on the path.
+    Standby,
```
ehaydenr, line 112: Should `PathStatus` we receive from the peer influence usability of paths? AFAICT the `usable()` method only takes into account `PathState`, which isn't updated when `PathStatus` is received.

quiche/src/lib.rs
```diff
@@ -8029,21 +8822,37 @@ pub mod testing {
...
-            conn.pkt_num_spaces[epoch].largest_rx_pkt_num,
+            conn.pkt_num_spaces.spaces.get(epoch, 0)?.largest_rx_pkt_num,
             hdr.pkt_num,
             hdr.pkt_num_len,
         );
 
-        let mut payload =
-            packet::decrypt_pkt(&mut b, pn, hdr.pkt_num_len, payload_len, aead)
-                .unwrap();
+        // We don't support multipath in this method.
```
ehaydenr, line 8843: Not supporting this is pretty unfortunate. I think this would be useful for tests. What would it take to support with with multipath?

toshiiw:
I tested how this multpath patch performs under partial network failures.

A test client opens 2 paths against the server and continuously send data over a QUIC stream  (the same test programs as 3 weeks ago).
Then the traffic of one of the paths is disabled by issuing `iptables -D INPUT -p udp -s 127.0.0.1 --sport 30002 -j DROP`.
Then, sooner or later, the client picks up the disabled path to send STREAM resends or pings, which will never reach the server, and eventually the QUIC connection is terminated with an idle timer.

I guess multipath should be able to handle such situation.
A few observations and random ideas:

- RTT based path selection might not work well with packet losses, as RTT is only updated upon successful transmit
- Is it a good idea to mark a path as stale after a few packet losses?
- Is there a workaround in a quiche application code? I guess counting packet losses in applications is cumbersome.

quiche/src/lib.rs
ehaydenr: To be more specific, I'd like to use this for implementing a test case for stream/path affinity: https://github.com/ehaydenr/quiche/commit/9f71ed79eab4056e213418ff86ca2450c96dc038#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R16631

qdeconinck: Oh right, I did not implement that before since I did not need that, but actually providing Multipath support in `decode_pkt()` is not very hard (just pushed a7b1ff8dcdf3d182dd51ffa354b8c9cf5fd8878a to fix that).

quiche/src/path.rs
qdeconinck: Taking into account the `PathStatus` into account is somewhat a scheduling decision that should be driven by the application. The default `get_send_path_id()` (called when using `send()` instead of `send_on_path()`) takes the `PathStatus` into account (it prefers all `Available` paths when possible, and only consider `Standby` ones when there only remains `Standby` paths).

A validated `Standby` path is by itself usable, there is some policy stating that "if it is possible, do not use that path".  Directly taking the `PathStatus` into account in the `usable()` method is probably not desirable, especially if all the remaining paths are `Standby` ones.

quiche/src/lib.rs
ehaydenr: Hm I tried using `decrypt_pkt`, but it panics on `find_scid_seq` being `None`. https://github.com/ehaydenr/quiche/commit/d7da76eccdc54bea8ce588d201f140d34752d6f9 🤔

qdeconinck: I think the issue is elsewhere, as with some minor changes I make the test passes :-) See my comments on your commit.

toshiiw:
> I tested how this multpath patch performs under partial network failures.
>     * Is there a workaround in a quiche application code? I guess counting packet losses in applications is cumbersome.

I noticed `pto_count` could be used as a loss counter, and the attached patch enables a multipath connection to recover from a failed path.
But the recovery is very slow (taking ~10 secs) and I'd love to see a better way to handle such a situation.

My understanding is that `mark_all_inflight_as_lost` must be called on a failed path to ensure that retransmits happen on a good path but closing a failed path is somewhat a slow operation.

[stale-patch.txt](https://github.com/cloudflare/quiche/files/10918169/stale-patch.txt)

apps/src/client.rs
```diff
@@ -600,3 +591,75 @@ pub fn connect(
...
+
+/// Generate a ordered list of 4-tuples on which the host should send packets,
+/// following a lowest-latency scheduling.
+fn lowest_latency_scheduler(
+    conn: &quiche::Connection,
+) -> impl Iterator<Item = (std::net::SocketAddr, std::net::SocketAddr)> {
+    use itertools::Itertools;
+    conn.path_stats()
+        .sorted_by_key(|p| p.rtt)
+        .map(|p| (p.local_addr, p.peer_addr))
```
toshiiw, line 664: When there is a closed path, I'd like to add something like `filter(|p| p.active)` here, but that will break probing of new paths.
It seems not enough information is marked as public in order to implement packet scheduler outside of the quiche library.

qdeconinck: Yeah, this is likely that the current content of `PathStats` might not be sufficient to address all the possible scheduling decisions. Might be useful to list all the fields of interest, but not sure if this PR will have all of them.

toshiiw: I guess I will at least need the following things, but I can locally patch the code for now and am not in a hurry.
The list would change as my project evolves, anyway. And I'm not sure `PathStats` is the right interface for all these things.

- a list of probing paths
- a list of ack eliciting paths
- "instantaneous" loss counter such as pto_count (for scheduling and path sanity checks, see the previous comment)

qdeconinck: @toshiiw I think if we want to stay in the current API philosophy of quiche, `PathStats` is fine.

Regarding the list of probing paths, this is already accessible since `PathStats` exposes the `PathValidationState`, so paths can be filtered through that field.

For the remaining points, it is still possible to extend the `PathStats` structure to also give access to, e.g., the `pto_count` (we already give access to some `Recovery` fields, so adding one is quite easy). For the ack eliciting, are you interested by paths requiring sending ack eliciting packets, or by paths having sent ack eliciting packets?

qdeconinck:
@masa-koz You are right on the fact that if a path experiences a PTO, it should send some packets there -- at least, to probe the path. Still, I'm not sure that the previously carried frames/data should always be retransmitted on the same path (it probably makes more sense to use an available "good" path instead).

masa-koz:
@qdeconinck The previously carried frames/data have not been declared as lost yet. Those frames are treated as lost in Recovery::on_loss_detection_timeout() for sending them in the probe packets. So, I think that those frame should be retransmitted on the same path.
I agree with you about the fact that it makes more sense to retransmit the frames on another "good" path. But the frames should be those which were included in the actually lost packets or be those which are marked as retransmitting before lost for some reason (for example, due to the request from the application), I think.

toshiiw:
AFAIK the multipath patch only has PTO to handle sudden 100% loss of a path, which doesn't work well because no newer packets are acked and usual loss detection never happens.

While I expect speedy recovery with multipath, there is a tradeoff with network usage etc. I'm happy to see duplicate retransmissions over available paths but other people might not.

It seems I've got this reliability issue (the former) and the availability issue (the latter) mixed up. I'll think again.

apps/src/client.rs
toshiiw: Or, `send_on_path` can have a `hint` bool argument (or another function wrapper) to notify if the function is allowed to send on a different path than specified.

ghedo:
Hey @qdeconinck, thanks for your continued work on this.

We've been working on testing and integrating this into our stack, so I've now started looking into rebasing this (there's a few more conflicts due to the recent keyupdate change), as well as doing some clean-up and starting to merge some of the changes (e.g. the initial refactor commit) to make iterating over this a bit easier. If you have some time it would be useful if you could do the rebase, and then split the refactor commit into a separate PR (if not I can do the myself) so we can merge that on its own.

Regarding the refactor specifically, what was the reason for splitting the crypto fields out of `PktNumSpace` and into a completely separate map? Why not just leave them in `PktNumSpace` and use the same map for lookups?

I think we might also need to think through KeyUpdate when multi-path is used.

Also, any thoughts on the changes to the draft that were recently discussed (e.g. path IDs)? Anything we can do right now to make those changes easier later?

qdeconinck:
Hey @ghedo, sorry for the delayed answer, I was away previous week.

Regarding the rebasing and isolation of "pure refactoring changes", I can find some time this week to perform this :-)

About the splitting of `PktNumSpace` into two separate maps, the motivation comes from the recent optimisations trying to minimise the number of paths lookups (like a45f66fd502fa94e740db57a150a921e73f6ffb2). With this organization, it is possible to keep the crypto context and the packet number space one into local variables without having Rust compiler complaining about ownership.

Finally, regarding recent discussions, I think it would make sense to have different keys on different paths in the long run. To that end, having this notion of "internal Path ID" instead of directly using CID sequence numbers would make sense (knowing that the CID is linked to that internal path ID by the peer proposing the CID). With the current code structure, I don't think it would introduce much changes (the `path_id` field of `ConnectionIdEntry` would be a `u64` instead of a `Option<usize>`), though I would have to try to implement such a variant to assess whether this is correct.

qdeconinck:
@ghedo Refactoring changes are isolated in #1493.

vanyingenzi:
Hello everyone,

I'm new to QUIC, and I'm commencing my master's thesis entitled "Enhancing the Performance of a Single QUIC Connection with Multi-Path QUIC."

While conducting measurements, I've noticed that the Multi-Path extension is experiencing correctness issues in my setup. I'm using loop-back addresses on a single host. Below is a script that reproduces the issue. The server has one endpoint, and the client has two endpoints.

**Observations**:
- The server validates the paths. However, it sends traffic on both paths, but the transfer ends prematurely with the client.
- The client terminates with a timeout error, where the duration of the timeout equals the time of the transfer.
- The occurrence of this error is non-deterministic, varying from one run to another. However, it *appears* to happen more frequently with larger files. I tested with 1GB and 8GB only.

I haven't delved deeply into the issue because I'm uncertain whether it's due to a misconfiguration on my part or if it's a bug in the actual source code.

Thank you in advance for your time.

```bash
#!/bin/bash

# Code partlty inspired by https://github.com/tumi8/quic-10g-paper

# Variables
QUICHE_REPO="https://github.com/qdeconinck/quiche.git"
QUICHE_COMMIT="d87332018d84fb7c429ad2ed34cbfdc6ee9477c8"
RUST_PLATFORM="x86_64-unknown-linux-gnu"
FILE_SIZE=8G
NB_RUNS=10

RED='\033[0;31m'
RESET='\033[0m'

echo_red() {
    echo -e "${RED}$1${RESET}"
}

get_unused_port(){
    local port
    port=$(shuf -i 2000-65000 -n 1)
    while netstat -atn | grep -q ":$port "; do
        port=$(shuf -i 2000-65000 -n 1)
    done
    echo "$port"
}

clone_mp_quiche() {
    if [ ! -d "./quiche" ]; then
        git clone --recursive "$QUICHE_REPO"
        cd quiche || exit
        git checkout "$QUICHE_COMMIT"
        RUSTFLAGS='-C target-cpu=native' cargo build --release
        cd ..
    fi
    if [ ! -f "./quiche-client" ]; then
        cp "quiche/target/release/quiche-client" .
    fi
    if [ ! -f "./quiche-server" ]; then
        cp "quiche/target/release/quiche-server" .
    fi
}

setup_rust() {
    # Rust
    if ! rustc --version 1>/dev/null 2>&1; then
        curl --proto '=https' --tlsv1.2 -sSf -o /tmp/rustup-init.sh https://sh.rustup.rs
        chmod +x /tmp/rustup-init.sh
        /tmp/rustup-init.sh -q -y --default-host "$RUST_PLATFORM" --default-toolchain stable --profile default
        source "$HOME/.cargo/env"
    else 
        echo "Rust is already installed"
    fi
}

setup_environment() {
    mkdir -p "$(pwd)/www" "$(pwd)/responses" "$(pwd)/logs"
    fallocate -l ${FILE_SIZE} "$(pwd)/www/${FILE_SIZE}B_file"
}

iteration_loop() {
    local server_pid server_port
    for iter in $(seq 1 ${NB_RUNS}); do
        echo "Testing Multi-Path QUIC correctness - Iteration $iter"
        # Run server
        env RUST_LOG=info ./quiche-server \
            --listen 127.0.0.1:19474 \
            --root "$(pwd)/www/" \
            --key "$(pwd)/quiche/apps/src/bin/cert.key" \
            --cert "$(pwd)/quiche/apps/src/bin/cert.crt" \
            --multipath \
            1>"$(pwd)/logs/server_${iter}.log" 2>&1 &
        server_pid=$!

        # Run client
        server_port=$(get_unused_port)
        client_port_1=$(get_unused_port)
        client_port_2=$(get_unused_port)
        env RUST_LOG=info ./quiche-client \
            --no-verify "https://127.0.0.1:19474/${FILE_SIZE}B_file" \
            --dump-responses "$(pwd)/responses/" \
            -A 127.0.0.1:${client_port_1} \
            -A 127.0.0.1:${client_port_2} \
            --multipath \
            1>"$(pwd)/logs/client_${iter}.log" 2>&1
        error_code=$?

        sleep 1
        
        kill -9 "$server_pid" 1>/dev/null 2>&1
        if [ $error_code -ne 0 ]; then
            echo_red "Error Client: $error_code"
            exit 1
        fi

        # Check if files are the same
        diff -q "$(pwd)/www/${FILE_SIZE}B_file" "$(pwd)/responses/${FILE_SIZE}B_file"
        if [ $? -ne 0 ]; then
            echo_red "Error: files are not the same"
            exit 1
        fi
    done
}

main() {
    # Version
    setup_rust
    [ $? -ne 0 ] && { echo_red "Error setting up rust"; exit 1; }
    clone_mp_quiche
    [ $? -ne 0 ] && { echo_red "Error cloning quiche"; exit 1; }
    setup_environment
    [ $? -ne 0 ] && { echo_red "Error setting up environment"; exit 1; }
    iteration_loop
}

main
```

PR #1312, opened 2022-09-06
quic: bound consecutive non ack-eliciting packet sends

ehaydenr:


quiche/src/lib.rs
```diff
@@ -3857,8 +3857,14 @@ impl Connection {
...
         self.emit_dgram = !dgram_emitted;
 
-        // Create PING for PTO probe if no other ack-eliciting frame is sent.
-        if self.paths.get(send_pid)?.recovery.loss_probes[epoch] > 0 &&
+        // Create PING for PTO probe if no other ack-eliciting frame is sent or if
+        // we've sent too many non ACK eliciting packets without having
+        // sent an ACK eliciting one
+        if self
+            .paths
+            .get_mut(send_pid)?
```
LPardue, line 3865: why do we need `get_mut`?

ehaydenr: thank you! - fixed

junhochoi (approved):

quiche/src/lib.rs
```diff
@@ -3857,8 +3857,10 @@ impl Connection {
         // Alternate trying to send DATAGRAMs next time.
         self.emit_dgram = !dgram_emitted;
 
-        // Create PING for PTO probe if no other ack-eliciting frame is sent.
-        if self.paths.get(send_pid)?.recovery.loss_probes[epoch] > 0 &&
+        // Create PING for PTO probe if no other ack-eliciting frame is sent or if
+        // we've sent too many non ACK eliciting packets without having
+        // sent an ACK eliciting one
+        if self.paths.get(send_pid)?.recovery.should_elicit_ack(epoch) &&
```
ghedo, line 3863: https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L3301 should be updated too I think?

Also, we can cache the value in a local `should_elicit_ack` variable or something, so we don't need to put this ultra-long method call in the condition itself (just for readability's sake).

quiche/src/recovery/mod.rs
```diff
@@ -158,6 +162,9 @@ pub struct Recovery {
 
     // BBR state.
     bbr_state: bbr::State,
+
+    /// How many non-ack-eliciting packets have been sent
```
ghedo, line 166:
```suggestion
    /// How many non-ack-eliciting packets have been sent.
```

quiche/src/lib.rs
```diff
@@ -14458,6 +14460,37 @@ mod tests {
...
+            assert!(len > 0);
+
+            let frames =
+                testing::decode_pkt(&mut pipe.client, &mut buf, len).unwrap();
+            if frames
+                .iter()
+                .any(|frame| matches!(frame, frame::Frame::Ping))
+            {
+                // found an explicity eliciting of an ACK
+                return;
```
ghedo, line 14488: Might be a good idea to check that this happens after 24 packets? e.g. can change the `_` in the for loop to a variable, and then do `assert_eq!(i, 24)` or something like that.

ghedo (changes requested):

quiche/src/lib.rs
ghedo: @ehaydenr I think the comment about https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L3301 still needs addressing? Can just move the `should_elicit_ack` variable up and re-use it there too.

ghedo (approved):

PR #1314, opened 2022-09-07
Clarify that token must be 16 bytes long

normanmaurer:
Motivation:

We should clarify that the token must be 16 bytes long as otherwise we might risk a crash

Modifications:

Add some more docs

Result:

Make it clear that the token must be 16 bytes


normanmaurer:
Just some small clarification as it was not clear to me without looking at the code how many bytes the token must have

ghedo (approved):

PR #1315, opened 2022-09-08
Make the condition straightforward

Banyc:


quiche/src/stream.rs
```diff
@@ -352,12 +352,12 @@ impl StreamMap {
             .entry(urgency)
             .or_insert_with(|| (BinaryHeap::new(), VecDeque::new()));
 
-        if !incr {
```
LPardue, line 355: In the priorities spec which this code is based around, incremental == false is the default. 

Since this code is quite internal any way, and the condition is a single bool, I don't see much compelling reason to change.

Banyc: Thanks for the clarification! Really appreciated!

Banyc:
close for incremental == false being the default and the condition being a single bool.

PR #1316, opened 2022-09-08
expose whether or not the dgram send queue is full

ehaydenr:
Useful for knowing when to "wake up" dgram writers. If it goes from full -> !full, we can re-attempt dgram sends.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4861,6 +4861,12 @@ impl Connection {
         self.dgram_send_queue.byte_size()
     }
 
+    /// Returns whether or not the DATAGRAM send queue is full.
+    #[inline]
+    pub fn dgram_send_queue_full(&self) -> bool {
+        self.dgram_send_queue.is_full()
+    }
+
     /// Sends data in a DATAGRAM frame.
     ///
     /// [`Done`] is returned if no data was written.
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -4861,6 +4861,12 @@ impl Connection {
         self.dgram_send_queue.byte_size()
     }
 
+    /// Returns whether or not the DATAGRAM send queue is full.
+    #[inline]
+    pub fn dgram_send_queue_full(&self) -> bool {
```
normanmaurer, line 4866: maybe prefix with `is_` to be more consistent with others ?
```suggestion
    pub fn is_dgram_send_queue_full(&self) -> bool {
```

LPardue:
So technically I think this could always be done if you compared the length of the queue (`dgram_send_queue_len`) to the configured length. No argument from me that `full()` is more ergonomic though. For parity, could you add `is_dgram_recv_queue_full` too?

I think you could also tweak the `dgram_multiple_datagrams` test too to add coverage of these methods. If you drop the queue sizes to match the number of datagrams being exchanged, you can add in a few assert!(full) and assert(!full) checks.

ehaydenr:
> So technically I think this could always be done if you compared the length of the queue (`dgram_send_queue_len`) to the configured length. 

Yep, this is what I've been doing, but since there's no way to query the configured length I've had to create my own variable and carry it around.

LPardue (approved):

PR #1319, opened 2022-09-12
fix SendBuf::len calculation

Banyc:


ghedo:
Can you provide more information on why this is needed please? And ideally a test case too.

Banyc:
Thanks for the reminder! I will complete it afterwork asap!

Banyc:
Previously, after marking `retransmit` to a slice of buffer in `SendBuf::data`, `SendBuf::len` adds up the total transmittable length of the buffer (`buf.len()`). `SendBuf::len` will be incorrect if a part of the buffer has already been marked as transmittable. Say the part is of length `l`. Before marking `retransmit`, `SendBuf::len` has already contain `l`. When marking `retransmit`, `SendBuf::len` will add up additional `l` and there will be two `l`s in `SendBuf::len`.

ghedo:
Thank you for the explanation and the patch! I merged this is dadb22c6bfd30a2c67db15162cb02a96fa1c2c0a with some small tweaks.

PR #1322, opened 2022-09-13
Fix server app close before 1RTT never getting sent

### LINKED ISSUE #1118: Server stall if application close during handshake
I have created a couple test cases that shows what works and what breaks: 05e4618eb33c72f5912e26c0d851ad628452e0b7

From [10.2.3](https://datatracker.ietf.org/doc/html/rfc9000#section-10.2.3), I see that application errors must not be sent during the handshake due to that potentially leaking application information. Although the RFC seems to recommending dropping the application data and sending a 0x1c connection_close, I would prefer to pass that information back to the user.

Stepping through `send_single`, It looks like it stalls because `is_closed` causes no Handshake acks to be sent, and since the `pkt_type` is stays Handshake, an application close will not be sent.

What I expected was for the handshake to complete normally, and have the application close be sent with the first 1RTT short header packet ~along with HANDSHAKE_DONE~.

EDIT:
I implemented a possible change for this behavior: ffdaeed1b35e1810210071be80153187d15b8d2f

kevinjohna6:
If a server does an app close after a connection is established, but before it has actually emitted the Handshake ack, the close will not be sent.
This is because, per spec, app close error codes must not be sent during the Handshake, however the app error prevents the handshake acknowledgement from being emitted.

This patch's test case may give a better example of how this issue may occur.

The resulting behavior of this patch is that the Handshake ack packet will be sent followed by a 1RTT with the connection_close.

Resolves #1118

I have been using this patch for a while without experiencing side effects, but there may be a better way of mitigating this issue.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3310,7 +3310,9 @@ impl Connection {
         // anything ACK eliciting.
         if self.pkt_num_spaces[epoch].recv_pkt_need_ack.len() > 0 &&
             (self.pkt_num_spaces[epoch].ack_elicited || ack_elicit_required) &&
-            !is_closing &&
+            (!is_closing ||
+                (pkt_type == Type::Handshake &&
+                    self.local_error().map_or(false, |le| le.is_app))) &&
             self.paths.get(send_pid)?.active()
         {
             let ack_delay =
@@ -12927,7 +12929,7 @@ mod tests {
     }
 
     #[test]
-    fn app_close() {
+    fn app_close_by_client() {
         let mut buf = [0; 65535];
 
         let mut pipe = testing::Pipe::default().unwrap();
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+    fn app_close_by_server_during_handshake() {
+        let mut pipe = testing::Pipe::default().unwrap();
+
+        // Client sends initial flight.
+        let flight = testing::emit_flight(&mut pipe.client).unwrap();
+        testing::process_flight(&mut pipe.server, flight).unwrap();
+
+        let flight = testing::emit_flight(&mut pipe.server).unwrap();
+
+        // both are unestablished
```
ghedo, line 12886:
```suggestion
        // Both connections are not established.
```

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+        assert!(!pipe.client.is_established() && !pipe.server.is_established());
+
+        testing::process_flight(&mut pipe.client, flight).unwrap();
+
+        // connection established for client
```
ghedo, line 12891:
```suggestion
        // Connection is established on the client.
```

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+        assert!(pipe.client.is_established());
+        // send after established
```
ghedo, line 12893:
```suggestion

        // Client sends after connection is established.
```

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+        pipe.client.stream_send(0, b"badauthtoken", true).unwrap();
+
+        let flight = testing::emit_flight(&mut pipe.client).unwrap();
+        testing::process_flight(&mut pipe.server, flight).unwrap();
+
+        // connection established for server but the Handshake ack has not been
+        // sent yet
```
ghedo, line 12900:
```suggestion
```

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+        assert!(pipe.server.is_established());
+        // close after established
```
ghedo, line 12902:
```suggestion

        // Server closes after connection is established.
```

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,58 @@ mod tests {
...
+        pipe.server
+            .close(true, 123, b"Invalid authentication")
+            .unwrap();
+
+        // Server sends Handshake ack and then 1RTT connection_close
```
ghedo, line 12907:
```suggestion
        // Server sends Handshake ACK and then 1RTT CONNECTION_CLOSE.
```

ghedo (dismissed):
LGTM, added some minor nits.

quiche/src/lib.rs
```diff
@@ -12871,6 +12873,59 @@ mod tests {
...
+        // Connection is established on the client.
+        assert!(pipe.client.is_established());
+
+        // Client sends after connection is established.
+        pipe.client.stream_send(0, b"badauthtoken", true).unwrap();
+
+        let flight = testing::emit_flight(&mut pipe.client).unwrap();
+        testing::process_flight(&mut pipe.server, flight).unwrap();
+
+        // Connection is established on the server but the Handshake ACK is not sent.
```
ghedo, line 12900:
```suggestion
        // Connection is established on the server but the Handshake ACK has not
        // been sent yet.
```

PR #1323, opened 2022-09-13
Create stream on stream_send even when tx_cap is 0

### LINKED ISSUE #1125: stream_send while tx_cap is 0 does not create a new stream
On both 0.10.0 and master calling stream_send with a new stream id while tx_cap is 0 does not internally create a new stream. (behavior introduced [here](https://github.com/cloudflare/quiche/commit/95ea1361794e533bdb75599246f2f92033720621#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R3458-R3469) and modified on master in [#1099](https://github.com/cloudflare/quiche/pull/1099/files#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R3647))
This is an issue in my application because, if a new stream is not created in quiche, `writable()` will never trigger further attempts to send that stream's data.

Moving `get_or_create_stream` before the tx_cap check passes my use case. See 94c9a06eb0b38584870d25c14abddc6eb60a0668 for my test and patch.

kevinjohna6:
Automatically create stream on stream_send even when tx_cap is 0. (It was also possible to do this by simply calling `stream_priority()`)

This allows an application to retry sending data on that stream when it is later marked as writable.

Resolves #1125 

This patch amends existing tests due a typo creating even numbered streams from the server.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4317,24 +4317,27 @@ impl Connection {
             self.blocked_limit = Some(self.max_tx_data);
         }
 
-        // Truncate the input buffer based on the connection's send capacity if
-        // necessary.
-        //
-        // When the cap is zero, the method returns Ok(0) *only* when the passed
-        // buffer is empty. We return Error::Done otherwise.
+        // Copy tx_cap here to use after borrow
         let cap = self.tx_cap;
+
+        // Get existing stream or create a new one.
+        // Even if the cap is zero, new streams are created internally so that
+        // they will eventually be in `writable()` and can be retried later.
+        let stream = self.get_or_create_stream(stream_id, true)?;
+
+        // When the cap is zero, the method returns Ok(0) *only* when fin is true
+        // AND the passed buffer is empty. We return Error::Done otherwise.
         if cap == 0 && !(fin && buf.is_empty()) {
             return Err(Error::Done);
         }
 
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -9849,8 +9852,9 @@ mod tests {
             pipe.server.stream_send(4, b"hello", false),
             Err(Error::Done)
         );
+
         assert_eq!(
-            pipe.server.stream_send(8, b"hello", false),
+            pipe.server.stream_send(9, b"hello", false),
```
LPardue, line 9857: I suspect the use of stream 8 was intentional. The server can't write on a client-initiated stream that hasn't been opened yet and the assertion tested that. Later assertions demonstrate that the server stream writes succeed after the client opened the stream.

kevinjohna6: Makes sense. I changed them to now just expect InvalidStreamState.

kevinjohna6:
I am still interested in including this change upstream, let me know if there is any further action I can take to help get it accepted.

ghedo:
Thanks for the contribution. The underlying bug was subsequently fixed by commit `c9311a189`: a stream is created before returning for exhausted connection-level send capacity and is made writable once capacity becomes available. Current `master` already contains this behavior, so closing this PR as already fixed.

PR #1325, opened 2022-09-16
h3: check and fail 0-length payload frames

LPardue:
HTTP/3 frames have a type, payload length, and frame-specific data.
There are some HTTP/3 fields that always contain fields, and as such,
they will always have a non-zero length payload length.

With this change, we add more robust checks for HTTP/3 frames that
contain fields but have a zero-length payload length. If this is
detected, its a critical error and the connection will be aborted with
the error H3_FRAME_ERROR.


LPardue:
NB: the diff is looks big because I've extracted some repetitive code out into test harness utility functions.

junhochoi (approved):

PR #1327, opened 2022-09-20
Added const quilifiers to some C API functions.

maxsharabayko:
The state-related C API functions do not modify the connection object, and thus can use a constant pointer.


ghedo (changes requested):

ghedo (approved):

ghedo:
Merged, thanks @maxsharabayko!

PR #1330, opened 2022-09-27
qlog: add send_at_time to PacketSent event

LPardue:
This allows us to record the time that the packet pacer would
send this packet at. Since the qlog crate only uses relative
time since log start, the packet send Instant is converted
into a qlog time base and we log the duration since log start.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3977,6 +3977,17 @@ impl Connection {
                 data: None,
             };
 
+            let send_at_time = match self.paths.get(send_pid) {
+                Ok(v) => {
+                    let send_time = v.recovery.get_packet_send_time();
+                    let dur = send_time.duration_since(q.start_time());
+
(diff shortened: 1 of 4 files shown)


quiche/src/lib.rs
```diff
@@ -3974,6 +3974,17 @@ impl Connection {
                 data: None,
             };
 
+            let send_at_time = match self.paths.get(send_pid) {
+                Ok(v) => {
+                    let send_time = v.recovery.get_packet_send_time();
```
junhochoi, line 3979: calling `get_packet_send_time()` here will give a previous packet sent time.
https://github.com/cloudflare/quiche/pull/1330/files#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41L4032 `recovery::on_packet_sent()` will update packet's `time_sent` field with the time to sent.

ghedo (approved):

PR #1332, opened 2022-09-27
lib: distinguish length field from payload length

Banyc:
Previously, at `recv_single`, the length field of the packet is named as `payload_len`. Although the decoding process correctly treats it as a length field, the `trace!` log and qlog are not.

According to `send_single`, `trace!` set `len` as the length of the actual plaintext payload. But `trace!` at `recv_single` set `len` as the length field of the packet.

Also according to `send_single` and qlog definition at <https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-00#section-5.1>, `payload_length` in `RawInfo` should be "the byte length of the entity's payload, without headers or trailers". But at `recv_single`, this `payload_length` in `RawInfo` takes the length field of the packet instead.

After the change, the `trace!` log and qlog are consistent with the decoding process. Also, the variable name is corrected.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2379,11 +2379,7 @@ impl Connection {
         let mut qlog_frames = vec![];
 
         let mut payload = packet::decrypt_pkt(
-            &mut b,
-            pn,
-            pn_len,
-            len_field,
-            aead,
+            &mut b, pn, pn_len, len_field, aead,
         )
         .map_err(|e| {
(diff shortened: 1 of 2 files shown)


Banyc:
Also, no functional code is changed.

Banyc:
I'm not so sure if the length field of a packet equals to the sum of the packet number length and the actual payload length. The change is based on that assumption. I'll close this if the assumption is false.

PR #1334, opened 2022-09-27
Set local_error if handshake fails with SSL_ERROR_SSL

normanmaurer:
Motivation:

We should set the local_error if the handshake fails with SSL_ERROR_SSL as we can't recover from it.

Modifications:

Set local_error when SSL_ERROR_SSL is received

Result:

Automatically close the connection on SSL_ERROR_SSL

### DIFF
--- a/quiche/src/tls.rs
+++ b/quiche/src/tls.rs
@@ -47,6 +47,7 @@ use crate::packet;
 
 const TLS1_3_VERSION: u16 = 0x0304;
 const TLS_ALERT_ERROR: u64 = 0x100;
+const INTERNAL_ERROR: u64 = 0x01;
 
 #[allow(non_camel_case_types)]
 #[repr(transparent)]
@@ -576,6 +577,7 @@ impl Handshake {
         let rc = unsafe { SSL_do_handshake(self.as_mut_ptr()) };
         self.set_ex_data::<Connection>(*QUICHE_EX_DATA_INDEX, std::ptr::null())?;
 
+        self.set_transport_error(ex_data, rc);
         self.map_result_ssl(rc)
     }
 
@@ -591,6 +593,7 @@ impl Handshake {
         let rc = unsafe { SSL_process_quic_post_handshake(self.as_mut_ptr()) };
         self.set_ex_data::<Connection>(*QUICHE_EX_DATA_INDEX, std::ptr::null())?;
 
+        self.set_transport_error(ex_data, rc);
         self.map_result_ssl(rc)
     }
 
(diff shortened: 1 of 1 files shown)


quiche/src/tls.rs
```diff
@@ -1093,6 +1095,24 @@ fn map_result_ptr<'a, T>(bssl_result: *const T) -> Result<&'a T> {
     }
 }
 
+fn set_local_error_if_needed(
```
tivrfoa, line 1098: Hi!

Why is this function defined outside of `impl Handshake`?

If it was an associated function (defined inside impl ...) you could then just call like:<br>
`self.set_local_error_if_needed(ex_data, rc);`

quiche/src/tls.rs
```diff
@@ -1093,6 +1095,24 @@ fn map_result_ptr<'a, T>(bssl_result: *const T) -> Result<&'a T> {
...
+) {
+    // SSL_ERROR_SSL
+    if ssl.get_error(bssl_result) == 1 {
+        // SSL_ERROR_SSL can't be recovered so ensure we set a
+        // local_error so the connection is closed.
+        // See https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html
+        if ex_data.local_error.is_none() {
+            *ex_data.local_error = Some(ConnectionError {
+                is_app: false,
+                error_code: TLS_ALERT_ERROR,
```
ghedo, line 1109: This  is not actually a TLS alert though, so it should be `0x01` (`INTERNAL_ERROR`) instead IMO.

quiche/src/tls.rs
ghedo: Yeah, better make this a method of `Handshake` (`map_result_ssl()` should probably be too, but that's a separate issue).

Also, maybe call this `set_transport_error()` or something.

ghedo (changes requested):

normanmaurer:
@ghedo I can open a PR to move `map_result_ssl ` as well

normanmaurer:
@ghedo FYI https://github.com/cloudflare/quiche/pull/1342

sildvauresberg (dismissed):

normanmaurer:
@ghedo should I rebase now that https://github.com/cloudflare/quiche/pull/1342 is merged ?

ghedo:
@normanmaurer Yeah, or I'll do it in a bit.

normanmaurer:
@ghedo done

ghedo (approved):

PR #1335, opened 2022-09-28
Fix the bug that quiche client v0.9.0 will be blocked when post data to nginx with quiche server v0.10.0+ (when proxy_request_buffering is on)

bachue:
I just found if the http3 client is quiche v0.9.0, and the server is nginx with quiche v0.10.0+(includes v0.14.0) and proxy_request_buffering is on, then POSTing some form data to the server would be blocked forever. But if `rb->rest` is increased by only 1, the bug would be fixed.


ghedo:
Superseded by https://github.com/cloudflare/quiche/commit/e434e42173f73bb533f6a39cd853eac752731fd9.

PR #1336, opened 2022-09-28
Add fn for processing undecrypted 0-rtt packets that can be reused

normanmaurer:
Motivation:

There was some code-duplication which can be removed by adding function.

Modifications:

Add process_undecrypted_0rtt_packets

Result:

Share code / cleanup


ghedo:
Merged, thanks @normanmaurer!

PR #1340, opened 2022-10-04
dgram: schedule both stream and dgram using the urgency

masa-koz:
This allows the application to set the priority of dgrams using dgram_send_priority() instead of dgram_send().
And this changes the way of sending stream and dgram to schedule them using the urgency.

### DIFF
--- a/quiche/src/dgram.rs
+++ b/quiche/src/dgram.rs
@@ -27,12 +27,13 @@
 use crate::Error;
 use crate::Result;
 
+use std::collections::BTreeMap;
 use std::collections::VecDeque;
 
 /// Keeps track of DATAGRAM frames.
 #[derive(Default)]
 pub struct DatagramQueue {
-    queue: Option<VecDeque<Vec<u8>>>,
+    queue: BTreeMap<u8, VecDeque<Vec<u8>>>,
     queue_max_len: usize,
     queue_bytes_size: usize,
 }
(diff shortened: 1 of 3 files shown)


quiche/src/dgram.rs
```diff
@@ -27,44 +27,46 @@
...
+    pub fn push(&mut self, data: Vec<u8>, urgency: u8) -> Result<()> {
         if self.is_full() {
             return Err(Error::Done);
         }
 
         self.queue_bytes_size += data.len();
         self.queue
-            .get_or_insert_with(Default::default)
+            .entry(urgency)
+            .or_insert_with(Default::default)
```
ehaydenr, line 58: I think this is equivalent to `.or_default()`

quiche/src/dgram.rs
```diff
@@ -27,44 +27,46 @@
...
             .push_back(data);
 
         Ok(())
     }
 
     pub fn peek_front_len(&self) -> Option<usize> {
-        self.queue.as_ref().and_then(|q| q.front().map(|d| d.len()))
+        self.queue.iter().next().and_then(|(_, q)| q.front().map(|d| d.len()))
```
ehaydenr, line 65: Now that it's stable, https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html#method.first_key_value is another option

quiche/src/dgram.rs
```diff
@@ -80,7 +82,23 @@ impl DatagramQueue {
...
+                } else {
+                    None
+                };
+                (data, clear)
+            } else {
+                (None, None)
+            };
+        
+        if let Some(urgency) = &clear {
+            self.queue.remove(urgency);
```
ehaydenr, line 99: what happens if the the BTreeMap has an entry for urgency 1, but the queue is empty? Should we move on to the next urgency? Seems like we should either loop to go to the next urgency (if any) or we should make sure this case never happens. This seems like a good scenario for a unit test

quiche/src/dgram.rs
```diff
@@ -89,11 +107,11 @@ impl DatagramQueue {
     }
 
     pub fn has_pending(&self) -> bool {
-        !self.queue.as_ref().map(|q| q.is_empty()).unwrap_or(true)
+        !self.queue.iter().next().map(|(_, q)| q.is_empty()).unwrap_or(true)
     }
 
     pub fn purge<F: Fn(&[u8]) -> bool>(&mut self, f: F) {
-        if let Some(q) = self.queue.as_mut() {
+        for (_, q) in self.queue.iter_mut() {
```
ehaydenr, line 114: now that we're looping, `self.queue_bytes_size` will end up reflecting the _last_ queue size, not the total. Maybe set to zero before the loop and accumulate each queue size as you loop?

PR #1341, opened 2022-10-05
quic: avoid panics for non-utf8 reasons/errors

ehaydenr:
If quiche receives non-utf8 reason bytes for conn/app close and attempts to write to qlog, it will panic.


junhochoi (dismissed):

LPardue (approved):
LGTM, thanks!

PR #1342, opened 2022-10-07
Move map_result_ssl(...) function to Handshake.

normanmaurer:
Motivation:

We should just move map_result_ssl(...) to Handshake to be more consistent

Modifications:

Move map_result_ssl(...)

Result:

Cleanup


ghedo (approved):

normanmaurer:
@ghedo thanks!

PR #1345, opened 2022-10-07
h3: fixup for with_transport() with early data

LPardue:


LPardue:
initial build failure was due to clippy ruling  `if !conn.is_server && !(conn.is_established() || conn.is_in_early_data())` is clearer, but I'd argue that is isn't clearer, so let's bind the client check to variable.

ghedo (approved):

PR #1349, opened 2022-10-10
qlog: add event for acked and dropped stream data

LPardue:


quiche/src/lib.rs
```diff
@@ -2591,6 +2591,21 @@ impl Connection {
...
+                                qlog::events::quic::DataMoved {
+                                    stream_id: Some(stream_id),
+                                    offset: Some(offset),
+                                    length: Some(length as u64),
+                                    from: Some(DataRecipient::Transport),
+                                    to: Some(DataRecipient::Dropped),
+                                    data: None,
+                                },
+                            );
+                            let now = time::Instant::now();
```
ghedo, line 2605: It looks like we already have a `now` variable used for qlog above?

LPardue: good catch, fixed

ghedo (approved):

PR #1355, opened 2022-10-24
Better support for Immediate Close during Handshake

LPardue:
Previously, when a quiche connection was closed using `close()` with
`app` true, the connection might not already have been in a state
where it was safe to emit the application CONNECTION_CLOSE frame. Because
calling `close()` puts the connection into a terminal state where
newly received packets are discarded rather than processed, it is
possible that a connection in this state cannot emit a suitable
CONNECTION_CLOSE. In some circumstances, this could have led to a
connection becoming idle and then silently closed.

Applications could have worked around this by carefully managing the
use of `close()` based on connection states. For example, by
manually converting to a connection level close (`app` false)
or by deferring the close until a later moment.

This change modifies the `close()` method so that it automatically
detects connection state and follows the recommendations in
https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.3-3. If
`close()` is called with `app` true, and it is not safe to emit,
the application error information is cleared and replaced with
a connection error of type 0x0c and empty reason.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -5569,7 +5569,11 @@ impl Connection {
     /// Closes the connection with the given error and reason.
     ///
     /// The `app` parameter specifies whether an application close should be
-    /// sent to the peer. Otherwise a normal connection close is sent.
+    /// sent to the peer. Otherwise a normal connection close is sent. If `app`
+    /// is true but the connection is not in a state that is safe to send an
+    /// application error (not established nor in early data), the error is
+    /// details are cleared in accordance with
+    /// https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.3-3.
     ///
     /// Returns [`Done`] if the connection had already been closed.
     ///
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -5545,7 +5545,11 @@ impl Connection {
     /// Closes the connection with the given error and reason.
     ///
     /// The `app` parameter specifies whether an application close should be
-    /// sent to the peer. Otherwise a normal connection close is sent.
+    /// sent to the peer. Otherwise a normal connection close is sent. If `app`
```
ghedo, line 5548: Nit: maybe make a new paragraph, like "Note that if `app` is true...".

quiche/src/lib.rs
```diff
@@ -5545,7 +5545,11 @@ impl Connection {
...
+    /// is true but the connection is not in a state that is safe to send an
+    /// application error (not established nor in early data), the error is
+    /// details are cleared in accordance with
+    /// https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.3-3.
```
ghedo, line 5552:
```suggestion
    /// [RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.3-3).
```

ghedo (changes requested):
Added a couple of nits, but lgtm overall.

quiche/src/lib.rs
```diff
@@ -5571,6 +5571,11 @@ impl Connection {
     /// The `app` parameter specifies whether an application close should be
     /// sent to the peer. Otherwise a normal connection close is sent.
     ///
+    /// If `app` is true but the connection is not in a state that is safe to
+    /// send an application error (not established nor in early data), the error
+    /// is details are cleared in accordance with [RFC
```
ghedo, line 5576: Was this supposed to say "error's details"?

ghedo (approved):

PR #1357, opened 2022-10-25
recovery: emit 'congestion_state_updated' qlog event

ghedo:


quiche/src/recovery/cubic.rs
```diff
@@ -425,6 +426,10 @@ fn has_custom_pacing() -> bool {
     false
 }
 
+fn state_str(r: &Recovery, now: Instant) -> &'static str {
+    reno::state_str(r, now)
```
LPardue, line 430: bit yuck to refer to reno here :D

maybe we could move states that are shared across CC's into a common type?

LPardue (approved):
LGTM modulo nit

Issue #635, opened 2020-09-13
Keep Alive + Ping 


victorstewart:
can we add a function to generate PING frames? specifically to the C api.

i need a way to send keep alives on idle connections, and drive the timeouts in my IO event loop.

i can submit a PR if you're all amenable to this.

ghedo:
Yeah, that seems like something that could be useful, thanks.

normanmaurer:
@ghedo is there any work planed on this ? If not I would like volunteer for adding support for PING. WDYT ?

normanmaurer:
talked with @ghedo offline ... I will work on a PR soonish..

lilongyan-huawei:
> talked with @ghedo offline ... I will work on a PR soonish..

@normanmaurer At your convenience, would you please tell me when you will submit the PR ？
I‘m also planning to implement this feature.

normanmaurer:
@lilongyan-huawei I didn't start yet... so if you want to work on this please do so. I think I will have no time for the next 2 weeks anyway.

normanmaurer:
@lilongyan-huawei I guess this was never started at your end ?

lilongyan-huawei:
> @lilongyan-huawei I guess this was never started at your end ?

have done. doing tests

normanmaurer:
@lilongyan-huawei thanks a lot... Looking forward to the PR

lilongyan-huawei:
您好,您的邮件我已收到，谢谢.

normanmaurer:
@ehaydenr I think you worked on this one ?

CMCDragonkai:
So currently there's no way to keep an idle QUIC connection alive?

I suppose we can just use the unreliable datagram extension to do this. But the QUIC protocol does specify the usage of PING frame to do this.

lilongyan-huawei:
您好,您的邮件我已收到，谢谢.

LPardue:
This is possible via send_ack_eliciting or send_ack_eliciting_on_path; see

https://docs.quic.tech/quiche/struct.Connection.html#method.send_ack_eliciting

Add in #1361 and should have closed this issue as resolved.

PR #1361, opened 2022-11-09
quic: API for app to elicit ACK from peer

ehaydenr:
- useful for keepalives
- ack-elicit is per-path to increase compatibility with multi-path work in #1310 
- not exactly a PING API since it doesn't send a PING if you would already send an ACK-eliciting packet. May or may not address #635 
- if this is interesting, we can update the C API too

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3887,9 +3887,11 @@ impl Connection {
         // Alternate trying to send DATAGRAMs next time.
         self.emit_dgram = !dgram_emitted;
 
-        // Create PING for PTO probe if no other ack-eliciting frame is sent or if
-        // we've sent too many non ACK eliciting packets without having
-        // sent an ACK eliciting one
+        // If no other ack-eliciting frame is sent, include a PING frame
+        // - if PTO probe needed; OR
+        // - if we've sent too many non ack-eliciting packets without having
+        // sent an ACK eliciting one; OR
+        // - the application requested an ack-eliciting frame be sent.
         let path = self.paths.get_mut(send_pid)?;
         if (ack_elicit_required || path.needs_ack_eliciting) &&
             !ack_eliciting &&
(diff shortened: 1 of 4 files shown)


normanmaurer:
@ehaydenr can you please also add ffi code and update the `quiche.h` file ?

quiche/src/lib.rs
```diff
@@ -3890,7 +3890,12 @@ impl Connection {
         // Create PING for PTO probe if no other ack-eliciting frame is sent or if
         // we've sent too many non ACK eliciting packets without having
         // sent an ACK eliciting one
```
LPardue, line 3892: Might worth updating this comment to cover the new condition, possibly we can break the long paragraph into some bullet points or something for clarity

quiche/src/lib.rs
```diff
@@ -4805,6 +4804,37 @@ impl Connection {
         MIN_CLIENT_INITIAL_LEN
     }
 
+    /// Queue a PING frame to be sent for the active path if no other
+    /// ACK-eliciting frame would otherwise be sent. See
+    /// [`send_ack_eliciting_on_path()`] for more detail.
+    ///
+    /// [`send_ack_eliciting_on_path()`]:
+    /// struct.Connection.html#method.send_ack_eliciting_on_path
```
LPardue, line 4812: Our usual pattern for public docs is a simple one line description on first line, then more detail on subsequent lines. So maybe something like 

```
/// Schedule an ack-eliciting packet on the active path.
///
/// QUIC packets might not contain ack-eliciting frames during normal operating conditions.
/// If the packet would already contain ack-eliciting frames, this method does not change any behavior.
/// However, if the packet would not ordinarily contain ack-eliciting frames, this method ensures that a PING frame
/// sent.
///
/// Calling this method multiple times before [`send()`] has no effect.
```

quiche/src/lib.rs
```diff
@@ -4805,6 +4804,37 @@ impl Connection {
...
+        if self.is_closed() || self.is_draining() {
+            return Ok(());
+        }
+        self.paths.get_active_mut()?.needs_ack_eliciting = true;
+        Ok(())
+    }
+
+    /// Queue a PING frame to be sent for the specified path if no other
+    /// ACK-eliciting frame would otherwise be sent. Calling this method
+    /// multiple times before calling send will have no effect.
```
LPardue, line 4823: My comment from above applies here too. However, this function can return an error if the user-provided path doesn't exist, so maybe we need to document that possibility in the API a little more explicitly.

quiche/src/lib.rs
```diff
@@ -4805,6 +4806,44 @@ impl Connection {
...
 
+    /// Schedule an ack-eliciting packet on the active path.
+    ///
+    /// QUIC packets might not contain ack-eliciting frames during normal
+    /// operating conditions. If the packet would already contain
+    /// ack-eliciting frames, this method does not change any behavior.
+    /// However, if the packet would not ordinarily contain ack-eliciting
+    /// frames, this method ensures that a PING frame sent.
+    ///
+    /// Calling this method multiple times before [`send()`] has no effect.
```
LPardue, line 4817: This should fix the warning

```suggestion
    /// Calling this method multiple times before [`send()`] has no effect.
    
    /// [`send()`]: struct.Connection.html#method.send
```

quiche/src/lib.rs
```diff
@@ -4805,6 +4806,46 @@ impl Connection {
...
+
```
LPardue, line 4818:
```suggestion
    ///
```

LPardue (approved):
LGTM

LPardue:
> @ehaydenr can you please also add ffi code and update the `quiche.h` file ?

Actually +1 Norman's request here, lets add it at the same time in this PR

ehaydenr:
whoops, thanks forgot about that. Just `ffi.rs` and `quiche.h`, right?

LPardue (approved):

PR #1365, opened 2022-12-01
h3: include minimum required capacity in Error::StreamBlocked

ehaydenr:


quiche/src/h3/mod.rs
```diff
@@ -1013,8 +1014,10 @@ impl Connection {
         if let Err(e) = conn.stream_send(stream_id, b"", false) {
             self.streams.remove(&stream_id);
 
+            // This should be impossible since we're only sending a 0 length
```
ehaydenr, line 1017: I think this branch is impossible, but maybe it won't be in the future?

PR #1380, opened 2023-01-12
Detect stateless reset

mhyumike:
This change provides a simple stateless reset detection that applications can prevent from using a connection that is supposed to be closed.

This PR completes the unfinished implementation on #1037

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2030,7 +2030,17 @@ impl Connection {
             ) {
                 Ok(v) => v,
 
-                Err(Error::Done) => left,
+                Err(Error::Done) => {
+                    // If the packet can't be processed or decrypted, check if it's a stateless
+                    // reset.
+                    if self.is_stateless_reset(&buf[len - left..len]) {
+                        trace!("{} packet is a stateless reset", self.trace_id);
+                        let now = time::Instant::now();
+                        let path = self.paths.get_active()?;
+                        self.draining_timer = Some(now + (path.recovery.pto() * 3));
+                    }
+                    left
+                },
 
                 Err(e) => {
                     // In case of error processing the incoming packet, close
@@ -2069,6 +2079,27 @@ impl Connection {
             }
         }
         Ok(())
+    }
+
+    /// Returns true if a QUIC packet is a stateless reset.
+    fn is_stateless_reset(&self, buf: &[u8]) -> bool {
+        // If the packet is too small, then we just throw it away
+        let buf_len = buf.len();
+        if buf_len < 21 {
+            return false;
+        }
+
+        match self.peer_transport_params.stateless_reset_token {
+            Some(token) => {
+                let token_len = 16;
+                if let Ok(token_from_buf) = buf[buf_len - token_len..buf_len].try_into() {
+                    return token == u128::from_be_bytes(token_from_buf)
+                }
+                // Failed to convert the slice to an array.
+                false
+            },
+            None => false,
+        }
     }
 
     /// Processes a single QUIC packet received from the peer.
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2031,14 +2031,14 @@ impl Connection {
                 Ok(v) => v,
 
                 Err(Error::Done) => {
-                    // If the packet can't be processed or decrypted, check if it's a stateless
-                    // reset.
+                    // If the packet can't be processed or decrypted, check if
+                    // it's a stateless reset.
                     if self.is_stateless_reset(&buf[len - left..len]) {
                         trace!("{} packet is a stateless reset", self.trace_id);
-                        let now = time::Instant::now();
-                        let path = self.paths.get_active()?;
-                        self.draining_timer = Some(now + (path.recovery.pto() * 3));
+
+                        self.closed = true;
                     }
+
                     left
                 },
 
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -2030,7 +2030,17 @@ impl Connection {
...
 
-                Err(Error::Done) => left,
+                Err(Error::Done) => {
+                    // If the packet can't be processed or decrypted, check if it's a stateless
+                    // reset.
+                    if self.is_stateless_reset(&buf[len - left..len]) {
+                        trace!("{} packet is a stateless reset", self.trace_id);
+                        let now = time::Instant::now();
+                        let path = self.paths.get_active()?;
+                        self.draining_timer = Some(now + (path.recovery.pto() * 3));
```
ghedo, line 2040: I don't think we need to drain the connection on stateless reset (if the peer sends a reset it's unlikely it will send any more packets), so you can just set `self.closed = true` here instead.

quiche/src/lib.rs
```diff
@@ -2071,6 +2081,27 @@ impl Connection {
...
 
+    /// Returns true if a QUIC packet is a stateless reset.
+    fn is_stateless_reset(&self, buf: &[u8]) -> bool {
+        // If the packet is too small, then we just throw it away
+        let buf_len = buf.len();
+        if buf_len < 21 {
+            return false;
+        }
+
+        match self.peer_transport_params.stateless_reset_token {
```
ghedo, line 2092: In theory we should check against the reset token of every active connection ID here, not just the initial one. Though it doesn't look like we have an API for iterating over all connection IDs currently, so the current code is probably fine for now, but let's add a TODO comment here for future reference.
```suggestion
        // TODO: we should iterate over all active destination connection IDs and check against their reset token.
        match self.peer_transport_params.stateless_reset_token {
```

quiche/src/lib.rs
```diff
@@ -2030,7 +2030,17 @@ impl Connection {
...
-                Err(Error::Done) => left,
+                Err(Error::Done) => {
+                    // If the packet can't be processed or decrypted, check if it's a stateless
+                    // reset.
+                    if self.is_stateless_reset(&buf[len - left..len]) {
+                        trace!("{} packet is a stateless reset", self.trace_id);
+                        let now = time::Instant::now();
+                        let path = self.paths.get_active()?;
+                        self.draining_timer = Some(now + (path.recovery.pto() * 3));
+                    }
```
ghedo, line 2041: Nit: empty line
```suggestion
                    }

```

quiche/src/lib.rs
```diff
@@ -2071,6 +2081,27 @@ impl Connection {
...
+
+        match self.peer_transport_params.stateless_reset_token {
+            Some(token) => {
+                let token_len = 16;
+                if let Ok(token_from_buf) = buf[buf_len - token_len..buf_len].try_into() {
+                    return token == u128::from_be_bytes(token_from_buf)
+                }
+                // Failed to convert the slice to an array.
+                false
+            },
```
ghedo, line 2100: Nit: empty line
```suggestion
            },

```

quiche/src/lib.rs
```diff
@@ -2071,6 +2081,27 @@ impl Connection {
         Ok(())
     }
 
+    /// Returns true if a QUIC packet is a stateless reset.
+    fn is_stateless_reset(&self, buf: &[u8]) -> bool {
+        // If the packet is too small, then we just throw it away
```
ghedo, line 2086:
```suggestion
        // If the packet is too small, then we just throw it away.
```

quiche/src/lib.rs
```diff
@@ -2071,6 +2081,27 @@ impl Connection {
...
+        if buf_len < 21 {
+            return false;
+        }
+
+        match self.peer_transport_params.stateless_reset_token {
+            Some(token) => {
+                let token_len = 16;
+                if let Ok(token_from_buf) = buf[buf_len - token_len..buf_len].try_into() {
+                    return token == u128::from_be_bytes(token_from_buf)
+                }
```
ghedo, line 2097: Nit: empty line
```suggestion
                }

```

quiche/src/lib.rs
```diff
@@ -2071,6 +2081,27 @@ impl Connection {
...
+        let buf_len = buf.len();
+        if buf_len < 21 {
+            return false;
+        }
+
+        match self.peer_transport_params.stateless_reset_token {
+            Some(token) => {
+                let token_len = 16;
+                if let Ok(token_from_buf) = buf[buf_len - token_len..buf_len].try_into() {
+                    return token == u128::from_be_bytes(token_from_buf)
```
ghedo, line 2096: RFC9000 calls for a constant-time comparison of the reset token (see https://www.rfc-editor.org/rfc/rfc9000#section-10.3.1-4) and I don't think it's the case here.

The ring crate we already depend on provides a constant-time comparison API that could be used here, see https://docs.rs/ring/0.16.20/ring/constant_time/fn.verify_slices_are_equal.html

ghedo (changes requested):
Thanks for the PR! I added a couple of comments and some minor nits, but generally looks good to me.

ghedo:
Also, please run `cargo +nightly fmt` to fix the failing build.

mhyumike:
Thanks for the review! All comments are addressed.

ghedo (approved):

ghedo:
Merged, thanks @mhyumike! I only tweaked the commit message a little to explain that additional changes will be needed at a later time to cover all stateless reset cases.

quiche/src/lib.rs
mhyumike: Thanks for the explanation and the code link! I'll be watching out for these cases.

Issue #1277, opened 2022-07-11
No ACK is sent when the cwnd is full


francoismichel:
Hello,

It seems that in `send_single()` in `lib.rs`, ACKs are only added to packets when the congestion window has remaining space.

This can lead to problems when both endpoints are sending: the receiver of new data won't be able to send the ACK frames
and it can lead to poor performances or connection interruption that I experienced during some tests.

The `left` variable should maybe not be capped to the `cwnd_available` of the path while ACKs have not been packed into the packet.

I can work on a PR to fix the issue.

What do you think ?

LPardue:
Hey @francoismichel . Thanks for the report. I think you're correct in your assessment that we are missing opportunities to send ACKs if cwnd is full. From RFC 9000, only ACK-only packets don't count against cwnd. So a better solution might be to have a branch of code in send_single() that is invoked when the cwnd_available = 0, that branch would only consider sending ACKs (transmission or retransmission).

francoismichel:
Just opened #1385 to fix this. It seems to work correctly.

It changes quite a few lines of code because of the renaming of the `left` variable and `cargo fmt`.

> So a better solution might be to have a branch of code in send_single() that is invoked when the cwnd_available = 0, that branch would only consider sending ACKs (transmission or retransmission).

I did not follow this exact approach to be able to correctly handle the cases where 0 < cwnd_available < ack_frame_size. Let me know if you wish to change anything.

PR #1385, opened 2023-01-18
Allow sending ACK only packets when the cwnd is full

francoismichel:
Addresses issue #1277.

Without this PR, having no space available in the congestion window prevents to send acknowledgements for the newly received packets. This can lead to situations with poor performance where no ACK is sent during a noticeable amount of time.

To implement this behaviour, we now only consider the available buffer size (`buf_cap`) when packing ACK frames.
When ACK frames are packed, reduce `left_for_congestion_controlled_packets` (new name for the `left` variable) to take the ACK frame size into account when packing the next non-ACK frames.

I added a new unit test verifying the inteded behaviour (`sends_ack_only_pkt_when_full_cwnd_and_ack_elicited`). This new test fails without the changes of this PR.

Do not hesitate to test, review and modify anything you want.

François

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3177,7 +3177,7 @@ impl Connection {
             }
         }
 
-        let mut buffer_capacity = b.cap();
+        let mut left = b.cap();
 
         let pn = self.pkt_num_spaces[epoch].next_pkt_num;
         let pn_len = packet::pkt_num_len(pn)?;
@@ -3261,8 +3261,8 @@ impl Connection {
         }
 
         // Make sure we have enough space left for the packet overhead.
-        match buffer_capacity.checked_sub(overhead) {
-            Some(v) => buffer_capacity = v,
+        match left.checked_sub(overhead) {
+            Some(v) => left = v,
 
             None => {
                 // We can't send more because there isn't enough space available
@@ -3280,23 +3280,13 @@ impl Connection {
         }
 
         // Make sure there is enough space for the minimum payload length.
-        if buffer_capacity < PAYLOAD_MIN_LEN {
+        if left < PAYLOAD_MIN_LEN {
             self.paths
                 .get_mut(send_pid)?
                 .recovery
                 .update_app_limited(false);
             return Err(Error::Done);
         }
-
-        // Limit output packet size by congestion window size.
-        let mut left_for_congestion_controlled_packets = cmp::min(
-            buffer_capacity,
-            self.paths
-                .get(send_pid)?
-                .recovery
-                .cwnd_available()
-                .saturating_sub(overhead),
-        );
 
         let mut frames: Vec<frame::Frame> = Vec::new();
 
@@ -3322,6 +3312,8 @@ impl Connection {
         packet::encode_pkt_num(pn, &mut b)?;
 
         let payload_offset = b.off();
+
+        let left_before_packing_ack_frame = left;
 
         // Create ACK frame.
         //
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -3177,11 +3177,7 @@ impl Connection {
             }
         }
 
-        let mut left = b.cap();
-
-        // Limit output packet size by congestion window size.
-        left =
-            cmp::min(left, self.paths.get(send_pid)?.recovery.cwnd_available());
+        let mut buffer_capacity = b.cap();
```
ghedo, line 3180: I think it would be cleaner to keep the `left` name, and simply move the `left = cmp::min(left, cwnd_available);` line to after the ACK frame is created? This way it seems the change would be smaller, and we wouldn't need the additional `consumed_by_acks` logic.

ghedo (changes requested):
Thanks! Looks good overall, but added a suggestion that could simplify things considerably.

quiche/src/lib.rs
francoismichel: Thanks for the review !
I still have a small question, if we put:

```
left = cmp::min(left, self.paths.get(send_pid)?.recovery.cwnd_available())
```

after the ACKs have been packed, how would we prevent the resulting packet to be larger than `cwnd_available` without using the `consumed_by_acks` logic ? We still need to subtract the bytes of the ACK frame from the available cwnd at some point to avoid packing too large packets. Or do we consider it is not that big of a deal ?

ghedo: I'm not sure I understand the question. `push_frame_to_pkt!()` will decrease `left` by the amount of ACK frame bytes written, so doing `cmp::min(left, cwnd_available)` afterwards will already take the ACK frame size into consideration? And if `cwnd_available` is zero then `left` becomes zero too. Am I missing something?

francoismichel: I was just thinking of the case where `cwnd_available` is both smaller than `b.cap()` and larger than 0. 

For instance if `b.cap()` is 1000, `cwnd_available` is 500 and we want to bundle an ACK frame of 10 bytes.
If I understand everything correctly, we start with `left` equal to `b.cap()` (=1000). We push the 10 bytes ACK frame into the packet so `left` is decreased by 10 and is now equal to 990. Then, we do `left = cmp::min(left, cwnd_available)`, so it becomes 500 because `cwnd_available` is equal to 500.

So it means that we can pack 500 more bytes of e.g. stream frames, resulting to a payload of 510 bytes (10 bytes for the ACK and 500 for the stream frames) while `available_cwnd` was 500, so the packet is too large for the available cwnd. I am not sure this is a big deal but this behaviour is diffrent from what quiche was doing before.

francoismichel:
Discussed offline with @qdeconinck and did a few changes to take the comments into  account.

Basically I've restored `left`'s name to avoid modifying too many lines.

Then I've removed the `consumed_by_acks` variable when packing ACKs , but still ensure
that the bytes consumed by the ACK frame are taken into account in the cwnd when packing the next frames.

It is a bit less explicit than before and therefore changes less code, but the behaviour should be the exact same. :-)

Thanks @qdeconinck and @ghedo for the discussion !

ghedo:
Thanks @francoismichel (and sorry for the delay, I got distracted by other things). This looks good to me, I'll merge shortly.

ghedo (approved):

ghedo:
Merged, thanks @francoismichel!

PR #1389, opened 2023-01-23
Add a way to dynamically link BoringSSL that is in a specified path

maskit:
This adds a way to dynamically link BoringSSL that is in a specified path. It enables users to use BoringSSL shared libraries (.so files) that users already have. The benefits are:
- Allows users to update BoringSSL without rebuilding Quiche
- Ensures that application and quiche use the same (compatible) BoringSSL (`quiche_conn_new_with_tls` requires it).

/cc @normanmaurer


normanmaurer (dismissed):

ghedo:
Thanks for the PR! Could you please rebase this to get the CI builds to pass?

ghedo (approved):

ghedo:
Merged, thanks @maskit!

Issue #1197, opened 2022-03-17
Logging with absolute Unix epoch timestamp instead of relative time in QLOG files


birkandenizer:
According to the official QLOG main schema [1], timestamps do not have to use the UNIX epoch timestamp as their reference. It is suggested that a reference timestamp such as "time since connection start in ms" can be used. I believe this is how it is implemented in the currently outputted sqlog files based on JSON-SEQ from quiche. 

I'm currently using quiche-client and quiche-server to generate network traces. However, I need access to the UNIX epoch timestamps of each event for my use case to correlate them with other events. My questions are:

1) Is it possible to get sqlog files with UNIX timestamps instead of relative timestamps that start from 0.0?

2) Is it correct to assume the creation time of a sqlog file as the first time a packet is received and saved? Thus use it as a reference point to convert the relative timestamps in the fields to absolute UNIX timestamps by summing them up together.

[1] https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-02#section-3.4.1

LPardue:
Hi! Thanks for asking, I think this all makes sense. Let me think on it some more over the weekend.

birkandenizer:
There are also some additional suggestions in the draft about my questions.

For my first question, relative approach that is currently used should include a "reference_time" field in "common_fields" [1]. It is said that ""reference_time" value is typically the first absolute timestamp". However, currently outputted sqlog files do not include such a field. I do not know if this is intentional or an oversight.

For my second question, in the currently outputted sqlog files, there is a "time_offset" field [2] that should indicate "how many milliseconds the starting time of the current trace should be offset". So my understanding from the created sqlog files is that since this "time_offset" is always 0.0, the creation time of the file is the same time when the first event is logged. Would that be a correct assumption?

[1] https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-02#section-3.4.1
[2] https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-02#section-3.3.1.1

LPardue:
I thought on this some more and ended up creating an associated issue on the qlog spec itself - https://github.com/quicwg/qlog/issues/198. We diverge from a strict interpretation of the spec.

[...]

PR #1390, opened 2023-01-23
Logging of reference_time as part of common_fields in qlog

birkandenizer:
This pull request aims to log the starting time of a connection. 

'Main logging schema for qlog' [1] suggest to specify a full 'reference_time' time stamp as part of 'common_fields' for logs that use relative style timestamps.

We first update the definition of reference_time to float64 to align better with 'Main logging schema for qlog' [1]. Then, we enable logging of reference_time to beginning of the connection.

[1] https://datatracker.ietf.org/doc/draft-ietf-quic-qlog-main-schema/

### DIFF
--- a/qlog/src/lib.rs
+++ b/qlog/src/lib.rs
@@ -605,7 +605,7 @@ pub struct CommonFields {
     pub group_id: Option<String>,
     pub protocol_type: Option<Vec<String>>,
 
-    pub reference_time: Option<f32>,
+    pub reference_time: Option<f64>,
     pub time_format: Option<String>,
     // TODO: additionalUserSpecifiedProperty
 }
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1883,7 +1883,15 @@ impl Connection {
                 time_offset: Some(0.0),
                 original_uris: None,
             }),
-            None,
+            Some(qlog::CommonFields {
+                group_id: None,
+                protocol_type: None,
+                reference_time: Some(time::SystemTime::now()
+                .duration_since(time::UNIX_EPOCH)
+                .expect("Time went backwards")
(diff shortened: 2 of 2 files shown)


LPardue:
hi @birkandenizer. Thanks for the PR. We talked about this previously on https://github.com/cloudflare/quiche/issues/1197, I don't think anything has changed. 

The reference time you're capturing here is SystemTime but the clock we use for events is a monotonic clock. These are different clocks and its incorrect to treat on as a reference to the other. Per https://github.com/quicwg/qlog/issues/198, I think the real solution is to add a monotonic clock type.

Note that the line that changes the `reference_time` field to f64 is correct in isolation, since it makes the code align with the definition in the spec https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#name-commonfields-definition, so I'd take that one line change on its own if you want to split it off.

birkandenizer:
Hey @LPardue, thank you for the response. I will do a separate pull request for f64 change.

I checked the [Rework the Timestamps section, add monotonic](https://github.com/quicwg/qlog/pull/290). It seems like a step in the right direction but unfortunately still does not achieve what I personally want. My goal is to correlate events in (s)qlog files to events that might be provided by other logging tools. Therefore, I will eventually need to see exactly when an event happened in quiche. Is there a way to do this without breaking consistency guarantees of time::Instant::now() in Rust?

LPardue:
I think this depends on how the other tools generate logs, what timestamp format they use, and how precise or accurate the clocks need to be.

I'm not opposed to supporting a different TimeFormat as a configuration option, in addition to our monotonic approach.

birkandenizer:
Thank you for your input. I will open a further pull request or issue if I decide to go that route.

Issue #1394, opened 2023-01-25
Final unknown/grease frames break stream state


g2p:
Found when trying to interop with an h3-based server.
https://github.com/hyperium/h3 can send grease frames when finishing a stream.

https://www.rfc-editor.org/rfc/rfc9114.html#extensions
https://www.rfc-editor.org/rfc/rfc9114.html#name-reserved-frame-types
    
This breaks quiche which skips the frame and attempts to read again from the now closed stream, then hangs indefinitely trying to finish the request.

Run:

    # From github.com/hyperium/h3
    cargo run --example server -- --listen 127.0.0.1:4433
    # From github.com/cloudflare/quiche
    cargo run --bin quiche-client https://127.0.0.1:4433/ --no-verify

The request from quiche-client hangs:
      
    [2023-01-25T22:46:02.515328714Z ERROR quiche_apps::common] HTTP/3 processing failed: TransportError(InvalidStreamState(0))
    [2023-01-25T22:46:12.524032857Z ERROR quiche_apps::common] connection timed out after 10.017217243s and only completed 0/1 requests

LPardue:
Thanks for the detailed report. I can reproduce the behaviour. Sending GREASE frames at this point is totally allowed by the spec; see https://www.rfc-editor.org/rfc/rfc9114.html#section-4.1-9. So I agree this is a bug in quiche. 


I appreciate you making #1395 and while that is one way to try and fix the problem, I don't see much value in bubbling up the unknown frame event. But its given me an idea for a different fix that I'll try out and offer as an alternative PR.

g2p:
Thank you for looking at it!

Yes, the PR was just me making sure I understood where the issue was, I'm not surprised if there's a better way to fix it.

LPardue:
@g2p PTAL at https://github.com/cloudflare/quiche/pull/1396/files. This isn't that different from what you had already, so thanks for the inspiration. It adds a test case to cover the scenario. I can also correctly receive the full response from hyperium server. If you'd like to verify the PR fixes the problem for you too, that would be good.

PR #1395, opened 2023-01-25
Fix handling of a final grease frame

g2p:
This fixes https://github.com/cloudflare/quiche/issues/1394.  It adds an event type to fix the interplay between process_readable_stream and handle_responses, so that unknown frames can bubble up outside of process_readable_stream.


ghedo:
Thanks for the PR (and the issue)! Looks like this has been superseded by #1396, so closing.

PR #1396, opened 2023-01-26
h3: exit process_readable_stream() gracefully

LPardue:
Generally for framed streams (request or push streams) it is expected
to keep looping around reading as much data as possbile and returning
an Event when something worth an application's attention happened, such
as HEADERS, GOAWAY, DATA (although that is a special case of its own).

Internally, we run this loop inside process_readable_stream() and apps
can invoke it via the poll() method. If a stream presents an event,
poll() would check if the stream was finished, mark it if so, then return
the event. The next time poll() was called, the marking would cause a
Finished event to happen. This all works fine as long as a frame
triggers an event that breaks process_readable_stream()'s loop.

However, there can be several types of frame that don't need to be
bubbled up, like GREASE frames or other extension frames. Previously,
one of these was the last frame on a stream, we'd erroneously try and
loop around again and read the next frame (which could never happen
because the stream had been fin'd). This would result in an
h3::Error(TransportError(InvalidStreamState(stream_id)) error getting
thrown up to the application, which is unexpected because everything was
read correctly.

With this change, if a frame is read correctly (meaning we read the
whole thing succesfully out of the stream) and its a frame that
generates no notable event, we'll check if the stream is finished.
If so, process_readable_stream() will exit gracefully with Error::Done
and the stream_finished() checks elsewhere will deal with correctly
generating the Finished events.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -2370,7 +2370,15 @@ impl Connection {
                     {
                         Ok(ev) => return Ok(ev),
 
-                        Err(Error::Done) => (),
+                        Err(Error::Done) => {
+                            // This might be a frame that is processed internally
+                            // without needing to bubble up to the user as an
+                            // event. Check whether the frame has FIN'd by QUIC
+                            // to prevent trying to read again on a closed stream.
+                            if conn.stream_finished(stream_id) {
+                                break;
+                            }
+                        },
 
                         Err(e) => return Err(e),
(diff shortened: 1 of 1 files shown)


ghedo:
Slightly reformatted the comment.

ghedo (approved):

g2p:
I can confirm this fixes #1394, with quiche-client and curl+quiche.

PR #1399, opened 2023-01-27
Add the ability to provide a custom BufFactory to a quiche connection

vkrasnov:
A BufFactory can override the generic buffer generation method with a custom method, eg. getting a buffer from a pool buffer.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -417,6 +417,8 @@ use std::collections::VecDeque;
 
 use smallvec::SmallVec;
 
+use range_buf::DefaultBufFactory;
+
 /// The current QUIC wire version.
 pub const PROTOCOL_VERSION: u32 = PROTOCOL_VERSION_V1;
 
@@ -1234,7 +1236,10 @@ impl Config {
 }
 
 /// A QUIC connection.
-pub struct Connection {
+pub struct Connection<F = DefaultBufFactory>
+where
+    F: BufFactory,
+{
     /// QUIC wire version used for the connection.
     version: u32,
 
@@ -1321,7 +1326,7 @@ pub struct Connection {
     lost_bytes: u64,
 
     /// Streams map, indexed by stream ID.
-    streams: stream::StreamMap,
+    streams: stream::StreamMap<F>,
 
     /// Peer's original destination connection ID. Used by the client to
     /// validate the server's transport parameter.
@@ -1469,6 +1474,21 @@ pub fn accept(
     Ok(conn)
 }
 
+/// Creates a new server-side connection, with a custom buffer generation
+/// method.
+///
+/// The buffers generated can be anything that can be drereferenced as a byte
+/// slice. See [`accept`] and [`BufFactory`] for more info.
+#[inline]
+pub fn accept_with_buf_factory<F: BufFactory>(
+    scid: &ConnectionId, odcid: Option<&ConnectionId>, local: SocketAddr,
+    peer: SocketAddr, config: &mut Config,
+) -> Result<Connection<F>> {
+    let conn = Connection::new(scid, odcid, local, peer, config, true)?;
+
+    Ok(conn)
+}
+
 /// Creates a new client-side connection.
 ///
 /// The `scid` parameter is used as the connection's source connection ID,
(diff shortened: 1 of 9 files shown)


quiche/src/lib.rs
```diff
@@ -1351,6 +1356,24 @@ pub fn accept(
     scid: &ConnectionId, odcid: Option<&ConnectionId>, local: SocketAddr,
     peer: SocketAddr, config: &mut Config,
 ) -> Result<Connection> {
+    println!("accept");
```
LPardue, line 1359: naughty

quiche/src/lib.rs
```diff
@@ -1351,6 +1356,24 @@ pub fn accept(
...
+/// method.
+///
+/// The buffers generated can be anything that can be drereferenced as a byte
+/// slice. See [`accept`] and [`BufFactory`] for more info.
+#[inline]
+pub fn accept_with_buf_factory<F: BufFactory>(
+    scid: &ConnectionId, odcid: Option<&ConnectionId>, local: SocketAddr,
+    peer: SocketAddr, config: &mut Config,
+) -> Result<Connection<F>> {
+    println!("accept with factory");
```
LPardue, line 1375: naughty

vkrasnov:
Any desire to review/merge this?

quiche/src/stream.rs
```diff
@@ -1144,11 +1158,16 @@ pub struct SendBuf {
 
     /// The error code received via STOP_SENDING.
     error: Option<u64>,
+
+    _bf: PhantomData<F>,
```
ehaydenr, line 1162: This can be removed. You're already using `F` in `data` above.

quiche/src/stream.rs
```diff
@@ -1144,11 +1158,16 @@ pub struct SendBuf {
...
 }
 
-impl SendBuf {
+impl<F> SendBuf<F>
```
ehaydenr, line 1165: If `BufFactory::Buf` is required to implement splitting and something like [chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) (but splits into iterator of `BufFactory::Buf`s), I think we can enable sending without alloc+copy. `SendBuf::write` would take an application-provided `BufFactory::Buf`.

quiche/src/stream.rs
```diff
@@ -642,12 +647,15 @@ impl StreamMap {
...
+pub struct Stream<F = DefaultBufFactory>
+where
+    F: BufFactory,
+{
     /// Receive-side stream buffer.
     pub recv: RecvBuf,
 
     /// Send-side stream buffer.
-    pub send: SendBuf,
+    pub send: SendBuf<F>,
```
ehaydenr, line 658: Why is only SendBuf parameterized with a BufFactory? Seems like RecvBuf should be parameterized as well. I'm also wondering if there are cases where you might want a different BufFactory for send vs. recv...

vkrasnov: Yes, that would be a follow up PR, RecvBuf is also important

normanmaurer:
I like the idea a lot! That said I think generally speaking I would love to be able to plugin a way to allocate a buffer that is used without the need to do any memory copy etc. Also I wonder if we could add the c API as well as part of the PR to ensure that we can make it work with the provided API.

vkrasnov:
> I like the idea a lot! That said I think generally speaking I would love to be able to plugin a way to allocate a buffer that is used without the need to do any memory copy etc. Also I wonder if we could add the c API as well as part of the PR to ensure that we can make it work with the provided API.

Yes, Eric mentioned it already, zerocopy write is definitely something that this change should enable. I am not sure this can be in anyway be exposed via the FFI I am afraid, it will have to rely on the default implementation, or maybe we can add builders for other kinds of bufs.

normanmaurer:
@vkrasnov hmm... I didn't check in detail but wouldnt it be just some sort of BufFactory C API wrapper that implements Drop for Buf ?

vkrasnov:
> @vkrasnov hmm... I didn't check in detail but wouldnt it be just some sort of BufFactory C API wrapper that implements Drop for Buf ?

So the C API is really problematic here as you can't really have generics in it, without doubling the interface surface area, in addition the factory is basically implemented as a global, stateless "class", so there really is no good way to pass it any function pointers from C or some such.

We could change it by using a dyn<Box<&BufferFactory>> instead, to eliminate the generic bounds, but that by itself would complicate this change a lot, since then the reference would have to be passed deep into the call stack, all the way to the place where RangeBufs are actually allocated, and in addition it will make it a tad slower, probably not critically so, but still have to make sure there is demand for it.

WDYT @ghedo , @ehaydenr

PR #1403, opened 2023-02-02
FFI: Change SendInfo to return in nanoseconds instead of timespec

FranzBusch:
# Motivation
Currently, we are returning a `timespec` in the `SendInfo` struct from the FFI interface. To construct that `timespec` we are copying over the bytes from a Rust `Instant` to the `timespec` struct. This runs into undefined behaviour since `Instant` is an opaque type and might not have the same layout as `timespec`. We have noticed this being a problem where Quiche is returning us `timespec`s with the nanosecond part being multiple days.

# Modification
This PR changes the way we return the "write at" information by changing it to a "write in" property. This allows wrapping libraries to use their native `Instant` types to get the current time and then add the returned nanoseconds value to it. This also aligns the interface with how we expose timeout information through the FFI layer.

# Result
We are no longer running into undefined behaviour.

cc @normanmaurer

### DIFF
--- a/quiche/src/ffi.rs
+++ b/quiche/src/ffi.rs
@@ -24,6 +24,7 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::time;
 use std::ffi;
 use std::ptr;
 use std::slice;
@@ -44,7 +45,6 @@ use libc::c_void;
 use libc::size_t;
 use libc::sockaddr;
 use libc::ssize_t;
-use libc::timespec;
 
 #[cfg(not(windows))]
 use libc::AF_INET;
@@ -725,7 +725,7 @@ pub struct SendInfo {
     to: sockaddr_storage,
     to_len: socklen_t,
 
-    at: timespec,
+    in_nanos: u64,
 }
 
 #[no_mangle]
@@ -743,7 +743,7 @@ pub extern fn quiche_conn_send(
             out_info.from_len = std_addr_to_c(&info.from, &mut out_info.from);
             out_info.to_len = std_addr_to_c(&info.to, &mut out_info.to);
 
-            std_time_to_c(&info.at, &mut out_info.at);
+            out_info.in_nanos = info.at.duration_since(time::Instant::now()).as_nanos() as u64;
 
             v as ssize_t
         },
@@ -1465,20 +1465,6 @@ fn std_addr_to_c(addr: &SocketAddr, out: &mut sockaddr_storage) -> socklen_t {
             sa_len as socklen_t
         },
     }
-}
-
-#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "windows")))]
-fn std_time_to_c(time: &std::time::Instant, out: &mut timespec) {
-    unsafe {
-        ptr::copy_nonoverlapping(time as *const _ as *const timespec, out, 1)
-    }
-}
-
-#[cfg(any(target_os = "macos", target_os = "ios", target_os = "windows"))]
-fn std_time_to_c(_time: &std::time::Instant, out: &mut timespec) {
-    // TODO: implement Instant conversion for systems that don't use timespec.
-    out.tv_sec = 0;
-    out.tv_nsec = 0;
 }
 
 #[cfg(test)]


quiche/src/ffi.rs
```diff
@@ -1467,20 +1467,6 @@ fn std_addr_to_c(addr: &SocketAddr, out: &mut sockaddr_storage) -> socklen_t {
...
-        ptr::copy_nonoverlapping(time as *const _ as *const timespec, out, 1)
-    }
-}
-
-#[cfg(any(target_os = "macos", target_os = "ios", target_os = "windows"))]
-fn std_time_to_c(_time: &std::time::Instant, out: &mut timespec) {
-    // TODO: implement Instant conversion for systems that don't use timespec.
-    out.tv_sec = 0;
-    out.tv_nsec = 0;
-}
```
FranzBusch, line 1482: A nice side effect of this change is that we now can implement pacing in Darwin & Windows libraries as well through the FFI interface

quiche/src/ffi.rs
```diff
@@ -743,7 +743,7 @@ pub extern fn quiche_conn_send(
             out_info.from_len = std_addr_to_c(&info.from, &mut out_info.from);
             out_info.to_len = std_addr_to_c(&info.to, &mut out_info.to);
 
-            std_time_to_c(&info.at, &mut out_info.at);
+            out_info.in_nanos = info.at.duration_since(time::Instant::now()).as_nanos() as u64;
```
normanmaurer, line 746: We could also just calculate the nanos (as you did here) and fill the `timespec` struct by hand... This would at least not break the API. WDYT ?

normanmaurer: most likely @ghedo has some opinions here :)

FranzBusch: I thought about this as well. I decided against this because it seems unnecessary to construct that timespec to just bridge over. We also would need to get the current time spec and then add the calculated nanos. IMO, that is all not worth the hassle and just returning nanos as we do with the timeout APIs works. Happy to hear more opinions on this!

normanmaurer: Yeah nanos are fine for me... Just wanted to bring up the suggestion :)

FranzBusch:
> Returning the amount of nanos sounds like a step backwards, the caller has to convert it to a timestamp at the earliest point possible anyway, because code that runs will otherwise affect the timing itself.

I understand your position here but there is no way to convert an `Instant` safely to a `timespec` as far as I know since the `Instant` type is opaque. We could change the internals of Quiche away from `Instant` and to something else but that just feels wrong. Furthermore, going to nanos here aligns it very nicely with the timeout methods where users also have to convert it to a point in time.

tbu-:
> but there is no way to convert an `Instant` safely to a `timespec` as far as I know since the `Instant` type is opaque.

I also couldn't find such a way, unfortunately.



> Furthermore, going to nanos here aligns it very nicely with the timeout methods where users also have to convert it to a point in time.

The Rust API recently started providing a `timeout_instant` method. I guess in order to address this issue.

ehaydenr:
Curious what folks think about changing `SendInfo` to not use `Instant` in the first place: #1409. `SendInfo` would contain the amount of time after "now" (whenever the SendInfo was generated) that the packet should be sent. This avoids FFI issues and allows the app to use whatever clock it wants

tbu-:
> Curious what folks think about changing `SendInfo` to not use `Instant` in the first place: #1409. `SendInfo` would contain the amount of time after "now" (whenever the SendInfo was generated) that the packet should be sent. This avoids FFI issues and allows the app to use whatever clock it wants

As I said above, I think this would a step backward. In the FFI case, it seems unavoidable, but in Rust it isn't. You can already use whatever clock you want by calculating `other_timekeeping::now() + (send_info.send_at - Instant::now())`, but by returning an `Instant`, it does not matter how long you take to process it, it's valid forever.

normanmaurer:
Did we have any "outcome" here ? Without a fix we cant implement pacing when using the C api

tbu-:
Another valid way to solve this is to remove the `at` field altogether and let quiche do the pacing. The API is already there, quiche can communicate timeouts to the user of the API. What's the advantage of the `at` field? Does it help because there's kernel support on Linux?

ghedo:
FTR, we just merged https://github.com/cloudflare/quiche/pull/1483 which should fix the immediate problem of invalid `timespec` values. That said, I think it would still be useful to move away from `Instant`, though it should be on both the Rust as well as C sides (e.g. using `Duration` in the Rust API) like what Eric was doing in https://github.com/cloudflare/quiche/pull/1409 (though not sure what the status of that PR is), but need to think about what the consequences will be.

FranzBusch:
Thanks for the update. Going to close this PR then

PR #1414, opened 2023-02-14
implement edge-triggered stream pseudo-iterators

ghedo:
Instead of cloning and returning a list of readable/writable streams, the new `stream_readable_next()` and `stream_writable_next()` return the first stream in the corresponding readable or writable queue, and remove it from said queue.

The end result is that each stream is only reported once, rather than being continuously reported until it becomes unreadable or unwritable, and as a consequence, applications will need to track partial read or written streams on their own.

In addition, a "low send watermark" is implemented through the existing `stream_writable()` method. When calling `stream_writable()`, if the stream doesn't at that time have enough capacity for the specified amount of bytes (i.e. `Ok(false)` is returned), then that stream will only be marked as writable again once its capacity grows enough to accomodate the amount of bytes requested.

This can be used in conjuction with `stream_writable_next()` to prevent streams from constantly being reported as writable, even though they don't have enough useful capacity. This is done automatically by the HTTP/3 implementation for example to ensure that request streams aren't constantly reported as writable until they have enough capacity for sending a HEADERS frame atomically.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4352,13 +4352,14 @@ impl Connection {
             self.blocked_limit = Some(self.max_tx_data);
         }
 
+        let cap = self.tx_cap;
+
         // Truncate the input buffer based on the connection's send capacity if
         // necessary.
         //
         // When the cap is zero, the method returns Ok(0) *only* when the passed
         // buffer is empty. We return Error::Done otherwise.
-        let cap = self.tx_cap;
-        if cap == 0 && !(fin && buf.is_empty()) {
+        if cap == 0 && !buf.is_empty() {
             return Err(Error::Done);
         }
 
@@ -4578,6 +4579,26 @@ impl Connection {
         Err(Error::InvalidStreamState(stream_id))
     }
 
+    /// Returns the next stream that has data to read.
+    ///
+    /// Note that once returned by this method, a stream ID will not be returned
+    /// again until it is "re-armed".
+    ///
+    /// The application will need to read all of the pending data on the stream,
+    /// and new data has to be received before the stream is reported again.
+    ///
+    /// This is unlike the [`readable()`] method, that returns the same list of
+    /// readable streams when called multiple times in succession.
+    ///
+    /// [`readable()`]: struct.Connection.html#method.readable
+    pub fn stream_readable_next(&mut self) -> Option<u64> {
+        let &stream_id = self.streams.readable.iter().next()?;
+
+        self.streams.mark_readable(stream_id, false);
+
+        Some(stream_id)
+    }
+
     /// Returns true if the stream has data that can be read.
     pub fn stream_readable(&self, stream_id: u64) -> bool {
         let stream = match self.streams.get(stream_id) {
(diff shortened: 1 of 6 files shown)


quiche/src/lib.rs
```diff
@@ -4600,6 +4601,28 @@ impl Connection {
...
+    ///
+    /// The application will need to read all of the pending data on the stream,
+    /// and new data has to be received before the stream is reported again.
+    ///
+    /// This is unlike the [`readable()`] method, that returns the same list of
+    /// readable streams when called multiple times in succession.
+    ///
+    /// [`readable()`]: struct.Connection.html#method.readable
+    pub fn stream_readable_next(&mut self) -> Option<u64> {
+        if let Some(&stream_id) = self.streams.readable.iter().next() {
```
ehaydenr, line 4617: consider early return syntax
```
let stream_id = self.streams.readable.iter().next()?;
```

quiche/src/lib.rs
```diff
@@ -4611,6 +4634,50 @@ impl Connection {
...
+    /// [`writable()`]: struct.Connection.html#method.writable
+    pub fn stream_writable_next(&mut self) -> Option<u64> {
+        // If there is not enough connection-level send capacity, none of the
+        // streams are writable.
+        if self.tx_cap == 0 {
+            return None;
+        }
+
+        for &stream_id in &self.streams.writable {
+            if let Some(stream) = self.streams.get(stream_id) {
```
ehaydenr, line 4658: if the stream isn't in the stream map, should we mark it not writable?

quiche/src/lib.rs
ghedo: Fixed, thanks!

quiche/src/lib.rs
ghedo: The problem is we can't remove elements from `self.streams.writable` while iterating it at the same time.

ehaydenr: Would it make sense to add a debug_assert then?

quiche/src/lib.rs
```diff
@@ -4374,13 +4374,14 @@ impl Connection {
...
+        let cap = self.tx_cap;
+
         // Truncate the input buffer based on the connection's send capacity if
         // necessary.
         //
         // When the cap is zero, the method returns Ok(0) *only* when the passed
         // buffer is empty. We return Error::Done otherwise.
-        let cap = self.tx_cap;
-        if cap == 0 && !(fin && buf.is_empty()) {
+        if cap == 0 && !buf.is_empty() {
```
LPardue, line 4384: Why is this line changed? It would seem to change behaviour compared to what the docs say.

ghedo: There is no actual behaviour change from the application POV. Also, the documentation talks about "0-length" buffers, not "0-length and final" ones when returning `Ok(0)` instead of `Done`, so it seems like this aligns better with that anyway.

In any case this is required to make sure the state of the stream is updated properly when doing 0-length writes (e.g. https://github.com/cloudflare/quiche/blob/master/quiche/src/h3/mod.rs#L1009), otherwise it doesn't get woken up later.

ghedo: That is, on `stream_send()` calls with 0-length buffers this makes it go a bit further in the code rather than return early, but the end result is the same.

ghedo: To add more details, the stream state wouldn't actually be created when doing a 0-length write because we would return early, so the code in the h3 code was useless. It wasn't a problem before, but now we actually need the state so that the stream can be marked as writable when it gets more capacity.

LPardue: cheers for the explanation, makes sense

LPardue (approved):

PR #1416, opened 2023-02-16
Don't send ACK if planning to send a PING

vkrasnov:
Because the ACK does not count towards the congestion window and the PING does, there can be a pathological case when the window is full then ACKs are emmited but PINGs are not, and quiche gets into an inifinite loop of sending ACKs, since the `ack_elicit_required` remains true, as no ack eliciting frame goes out.


ehaydenr:
looks like I stumbled across the same thing in #1418 and only seeing your PR now

ghedo:
Replaced by https://github.com/cloudflare/quiche/pull/1419 (but thanks for the patch, and for finding the bug!).

PR #1418, opened 2023-02-20
quic: don't optimistically bundle ACK when ack-eliciting required

ehaydenr:
MAX_OUTSTANDING_NON_ACK_ELICITING was introduced to avoid the sent queue from getting too large and to solicit additional RTT samples. As part of that change, we optimistically bundled an ACK frame if we intended on including a PING if no other ACK-eliciting frame was included in the packet. This was "optimistic" because there was no check for whether or not the congestion window would be too full for the PING frame.

When ACK-only packets that ignore congestion window were introduced, the MAX_OUTSTANDING_NON_ACK_ELICITING case wasn't covered. If the number of outstanding non-ack-eliciting packets exceeded the maximum and the congestion window was full, the connection would produce ACK-only packets whenever `Connection::send` was called. If `Connection::send` is called until it returns `Error::Done`, it will be called forever. This may lead to packet send surges, massive sent queues, or even an infinite loop depending on the application.

#1312 and #1385

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3318,13 +3318,9 @@ impl Connection {
         let left_before_packing_ack_frame = left;
 
         // Create ACK frame.
-        //
-        // If we think we may explicitly elicit an ACK via PING later, go ahead
-        // and generate an ACK (if there's anything to ACK) since we're going to
-        // send a packet with PING anyways - even if we haven't received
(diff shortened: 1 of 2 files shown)


ghedo:
Replaced by https://github.com/cloudflare/quiche/pull/1419 (but thanks anyway!).

PR #1419, opened 2023-02-21
only send opportunistic ACK if there is enough cwnd for PING

ghedo:
Because the ACK does not count towards the congestion window and the PING does, there can be a pathological case when the window is full then ACKs are emmited but PINGs are not, and quiche gets into an inifinite loop of sending ACKs, since the ack_elicit_required remains true, as no ack eliciting frame goes out.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3307,14 +3307,17 @@ impl Connection {
 
         let payload_offset = b.off();
 
+        let cwnd_available =
+            path.recovery.cwnd_available().saturating_sub(overhead);
+
         let left_before_packing_ack_frame = left;
 
         // Create ACK frame.
         //
-        // If we think we may explicitly elicit an ACK via PING later, go ahead
-        // and generate an ACK (if there's anything to ACK) since we're going to
-        // send a packet with PING anyways - even if we haven't received
-        // anything ACK eliciting.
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -3336,21 +3338,25 @@ impl Connection {
...
             };
 
-            // ACK-only packets are not congestion controlled so ACKs must be
-            // bundled considering the buffer capacity only, and not
-            // the available cwnd.
-            if push_frame_to_pkt!(b, frames, frame, left) {
-                pkt_space.ack_elicited = false;
+            // When a PING frame needs to be sent, avoid sending the ACK if
+            // there is not enough cwnd available for both.
+            if !ack_elicit_required || frame.wire_len() < cwnd_available {
```
ehaydenr, line 3343: should this be ACK frame wire length + PING frame wire length?

ghedo: Ok, updated.

LPardue:
I suspect that this code will be made complicated when we add support for ack-frequency. But since that's a future problem to solve, this LGTM.

LPardue (approved):

ghedo:
Will look into adding more tests later, but merging this now so that we are unblocked by the bug.

PR #1421, opened 2023-02-23
h3: better estimate QPACK encoded output size

LPardue:
The HTTP/3 and QPACK encoding process relies the HTTP/3 layer creating
an output buffer and passing this to the QPACK layer for it to encode
into. Previously, the calculation of the buffer size used a simple
approach for estimating the correct size of the buffer. In some cases
this estimate could be wrong, for instance when Huffman-encoding a
literal value in the low or extended ASCII ranges. This could lead to an
underestimation of the buffer size, which cause the QPACK encode to fail
leading to an h3::InternalError that would bubble up to an application
attempting to send requests or responses.

With this change, we move the check of a header's static representation
type earlier in the encode process, and use the type to provide a better
estimate of the required buffer sizes. The additional upshot is that
fully-indexed types (where the name and value are both in the static table)
require less storage and we no longer over estimate.

### DIFF
--- a/quiche/src/h3/qpack/mod.rs
+++ b/quiche/src/h3/qpack/mod.rs
@@ -98,8 +98,11 @@ mod tests {
             h3::Header::new(b":method", b"GET"),
         ];
 
+        let lookups: Vec<encoder::StaticLookup> =
+            headers.iter().map(|h| encoder::lookup_static(h)).collect();
+
         let mut enc = Encoder::new();
-        assert_eq!(enc.encode(&headers, &mut encoded), Ok(240));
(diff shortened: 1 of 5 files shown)


quiche/src/h3/qpack/mod.rs
```diff
@@ -143,20 +151,73 @@ mod tests {
...









")];
```
LPardue, line 182: note: this looks like an empty string, but it is actually the <RS> character `` I thought Github fixed that after https://www.bleepingcomputer.com/news/security/invisible-characters-could-be-hiding-backdoors-in-your-javascript-code/ but guess not

LPardue:
Closed in favour of https://github.com/cloudflare/quiche/pull/1423

PR #1422, opened 2023-02-23
Downgrade epoch to Initial if remote peer can't decrypt yet

normanmaurer:
Motivation:

We should downgrade the epoch to Initial if the remote peer cant decrypt yet, failing to do so might cause the remote peer to drop the frames on the floor and trying to resend.

Modifications:

Check if we need to downgrade to Intial and if so do it.

Result:

Remote peer can always handle frame

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -6224,10 +6224,22 @@ impl Connection {
                 crypto::Level::OneRTT => packet::Epoch::Application,
             };
 
-            if epoch == packet::Epoch::Application && !self.is_established() {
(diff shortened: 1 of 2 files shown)


normanmaurer:
Will work on a unit test

normanmaurer:
@ghedo this has a unit test now which failed before the change. I used the `#[cfg(test)]` macro to expose stuff only for testing. If you have an other preference on how to do this let me know.

ghedo (approved):
Change lgtm overall, but I refactored the code slightly.

ghedo:
Merged, thanks @normanmaurer!

PR #1425, opened 2023-02-27
stream: delete unreachable logic

wangfuyu:
if buf.fin() && buf.is_empty() && self.fin_off.is_some() true, then self.fin_off.is_some() && buf.is_empty() MUST be true, which means that we should return Ok(()) before this command.


ghedo:
Merged, thanks @wangfuyu!

Smithad499:
The application meets requirements for listing, Thank you.

PR #1429, opened 2023-03-02
docs: Clarify using Config, especially params that can be zero

### LINKED ISSUE #1250: Clarify which configuration options are mandatory
Using `quiche::Config::default()` will almost never work. At least the following options have to be set:
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_streams_bidi
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_streams_uni
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_data
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_stream_data_bidi_local
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_stream_data_bidi_remote
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_initial_max_stream_data_uni

In most cases, the ALPN and certificates have to be set as well:
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.set_application_protos
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.load_verify_locations_from_file (for clients)
- https://docs.rs/quiche/latest/quiche/struct.Config.html#method.load_cert_chain_from_pem_file / https://docs.rs/quiche/latest/quiche/struct.Config.html#method.load_priv_key_from_pem_file (for servers)

It would be great to document these prominently somewhere, as well as clarify the existing documentation which says
> When set to a non-zero value quiche will only allow at most v bytes of incoming stream data to be buffered for each locally-initiated bidirectional stream.
> The default value is 0.

That misleadingly implies that quiche will allow an unlimited number of bytes when it's set to 0, which is not the case; instead it allows no bytes.

LPardue:
Closes #1250

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,7 @@ The first step in establishing a QUIC connection using quiche is creating a
 let mut config = quiche::Config::new(quiche::PROTOCOL_VERSION)?;
 config.set_application_protos(&[b"h3"]);
 
-// additional configuration for application or use case
-
+// Additional configuration specific to application and use case...
 ```
 
 The [`Config`] object controls important aspects of the QUIC connection such as
--- a/README.md
+++ b/README.md
@@ -81,14 +81,15 @@ config.set_application_protos(&[b"example-proto"]);
 // Additional configuration specific to application and use case...
 ```
 
-The [`Config`] object controls important aspects of the QUIC connection such as
-QUIC version, ALPN IDs, flow control, congestion control, idle timeout and other
-properties or features.
+The [`Config`] object controls important aspects of the QUIC connection such
+as QUIC version, ALPN IDs, flow control, congestion control, idle timeout
+and other properties or features.
 
-QUIC is a general-purpose transport protocol and there are several configuration
-properties where there is no reasonable default value. For example, the
-permitted number of concurrent streams of any particular type is dependent on
-the application running over QUIC, and other use-case specific concerns.
+QUIC is a general-purpose transport protocol and there are several
+configuration properties where there is no reasonable default value. For
+example, the permitted number of concurrent streams of any particular type
+is dependent on the application running over QUIC, and other use-case
+specific concerns.
 
 quiche defaults several properties to zero, applications most likely need
 to set these to something else to satisfy their needs using the following:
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -949,6 +953,10 @@ impl Config {
     /// application) and will allow more data to be received as the buffer is
     /// consumed by the application.
     ///
+    /// When set to zero, either explicitly or via the default, quiche will not
+    /// give any flow control to the peer, preventing if from sending any stream
```
jyn514, line 957:
```suggestion
    /// give any flow control to the peer, preventing it from sending any stream
```
(and the same for all the following comments)

LPardue: derp, thanks

jyn514 (approved):

README.md
```diff
@@ -69,17 +69,45 @@ production)
...
+### Configuring connections
 
 The first step in establishing a QUIC connection using quiche is creating a
-configuration object:
+[`Config`] object:
 
 ```rust
-let config = quiche::Config::new(quiche::PROTOCOL_VERSION)?;
+let mut config = quiche::Config::new(quiche::PROTOCOL_VERSION)?;
+config.set_application_protos(&[b"h3"]);
```
ghedo, line 79: This overview is supposed to be somewhat generic, so using the h3 ALPN here makes it sound like everyone has to use it regardless of what protocol runs on top of QUIC. Maybe change it to `example-proto` or something?

README.md
```diff
@@ -69,17 +69,45 @@ production)
...
+
+// additional configuration for application or use case
```
ghedo, line 81:
```suggestion
// Additional configuration specific to application and use case...
```

README.md
```diff
@@ -69,17 +69,45 @@ production)
...
+
```
ghedo, line 82:
```suggestion
```

ghedo (changes requested):

README.md
LPardue: yep

ghedo (approved):

PR #1430, opened 2023-03-03
ranges: simplify removing first element of list

ghedo:
This API used to require an unstable feature, but it's now been stabilized.


LPardue (approved):
Strong & stable

PR #1431, opened 2023-03-03
fix #1424 -- pn are now encoded in 3 bytes when 2^16 <= pn < 2^24-1

### LINKED ISSUE #1424: [bug?] Clarifying pkt_num encoding/decoding
https://github.com/cloudflare/quiche/blob/7d8c82340e45ea80009ac5511523710f6c85163b/quiche/src/packet.rs#L714-L730

If I didn't read it wrong, packet numbers should be 1 to 4 bytes long according to the RFC , but it looks like this code will never correctly encode 3-bytes long packet numbers, since `pkt_num_len` never returns 3.

The bug should not lead to any issue in between two quiche endpoints. However, this should lead quiche to read 4 bytes instead of 3 against another QUIC implementation since `pkt_num_len` is used in the receive codepath as well, which *should* lead to a decryption error of a valid packet.

frochet:


LPardue (dismissed):
LGTM

quiche/src/packet.rs
```diff
@@ -560,6 +560,9 @@ pub fn pkt_num_len(pn: u64) -> Result<usize> {
         1
     } else if pn < u64::from(u16::MAX) {
         2
+
```
ghedo, line 563:
```suggestion
```

ghedo (changes requested):

ghedo (approved):

ghedo:
Merged, thanks @frochet!

PR #1432, opened 2023-03-06
make dgram peek infallible

ehaydenr:


LPardue:
This doesn't seem an unreasonable change but we might rework the API to better support multiple datagram queues, in which case we'd probably redesign things anyway and fold this into that in one step.

PR #1441, opened 2023-03-11
Remove code duplication around dgram sending

tbu-:


tbu-:
This PR is now over a year old. I'd rebase it if there's a chance that someone takes a look at it.

PR #1442, opened 2023-03-13
replace deprecated number constants by built-in primitive type constants

cjd0901:
[std::u8](https://doc.rust-lang.org/stable/std/u8/index.html) has been deprecated in the standard library, as have [std::u16](https://doc.rust-lang.org/stable/std/u16/index.html), [std::u32](https://doc.rust-lang.org/stable/std/u32/index.html), [std::u64](https://doc.rust-lang.org/stable/std/u64/index.html), [std::usize](https://doc.rust-lang.org/stable/std/usize/index.html), etc. So I replaced these deprecated constants by built-in primitive type constants. Then these are all I can find, please add if there is something missing.


ghedo (approved):

ghedo:
Merged, thanks @cjd0901!

PR #1445, opened 2023-03-16
optimize some SendBuf logic

wangfuyu:
If data is empty, we don't need to call ack_off()

### DIFF
--- a/quiche/src/stream.rs
+++ b/quiche/src/stream.rs
@@ -1320,11 +1320,6 @@ impl SendBuf {
             self.fin_off = Some(max_off);
         }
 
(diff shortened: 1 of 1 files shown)


quiche/src/stream.rs
```diff
@@ -1192,11 +1192,6 @@ impl SendBuf {
             self.fin_off = Some(max_off);
         }
 
-        // Don't queue data that was already fully acked.
-        if self.ack_off() >= max_off {
-            return Ok(data.len());
-        }
```
wangfuyu, line 1198: @ghedo I think here would never happen, right? SendBuf.write only be called by application.

PR #1446, opened 2023-03-16
Update checking for delivery rate app_limited flag correctly

lohith-bellad:
Delivery rate app_limited flag should be updated based on checking buffered data size in send buffer.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1261,6 +1261,9 @@ pub struct Connection {
     /// Number of stream data bytes that can be buffered.
     tx_cap: usize,
 
+    // Number of bytes buffered in the send buffer.
+    tx_buffered: usize,
+
     /// Total number of bytes sent to the peer.
     tx_data: u64,
 
@@ -1738,6 +1741,8 @@ impl Connection {
             almost_full: false,
 
             tx_cap: 0,
+
+            tx_buffered: 0,
 
             tx_data: 0,
             max_tx_data: 0,
@@ -2694,6 +2699,9 @@ impl Connection {
                         };
 
                         stream.send.ack_and_drop(offset, length);
+
+                        self.tx_buffered =
+                            self.tx_buffered.saturating_sub(length);
 
                         qlog_with_type!(QLOG_DATA_MV, self.qlog, q, {
                             let ev_data = EventData::DataMoved(
@@ -4502,6 +4510,8 @@ impl Connection {
 
         self.tx_data += sent as u64;
 
+        self.tx_buffered += sent;
+
         qlog_with_type!(QLOG_DATA_MV, self.qlog, q, {
             let ev_data = EventData::DataMoved(qlog::events::quic::DataMoved {
                 stream_id: Some(stream_id),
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -6453,17 +6460,21 @@ impl Connection {
...
 
-                    let (lost_packets, lost_bytes) = p.recovery.on_ack_received(
-                        &ranges,
-                        ack_delay,
-                        epoch,
-                        handshake_status,
-                        now,
-                        &self.trace_id,
-                    )?;
+                    let (lost_packets, lost_bytes, acked_bytes) =
```
ghedo, line 6463: The counter is increased by the number of stream data bytes buffered, but this counts _all_ bytes of a packet (including other frames, packet headers, ...) so it's always going to decrease the counter by more bytes than were actually buffered.

Instead I think the counter needs to be decreased only when STREAM frames specifically are acked, and only decrease by the actual amount of stream data that was acked, see https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L2689

ghedo (changes requested):

ghedo:
Also, I think we need to do the same thing for CRYPTO frames as well, otherwise the connection will always be app_limited during the handshake?

quiche/src/lib.rs
ghedo: I think for DATAGRAMs it's easier because we already have the total queue size, but yeah, I guess the check in `delivery_rate_check_if_app_limited()` should use something like `self.tx_buffered + self.dgram_send_queue.byte_size()`.

lohith-bellad:
> 

I think CRYPTO frames are fine, even though during handshake if conn is aapp_limited it will not make BBR reduce cwnd or pacing rate.

quiche/src/lib.rs
```diff
@@ -6453,17 +6460,21 @@ impl Connection {
...
+                            handshake_status,
+                            now,
+                            &self.trace_id,
+                        )?;
 
                     self.lost_count += lost_packets;
                     self.lost_bytes += lost_bytes as u64;
+
+                    self.tx_buffered =
+                        self.tx_buffered.saturating_sub(acked_bytes);
```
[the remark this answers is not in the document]
lohith-bellad: ah yes, sure will do it.

quiche/src/lib.rs
lohith-bellad: thanks will address these things.

quiche/src/recovery/mod.rs
```diff
@@ -420,7 +420,7 @@ impl Recovery {
         &mut self, ranges: &ranges::RangeSet, ack_delay: u64,
         epoch: packet::Epoch, handshake_status: HandshakeStatus, now: Instant,
         trace_id: &str,
-    ) -> Result<(usize, usize)> {
+    ) -> Result<(usize, usize, usize)> {
```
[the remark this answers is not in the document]
lohith-bellad: not using acked_bytes from on_ack_received()

Issue #1273, opened 2022-07-04
Error handling key updates?


adewes:
When sending data through a QUIC stream the connection always fails at the 99.999th packet received, giving me "dropped invalid packet" errors. It seems this is due to a [key update](https://github.com/lucas-clemente/quic-go/commit/a09c0453244227768bbb9bba96028eeabf6dcc10) initiated by the server (based on `quic-go`). I couldn't figure out how to fix this, is Quiche maybe not handling key updates correctly in this case? I use the following client configuration:

```rust
config.set_max_idle_timeout(300000);
config.set_max_recv_udp_payload_size(1350);
config.set_max_send_udp_payload_size(1350);
config.set_initial_max_data(10_000_000);
config.set_initial_max_stream_data_bidi_local(10_000_000);
config.set_initial_max_stream_data_bidi_remote(10_000_000);
config.set_initial_max_streams_bidi(1000);
config.set_initial_max_streams_uni(1000);
config.set_disable_active_migration(false);
```

Here's the log output leading to the error. After receiving the invalid packets the connection eventually times out. Log entries with name "xxxxxxxxxxxxx" are from my client code.

```
[2022-07-04T13:03:49Z TRACE xxxxxxxxxxxxx] sending 49 bytes out...
[2022-07-04T13:03:49Z TRACE xxxxxxxxxxxxx] written 49
[2022-07-04T13:03:49Z TRACE xxxxxxxxxxxxx] received command...
[2022-07-04T13:03:49Z TRACE xxxxxxxxxxxxx] sending 16 bytes on stream 20
[2022-07-04T13:03:49Z TRACE quiche] 0000000000000000000000000000000000000000 tx pkt Short dcid=22bd8f18 key_phase=false len=24 pn=100005
[2022-07-04T13:03:49Z INFO  xxxxxxxxxxxxx::tests] Echoing 16 TCP bytes...
[2022-07-04T13:03:49Z TRACE quiche] 0000000000000000000000000000000000000000 tx frm STREAM id=20 off=147148 len=16 fin=false
[2022-07-04T13:03:49Z INFO  xxxxxxxxxxxxx::tests] Reading data for echo...
[...]

PR #1456, opened 2023-03-25
handle peer-initiated key update

### LINKED ISSUE #1115: Connections stop working with AES-ciphers because of missing Key Update

ghedo:
This adds support for handling key update initiated by a peer. Based on the changes from #1121 with some additional modifications.

In addition to the included test case, this was also tested using the QUIC interop runner.

Fixes #1115, #1273, #1121.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -560,6 +560,9 @@ pub enum Error {
 
     /// Not enough available identifiers.
     OutOfIdentifiers,
+
+    /// Error in key update.
+    KeyUpdate,
 }
 
 impl Error {
@@ -572,6 +575,7 @@ impl Error {
             Error::FlowControl => 0x3,
             Error::StreamLimit => 0x4,
             Error::FinalSize => 0x6,
+            Error::KeyUpdate => 0xe,
             _ => 0xa,
         }
     }
@@ -597,6 +601,7 @@ impl Error {
             Error::StreamReset { .. } => -16,
             Error::IdLimit => -17,
             Error::OutOfIdentifiers => -18,
+            Error::KeyUpdate => -19,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -2532,6 +2580,67 @@ impl Connection {
...
+                crypto_open: open_prev,
+                pn_on_update: pn,
+                update_acked: false,
+                timer: now + (recv_path.recovery.pto() * 3),
+            });
+
+            self.key_phase = !self.key_phase;
+
+            let trigger = Some(
+                qlog::events::security::KeyUpdateOrRetiredTrigger::RemoteUpdate,
```
LPardue, line 2616: need to move this withing the qlog_with_type block or add an explicit feature check

quiche/src/lib.rs
```diff
@@ -9032,6 +9169,89 @@ mod tests {
...
+                .as_ref()
+                .unwrap()
+                .update_acked
+        );
+
+        // Server sends message with the new key
+        assert_eq!(pipe.server.stream_send(4, b"world", true), Ok(5));
+        assert_eq!(pipe.advance(), Ok(()));
+
+        // Ensure update key is completed and client can decrypts packet
```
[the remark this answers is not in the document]
ghedo: Fixed, thanks.

quiche/src/lib.rs
ghedo: Fixed thanks.

LPardue (approved):

PR #1470, opened 2023-03-30
Improved static qpack lookup

vkrasnov:
This change improves the QPACK static table lookup by avoiding iterating over all of the 99 entries for each header. Instead each lookup now consists of finding the table for the proper name length, then iterating over the possible names and if found iterating over the list of values for that name.

In the best case this will result in 0 iterations, in the worst case for an unmatched header it will result in 6 iterations for a name of length 7, and in 18 iteration for the worst succesful match for ":status: 500".

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -576,13 +576,17 @@ pub trait NameValue {
     fn value(&self) -> &[u8];
 }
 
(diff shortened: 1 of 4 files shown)


vkrasnov:
@LPardue I don't think so, that would require an allocation for each header

ghedo (changes requested):

ghedo (approved):

PR #1471, opened 2023-03-31
Faster Huffman encoding

vkrasnov:
Slightly better optimization of the Huffman encoding method, that emits bits in 8 byte chunks, instead of the current byte by byte implementation. The result is about twice as fast.


vkrasnov:
I wrote a short benchmark that encodes a set of actual headers I got from the cloudflare blog:
https://gist.github.com/rust-play/6dd13945e7eb2895ca0b62f85f5809d9

The results are:

Apple M1:

Master: 2.87 s
#1470 : 1.17 s
#1470  + this : 0.83 s

AMD EPYC 7713:

Master: 5.84
#1470: 1.84
#1470  + this: 1.28

So overall we are looking at about a 4x faster header encoding.

LPardue (approved):

ghedo (approved):

PR #1473, opened 2023-04-03
Fix BBR cwnd recovery logic and pacing

lohith-bellad:
Fix BBR cwnd recovery logic during congestion episode.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2950,10 +2950,6 @@ impl Connection {
 
         self.ack_eliciting_sent = false;
 
-        // Reset pacer and start a new burst when a valid
-        // packet is received.
-        self.paths.get_mut(recv_pid)?.recovery.pacer.reset(now);
-
         Ok(read)
     }
 
--- a/quiche/src/recovery/bbr/mod.rs
+++ b/quiche/src/recovery/bbr/mod.rs
@@ -255,12 +255,13 @@ impl State {
 fn bbr_enter_recovery(r: &mut Recovery, now: Instant) {
     r.bbr_state.prior_cwnd = per_ack::bbr_save_cwnd(r);
 
-    r.congestion_window = r.bytes_in_flight +
-        r.bbr_state.newly_acked_bytes.max(r.max_datagram_size);
+    r.congestion_window = r.bytes_in_flight.max(r.max_datagram_size);
     r.congestion_recovery_start_time = Some(now);
 
     r.bbr_state.packet_conservation = true;
     r.bbr_state.in_recovery = true;
+
+    r.bbr_state.newly_lost_bytes = 0;
 
     // Start round now.
     r.bbr_state.next_round_delivered = r.delivery_rate.delivered();
@@ -308,7 +309,7 @@ fn on_packets_acked(
     });
 
     if let Some(pkt) = packets.last() {
-        if !r.in_congestion_recovery(pkt.time_sent) {
+        if !r.in_congestion_recovery(pkt.time_sent) && r.bbr_state.in_recovery {
             // Upon exiting loss recovery.
             bbr_exit_recovery(r);
         }
--- a/quiche/src/recovery/pacer.rs
+++ b/quiche/src/recovery/pacer.rs
@@ -123,7 +123,7 @@ impl Pacer {
     }
 
     /// Resets the pacer for the next burst.
-    pub fn reset(&mut self, now: Instant) {
+    fn reset(&mut self, now: Instant) {
         self.used = 0;
 
         self.last_update = now;


ghedo (changes requested):
Added some comments as I'm not sure I understand what this does.

quiche/src/recovery/bbr/mod.rs
```diff
@@ -255,13 +255,14 @@ impl State {
 fn bbr_enter_recovery(r: &mut Recovery, now: Instant) {
     r.bbr_state.prior_cwnd = per_ack::bbr_save_cwnd(r);
 
-    r.congestion_window = r.bytes_in_flight +
-        r.bbr_state.newly_acked_bytes.max(r.max_datagram_size);
+    r.congestion_window = r.bytes_in_flight.max(r.max_datagram_size);
```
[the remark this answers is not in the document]
lohith-bellad: When we see ACKs, the sequence of CC operations called are congestion_event if there are losses and on_packets_acked. From congestion_event we get to know lost_bytes and from on_packets_acked we get to know acked_bytes. The acked_bytes addition to cwnd the above function was doing is https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/mod.rs#L300 which is from previous ACK episode. So I have removed that and now cwnd is set to in_flight and we we inflate the cwnd by acked_bytes in https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/per_ack.rs#L176 anyways.

quiche/src/recovery/pacer.rs
```diff
@@ -119,7 +122,11 @@ impl Pacer {
 
         self.capacity = capacity;
 
-        self.rate = rate;
+        if capacity > 32000 {
```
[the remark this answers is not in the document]
lohith-bellad: I was targeting mainly large GSO send batches. But actually we dont need that check.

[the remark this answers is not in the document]
lohith-bellad: updated  the patch

ghedo (approved):

quiche/src/lib.rs
```diff
@@ -2950,10 +2950,6 @@ impl Connection {
 
         self.ack_eliciting_sent = false;
 
-        // Reset pacer and start a new burst when a valid
-        // packet is received.
-        self.paths.get_mut(recv_pid)?.recovery.pacer.reset(now);
```
junhochoi, line 2955: Sorry for commenting already merged comment.

This was added (AFAICT) because app calls quiche recv() and send() sequentially,
recv() may process ACK and update congestion control status, might lead to pacing rate change. 

Also, this change also impacts other CC with pacing (e.g. cubic + pacing). I think this need to be
a separate change from this PR and verify it has no negative impact.

quiche/src/recovery/bbr/mod.rs
```diff
@@ -255,13 +255,14 @@ impl State {
...
     r.congestion_recovery_start_time = Some(now);
 
     r.bbr_state.packet_conservation = true;
     r.bbr_state.in_recovery = true;
 
+    r.bbr_state.newly_lost_bytes = 0;
```
junhochoi, line 264: `newly_lost_bytes` is used in [ack processing](https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/per_ack.rs#L164), which is called in [`on_packets_acked()` callback and cleared right after.] (https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/mod.rs#L320-L322)
If we clear it here, it won't be used I think

quiche/src/recovery/bbr/mod.rs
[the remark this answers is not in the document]
junhochoi: I see. It looks like linux or quiche does the calculation during ack processing.

To prevent the current issue (using previous acked_bytes), I think we can move the part calling `bbr_enter_recovery()` to `on_packets_acked()` and leave the cwnd calculation to follow the spec better.

quiche/src/lib.rs
lohith-bellad: Yes, recv() and send() are called  sequentially and recv() may ultimately lead to pacing rate changes which is taken care in pacer methods. If resetting pacer is needed then its called from pacer methods when CC updates the pacer. We dont need to reset it here I think. This has been tested in various RTT environments.

junhochoi: Since the pacer module assumes the current library behavior (https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/pacer.rs#L32-L34), the pacer status need to be cleared before we prepare a new series of send(). Without it, a small send().. -> recv() (let's say it's not ack) -> send().. may skew the pacer, because 2nd send() will continue from the previous pacer.used which is not an intended behavior. Not sure this was a concern for you.

Note that we already have some protection for [such case](https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/pacer.rs#L177), this may not be noticeable.

quiche/src/recovery/bbr/mod.rs
junhochoi: Probably we need to understand what tcp (or other) does. their in_flight is including lost packets already? (if not, the algorithm is correct)

I understand lost_bytes can be used twice. If so, instead of doing L264, we need to remove https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/per_ack.rs#L166-L171 . lost_bytes is used in other places, so I don't think just clearing up here is not a good idea. I'll update my PR (#1478) to reflect that.

lohith-bellad: I meant to say, lost_bytes are counted twice only when we enter the recovery episode, not when we are in recovery. 
https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/per_ack.rs#L166-L171 is needed because we may see losses while we are in recovery.

quiche/src/lib.rs
lohith-bellad: pacer.used is always cleared when we hit the capacity or see packet with size less than previous packet size to match GSO requirements. Are you saying if we end up in a situation where cwnd allows to send just one packet and pacer.used is not cleared in that case ?

junhochoi: Yes, e.g. send a full sized packet, recv, send a full size packet again. Hard to control from the app, but it could happen. (I've seen it)

quiche/src/recovery/bbr/mod.rs
junhochoi: Ahhh right... Then I think we need to add lost_bytes back in bbr_enter_recovery().
See if it makes sense:
https://github.com/cloudflare/quiche/pull/1478/files#diff-89fb0e46580c044508e7c118b4124b0f151e2a4437e66029fb897cdb18c50a12R263

lohith-bellad: The current is doing the same thing right ? Why do you want to add and then subtract ?

quiche/src/lib.rs
lohith-bellad: Ok got it, thanks for clarification.

quiche/src/recovery/bbr/mod.rs
lohith-bellad: When we are entering recovery lost_bytes is made zero since itss already calculated in in_flight. In recovery phase we are not making lost_bytes zero so it is counted.

PR #1474, opened 2023-04-04
Reuse the same Vec for newly_acked

vkrasnov:
Currently every time an ack arrives quiche allocates a new Vec to hold the newly acked packet descriptors. This is rather expensive, and can be avoided by reusing the same Vec. Sadly due to the way Recovery interface is implemented it can't be integrated in Recovery without significant rewrite, and has to be provided externally. Still the performance benefits are pretty significant.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1394,6 +1394,9 @@ pub struct Connection {
     /// Whether the connection should prevent from reusing destination
     /// Connection IDs when the peer migrates.
     disable_dcid_reuse: bool,
+
+    /// A resusable buffer used by Recovery
+    newly_acked: Vec<recovery::Acked>,
 }
 
 /// Creates a new server-side connection.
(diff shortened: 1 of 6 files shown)


quiche/src/lib.rs
```diff
@@ -1394,6 +1394,9 @@ pub struct Connection {
     /// Whether the connection should prevent from reusing destination
     /// Connection IDs when the peer migrates.
     disable_dcid_reuse: bool,
+
+    /// A resusable buffer used by Recovery
+    newly_acked: Vec<recovery::Acked>,
```
ghedo, line 1399: Why is this not in `Recovery` rather than `Connection`?

vkrasnov: I explained in the commit message. Because all recovery methods take &mut Recovery, you can't use a field in Recovery

ghedo (approved):

PR #1478, opened 2023-04-06
bbr: fix cwnd calculation when recovery starts

junhochoi:
Alternative to #1473. Also fixed the test which I believe it's correct.

Previously `bbr_enter_recovery()` used old `newly_acked_bytes` to set
a new cwnd when the recovery starts. To prevent this from happening,
`bbr_enter_recovery()` is called after `newly_acked_bytes` is updated
in a current ack processing. Also `bbr_modulate_cwnd_for_recovery()`
use the same cwnd calculation to match the same in `bbr_enter_recovery()`.

@lohith-bellad please feel free to take a look if it makes sense. I didn't test it much

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2955,6 +2955,10 @@ impl Connection {
 
         self.ack_eliciting_sent = false;
 
+        // Reset pacer and start a new burst when a valid
+        // packet is received.
+        self.paths.get_mut(recv_pid)?.recovery.pacer.reset(now);
+
         Ok(read)
     }
 
--- a/quiche/src/recovery/bbr/mod.rs
+++ b/quiche/src/recovery/bbr/mod.rs
@@ -255,13 +255,12 @@ impl State {
 fn bbr_enter_recovery(r: &mut Recovery, now: Instant) {
     r.bbr_state.prior_cwnd = per_ack::bbr_save_cwnd(r);
 
-    r.congestion_window = r.bytes_in_flight.max(r.max_datagram_size);
+    r.congestion_window = r.bytes_in_flight +
+        r.bbr_state.newly_acked_bytes.max(r.max_datagram_size);
     r.congestion_recovery_start_time = Some(now);
 
     r.bbr_state.packet_conservation = true;
     r.bbr_state.in_recovery = true;
-
-    r.bbr_state.newly_lost_bytes = 0;
 
     // Start round now.
     r.bbr_state.next_round_delivered = r.delivery_rate.delivered();
@@ -311,7 +310,7 @@ fn on_packets_acked(
         });
 
     if let Some(pkt) = packets.last() {
-        if !r.in_congestion_recovery(pkt.time_sent) && r.bbr_state.in_recovery {
+        if !r.in_congestion_recovery(pkt.time_sent) {
             // Upon exiting loss recovery.
             bbr_exit_recovery(r);
         }
--- a/quiche/src/recovery/bbr/mod.rs
+++ b/quiche/src/recovery/bbr/mod.rs
@@ -252,12 +252,15 @@ impl State {
 }
 
 // When entering the recovery episode.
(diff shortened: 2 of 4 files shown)


quiche/src/lib.rs
```diff
@@ -2955,6 +2955,10 @@ impl Connection {
 
         self.ack_eliciting_sent = false;
 
+        // Reset pacer and start a new burst when a valid
+        // packet is received.
+        self.paths.get_mut(recv_pid)?.recovery.pacer.reset(now);
+
```
lohith-bellad, line 2961: Why do you want to reset pacer here ?

lohith-bellad (changes requested):

quiche/src/lib.rs
lohith-bellad: https://github.com/cloudflare/quiche/pull/1473#discussion_r1160125794

quiche/src/recovery/bbr/mod.rs
```diff
@@ -310,9 +308,14 @@ fn on_packets_acked(
...
+    // Upon entering Fast Recovery.
+    if !r.bbr_state.in_recovery && r.congestion_recovery_start_time.is_some() {
+        bbr_enter_recovery(r);
+    }
+
     if let Some(pkt) = packets.last() {
-        if !r.in_congestion_recovery(pkt.time_sent) && r.bbr_state.in_recovery {
-            // Upon exiting loss recovery.
+        // Upon exiting loss recovery.
+        if !r.in_congestion_recovery(pkt.time_sent) {
```
lohith-bellad, line 318: This check leads to calling bbr_exit_recovery even though we are not in recovery!

quiche/src/lib.rs
junhochoi: See my comment in above PR.

quiche/src/recovery/bbr/mod.rs
junhochoi: I'd like to understand which case can create such condition. in_congestion_recovery() checks should return `false` during non-recovery phase.

junhochoi: Thought it again and it's correct.

lohith-bellad: Isn't we should call bbr_exit_recovery() only if we are in recovery phase? If so then above code will call bbr_exit_recovery() even  though you are not in recovery.

junhochoi: Oh, I mean I updated the code as you proposed originally. :)

quiche/src/recovery/bbr/mod.rs
```diff
@@ -252,17 +252,19 @@ impl State {
...
+fn bbr_enter_recovery(r: &mut Recovery) {
     r.bbr_state.prior_cwnd = per_ack::bbr_save_cwnd(r);
 
-    r.congestion_window = r.bytes_in_flight.max(r.max_datagram_size);
-    r.congestion_recovery_start_time = Some(now);
+    // Temporarily add lost bytes, because bytes_in_flight
+    // doesn't have lost bytes at this point. cwnd will be
+    // calculated again in bbr_modulate_cwnd_for_recovery().
+    r.congestion_window = r.bytes_in_flight +
+        r.bbr_state.newly_acked_bytes.max(r.max_datagram_size) +
```
lohith-bellad, line 262: The current code is doing the same thing. Since, we don't have the acked_bytes  at this point, cwnd is set correctly in bbr_modulate_cwnd_for_recovery() function. At any case bbr_modulate_cwnd_for_recovery is called after entering the recovery.

ghedo:
Thanks for the contribution. This PR changes the legacy BBR and pacer implementation that was removed by commit `bda63e05c` through #2297. The patch no longer applies to the active congestion-control implementation, so closing this PR as obsolete.

PR #1490, opened 2023-04-24
Rewrite congestion and recovery code

vkrasnov:
This is a significant change that replaces most of the congestion and recovery control implementation with an implementation that mostly ports the one from google/quiche. IMHO The change nicely breaks up congestion, recovery and pacing into separate, easily mainatinable and extensible pieces, and is also more idiomatic. The advantage of porting from google/quiche is that at least the cubic and pacing codes are very well tested, and require minimal review. The new pacing implementaion is also more accurate.


vkrasnov:
Survived by #1742

PR #1491, opened 2023-04-24
Vlad/circ buffer

vkrasnov:


quiche/src/stream.rs
```diff
@@ -1572,7 +1136,7 @@ impl RangeBuf {
     /// Creates a new `RangeBuf` from the given slice.
     pub fn from(buf: &[u8], off: u64, fin: bool) -> RangeBuf {
         RangeBuf {
-            data: Arc::new(Vec::from(buf)),
+            data: buf.to_vec().into(),
```
ehaydenr, line 1139: Thoughts on how this alloc/copy might be avoidable? I get we need something to track what hasn't been ACKed by the peer so we can retransmit if lost, but I wonder if there's a way to leave the data inside the SendBuf circular buffer instead (avoiding alloc+copy). I recognize this would complicate and grow the scope of SendBuf.

quiche/src/stream/send_buf.rs
```diff
@@ -0,0 +1,343 @@
...
+use crate::Result;
+
+/// Send-side stream buffer.
+///
+/// Stream data scheduled to be sent to the peer and is buffered in a circular
+/// buffer.
+#[derive(Debug, Default)]
+pub struct SendBuf {
+    /// A circular buffer that stores the data that should be sent to the peer
+    pub(super) data: VecDeque<u8>,
```
ehaydenr, line 15: I'm trying to understand the implications here. If I'm understanding correctly, previously, an app write would result in a RangeBuf. Now, we write directly to the VecDeque which will alloc when growth is necessary, but otherwise avoids the allocation. The allocation _does_ eventually happen when the RangeBuf is created to track what needs to be retransmitted. See my previous comment about how this might be avoided. As-is, are we just moving the alloc/copy to a later point in time?

ehaydenr: I'm now seeing #1399 which I think probably makes more sense

PR #1492, opened 2023-04-26
Flush keylog write straight away

gjcairo:
When debugging a QUIC packet exchange, enabling keylog allows to retrieve the encryption keys and set them up in, for example, Wireshark.

Realised today that keylog file was not being written until the connection was closed, which can be an inconvenience when debugging. Looking at the code, it seems that the keys are being written but not flushed out. A connection close/dealloc must be causing the flush currently, hence why the file is only written on connection close.

This PR adds a flush right after the write to make sure that the keylog file is made available as soon as possible.

### DIFF
--- a/quiche/src/tls.rs
+++ b/quiche/src/tls.rs
@@ -1106,6 +1106,7 @@ extern fn keylog(ssl: *mut SSL, line: *const c_char) {
         full_line.push(b'\n');
 
         keylog.write_all(&full_line[..]).ok();
+        keylog.flush().ok();
     }
 }
 


normanmaurer:
Nice one... Thanks @gjcairo

quiche/src/tls.rs
```diff
@@ -1106,6 +1106,7 @@ extern fn keylog(ssl: *mut SSL, line: *const c_char) {
         full_line.push(b'\n');
 
         keylog.write_all(&full_line[..]).ok();
+        keylog.flush().ok();
```
FranzBusch, line 1109: This is specifically required for the FFI layer because we are wrapping a `File` into a `BufWriter` which causes these writes not to be written out immediately without this `flush`.

ghedo (approved):

ghedo:
Merged, thanks @gjcairo!

FranzBusch:
@ghedo Thanks! Can we get a release tagged with this?

gjcairo:
Hey @ghedo , just a friendly ping to see if we can get another release tagged? It's been over a month since the last one and there have been 14 commits merged since, so maybe it's a good time.

PR #1493, opened 2023-04-27
refactoring to prepare multipath support without side effects

qdeconinck:
- packet number space map
- spaced packet number
- `PathEvent::Closed` now includes error code and reason
- function refactoring in lib.rs and frame.rs

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -719,7 +719,7 @@ fn handle_path_events(client: &mut Client) {
                 );
             },
 
-            quiche::PathEvent::Closed(local_addr, peer_addr) => {
+            quiche::PathEvent::Closed(local_addr, peer_addr, _err, _reason) => {
                 info!(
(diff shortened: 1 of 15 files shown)


apps/src/bin/quiche-server.rs
```diff
@@ -716,7 +716,7 @@ fn handle_path_events(client: &mut Client) {
                 );
             },
 
-            quiche::PathEvent::Closed(local_addr, peer_addr) => {
+            quiche::PathEvent::Closed(local_addr, peer_addr, _err, _reason) => {
                 info!(
                     "{} Path ({}, {}) is now closed and unusable",
```
LPardue, line 721: may as well print the `err` and `reason`

apps/src/client.rs
```diff
@@ -428,7 +428,7 @@ pub fn connect(
                     );
                 },
 
-                quiche::PathEvent::Closed(local_addr, peer_addr) => {
```
LPardue, line 431: may as well print the `err` and `reason`

PR #1500, opened 2023-05-02
WIP: test pinning serde_with to 2.3.2

LPardue:


LPardue:
The test proved it wasn't serde_with. Doctor Quincy moves on.

PR #1502, opened 2023-05-03
Adds Support for Preferred Addressing

Clendenin:
This PR adds support for Preferred Addressing as described in
https://datatracker.ietf.org/doc/html/rfc9000#name-servers-preferred-address

Based on this PR: https://github.com/cloudflare/quiche/pull/1208

Addresses issue: https://github.com/cloudflare/quiche/issues/828

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1735,10 +1735,12 @@ impl Connection {
             reset_token,
         );
 
-        // If we aren't using zero length connection id assign an SCID to the preferred address
-        // transport argument if one is used.
+        // If we aren't using zero length connection id assign an SCID to the
+        // preferred address transport argument if one is used.
         if !scid.is_empty() {
-            Self::assign_new_scid_to_preferred_address(config, is_server, &mut ids)?;
+            Self::assign_new_scid_to_preferred_address(
+                config, is_server, &mut ids,
+            )?;
         }
 
         let mut conn = Connection {
@@ -1937,7 +1939,9 @@ impl Connection {
     }
 
     /// Assign a new SCID to the preferred address transport argument
-    fn assign_new_scid_to_preferred_address(config: &mut Config, is_server: bool, ids: &mut ConnectionIdentifiers) -> Result<()> {
+    fn assign_new_scid_to_preferred_address(
+        config: &mut Config, is_server: bool, ids: &mut ConnectionIdentifiers,
+    ) -> Result<()> {
         // Only servers support sending a preferred address in the transport
         // paramters
         if !is_server &&
@@ -1965,7 +1969,7 @@ impl Connection {
             // This should be sequence number one by RFC
             debug_assert_eq!(seq, 1);
         }
-        return Ok(())
+        return Ok(());
     }
 
     /// Sets keylog output to the designated [`Writer`].
@@ -6400,7 +6404,8 @@ impl Connection {
             .set_source_conn_id_limit(peer_params.active_conn_id_limit);
 
         if let Some(preferred_address) = &peer_params.preferred_address_params {
-            // Client should not send preferred address argument, zero-length connection ID must not be sent. 
+            // Client should not send preferred address argument, zero-length
+            // connection ID must not be sent.
             if self.is_server || preferred_address.connection_id.is_empty() {
                 return Err(Error::InvalidTransportParam);
             }
@@ -8320,14 +8325,20 @@ pub mod testing {
                     &client_scid,
                     client_addr,
                     server_addr,
-                    &mut self.client_config.take().unwrap_or(PipeBuilder::default_test_config()?),
+                    &mut self
+                        .client_config
+                        .take()
+                        .unwrap_or(PipeBuilder::default_test_config()?),
                 )?,
                 server: accept(
                     &server_scid,
                     None,
                     server_addr,
                     client_addr,
-                    &mut self.server_config.take().unwrap_or(PipeBuilder::default_test_config()?),
+                    &mut self
+                        .server_config
+                        .take()
+                        .unwrap_or(PipeBuilder::default_test_config()?),
                 )?,
             })
         }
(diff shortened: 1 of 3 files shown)


LPardue:
There's a few clippy nits and nightly rustfmt nits (`cargo +nightly fmt`).

quiche/src/lib.rs
```diff
@@ -742,6 +751,26 @@ impl Config {
         Self::with_tls_ctx(version, tls::Context::new()?)
     }
 
+    /// Creates a default config object using commonly used values
+    /// Useful in testing
+    pub fn default() -> Result<Self> {
```
LPardue, line 756: We shouldn't do this, there is no sensible default for the QUIC transport because the values depend on the application protocol using QUIC. Making this a public function is dangerous as it might encourage bad practice. 

If you'd like a convenience or utility function to make the tests easier, please put a config creation function inside the test module

quiche/src/lib.rs
```diff
@@ -1710,13 +1758,39 @@ impl Connection {
...
+
+                // This should be sequence number one by RFC
+                debug_assert_eq!(seq, 1);
+            }
+        }
+
+        // Only servers support sending a preferred address in the transport paramters
+        if !is_server && config.local_transport_params.preferred_address_params.is_some() {
+            return Err(Error::InvalidTransportParam)
+        }
```
LPardue, line 1792: We can do this check inside the `if let Some block` above. Then it might be a good candidate to extract into its own function anyway.

quiche/src/lib.rs
```diff
@@ -6264,9 +6338,51 @@ impl Connection {
     pub fn is_server(&self) -> bool {
         self.is_server
     }
+    
+    /// Returns the server's preferred address which is a tuple with a
+    /// V4 socket address, a connection ID and a stateless reset token.
+    pub fn server_preferred_address_v4(
+        &self,
+    ) -> Result<Option<(SocketAddrV4, ConnectionId, u128)>> {
```
LPardue, line 6346: So I'm wondering what the utility of the Ok(None) is? 

I imagine a case where a client can call the `_v4` method, sees there isn't anything so it checks the `_v6` version. 

But maybe it would be easier to just return a copy of the parameter directly to the client and let it figure this all out (i.e. it can do its own pattern matching on the things it really cares about). In other words, just return `Option<PreferredAddressParams>`.

And I'm not too sure the value of ClientOnlyOperation error. If we just returned None to a server always and documented that, would server app react to the situation any differently?

quiche/src/lib.rs
```diff
@@ -6374,6 +6490,13 @@ impl Connection {
         self.ids
             .set_source_conn_id_limit(peer_params.active_conn_id_limit);
 
+        if let Some(preferred_address) = &peer_params.preferred_address_params {
+            // Zero-length connection ID must not be sent.
+            if preferred_address.connection_id.is_empty() {
+                return Err(Error::InvalidTransportParam);
+            }
+        }
+
```
LPardue, line 6499: Should you check for this connection being a server here too?

quiche/src/lib.rs
```diff
@@ -7961,13 +8134,143 @@ impl TransportParams {
...
+            cid,
+            stateless_reset_token,
+        ))
+    }
+
+    fn decode_v4(b: &mut octets::Octets) -> Result<Option<SocketAddrV4>> {
+        let mut addr_v4 = None;
+        let ip_v4_b = <[u8; 4]>::try_from(b.get_bytes(4)?.slice(4)?)
+            .map_err(|_| Error::InvalidTransportParam)?;
+        let ip_v4 = Ipv4Addr::from(ip_v4_b);
```
LPardue, line 8192: This might be clearer by just calling b.slice(4)? directly like 

```
let ip_v4 = Ipv4Addr::from(b.slice(4));
b.advance(4);
```

but with some actual error handling and mapping.

quiche/src/lib.rs
```diff
@@ -7961,13 +8134,143 @@ impl TransportParams {
...
+
+        b.put_varint(preferred_address_params.connection_id.len() as u64)?;
+        b.put_bytes(&preferred_address_params.connection_id)?;
+
+        b.put_bytes(&preferred_address_params.stateless_reset_token.to_be_bytes())?;
+
+        Ok(b.off())
+    }
+
+    fn address_v4(&self) -> Option<(SocketAddrV4, ConnectionId, u128)> {
```
LPardue, line 8261: Related to earlier comments, I'm not sure the of the value of these methods that wrap access to the fields in the parameter.

LPardue:
Thanks for the PR. The scope of this change is limited to adding an API and doing some of the required protocol checks to ensure the Transport Paramters are exchanged properly. The tests comprehensively cover that, which is good.

Unless I missed it, this PR doesn't test that preferred address would actually work between clients and servers talking IP. And as noted in one of the review comments, the API seems to make assumptions about how applications would actually make this work in practice. Without a concrete example, its harder to know if there are interoperability issues or if the API is the right shape for clients.

Updating the quiche-apps would probably blow the scope of this PR, making it harder to review. Perhaps we could make a separate follow-on PR that adds the preferred address feature to the quiche-apps so that we can get a view of how all the pieces would fit together.

Clendenin:
> Thanks for the PR. The scope of this change is limited to adding an API and doing some of the required protocol checks to ensure the Transport Paramters are exchanged properly. The tests comprehensively cover that, which is good.
> 
> Unless I missed it, this PR doesn't test that preferred address would actually work between clients and servers talking IP. And as noted in one of the review comments, the API seems to make assumptions about how applications would actually make this work in practice. Without a concrete example, its harder to know if there are interoperability issues or if the API is the right shape for clients.
> 
> Updating the quiche-apps would probably blow the scope of this PR, making it harder to review. Perhaps we could make a separate follow-on PR that adds the preferred address feature to the quiche-apps so that we can get a view of how all the pieces would fit together.

Thank you for the feedback @LPardue, it is appreciated :), I will work to free up some time so I can go through and address your comments.

quiche/src/lib.rs
```diff
@@ -7867,7 +7963,26 @@ impl TransportParams {
...
+                /// Stateless Reset Token: 16 bytes
+                const PREFERRED_ADDRESS_PARAM_MAX_SIZE: usize = 61;
+
+                let mut buffer = [0; PREFERRED_ADDRESS_PARAM_MAX_SIZE];
+                let written_to_buffer = PreferredAddressParams::encode(
+                    preferred_address_params,
+                    &mut buffer,
+                )?;
+                TransportParams::encode_param(&mut b, 0x000d, written_to_buffer)?;
+                b.put_bytes(&buffer[0..written_to_buffer])?;
```
LPardue, line 7983: I guess we have to alloc anyway, so maybe just do that inside `PreferredAddressParams::encode` and return the buffer there?

quiche/src/lib.rs
```diff
@@ -7867,7 +7963,26 @@ impl TransportParams {
             TransportParams::encode_param(&mut b, 0x000c, 0)?;
         }
 
-        // TODO: encode preferred_address
+        if is_server {
+            if let Some(preferred_address_params) = &tp.preferred_address_params {
+                /// IPv4: 4 bytes
```
LPardue, line 7968: nit: these should use double quotes `//`

PR #1503, opened 2023-05-03
Properly clear dgram_event_triggered

vkrasnov:


LPardue (approved):
This is correct in the purest sense but FWIW we plan to remove the API anyway https://github.com/cloudflare/quiche/pull/1291

PR #1508, opened 2023-05-10
Include the BoringSSL lib path to the rpath when linking with a specified path.

brbzull0:
If a custom BoringSSLl is used then this may be outside the system library search path which would make the libquiche to use the system one which may not be what the user wants.

This is needed so the build [adds](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-arg) a directory to the runtime library search path which is needed for the generated quiche shared object, otherwise this is unknown by the lib and forces users to have the `LD_LIBRARY_PATH` set with a custom bssl path.

### DIFF
--- a/quiche/src/build.rs
+++ b/quiche/src/build.rs
@@ -231,6 +231,8 @@ fn main() {
             cfg.build_target("crypto").build().display().to_string()
         });
 
+        println!("cargo:rustc-link-arg=-Wl,-rpath,{}", bssl_dir);
+
         let build_path = get_boringssl_platform_output_path();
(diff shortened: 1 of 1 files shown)


brbzull0:
Seems GNU and Clang can use the same syntax for the `-rpath`, using the `,` and not the `=` , can't test(macos) this locally so this is purely base on Docs. If this does not work, I'll get someone to help me test this on `macos`.
Thanks.

ghedo (approved):

ghedo:
Merged, thanks @brbzull0!

PR #1515, opened 2023-05-16
Make BBR less aggresive

lohith-bellad:
This changes does two things,
1. Avoids pacer skew that might show up when cwnd restricts us to clear pacer.used.
2. Make BBR less aggressive. In probe_bw phase instead of keeping inflight two times the BDP, we keep 1.25 times BDP. This has been tested in different RTT (<1ms, 8msec, 45msec) environments. And this greatly redcues BBR aggressiveness and retransmissions.

### DIFF
--- a/quiche/src/recovery/bbr/per_ack.rs
+++ b/quiche/src/recovery/bbr/per_ack.rs
@@ -270,7 +270,7 @@ fn bbr_enter_probe_bw(r: &mut Recovery, now: Instant) {
 
     bbr.state = BBRStateMachine::ProbeBW;
     bbr.pacing_gain = 1.0;
-    bbr.cwnd_gain = 2.0;
+    bbr.cwnd_gain = 1.25;
 
     // cycle_index will be one of (1, 2, 3, 4, 5, 6, 7). Since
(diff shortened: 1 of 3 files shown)


quiche/src/recovery/bbr/per_ack.rs
```diff
@@ -270,7 +270,7 @@ fn bbr_enter_probe_bw(r: &mut Recovery, now: Instant) {
 
     bbr.state = BBRStateMachine::ProbeBW;
     bbr.pacing_gain = 1.0;
-    bbr.cwnd_gain = 2.0;
+    bbr.cwnd_gain = 1.25;
```
junhochoi, line 273: This change diverges from original [BBR algorithm (draft-00)](https://datatracker.ietf.org/doc/html/draft-cardwell-iccrg-bbr-congestion-control-00#section-4.3.4.3). It will make cwnd growth less aggressive, but I believe it will take longer to find a max BDP in a variable high bandwidth/high rtt link. I think it would be great to discuss in bbr-dev@ mailing list.

lohith-bellad:
Not needed!

PR #1519, opened 2023-05-22
Add API to load certificate and private key from memory buffers

### LINKED ISSUE #1511: Add `quiche_config_load_priv_key` to FFI to be able to load a priv key without writing it to disk
Currently, the only way to load a priv key using the FFI is to call `quiche_config_load_priv_key_from_pem_file`. Unfortunately, this requires writing the key to a temporary file on disk which opens all sorts of security issues (access privileges, hardware tampering to recover the deleted key...).

Adding an extra `quiche_config_load_priv_key` to FFI that would map to BoringSSL's `SSL_CTX_use_PrivateKey` would solve this issue.

lorban:
This exposes BoringSSL's  `SSL_CTX_use_certificate` and `SSL_CTX_use_PrivateKey` calls so that it's possible to pass Quiche a certificate and a private key via in-memory buffers instead of having to write files on disk containing their PEM representations.

Fixes #1511.

### DIFF
--- a/quiche/src/tls.rs
+++ b/quiche/src/tls.rs
@@ -164,6 +164,13 @@ struct SSL_PRIVATE_KEY_METHOD {
     ) -> ssl_private_key_result_t,
 }
 
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+struct CBS {
+    data: *const u8,
+    len: libc::size_t,
+}
+
 lazy_static::lazy_static! {
     /// BoringSSL Extra Data Index for Quiche Connections
     pub static ref QUICHE_EX_DATA_INDEX: c_int = unsafe {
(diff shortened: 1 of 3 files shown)


pkropachev:
@LPardue, @ghedo, Lucas, Alessandro, could you please consider integrating this code?

quiche/src/tls.rs
```diff
@@ -251,6 +257,32 @@ impl Context {
...
+            let pkey = EVP_parse_private_key(&cbs);
+            let rc = SSL_CTX_use_PrivateKey(self.as_mut_ptr(), pkey);
+            EVP_PKEY_free(pkey);
+            rc
+        })
+    }
+
+    pub fn use_cert(&mut self, cert: *const u8, cert_len: libc::size_t) -> Result<()> {
+        map_result(unsafe {
+            let x = d2i_X509(
```
pkropachev, line 272: Does it make sense to use PEM format instead of DER to inline with `use_certificate_chain_file()` that takes path to the PEM file?

lorban: I'm not too versed in how crypto code works, but my understanding is that DER is the in-memory, machine-readable representation while PEM is the on-disk, human-readable representation.

If there is a consensus that using PEM instead of DER here would be a better idea, then I'm not opposed to that.

pkropachev: Actually, for me DER is ok, this is binary format, it doesn't require encode/decode to Base64 (like for PEM - text format). In case generating private key and certificate using SSL API in code, then I think DER format is more acceptable.

PR #1526, opened 2023-06-02
add method to expose all source CIDs

ghedo:
An application can track these on its own, but quiche already has this information, so it can easily be exposed.


vkrasnov (approved):

LPardue (approved):

LPardue:
no C API?

ghedo:
> no C API?

Lol, exposing iterators is a pain... we can look into that later, but for now we need this change ASAP.

PR #1533, opened 2023-06-06
Introduce session_and_params methods

normanmaurer:
Motivation:

At the moment the session and transport params are serialized into one buffer using their length as a prefix. While this works it would be better to keep them split for various reasons. One is that it doesnt make sense to encode them first into one just to split them again later. The other is that users might want to implement session re-use / caching themselves by using BoringSSL callbacks in which case they need to re-implement the encoding again which might break later as its an implementation details.

Modifications:

- Add session_and_params(...) and set_session_and_params(...) methods
- Deprecate the old methods

Result:

Improve API.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1236,11 +1236,18 @@ pub struct Connection {
     /// TLS handshake state.
     handshake: tls::Handshake,
 
-    /// Serialized TLS session buffer.
+    /// Serialized TLS session buffer that combines the session and transport
+    /// params.
     ///
     /// This field is populated when a new session ticket is processed on the
     /// client. On the server this is empty.
     session: Option<Vec<u8>>,
+
+    /// Serialized TLS session buffer.
+    ///
+    /// This field is populated when a new session ticket is processed on the
+    /// client. On the server this is empty.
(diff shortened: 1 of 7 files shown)


quiche/src/lib.rs
```diff
@@ -2008,19 +2017,38 @@ impl Connection {
...
+    /// On the client, this can be used to offer the given serialized session,
+    /// as returned by [`session_and_params()`], for resumption.
+    ///
+    /// This must only be called immediately after creating a connection, that
+    /// is, before any packet is sent or received.
+    ///
+    /// [`session_and_params()`]: struct.Connection.html#method.session_and_params
+    #[inline]
+    pub fn set_session_and_params(
+        &mut self, session_and_params: (&[u8], &[u8]),
```
normanmaurer, line 2046: I used a tuple but I was wondering if a struct would be better to make it easier to understand what is the session and what is the params. WDYT ?

normanmaurer: @ghedo WDYT ?

ghedo: Why not just separate arguments?

normanmaurer: I wanted to match it with what is returned by `session_and_param`.

PR #1537, opened 2023-06-08
remove per-stream application data

ghedo:
This was originally added for curl, but it was never used. The Rust API is also hard to use, because the app data reference borrows the whole connection object.


LPardue:
makes my life easier

LPardue (approved):

PR #1540, opened 2023-06-12
prioritized readable and writable iterators

LPardue:
This change modifies the internal implementaion of
readable and writable stream collections so that they
better respect the ordering expected from RFC 9218.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -400,10 +400,13 @@ use qlog::events::EventImportance;
 use qlog::events::EventType;
 #[cfg(feature = "qlog")]
 use qlog::events::RawInfo;
+use stream::StreamPriorityKey;
 
 use std::cmp;
 use std::convert::TryInto;
 use std::time;
+
+use std::sync::Arc;
 
 use std::net::SocketAddr;
 
@@ -4403,6 +4406,7 @@ impl Connection {
         }
 
         let local = stream.local;
+        let priority_key = Arc::clone(&stream.priority_key);
 
         #[cfg(feature = "qlog")]
         let offset = stream.recv.off_front();
@@ -4419,7 +4423,7 @@ impl Connection {
                     self.streams.collect(stream_id, local);
                 }
 
-                self.streams.remove_readable(stream_id);
+                self.streams.remove_readable(&priority_key);
                 return Err(e);
             },
         };
@@ -4435,7 +4439,7 @@ impl Connection {
         }
 
         if !readable {
-            self.streams.remove_readable(stream_id);
+            self.streams.remove_readable(&priority_key);
         }
 
         if complete {
@@ -4541,6 +4545,8 @@ impl Connection {
 
         let was_flushable = stream.is_flushable();
 
+        let priority_key = Arc::clone(&stream.priority_key);
+
         // Truncate the input buffer based on the connection's send capacity if
         // necessary.
         //
@@ -4554,7 +4560,7 @@ impl Connection {
                 // again once the capacity increases.
                 //
                 // Since the stream is writable already, mark it here instead.
-                self.streams.insert_writable(stream_id);
+                self.streams.insert_writable(&priority_key);
             }
 
             return Err(Error::Done);
@@ -4570,7 +4576,7 @@ impl Connection {
             Ok(v) => v,
 
             Err(e) => {
-                self.streams.remove_writable(stream_id);
+                self.streams.remove_writable(&priority_key);
                 return Err(e);
             },
         };
@@ -4606,7 +4612,7 @@ impl Connection {
         }
 
         if !writable {
-            self.streams.remove_writable(stream_id);
+            self.streams.remove_writable(&priority_key);
         } else if was_writable && blocked_by_cap {
             // When `stream_writable_next()` returns a stream, the writable
             // mark is removed, but because the stream is blocked by the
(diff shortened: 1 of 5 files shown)


quiche/src/stream.rs
```diff
@@ -943,14 +946,20 @@ impl Iterator for StreamIter {
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        self.streams.pop()
+        let ret = self.streams.get(self.index).cloned();
```
ghedo, line 949: Somewhat simpler would be:
```rust
let v = self.streams.get(self.index)?;
self.index += 1;
Some(*v)
```
Not 100% sure it compiles though.

quiche/src/stream.rs
```diff
@@ -421,11 +432,42 @@ impl StreamMap {
...
+
+            self.writable_prioritized.insert(Arc::clone(priority_key));
         } else {
             self.writable.remove(&stream_id);
+
+            self.writable_prioritized_remove(priority_key);
+        }
+    }
+
+    fn writable_prioritized_remove(
```
ghedo, line 450:
```suggestion
    fn remove_writable(
```

quiche/src/stream.rs
```diff
@@ -421,11 +432,42 @@ impl StreamMap {
...
+    ) {
+        let mut c = {
+            let ptr = Arc::as_ptr(priority_key);
+            unsafe { self.writable_prioritized.cursor_mut_from_ptr(ptr) }
+        };
+
+        c.remove();
+    }
+
+    pub fn writable_prioritized_update(
```
ghedo, line 461:
```suggestion
    pub fn update_priority(
```
(we can use this method in the future for readable/flushable too)

quiche/src/stream.rs
```diff
@@ -741,17 +844,99 @@ pub fn is_bidi(stream_id: u64) -> bool {
...
+#[derive(Debug)]
+pub struct StreamPriorityKey {
+    pub urgency: u8,
+    pub incremental: bool,
+    pub id: u64,
+
+    pub writable: RBTreeAtomicLink,
+}
+
+impl Clone for StreamPriorityKey {
```
ghedo, line 857: Can't we just derive this?

quiche/src/lib.rs
```diff
@@ -4670,10 +4680,26 @@ impl Connection {
             return Ok(());
         }
 
+        let is_writable = stream.is_writable();
+
         stream.urgency = urgency;
         stream.incremental = incremental;
 
-        // TODO: reprioritization
+        let old_priority_key = stream.priority_key.clone();
```
ghedo, line 4688: I wonder if something like the following would work:
```rust
let new_priority_key = Arc::new(...);
let old_priority_key = std::mem::replace(&mut stream.priority_key, new_priority_key);

self.streams.writable_prioritized_update(
  &old_priority_key,
  &stream.priority_key,
  is_writable,
);
```
Looks a little cleaner (if it works).

ghedo (changes requested):
Added a few (minor) comments.

quiche/src/stream.rs
```diff
@@ -421,11 +432,42 @@ impl StreamMap {
...
+
+    pub fn writable_prioritized_update(
+        &mut self, old: &Arc<StreamPriorityKey>, new: &Arc<StreamPriorityKey>,
+        writable: bool,
+    ) {
+        if old.writable.is_linked() {
+            self.writable_prioritized_remove(old);
+        }
+
+        if writable {
```
ghedo, line 469: If the old key was linked then we can assume the new one should be inserted too? This way we wouldn't need the additional `writable` argument.

quiche/src/lib.rs
LPardue: yeah this works

quiche/src/stream.rs
LPardue: yep

quiche/src/stream.rs
```diff
@@ -678,6 +775,12 @@ impl Stream {
             local,
             urgency: DEFAULT_URGENCY,
             incremental: true,
+            priority_key: Arc::new(StreamPriorityKey {
```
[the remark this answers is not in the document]
LPardue: done

quiche/src/stream.rs
```diff
@@ -421,11 +432,42 @@ impl StreamMap {
     /// to when an existing stream becomes writable (or stops being writable).
     ///
     /// If the stream was already in the list, this does nothing.
-    pub fn mark_writable(&mut self, stream_id: u64, writable: bool) {
+    pub fn mark_writable(
```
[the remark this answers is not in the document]
LPardue: Seemed easier to split them in a separate PR, see https://github.com/cloudflare/quiche/pull/1553

quiche/src/stream.rs
ghedo: @LPardue it seems this comment still applies. Also, probably need to add the same logic for `readable` too.

quiche/src/lib.rs
```diff
@@ -4435,7 +4439,7 @@ impl Connection {
         }
 
         if !readable {
-            self.streams.remove_readable(stream_id);
+            self.streams.remove_readable(&priority_key);
```
ghedo, line 4442: I suppose we need to do the shuffling for `readable` too? like:
```
        if incremental && readable {
            // Shuffle the incremental stream to the back of the the queue.
            self.streams.remove_readable(&priority_key);
            self.streams.insert_readable(&priority_key);
        }
```

ghedo (approved):

PR #1555, opened 2023-07-04
h3: allow applications to set additional settings

francoismichel:
Using additional settings, an application can configure and set H3 settings that are not explicitly handled by quiche.

This can be useful to develop some apps atop quiche's h3 without needing quiche to explicitly support it.

For instance, I use it in my webtransport implementation atop quiche: https://github.com/francoismichel/webtransport-quiche/blob/269d8fcba3e874c1984ec65c6dfd653cce4c15eb/src/lib.rs#L97C13-L97C13

Let me know your thoughts and comments.

Cheers,

François

### DIFF
--- a/quiche/src/h3/frame.rs
+++ b/quiche/src/h3/frame.rs
@@ -70,6 +70,7 @@ pub enum Frame {
         connect_protocol_enabled: Option<u64>,
         h3_datagram: Option<u64>,
         grease: Option<(u64, u64)>,
+        additional_settings: Option<Vec<(u64, u64)>>,
         raw: Option<Vec<(u64, u64)>>,
     },
 
@@ -181,6 +182,7 @@ impl Frame {
                 connect_protocol_enabled,
                 h3_datagram,
                 grease,
+                additional_settings,
                 ..
             } => {
                 let mut len = 0;
@@ -217,6 +219,13 @@ impl Frame {
                     len += octets::varint_len(val.1);
                 }
 
+                if let Some(vals) = additional_settings {
+                    for val in vals {
+                        len += octets::varint_len(val.0);
+                        len += octets::varint_len(val.1);
+                    }
+                }
+
                 b.put_varint(SETTINGS_FRAME_TYPE_ID)?;
                 b.put_varint(len as u64)?;
 
@@ -250,6 +259,13 @@ impl Frame {
                 if let Some(val) = grease {
                     b.put_varint(val.0)?;
                     b.put_varint(val.1)?;
+                }
+
+                if let Some(vals) = additional_settings {
+                    for val in vals {
+                        b.put_varint(val.0)?;
+                        b.put_varint(val.1)?;
+                    }
                 }
             },
 
@@ -457,10 +473,11 @@ impl std::fmt::Debug for Frame {
                 max_field_section_size,
                 qpack_max_table_capacity,
                 qpack_blocked_streams,
+                additional_settings,
                 raw,
                 ..
             } => {
-                write!(f, "SETTINGS max_field_section={max_field_section_size:?}, qpack_max_table={qpack_max_table_capacity:?}, qpack_blocked={qpack_blocked_streams:?} raw={raw:?}")?;
+                write!(f, "SETTINGS max_field_section={max_field_section_size:?}, qpack_max_table={qpack_max_table_capacity:?}, qpack_blocked={qpack_blocked_streams:?} raw={raw:?}, additional_settings={additional_settings:?}")?;
             },
 
             Frame::PushPromise {
@@ -525,6 +542,7 @@ fn parse_settings_frame(
     let mut connect_protocol_enabled = None;
     let mut h3_datagram = None;
     let mut raw = Vec::new();
+    let mut additional_settings: Option<Vec<(u64, u64)>> = None;
 
     // Reject SETTINGS frames that are too long.
     if settings_length > MAX_SETTINGS_PAYLOAD_SIZE {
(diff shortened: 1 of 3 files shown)


quiche/src/h3/frame.rs
```diff
@@ -567,8 +585,11 @@ fn parse_settings_frame(
             0x0 | 0x2 | 0x3 | 0x4 | 0x5 =>
                 return Err(super::Error::SettingsError),
 
-            // Unknown Settings parameters must be ignored.
-            _ => (),
+            // Unknown Settings parameters go into additional_settings.
+            _ => match &mut additional_settings {
+                Some(s) => s.push((identifier, value)),
+                None => additional_settings = Some(vec![(identifier, value)]),
+            },
```
LPardue, line 592: Couple of points:

1. you can use get_or_insert to make the code cleaner like

```
_ => {
  let s: &mut Vec<(u64, u64)> = additional_settings.get_or_insert_default();
  s.push((identifier, value));
}
```

2. Is it really useful to capture grease settings types? I don't think we should waste effort trying to detect grease values but something to think about how an application will handle this.

quiche/src/h3/mod.rs
```diff
@@ -515,6 +515,9 @@ pub struct Config {
     qpack_max_table_capacity: Option<u64>,
     qpack_blocked_streams: Option<u64>,
     connect_protocol_enabled: Option<u64>,
+    /// additional settings are settings that are not part of the H3
+    /// settings explicitly handled above
```
LPardue, line 519: and grease, if you keep that

quiche/src/h3/mod.rs
```diff
@@ -565,6 +569,35 @@ impl Config {
             self.connect_protocol_enabled = None;
         }
     }
+
+    /// Sets H3 additional settings.
```
LPardue, line 573:
```suggestion
    /// Sets additional HTTP/3 settings.
```

quiche/src/h3/mod.rs
```diff
@@ -565,6 +569,35 @@ impl Config {
...
+    ///
+    /// The default value is no additional settings.
+    /// The `additional_settings` parameter must not contain settings already
+    /// handled by default in this library.
```
LPardue, line 577: We should just list those settings here using the IANA names e.g. SETTINGS_QPACK_MAX_TABLE_CAPACITY, etc.

However, this raises an interesting future compatibility problem, where an application used additional settings prior to the library adding first-class support for them. Raising an error at runtime is sort of annoying in that situation, I wonder if there is some elegant way to make this a compiler error.

quiche/src/h3/mod.rs
```diff
@@ -565,6 +569,35 @@ impl Config {
...
+            frame::SETTINGS_MAX_FIELD_SECTION_SIZE,
+            frame::SETTINGS_QPACK_BLOCKED_STREAMS,
+            frame::SETTINGS_ENABLE_CONNECT_PROTOCOL,
+            frame::SETTINGS_H3_DATAGRAM,
+        ];
+        if additional_settings
+            .iter()
+            .any(|(identifier, _)| explicit_settings.contains(identifier))
+        {
+            return Err(Error::SettingsError);
```
LPardue, line 596: So although we don't currently do duplicate settings checks on receive side, it would be nice to prevent the caller from mistakenly trying to send duplicate additional settings (see https://www.rfc-editor.org/rfc/rfc9114.html#section-7.2.4-6). That should also be documented in the doc as a failure condition, and added as a test.

LPardue (commented):
Looks good overall. However, there's a few minor points we should think about and address

quiche/src/h3/mod.rs
francoismichel: Maybe we could use a hashmap as input to prevent duplicate additional settings and avoid this additional failure condition.

quiche/src/h3/frame.rs
francoismichel: Concerning your point 2: 
The notion of "additional params" is specific to the sending-side.
On the receiving-side, the application retrieves the settings using the `.peer_settings_raw()` method.
I think that the behaviour on the receiving side is thus the same as before since the grease settings were already part of the "raw settings" retrieved by the application before.

quiche/src/h3/mod.rs
francoismichel: Done in commit [d67f4ff](https://github.com/cloudflare/quiche/pull/1555/commits/d67f4ff5390a63a45ecfdec59b68241aebf29665).

quiche/src/h3/mod.rs
francoismichel: I think I'll keep vectors and chck the presence of duplicates to keep the settings ordering

francoismichel: Done in commit 1147a80.

LPardue:
Can you also add coverage for this field in the [to_qlog](https://github.com/cloudflare/quiche/blob/master/quiche/src/h3/frame.rs#L317) method please? Basically copy the `grease` block in that func but loop over all values.

francoismichel:
Done, I tested with the client and server, additional settings appear in the qlogs at when created and parsed, so it should be fine.

I don't know if we should add a way to specify additional_settings in `apps/src/common.rs:make_h3_config()` or something ?

LPardue:
> I don't know if we should add a way to specify additional_settings in apps/src/common.rs:make_h3_config() or something ?

Lets punt on that for now, it seems limited value for the example apps since they don't need to send anything additional.

LPardue (approved):

PR #1557, opened 2023-07-07
nginx: sync patch

ghedo:
Haven't done this in a while... sync with internal repo.

### DIFF
--- a/nginx/README.md
+++ b/nginx/README.md
@@ -159,6 +159,26 @@ Sets the per-stream incoming flow control limit.
 
 Sets the timeout of inactivity after which the connection is closed.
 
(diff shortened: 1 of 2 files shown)


LPardue (dismissed):

junhochoi:
Good to see some familiar code!
I'd think that readme also need some update, such as pacing requires fq to work.

ghedo:
@junhochoi yeah, good point. I think we should probably just add a config option to enable pacing and leave it disabled by default.

ghedo:
Added `http3_congestion_control` and `http3_pacing` config options.

junhochoi (approved):

Issue #1559, opened 2023-07-07
quiche/src/tls.rs uses an unsafe bindings strategy


davidben:
https://github.com/cloudflare/quiche/blob/master/quiche/src/tls.rs#L1255 contains a pile of hand-written bindings for BoringSSL functions. BoringSSL is not ABI-stable, so this is not a supported way to use the library. This will result in subtle memory errors, with no checking that you've gotten it right.

Indeed, there are already mistakes:
https://github.com/cloudflare/quiche/blob/master/quiche/src/tls.rs#L1442C8-L1443

BoringSSL uses `size_t`, not `int`, for those functions. That means this crate only works correctly on calling conventions where `int` and `size_t` are compatible. Specifically, on 64-bit platforms, you're relying on the upper half of the 64-bit register having a defined value when it is used to pass in a 32-bit parameter. You're in luck that most calling conventions require this, but this is not guaranteed.

This should be replaced with something robust, e.g. bindgen.

ghedo:
> This should be replaced with something robust, e.g. bindgen.

Yes, though the main complication is that we don't enforce a specific version of BoringSSL and we don't necessarily have access to the headers for that version (e.g. an application can provide its own BoringSSL build). Need to figure out a way to work around that.

davidben:
Not having headers seems unlikely to work well, given how we expect BoringSSL to be used. You also don't usually expect to be able to use a library in C without its headers.

Perhaps you all could transition to requiring the application provide both the build and the headers?

davidben:
In parallel, I suppose I can start by upstreaming some changes to gradually take the hand-transcribed bindings closer to what bindgen would have generated. There are quite a few mistakes. :-)

davidben:
Any news on this? This is a memory safety vulnerability waiting to happen if we, e.g., ever extend our existing structs.

PR #1561, opened 2023-07-11
Fix a pile of incorrect bindings in tls.rs

davidben:
See issue #1559. This does not fix the crate's unsafe bindings strategy, but corrects some of the mistakes that snuck in as a result of this unsafe strategy.

One thing to call out is that sk_num and sk_value's index types were wrong. This would have caused a problem were the crate not saved by most 64-bit calling conventions defining the upper half of the register when a 64-bit register is used to pass a 32-bit value. (But I have not carefully checked all calling conventions for this. The real fix is to stop writing bindings this way. This bindings strategy is even less safe than calling the library from C, which is quite a low bar to fail!)

Beyond that, the issues are mostly const-ness, though one notable issue is that function pointers need to be imported as Option. Also the return value for set_min_proto_version and set_max_proto_version was being silently ignored.

### DIFF
--- a/quiche/src/tls.rs
+++ b/quiche/src/tls.rs
@@ -705,7 +705,7 @@ impl Handshake {
                 map_result_ptr(SSL_get0_peer_certificates(self.as_ptr())).ok()?;
 
             let num = sk_num(chain);
-            if num <= 0 {
+            if num == 0 {
                 return None;
             }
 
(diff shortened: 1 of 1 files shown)


ghedo (approved):

ghedo:
Merged, thanks! Just fixed a couple of clippy warnings.

PR #1564, opened 2023-07-18
lib: migrate flushable to RBtree

LPardue:
Rather than use a custom BTree format, use the same RBtree that is used
for readable and writable stream tracking and prioritization.

~Note that this appears to break some path related tests, hence the draft status.~ Test errors were only observed in local dev due to one missing change that was fixed during final pass before opening PR.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3296,13 +3296,8 @@ impl Connection {
                         // set.
                         if (stream.is_flushable() || empty_fin) && !was_flushable
                         {
-                            let urgency = stream.urgency;
-                            let incremental = stream.incremental;
-                            self.streams.push_flushable(
-                                stream_id,
-                                urgency,
-                                incremental,
-                            );
+                            let priority_key = Arc::clone(&stream.priority_key);
+                            self.streams.insert_flushable(&priority_key);
                         }
 
                         self.stream_retrans_bytes += length as u64;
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -4001,15 +3996,14 @@ impl Connection {
             path.active() &&
             !dgram_emitted
         {
-            while let Some(stream_id) = self.streams.peek_flushable() {
+            for stream_id in self.streams.flushable() {
```
ghedo, line 3999: For the purposes of this code we don't actually need the full iterator since we only end up using only one stream anyway, so I think we should keep the `peek_flushable()` API, which should be easy to implement on top of intrusive rbtree I think.

quiche/src/lib.rs
```diff
@@ -4001,15 +3996,14 @@ impl Connection {
...
                 let stream = match self.streams.get_mut(stream_id) {
                     // Avoid sending frames for streams that were already stopped.
                     //
                     // This might happen if stream data was buffered but not yet
                     // flushed on the wire when a STOP_SENDING frame is received.
                     Some(v) if !v.send.is_stopped() => v,
                     _ => {
-                        self.streams.remove_flushable();
```
ghedo, line 4012: This should stay?

ghedo (changes requested):
Nice! Added a couple of comments.

quiche/src/lib.rs
LPardue: Sure

ghedo (approved):

PR #1567, opened 2023-07-20
Fix broken send_on_path_test

PSUdaemon:
The function declaration of `sort()` is `fn sort(&mut self) -> ()` and so these tests were simply asserting `() == ()`. After correcting them they failed and so a change to `paths_iter()` is needed.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -5940,15 +5940,9 @@ impl Connection {
             sockaddrs: self
                 .paths
                 .iter()
-                .filter_map(|(_, p)| {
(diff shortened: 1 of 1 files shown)


ghedo:
So, it turns out the `paths_iter()` condition was right, but the test was broken, which was hidden by the `sort()` bug you fixed. I pushed a commit to fix it, baiscally the idea is that the packets that are created in the test need to actually be sent to the server, so it can reply to the PATH_CHALLENGEs, which in turn makes the client update the paths' state correctly.

ghedo:
Anyway, good catch, and thanks for the fix!

ghedo (approved):

PR #1568, opened 2023-07-20
remove unnecessary argument to decode_pkt

ehaydenr:
slices already encode their length, no need for extra len argument


ghedo (approved):

ghedo:
Merged, thanls!

PR #1574, opened 2023-07-28
Correctly account for datagram queue in delivery_rate_check_if_app_limited

toshiiw:
This patch fixes the calculation and avoids app_limited to be incorrectly set true.


ghedo (approved):

ghedo:
Sorry for the delay, this is merged now. Thanks @toshiiw!

PR #1575, opened 2023-07-28
quic: add peer_transport_params()

ehaydenr:
Currently, transport parameters are available through `stats()`. APIs that used to offer access to certain transport parameters have been removed in favor of using `stats()`. While this does work, it's wasteful to return such a large structure when you may only care about a single TP. This change makes `TransportParameters` public, adds a method `peer_transport_params()` that returns a reference, and removes transport parameter fields from `Stats`.

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -516,7 +516,6 @@ typedef struct {
 
     // The number of known paths for the connection.
     size_t paths_count;
-
 } quiche_stats;
 
 // Collects and returns statistics about the connection.
@@ -563,6 +562,8 @@ typedef struct {
     ssize_t peer_max_datagram_frame_size;
 } quiche_transport_params;
 
+// Returns the peer's transport parameters in |out|. Returns false if we have
+// not yet processed the peer's transport parameters.
 bool quiche_conn_peer_transport_params(const quiche_conn *conn, quiche_transport_params *out);
 
 typedef struct {
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -517,6 +517,12 @@ typedef struct {
     // The number of known paths for the connection.
     size_t paths_count;
 
+} quiche_stats;
+
+// Collects and returns statistics about the connection.
+void quiche_conn_stats(const quiche_conn *conn, quiche_stats *out);
+
+typedef struct {
     // The maximum idle timeout.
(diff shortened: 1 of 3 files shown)


quiche/src/ffi.rs
```diff
@@ -1091,22 +1091,25 @@ pub struct Stats {
...
-    peer_initial_max_streams_uni: u64,
-    peer_ack_delay_exponent: u64,
-    peer_max_ack_delay: u64,
-    peer_disable_active_migration: bool,
-    peer_active_conn_id_limit: u64,
-    peer_max_datagram_frame_size: ssize_t,
     paths: [PathStats; 8],
 }
 
+pub struct TransportParams {
```
ghedo, line 1097: You also need to update [quiche.h](https://github.com/cloudflare/quiche/blob/master/quiche/include/quiche.h) with both type (something like `quiche_transport_params`) and function.

ghedo (changes requested):

quiche/src/ffi.rs
ehaydenr: Oops, updated

quiche/include/quiche.h
```diff
@@ -517,6 +517,12 @@ typedef struct {
     // The number of known paths for the connection.
     size_t paths_count;
```
ghedo, line 519:
```suggestion
```

quiche/include/quiche.h
```diff
@@ -555,10 +561,9 @@ typedef struct {
 
     // DATAGRAM frame extension parameter, if any.
     ssize_t peer_max_datagram_frame_size;
-} quiche_stats;
+} quiche_transport_params;
 
-// Collects and returns statistics about the connection.
-void quiche_conn_stats(const quiche_conn *conn, quiche_stats *out);
+bool quiche_conn_peer_transport_params(const quiche_conn *conn, quiche_transport_params *out);
```
ghedo, line 566:
```suggestion
// Returns the peer's transport parameters in |out|. Returns false if we have
// not yet processed the peer's transport parameters.
bool quiche_conn_peer_transport_params(const quiche_conn *conn, quiche_transport_params *out);
```

ghedo (approved):

ghedo (approved):

ghedo:
Merged (with some small fixes), thanks @ehaydenr!

PR #1592, opened 2023-08-18
Fix goaway frame type id

wangfuyu:
RFC9114 7.2.6. GOAWAY
The GOAWAY frame (type=0x07) is used to initiate graceful shutdown of an HTTP/3 connection by either endpoint. GOAWAY allows an endpoint to stop accepting new requests or pushes while still finishing processing of previously received requests and pushes. This enables administrative actions, like server maintenance. GOAWAY by itself does not close a connection.


LPardue (approved):

LPardue:
Merged, thanks!

Smithad499 (commented):
Adrian Smith original owner

PR #1598, opened 2023-08-23
Fix BBRv1 idle_restart flag

lohith-bellad:
For Web traffic which is sending packets for short duration and then idle for some duration, we should not enter probe RTT phase when idle. This leads to unnecessary reduction in cwnd and sometimes stalls.

When probe RTT timer has expired, idle_restart flag decides whether to enter probe RTT or not. In quiche we reduce the inflight after updating the BBR model and state, this leads to inflight not being zero. Since inflight is not zero we do not set idle_restart flag and we always enter probe RTT phase for web kind of traffic.

Here we check if inflight is less than 4 packets and if app_limited flag is also set, we set idle_restart flag. The idea being since inflight is low we would have already captured min_rtt and dont need probe RTT phase.

### DIFF
--- a/quiche/src/recovery/bbr/mod.rs
+++ b/quiche/src/recovery/bbr/mod.rs
@@ -290,9 +290,9 @@ fn reset(r: &mut Recovery) {
 }
 
 fn on_packet_sent(r: &mut Recovery, sent_bytes: usize, _now: Instant) {
-    r.bytes_in_flight += sent_bytes;
-
     per_transmit::bbr_on_transmit(r);
+
+    r.bytes_in_flight += sent_bytes;
 }
 
 fn on_packets_acked(


quiche/src/recovery/bbr/per_transmit.rs
```diff
@@ -36,7 +36,9 @@ pub fn bbr_on_transmit(r: &mut Recovery) {
 
 // 4.3.4.4.  Restarting From Idle
 fn bbr_handle_restart_from_idle(r: &mut Recovery) {
-    if r.bytes_in_flight == 0 && r.delivery_rate.app_limited() {
```
junhochoi, line 39: Interesting bug. So bbr_on_transmit() is called after bytes_in_flight is increased (https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/mod.rs#L293), there is no chance bytes_in_flight here is 0.

What about just moving https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/bbr/mod.rs#L293 after calling bbr_on_transmit() (L295)?

In linux tcp bbr, it is [checked](https://github.com/torvalds/linux/blob/master/net/ipv4/tcp_bbr.c#L336) when CA_EVENT_TX_START (first transmission when no packet is in flight), so I think it's good to keep the definition (checking bytes_in_flight == 0, not a small number) here.

lohith-bellad: Thats actually much better. Thank you for reviewing the code change.

junhochoi (approved):

ghedo (approved):

PR #1600, opened 2023-08-25
Fix doc reference to PathEvent::New

PSUdaemon:


PSUdaemon:
The failure looks like it will be fixed with #1599 and will affect all builds.

ghedo (approved):

ghedo:
Merged, thanks!

PR #1603, opened 2023-08-29
apps: update to nix 0.27

ghedo:
One notable change is that functions like `setsockopt()` now require `AsFd` instead of the raw fd, but mio doesn't implement that, so we need to convert fds explicitly.


ghedo:
Merged without approval to unblock broken builds.

PR #1607, opened 2023-09-05
lib: skip dgram frame size calc when none to send

LPardue:
Previously, we calculated the value of the largest DATAGRAM frame that
could be sent every time we wanted to send a packet. The calculation is
controlled by peer TPs, which can be sent even if the peer doesn't
really want us to use the DATAGRAM on the connection (e.g. conventional
HTTP/3 serving).

We only need the DATAGRAM size when we have datagrams to actually send.
This change implements a check of the internal DATAGRAM send queue to
avoid the calculation.

### DIFF
--- a/quiche/src/dgram.rs
+++ b/quiche/src/dgram.rs
@@ -103,6 +103,10 @@ impl DatagramQueue {
         self.len() == self.queue_max_len
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
     pub fn len(&self) -> usize {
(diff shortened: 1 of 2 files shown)


LPardue:
The other optimization would be to store the result of the size calculation, rather than do it for every packet. However, `dgram_max_writable_len` internally factors in DCID and UDP payload size, which could change during a connection (or be different for different paths?). I think a bigger refactor could be done to recalc the sizes when either of those things changes but I'm punting on that for now.

LPardue:
Fails due to https://www.githubstatus.com/incidents/c8zjb9351hlc :disappointed:

ghedo (approved):

PR #1612, opened 2023-09-07
stream: split RecvBuf and SendBuf into submodules

ghedo:


quiche/src/stream/mod.rs
```diff
@@ -789,1851 +772,259 @@ impl Default for StreamPriorityKey {
             flushable: Default::default(),
         }
     }
-}
-
-impl PartialEq for StreamPriorityKey {
```
ghedo, line 794: FTR, it seems git is being dumb here when generating the diff, none of the priority code actually moved (you can see this more clearly with `git show --minimal` locally).

LPardue (approved):

PR #1619, opened 2023-09-17
Make quiche-client respect concurrency limits passed via CLI

evanrittenhouse:
Tested by spinning up a quiche-server instance and then running:
```shell
RUST_LOG=trace cargo run --bin quiche-client -- https://127.0.0.1:4433 --no-verify --requests 50 --concurrent-requests <concurrent_requests>
```
with `concurrent_requests` set to 1, 5,and 50. I then ensured that the `trace` log I added got printed in the expected cases. I also:
1.  added a print to check `self.reqs_in_flight` and ensured that it was always less than or equal to the `concurrent_requests` limit.
2. Ensured that 50/50 responses were received.

### DIFF
--- a/apps/src/common.rs
+++ b/apps/src/common.rs
@@ -29,8 +29,6 @@
 //! This module provides some utility functions that are common to quiche
 //! applications.
 
-use std::cmp;
-
 use std::io::prelude::*;
 
 use std::collections::HashMap;
@@ -1126,6 +1124,7 @@ impl HttpConn for Http3Conn {
         &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
         max_reqs_concurrent: &u64,
     ) {
+        debug!("EVAN sending requests");
         let mut reqs_done = 0;
 
         let max_reqs_concurrent =
@@ -1134,7 +1133,10 @@ impl HttpConn for Http3Conn {
         // First send headers.
         for req in self.reqs.iter_mut().skip(self.reqs_hdrs_sent) {
             if self.reqs_in_flight >= max_reqs_concurrent {
-                debug!("too many reqs in flight, retry later...");
+                debug!(
+                    "too many reqs in flight, retry later... [{}]",
+                    self.reqs_in_flight
+                );
                 break;
             }
 
@@ -1145,6 +1147,10 @@ impl HttpConn for Http3Conn {
             ) {
                 Ok(v) => {
                     self.reqs_in_flight += 1;
+                    debug!(
+                        "EVAN incremented reqs_in_flight: [{}]",
+                        self.reqs_in_flight
+                    );
                     v
                 },
 
@@ -1239,6 +1245,7 @@ impl HttpConn for Http3Conn {
         &mut self, conn: &mut quiche::Connection, buf: &mut [u8],
         req_start: &std::time::Instant,
     ) {
+        debug!("EVAN entered handle_responses");
         loop {
             match self.h3_conn.poll(conn) {
                 Ok((stream_id, quiche::h3::Event::Headers { list, .. })) => {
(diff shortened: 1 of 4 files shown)


apps/src/common.rs
```diff
@@ -1162,6 +1184,9 @@ impl HttpConn for Http3Conn {
             reqs_done += 1;
         }
         self.reqs_hdrs_sent += reqs_done;
+        if self.reqs_in_flight > 0 {
+            self.reqs_in_flight -= 1;
+        }
```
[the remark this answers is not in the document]
evanrittenhouse: Moved

apps/src/common.rs
```diff
@@ -1276,6 +1298,10 @@ impl HttpConn for Http3Conn {
 
                 Ok((_stream_id, quiche::h3::Event::Finished)) => {
                     self.reqs_complete += 1;
+                    if self.reqs_in_flight > 0 {
```
LPardue, line 1301: Nit: You can use saturating_sub() here

evanrittenhouse: TIL that exists, thanks

apps/src/common.rs
```diff
@@ -29,6 +29,8 @@
 //! This module provides some utility functions that are common to quiche
 //! applications.
 
+use std::cmp;
```
LPardue, line 32: no need to use this, there's only 2 instances in the whole file so using the full def is fine.

apps/src/common.rs
```diff
@@ -438,8 +442,12 @@ impl Http09Conn {
 
 impl HttpConn for Http09Conn {
     fn send_requests(
-        &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
+        &mut self, conn: &mut quiche::Connection, client_args: &ClientArgs,
     ) {
+        let ClientArgs {
+            dump_response_path: target_path,
+            ..
+        } = client_args;
```
LPardue, line 450: This destructuring isn't helping much. Suggest you drop it and just access the fields where you need them e.g. client_args.dump_response_path

apps/src/common.rs
```diff
@@ -1114,18 +1125,35 @@ impl Http3Conn {
...
 impl HttpConn for Http3Conn {
     fn send_requests(
-        &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
+        &mut self, conn: &mut quiche::Connection, client_args: &ClientArgs,
     ) {
+        let ClientArgs {
+            dump_response_path: target_path,
+            reqs_concurrent,
+            ..
+        } = client_args;
```
LPardue, line 1134: Same comment here about destructuring, it causes some of the later code to use refs when it doesn't (and vice versa). Easier to just drop it for clarity.

apps/src/common.rs
```diff
@@ -1276,6 +1304,8 @@ impl HttpConn for Http3Conn {
 
                 Ok((_stream_id, quiche::h3::Event::Finished)) => {
                     self.reqs_complete += 1;
+                    let _ = self.reqs_in_flight.saturating_sub(1);
```
LPardue, line 1307:
```suggestion
                    self.reqs_in_flight = self.reqs_in_flight.saturating_sub(1);
```

Need to assign the value or else it won't do anything

quiche/src/lib.rs
```diff
@@ -7401,6 +7401,11 @@ impl Connection {
 
         Ok(pid)
     }
+
+    /// Get the peer's current max number of bidirectional streams.
+    pub fn get_peer_max_streams_bidirectional(&self) -> u64 {
+        self.streams.max_streams_bidi()
+    }
```
LPardue, line 7408: Sorry I should have explained more clearly. You don't want to do this since the number doesn't do what it sounds like. We should be checking `peer_streams_left_bidi()` which is already available.

apps/src/common.rs
```diff
@@ -1114,18 +1125,35 @@ impl Http3Conn {
...
-        &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
+        &mut self, conn: &mut quiche::Connection, client_args: &ClientArgs,
     ) {
+        let ClientArgs {
+            dump_response_path: target_path,
+            reqs_concurrent,
+            ..
+        } = client_args;
+
+        let max_streams_bidi = conn.get_peer_max_streams_bidirectional();
```
LPardue, line 1136: Due to the unique way that QUIC does things, this number doesn't match your expectations. Instead peer_streams_left_bidi will let you know how many more streams you _could_ open. And so you'll want to use that number in combination with the number of requests you want to make (not the number you have made) when deciding later to break or not.

LPardue (changes requested):

PR #1629, opened 2023-10-10
Flush qlog streamer before closing the connection

FranzBusch:
# Motivation
I was debugging some connection problems and enabled qlog; however, often I did not get output. My use-case is going through the FFI layer which instantiates a `BufferedWriter` for the file. After looking through the code a bit it appears that the `QLogStreamer` is not flushed on all close paths which can lead to missing logs.

# Modification
Make sure that on every close path the streamer is flushed and closed as well.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1625,21 +1625,6 @@ macro_rules! push_frame_to_pkt {
     }};
 }
 
-/// Conditional qlog actions.
-///
-/// Executes the provided body if the qlog feature is enabled and quiche
-/// has been configured with a log writer.
-macro_rules! qlog_with {
-    ($qlog:expr, $qlog_streamer_ref:ident, $body:block) => {{
-        #[cfg(feature = "qlog")]
-        {
-            if let Some($qlog_streamer_ref) = &mut $qlog.streamer {
-                $body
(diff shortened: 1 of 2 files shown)


FranzBusch:
@ghedo @LPardue I have updated the code to implement `Drop` for the `QLogStreamer` but I also manually set the streamer to `None` in some paths so that we flush as soon as possible. That's often really helpful when debugging something with QLog. The `Drop` implementation makes sure that we are going to flush in all the remaining cases.

quiche/src/lib.rs
```diff
@@ -7436,6 +7430,12 @@ impl Connection {
 
         Ok(pid)
     }
+
+    // Closes the connection and drops the QLogStreamer.
+    fn close_connection_and_drop_qlog_streamer(&mut self) {
+        self.qlog.streamer = None;
```
LPardue, line 7436:
```suggestion
        #[cfg(feature = "qlog")]
        self.qlog.streamer = None;
```

This object is optional based on a feature, so we need to check for that

LPardue (approved):
LGTM now, thanks.

We'll probably hold on merging just this moment in order to batch up some other qlog-reated changes.

PR #1630, opened 2023-10-12
Various fuzzing updates

ghedo:
See commit list (you can probably skip the last one...).


ghedo:
FTR, the GitGuardian report above is a false positive (the cert/key pair is just used for fuzzing and is self-signed).

LPardue (approved):

PR #1634, opened 2023-10-18
Introduces preferred addressing to the Quiche apps server and client examples

Clendenin:
This PR:
* Introduces the preferred addressing feature, proposed here https://github.com/cloudflare/quiche/pull/1502, into the Quiche server app and client app examples
* Is a superset of that PR since it is dependent on it. Therefore it contains all the commits from it. All new code exclusively relevant to this PR can be found in the large squashed commit found here: c20995820550c601b13cc92ab6edb1c1cb477547 and any subsequent commits to resolve merge/rebase issues.
* Based on this original PR by another contributor: https://github.com/cloudflare/quiche/pull/1208
* Addresses issue: https://github.com/cloudflare/quiche/issues/828

To reproduce:
* Run server by: `RUST_LOG=trace cargo run --bin quiche-server -- --cert apps/src/bin/cert.crt --key apps/src/bin/cert.key --enable-active-migration --max-active-cids 4 --preferred-address 127.0.0.1:4435`
* Run client by: `RUST_LOG=trace cargo run --bin quiche-client -- --no-verify --enable-active-migration --perform-migration --max-active-cids 4 https://127.0.0.1:4433`
* The Quiche Client App initially migrates to the servers preferred address (`127.0.0.1:4435`). It then performs a subsequent migration to a new port of its own.
* Confirm migration by seeing path validations, examples below, and other logs stating the number of written bytes on your migrated path.
Screen Shot 2023-10-18 at 9 31 33 AM
Screen Shot 2023-10-18 at 9 30 10 AM
Screen Shot 2023-10-18 at 9 31 00 AM

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -821,7 +821,6 @@ fn validate_token<'a>(
 
 fn handle_path_events(client: &mut Client) {
     while let Some(qe) = client.conn.path_event_next() {
-        trace!("There was an event");
         match qe {
             // The Client may be trying to probe the new preferred path given to
             // it by the server.
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -225,7 +225,6 @@ fn main() {
             false => clients.values().filter_map(|c| c.conn.timeout()).min(),
         };
 
-        trace!("Polling the event loop for I/O events");
         let mut poll_res = poll.poll(&mut events, timeout);
         while let Err(e) = poll_res.as_ref() {
             if e.kind() == std::io::ErrorKind::Interrupted {
@@ -736,8 +735,6 @@ fn main() {
         // Part 3/3:
         // Garbage collect closed connections.
         clients.retain(|_, ref mut c| {
-            trace!("Collecting garbage");
-
             if c.conn.is_closed() {
                 info!(
                     "{} connection collected {:?} {:?}",
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -92,7 +92,7 @@ fn main() {
         // Set SO_TXTIME socket option on the listening UDP socket for pacing
         // outgoing packets.
         if !args.disable_pacing {
-            match set_txtime_sockopt(&socket) {
+            match set_txtime_sockopt(socket) {
                 Ok(_) => {
                     pacing = true;
                     debug!("successfully set SO_TXTIME socket option");
@@ -696,7 +696,7 @@ fn main() {
                 }
 
                 if let Err(e) = send_to(
-                    &socket,
+                    socket,
                     &out[..total_write],
                     &dst_info.unwrap(),
                     client.max_datagram_size,
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -249,15 +249,15 @@ fn main() {
         // until there are no more packets to read.
         for event in &events {
             // Match the event to the correct socket
-            let socket = match event.token() {
+            let (socket, socket_name) = match event.token() {
                 mio::Token(0) => {
                     trace!("Event for starting socket occurred");
-                    &starting_socket
+                    (&starting_socket, "starting")
                 },
 
                 mio::Token(1) => {
                     trace!("Event for preferred socket occured: {:?}", event);
-                    &preferred_socket
+                    (&preferred_socket, "preferred")
                 },
 
                 _ => unreachable!(),
@@ -265,7 +265,7 @@ fn main() {
 
             let local_addr = socket.local_addr().unwrap();
 
-            trace!("Read loop starting up for {:?}", socket);
+            trace!("Read loop starting up for {:?} socket {:?}", socket_name, socket);
             'read: loop {
                 let (len, from) = match socket.recv_from(&mut buf) {
                     Ok(v) => v,
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -30,7 +30,9 @@ extern crate log;
 use std::io;
 
 use std::net::SocketAddr;
-use std::net::{self,};
+use std::net::{
+    self,
+};
 
 use std::io::prelude::*;
 
@@ -620,11 +622,7 @@ fn main() {
                 while client.conn.scids_left() > 0 {
                     trace!("Generating new source cids");
                     let (scid, reset_token) = generate_cid_and_reset_token(&rng);
-                    if client
-                        .conn
-                        .new_scid(&scid, reset_token, false)
-                        .is_err()
-                    {
+                    if client.conn.new_scid(&scid, reset_token, false).is_err() {
                         break;
                     }
 
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -29,7 +29,8 @@ extern crate log;
 
 use std::io;
 
-use std::net;
+use std::net::SocketAddr;
+use std::net::{self,};
 
 use std::io::prelude::*;
 
@@ -65,37 +66,61 @@ fn main() {
     let conn_args = CommonArgs::with_docopt(&docopt);
     let args = ServerArgs::with_docopt(&docopt);
 
+    let preferred_address: SocketAddr = args
+        .preferred_address
+        .parse()
+        .expect("Unable to parse preferred address.");
+
     // Setup the event loop.
     let mut poll = mio::Poll::new().unwrap();
     let mut events = mio::Events::with_capacity(1024);
 
     // Create the UDP listening socket, and register it with the event loop.
-    let mut socket =
+    let mut starting_socket =
         mio::net::UdpSocket::bind(args.listen.parse().unwrap()).unwrap();
 
-    // Set SO_TXTIME socket option on the listening UDP socket for pacing
-    // outgoing packets.
-    if !args.disable_pacing {
-        match set_txtime_sockopt(&socket) {
-            Ok(_) => {
-                pacing = true;
-                debug!("successfully set SO_TXTIME socket option");
-            },
-            Err(e) => debug!("setsockopt failed {:?}", e),
-        };
+    // Create another preferred UDP listening socket that can be used for server
+    // migration.
+    let mut preferred_socket =
+        mio::net::UdpSocket::bind(preferred_address).unwrap();
+
+    let sockets = vec![&starting_socket, &preferred_socket];
+
+    for socket in sockets {
+        // Set SO_TXTIME socket option on the listening UDP socket for pacing
+        // outgoing packets.
+        if !args.disable_pacing {
+            match set_txtime_sockopt(&socket) {
+                Ok(_) => {
+                    pacing = true;
+                    debug!("successfully set SO_TXTIME socket option");
+                },
+                Err(e) => debug!("setsockopt failed {:?}", e),
+            };
+        }
     }
 
-    info!("listening on {:}", socket.local_addr().unwrap());
+    info!("listening on {:}", starting_socket.local_addr().unwrap());
+    info!("listening on {:}", preferred_socket.local_addr().unwrap());
 
+    // Register starting socket with event loop
     poll.registry()
-        .register(&mut socket, mio::Token(0), mio::Interest::READABLE)
+        .register(&mut starting_socket, mio::Token(0), mio::Interest::READABLE)
+        .unwrap();
+    // Register preferred socket with event loop
+    poll.registry()
+        .register(
+            &mut preferred_socket,
+            mio::Token(1),
+            mio::Interest::READABLE,
+        )
         .unwrap();
 
     let max_datagram_size = MAX_DATAGRAM_SIZE;
     let enable_gso = if args.disable_gso {
         false
     } else {
-        detect_gso(&socket, max_datagram_size)
+        detect_gso(&starting_socket, max_datagram_size)
     };
 
     trace!("GSO detected: {}", enable_gso);
@@ -128,6 +153,8 @@ fn main() {
     config.set_max_stream_window(conn_args.max_stream_window);
 
     config.enable_pacing(pacing);
+
+    let rng = SystemRandom::new();
 
     let mut keylog = None;
 
@@ -163,20 +190,21 @@ fn main() {
         config.enable_dgram(true, 1000, 1000);
     }
 
-    let rng = SystemRandom::new();
     let conn_id_seed =
         ring::hmac::Key::generate(ring::hmac::HMAC_SHA256, &rng).unwrap();
 
     let mut next_client_id = 0;
+    // Connection IDs mapped to client IDs
     let mut clients_ids = ClientIdMap::new();
+
+    // The clients being served by this server
     let mut clients = ClientMap::new();
 
     let mut pkt_count = 0;
 
     let mut continue_write = false;
 
-    let local_addr = socket.local_addr().unwrap();
-
+    trace!("Main server loop starting up");
     loop {
         // Find the shorter timeout from all the active connections.
         //
@@ -187,6 +215,7 @@ fn main() {
             false => clients.values().filter_map(|c| c.conn.timeout()).min(),
         };
 
+        trace!("Polling the event loop for I/O events");
         let mut poll_res = poll.poll(&mut events, timeout);
         while let Err(e) = poll_res.as_ref() {
             if e.kind() == std::io::ErrorKind::Interrupted {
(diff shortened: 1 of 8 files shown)


quiche/src/lib.rs
```diff
@@ -8306,7 +8426,26 @@ impl TransportParams {
...
+                /// Stateless Reset Token: 16 bytes
+                const PREFERRED_ADDRESS_PARAM_MAX_SIZE: usize = 61;
+
+                let mut buffer = [0; PREFERRED_ADDRESS_PARAM_MAX_SIZE];
+                let written_to_buffer = PreferredAddressParams::encode(
+                    preferred_address_params,
+                    &mut buffer,
+                )?;
+                TransportParams::encode_param(&mut b, 0x000d, written_to_buffer)?;
+                b.put_bytes(&buffer[0..written_to_buffer])?;
```
Clendenin, line 8446: I guess we have to alloc anyway, so maybe just do that inside PreferredAddressParams::encode and return the buffer there? - Lucas

quiche/src/lib.rs
```diff
@@ -8306,7 +8426,26 @@ impl TransportParams {
             TransportParams::encode_param(&mut b, 0x000c, 0)?;
         }
 
-        // TODO: encode preferred_address
+        if is_server {
+            if let Some(preferred_address_params) = &tp.preferred_address_params {
+                /// IPv4: 4 bytes
+                /// Port: 2 bytes
+                /// IPv6: 16 bytes
```
Clendenin, line 8433: nit: these should use double quotes //

apps/src/client.rs
```diff
@@ -485,15 +520,79 @@ pub fn connect(
...
+                // Manually track the new Destination Connection ID and reset
+                // token of the preferred address.
+                conn.new_destination_cid(
+                    preferred_address_params.connection_id,
+                    1,
+                    preferred_address_params.stateless_reset_token,
+                    0,
+                    &mut SmallVec::new(),
+                )
+                .unwrap();
```
Clendenin, line 544: Switch from manually adding to an automatic process to simply use - Lucas

apps/src/bin/quiche-server.rs
```diff
@@ -65,37 +68,61 @@ fn main() {
...
-            Ok(_) => {
-                pacing = true;
-                debug!("successfully set SO_TXTIME socket option");
-            },
-            Err(e) => debug!("setsockopt failed {:?}", e),
-        };
+    // Create another preferred UDP listening socket that can be used for server
+    // migration.
+    let mut preferred_socket =
+        mio::net::UdpSocket::bind(preferred_address).unwrap();
```
Clendenin, line 87: Can this preferred socket take new connections not just migrated connections?
If a new connection starts here, is the preferred address provided over the transport params? What does this cause? - Lucas

Clendenin: Error Result if preferred address is same as current address - fail on startup - Lucas

quiche/src/cid.rs
```diff
@@ -425,10 +425,6 @@ impl ConnectionIdentifiers {
     ///
     /// Returns an error if the provided Connection ID or its metadata are
     /// invalid.
-    ///
-    /// Returns a list of tuples (DCID sequence number, Path ID), containing the
-    /// sequence number of retired DCIDs that were linked to their respective
-    /// Path ID.
```
LPardue, line 431: This comment has been deleted but there was no code change. Is the comment actually wrong (and we can just fix that up in a quick standalone PR) or is there something else at hand?

Clendenin: The comment was out of date. iirc the function was changed but the comment was never updated.

apps/src/client.rs
```diff
@@ -432,11 +461,17 @@ pub fn connect(
                 quiche::PathEvent::New(..) => unreachable!(),
 
                 quiche::PathEvent::Validated(local_addr, peer_addr) => {
+                    trace!("PathEvent::Validated");
                     info!(
                         "Path ({}, {}) is now validated",
                         local_addr, peer_addr
                     );
-                    conn.migrate(local_addr, peer_addr).unwrap();
+                    if migrated {
```
LPardue, line 469: Would adding logs for these option help?

quiche/src/lib.rs
```diff
@@ -8388,13 +8554,144 @@ impl TransportParams {
...
+
+        let cid = b.get_bytes_with_varint_length()?.to_vec().into();
+        let stateless_reset_token = u128::from_be_bytes(
+            b.get_bytes(16)?.to_vec().try_into().unwrap(), /* Since a 128 bit
+                                                            * vector has just
+                                                            * successfully
+                                                            * been created
+                                                            * above, this
+                                                            * should never
+                                                            * panic. */
```
LPardue, line 8605: Please convert into a comment on the line above

quiche/src/lib.rs
```diff
@@ -8388,13 +8554,144 @@ impl TransportParams {
...
+        let mut addr_v4 = None;
+        let ip_v4_b = <[u8; 4]>::try_from(b.slice(4)?)
+            .map_err(|_| Error::InvalidTransportParam)?;
+        b.skip(4)?;
+        let ip_v4 = Ipv4Addr::from(ip_v4_b);
+
+        let port_v4 = b.get_u16()?;
+
+        if ip_v4.is_unspecified() ^ (port_v4 == 0) {
+            return Err(Error::InvalidTransportParam);
```
LPardue, line 8626: I think this is trying to enforce https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.32.1 ?

> Servers MAY choose to only send a preferred address of one address family by sending an all-zero address and port (0.0.0.0:0 or [::]:0) for the other family. IP addresses are encoded in network byte order.

My concern is that this is called in the top-level `[decode()](https://github.com/cloudflare/quiche/pull/1634/files#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R8589)` function and would early abort something that is not technicall invalid

apps/src/client.rs
```diff
@@ -485,15 +520,79 @@ pub fn connect(
...
+                    &mut SmallVec::new(),
+                )
+                .unwrap();
+
+                // Probe the path. If it successfully validates then the event
+                // handler will see the `PathEvent::Validated`
+                // event occur and begin migration. The server will see
+                // `PathEvent::PeerMigrated` upon successful
+                // migration.
+                if let Some(addr_v4) = preferred_address_params.addr_v4 {
```
LPardue, line 551: The selection of which preffered address to use is subjective - this is just an example. There's also bit of duplication across the branches that can be reduced. Consider extracting into a function like "select_and_probe_preferred_address"

apps/src/client.rs
```diff
@@ -485,15 +520,79 @@ pub fn connect(
...
+                    )
+                    .unwrap();
+                }
+
+                server_preferred_address_probed = true;
+            }
+        }
+
+        // Perform client side connection migration
+        if args.perform_migration &&
```
LPardue, line 580: Also consider making this its own function

quiche/src/cid.rs
LPardue: In that case, please create a separate PR with the change

quiche/src/frame.rs
```diff
@@ -1012,14 +1012,15 @@ impl Frame {
...
+            Frame::ApplicationClose { error_code, reason } => {
                 QuicFrame::ConnectionClose {
                     error_space: Some(ErrorSpace::ApplicationError),
                     error_code: Some(*error_code),
                     error_code_value: None, // raw error is no different for us
                     reason: Some(String::from_utf8_lossy(reason).into_owned()),
                     trigger_frame_type: None, // don't know trigger type
-                },
+                }
+            },
```
LPardue, line 1023: This seems like an unrelated formatting change. Let's drop it to simplify the PR

quiche/src/lib.rs
```diff
@@ -831,8 +840,7 @@ impl Config {
         Self::with_tls_ctx(version, tls::Context::new()?)
     }
 
-    /// Creates a config object with the given version and
-    /// [`SslContextBuilder`].
+    /// Creates a config object with the given version and [`SslContext`].
```
LPardue, line 843: unrelated change AFAICT, so better as a separate PR

quiche/src/lib.rs
```diff
@@ -1339,6 +1347,25 @@ impl Config {
...
         self.disable_dcid_reuse = v;
     }
+
+    /// Set the server's preferred address.
+    pub fn set_preferred_address(
+        &mut self, v4: Option<SocketAddrV4>, v6: Option<SocketAddrV6>,
+        connection_id: ConnectionId<'static>, stateless_reset_token: u128,
+    ) -> Result<()> {
+        if (v4.is_none() && v6.is_none()) || connection_id.is_empty() {
+            return Err(Error::TlsFail);
```
LPardue, line 1357: InvalidTransportParam is probably more intuitive for this error

quiche/src/lib.rs
```diff
@@ -1339,6 +1347,25 @@ impl Config {
     pub fn set_disable_dcid_reuse(&mut self, v: bool) {
         self.disable_dcid_reuse = v;
     }
+
+    /// Set the server's preferred address.
```
LPardue, line 1351: Let's add some further description of how we expect callers to use this function, with a link to the appropriate RFC section. Right now, it's too easy to do the wrong thing (v4 and v6 None, or empty CID) and get an error result that doesn't explain why.

quiche/src/lib.rs
```diff
@@ -1848,13 +1875,21 @@ impl Connection {
...
-        let ids = cid::ConnectionIdentifiers::new(
+        let mut ids = cid::ConnectionIdentifiers::new(
             config.local_transport_params.active_conn_id_limit as usize,
             scid,
             active_path_id,
             reset_token,
         );
 
+        // If we aren't using zero length connection id assign an SCID to the
+        // preferred address transport argument if one is used.
```
LPardue, line 1886:
```suggestion
        // If we aren't using a zero-length SCID, assign a new one to server's preferred address.
```

quiche/src/lib.rs
```diff
@@ -8388,13 +8554,144 @@ impl TransportParams {
...
+    }
+
+    /// Encode the preferred address parameters into a buffer.
+    ///
+    /// # Arguments
+    ///
+    /// * `preferred_address_params` - A reference to the preferred address
+    ///   parameters to be encoded.
+    /// * `buffer` - A reference to a buffer to encode the
+    ///   `preferred_address_params` into.
```
LPardue, line 8662: This comment doesn't seem particularly useful, especially since it's internal code

quiche/src/lib.rs
```diff
@@ -9009,6 +9425,445 @@ mod tests {
         );
     }
 
+    #[test]
```
LPardue, line 9428: Can we add some tests for contructing different types of bad server pereferred address too. I want to ensure that such things are caught as early as possible and that any future changes can be caught in a regression test before they hit apps.

apps/src/bin/quiche-server.rs
```diff
@@ -65,37 +68,61 @@ fn main() {
...
+                    debug!("successfully set SO_TXTIME socket option");
+                },
+                Err(e) => debug!("setsockopt failed {:?}", e),
+            };
+        }
     }
 
-    info!("listening on {:}", socket.local_addr().unwrap());
+    info!("listening on {:}", starting_socket.local_addr().unwrap());
+    info!("listening on {:}", preferred_socket.local_addr().unwrap());
```
LPardue, line 106: may as well update the log to state the kind of each socket

apps/src/bin/quiche-server.rs
```diff
@@ -187,129 +217,114 @@ fn main() {
             false => clients.values().filter_map(|c| c.conn.timeout()).min(),
         };
 
+        trace!("Polling the event loop for I/O events");
```
LPardue, line 220: this will be too spammy

apps/src/bin/quiche-server.rs
```diff
@@ -187,129 +217,114 @@ fn main() {
...
-            if let Some(target_path) = conn_args.dump_packet_path.as_ref() {
-                let path = format!("{target_path}/{pkt_count}.pkt");
+            let local_addr = socket.local_addr().unwrap();
 
-                if let Ok(f) = std::fs::File::create(path) {
-                    let mut f = std::io::BufWriter::new(f);
-                    f.write_all(pkt_buf).ok();
-                }
-            }
+            trace!("Read loop starting up for {:?}", socket);
```
LPardue, line 253: are we able to qualify which socket is being used here, or elsewhere? I think that will be useful to avoid any confusion

apps/src/bin/quiche-server.rs
```diff
@@ -187,129 +217,114 @@ fn main() {
...
-                if !args.no_retry {
-                    // Token is always present in Initial packets.
-                    let token = hdr.token.as_ref().unwrap();
+                // Lookup a connection based on the packet's connection ID. If
+                // there is no connection matching, create a new
+                // one.
+                let client = if !clients_ids.contains_key(&hdr.dcid) &&
+                    !clients_ids.contains_key(&conn_id)
+                {
+                    trace!("A new connection!");
```
LPardue, line 316: If we want to keep a log line for this, lets change what we print, for example to state the ID of the new connection, the source IP or something. It might be interesting to upgrade it to info level too

apps/src/bin/quiche-server.rs
```diff
@@ -326,289 +341,383 @@ fn main() {
...
-                    max_send_burst: MAX_BUF_SIZE,
-                };
+                    // Track the preferred address connection ID with the client
+                    // it was given to
+                    clients_ids.insert(preferred_address_scid, client_id);
+
+                    // Create the new connection!
+                    #[allow(unused_mut)]
+                    // Automatically provisions SCID for preferred address.
+                    let mut conn = quiche::accept(
```
LPardue, line 440: Is this a hangover? if we don't need the `mut` then lets not declare it as such

apps/src/bin/quiche-server.rs
```diff
@@ -326,289 +341,383 @@ fn main() {
...
-            };
+                    next_client_id += 1;
 
-            let recv_info = quiche::RecvInfo {
-                to: local_addr,
-                from,
-            };
+                    clients.get_mut(&client_id).unwrap()
+                } else {
+                    trace!("We've seen this connection before!");
```
LPardue, line 490: seems quite spammy

apps/src/bin/quiche-server.rs
```diff
@@ -690,7 +799,10 @@ fn validate_token<'a>(
 
 fn handle_path_events(client: &mut Client) {
     while let Some(qe) = client.conn.path_event_next() {
+        trace!("There was an event");
```
LPardue, line 802: Seems not useful in it's current form :smile:

apps/src/bin/quiche-server.rs
```diff
@@ -744,6 +856,7 @@ fn handle_path_events(client: &mut Client) {
             },
 
             quiche::PathEvent::PeerMigrated(local_addr, peer_addr) => {
+                trace!("PathEvent::PeerMigrated");
```
LPardue, line 859: seems duplicated by the info log below

quiche/src/lib.rs
Clendenin: I believe my original intention was to have these serve as "Documentation Comments" to the constant below. This is what it looks like when the documentation for the co
Screenshot 2024-07-30 at 11 57 40 AM
nstant is pulled up in my editor:

apps/src/bin/quiche-server.rs
Clendenin: Yeah that is a great question.
For the example app the app will accept connections starting on the servers preferred address. The server will still encode a "preferred address" transport parameter into the transport parameters it initially sends to the client. The client will receive the transport parameters and begin decoding them. It will see an encoded preferred address transport parameter and begin decoding it. The example app will then initiate `probe_path()` on this "new" preferred address. The `probe_path()` logic sees that this is an existing path. It will generate a Path_Challenge frame and then it does not do path migration since there in no migration to perform.

This separation of concerns makes sense to me. The probe_path / connection migration code should be resilient enough to not attempt migrations to the path it is already on.

Screenshot 2024-08-01 at 2 07 11 PM

apps/src/client.rs
```diff
@@ -264,14 +287,17 @@ pub fn connect(
...
-            trace!("timed out");
+            trace!("Event loop reporting no events - timed out.");
 
             conn.on_timeout();
         }
 
         // Read incoming UDP packets from the socket and feed them to quiche,
         // until there are no more packets to read.
         for event in &events {
+            trace!("Event occured: {:?}", event);
```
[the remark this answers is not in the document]
Clendenin: I am confused. Isn't this file just to demonstrate how to use the code? I left this in because it is helpful to see what event occurred and when.

apps/src/client.rs
Clendenin: added 👍

quiche/src/lib.rs
```diff
@@ -6055,6 +6136,18 @@ impl Connection {
         Ok(dcid_seq)
     }
 
+    /// Returns the server's preferred address transport parameters which can be
+    /// used to perform a connection migration.
+    pub fn server_preferred_address_params(
+        &mut self,
```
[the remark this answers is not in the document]
Clendenin: The internal call to `new_destination_cid()` requires it be mutable

quiche/src/lib.rs
Clendenin: Done 👍

quiche/src/lib.rs
Clendenin: If the IPv4 address is unspecified (all 0s) and the associated port is also unspecified (all 0s) then the XOR operation will fail and this `decode_v4` function will return Ok(None)

apps/src/client.rs
Clendenin: I coalesced the branches into one to dry the code.

apps/src/client.rs
Clendenin: I don't quite follow, what is your reasoning to break this out into a function?

quiche/src/frame.rs
Clendenin: Removed

quiche/src/lib.rs
Clendenin: Removed

quiche/src/lib.rs
Clendenin: Updated 👍

quiche/src/lib.rs
Clendenin: Good call! Added 👍

quiche/src/lib.rs
Clendenin: Updated 👍

quiche/src/lib.rs
Clendenin: removed

apps/src/bin/quiche-server.rs
Clendenin: Done 👍

apps/src/bin/quiche-server.rs
Clendenin: removed 👍 , along with another one that was also spammy.

apps/src/bin/quiche-server.rs
Clendenin: Added the socket name to the trace 👍

apps/src/bin/quiche-server.rs
Clendenin: Further down we have
`trace!("New connection: dcid={:?} scid={:?}", hdr.dcid, scid);`
I have changed that trace!() to:
`info!("A new connection has occurred. dcid={:?} scid={:?}", hdr.dcid, scid);`
and removed:
`trace!("A new connection!");`
I have also updated a later trace!() logging repeat connections to log out:
`info!("A repeat connection has occurred. Client ID={:?}", client_id);`

apps/src/bin/quiche-server.rs
Clendenin: `conn.set_keylog(Box::new(keylog));` needs a mutable ref

apps/src/bin/quiche-server.rs
Clendenin: Updated as mentioned above. It is not too bad. Only shows up a couple of times.

apps/src/bin/quiche-server.rs
Clendenin: lol true, removed 👍

apps/src/bin/quiche-server.rs
Clendenin: True, removed 👍

quiche/src/lib.rs
Clendenin: Yes. I refactored the existing `server_preferred_address_rejects_zero_len_cid()` test and created a new one called `server_preferred_address_rejects_empty_v4_and_empty_v6_preferred_addresses()`

quiche/src/lib.rs
Clendenin: Lucas: "Switch these to regular //"

apps/src/bin/quiche-server.rs
Clendenin: Lucas: "Add a comment to explicitly state that this is intentional and not a bug"
Lucas to follow up with working group to further discuss.

quiche/src/lib.rs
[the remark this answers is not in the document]
Clendenin: Lucas: "Double check to see if you can remove this"

apps/src/client.rs
Clendenin: Lucas: "Put this into a function to make it more obvious to others"

apps/src/client.rs
Clendenin: Lucas: "Split it out to enforce soc"

kckeiks:
@Clendenin thank you for including my commits 👍

PR #1636, opened 2023-10-24
Consider only data in the network, not data buffered in host due to pacing at lower layers

lohith-bellad:
When using packet pacing at the lower layers there can be less data in the network than in_flight data. This is due to buffering the packets at lower layers until the sent time of the packet. When using BBR as the CC algorithm and when we are checking the condition to exit the drain phase we need to consider only the data which is in the network and not the data buffered in the host. If we consider the inflight(host buffered
+ in the network) data then we exit the drain phase later than expected leading to low throughput.

### DIFF
--- a/quiche/src/recovery/bbr/mod.rs
+++ b/quiche/src/recovery/bbr/mod.rs
@@ -605,7 +605,7 @@ mod tests {
         }
 
         // Stop at right before filled_pipe=true.
-        for _ in 0..5 {
+        for _ in 0..6 {
             let pkt = Sent {
                 pkt_num: pn,
                 frames: smallvec![],
@@ -640,9 +640,9 @@ mod tests {
 
         let mut acked = ranges::RangeSet::default();
 
-        // We sent 5 packets, but ack only one, to stay
+        // We sent 6 packets, but ack only one, to stay
         // in Drain state.
-        acked.insert(0..pn - 4);
+        acked.insert(0..pn - 5);
 
         assert_eq!(
             r.on_ack_received(
--- a/quiche/src/recovery/bbr/per_ack.rs
+++ b/quiche/src/recovery/bbr/per_ack.rs
@@ -257,11 +257,18 @@ fn bbr_check_drain(r: &mut Recovery, now: Instant) {
     }
 
     if r.bbr_state.state == BBRStateMachine::Drain &&
-        r.bytes_in_flight <= bbr_inflight(r, 1.0)
+        bbr_bytes_in_net(r, now) <= bbr_inflight(r, 1.0)
     {
         // we estimate queue is drained
         bbr_enter_probe_bw(r, now);
     }
+}
+
+fn bbr_bytes_in_net(r: &mut Recovery, now: Instant) -> usize {
+    let buffered = r.get_host_buffered(now);
+    let in_flight = r.bytes_in_flight;
+
+    in_flight.saturating_sub(buffered)
 }
 
 // 4.3.4.3.  Gain Cycling Algorithm
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -400,6 +400,24 @@ impl Recovery {
         trace!("{} {:?}", trace_id, self);
     }
 
+    // Limit data buffered in host send buffer to avoid bufferbloat
+    // at the sender.
+    pub fn get_host_buffered(&mut self, now: Instant) -> usize {
+        let mut buffered: usize = 0;
+
+        for epoch in 0..packet::Epoch::count() {
+            for bytes in self.sent[epoch].iter()
+                // Skip packets that have already been sent, acked or lost, or that are not in-flight.
+                .filter(|p| p.in_flight && p.time_acked.is_none() && p.time_lost.is_none() && p.time_sent > now)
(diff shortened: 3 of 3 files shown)


quiche/src/recovery/bbr/per_ack.rs
```diff
@@ -256,14 +256,19 @@ fn bbr_check_drain(r: &mut Recovery, now: Instant) {
         bbr_enter_drain(r);
     }
 
+    println!("est = {:?}", bbr_inflight(r, 1.0));
```
LPardue, line 259: nit: leftover

lohith-bellad: fixed it!

quiche/src/recovery/mod.rs
```diff
@@ -400,6 +400,22 @@ impl Recovery {
         trace!("{} {:?}", trace_id, self);
     }
 
+    // Limit data buffered in host send buffer to avoid bufferbloat
+    // at the sender.
+    pub fn get_host_buffered(&mut self, now: Instant) -> usize {
+        let mut buffered: usize = 0;
+
+        for epoch in 0..packet::Epoch::count() {
+            for pkt in self.sent[epoch].iter() {
```
ghedo, line 409: This needs to exclude packets that have already been acked or lost as they are not immediately removed from the list, as well as include only packets that are actually "in-flight".

I think the inner loop can be replaced with something like:
```rust
            buffered += sent
                // Skip packets that have already been sent, acked or lost, or that are not in-flight.
                .filter(|p| p.in_flight && p.time_acked.is_none() && p.time_lost.is_none() && p.time_sent > now)
                .map(|x| x + p.size);
```
(not entirely sure it actually compiles though)

quiche/src/recovery/mod.rs
```diff
@@ -2262,6 +2278,174 @@ mod tests {
...
+        // Send the third packet out.
+        let p = Sent {
+            pkt_num: 3,
+            frames: smallvec![],
+            time_sent: now,
+            time_acked: None,
+            time_lost: None,
+            size: 5000,
+            ack_eliciting: true,
+            in_flight: true,
```
ghedo, line 2427: We should also add checks for a packet that is not in-flight.

quiche/src/recovery/mod.rs
```diff
@@ -2262,6 +2278,174 @@ mod tests {
...
+            HandshakeStatus::default(),
+            now,
+            "",
+        );
+
+        assert_eq!(r.sent[packet::Epoch::Application].len(), 2);
+        assert_eq!(r.bytes_in_flight, 12000);
+        assert_eq!(r.get_host_buffered(now), 0);
+
+        // Send the third packet out.
```
ghedo, line 2418:
```suggestion
        // Send the fourth packet out.
```

quiche/src/recovery/mod.rs
```diff
@@ -2262,6 +2278,174 @@ mod tests {
...
+        let mut cfg = crate::Config::new(crate::PROTOCOL_VERSION).unwrap();
+        cfg.set_cc_algorithm(CongestionControlAlgorithm::CUBIC);
+
+        let mut r = Recovery::new(&cfg);
+
+        let mut now = Instant::now();
+
+        assert_eq!(r.sent[packet::Epoch::Application].len(), 0);
+
+        // send out first packet (a full initcwnd).
```
ghedo, line 2293:
```suggestion
        // Send out the first packet (a full initcwnd).
```

ghedo (changes requested):

quiche/src/recovery/bbr/mod.rs
```diff
@@ -303,10 +303,10 @@ fn on_packets_acked(
         packets.drain(..).fold(0, |acked_bytes, p| {
             r.bbr_state.prior_bytes_in_flight = r.bytes_in_flight;
 
-            per_ack::bbr_update_model_and_state(r, &p, now);
-
             r.bytes_in_flight = r.bytes_in_flight.saturating_sub(p.size);
 
+            per_ack::bbr_update_model_and_state(r, &p, now);
+
```
junhochoi, line 309: This might change the result of functions using bytes_in_flight in bbr_update_model_and_state(). Is this a necessary change?

quiche/src/recovery/mod.rs
```diff
@@ -2262,6 +2280,204 @@ mod tests {
             now + Duration::from_secs_f64(12000.0 / pacing_rate as f64)
         );
     }
+
+    #[test]
+    fn test_host_buffered() {
+        let mut cfg = crate::Config::new(crate::PROTOCOL_VERSION).unwrap();
+        cfg.set_cc_algorithm(CongestionControlAlgorithm::CUBIC);
```
junhochoi, line 2287: Should we test it with BBR CC?

junhochoi:
It seems this PR ports some of linux bbr code (`bbr_packets_in_net_at_edt()`)? If so, it is used in 2 places including this change, you might want to take a look at https://github.com/torvalds/linux/blob/611da07b89fdd53f140d7b33013f255bf0ed8f34/net/ipv4/tcp_bbr.c#L569C2-L569C2 too (bbr_is_next_cycle_phase).

PR #1640, opened 2023-10-31
crypto: avoid unnecessary allocations when generating keys

ghedo:
In some cases we are turning Vec values into slices to pass them to functions that will then turn the slice back into a Vec.

Instead just pass `Vec`s around to avoid some unnecessary allocations.

### DIFF
--- a/quiche/src/crypto.rs
+++ b/quiche/src/crypto.rs
@@ -138,28 +138,29 @@ pub struct Open {
 
 impl Open {
     pub fn new(
-        alg: Algorithm, key: &[u8], iv: &[u8], hp_key: &[u8], secret: &[u8],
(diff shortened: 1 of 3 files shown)


rushilmehra (approved):

quiche/src/crypto.rs
```diff
@@ -155,7 +156,7 @@ impl Open {
         Ok(Open {
             alg: aead,
 
-            secret: Vec::from(secret),
```
evanrittenhouse, line 158: If we're passing a `&[u8]` as the parameter, wouldn't we still need `Vec::from(secret)`? I think we could switch `secret` to `&[u8]` and then pass `&Vec<u8>` as the parameter

ghedo: Yeah, there were some changes missing.

LPardue (approved):

evanrittenhouse (approved):

Issue #1513, opened 2023-05-16
Wrong RTT being calculated when pacing is turned on.


kalradivyanshu:
We were debugging some issues, and found a very peculiar behaviour in quiche. Quiche underreports RTT by a lot if pacing is turned on. We have a server in mumbai and one in delhi, the mumbai server is the quiche server run with:
```
QLOGDIR=./logs RUST_LOG=info cargo run --bin quiche-server --release -- --cert apps/src/bin/cert.crt --key apps/src/bin/cert.key --listen 0.0.0.0:4434 --cc-algorithm bbr --root apps/src/bin/root
```
The ping command shows the ping as 30 - 32ms:
![image](https://github.com/cloudflare/quiche/assets/12642750/c2a98ee6-a629-4809-a817-6f21180a2311)

running the server with `--disable-pacing` and then printing the path min_rtt every 200ms shows:
![image](https://github.com/cloudflare/quiche/assets/12642750/108135e4-1eac-4d47-83ba-28a46f8b5b57)

we printed this by add the following snippet:
```rust
 if last_stat_at.elapsed() > Duration::from_millis(200) {
    for stat in client.conn.path_stats() {
        log::info!("path_min_rtt: {:?}", stat.min_rtt);
    }
    last_stat_at = std::time::Instant::now();
}
```

But when we remove `--disable-pacing` the RTT is reported as 1.1 ms or even lower:

![image](https://github.com/cloudflare/quiche/assets/12642750/e03b43df-e626-4b48-b84f-5b3b261a43df)

The SO_TXTIME socket is being set:
![image](https://github.com/cloudflare/quiche/assets/12642750/b70dd64a-53ea-4f69-84dd-5a699cbae6ca)

Steps to reproduce:
1. Clone https://github.com/cloudflare/quiche in two servers with 20ms or more RTT between them.
2. run `QLOGDIR=./logs RUST_LOG=info cargo run --bin quiche-server --release -- --cert apps/src/bin/cert.crt --key apps/src/bin/cert.key --listen 0.0.0.0:4434 --cc-algorithm bbr --root apps/src/bin/root` in one of them
3. run `cargo run --bin quiche-client -- https://ip:4434/index.html --no-verify` in the other
4. check the value of rtt min either via logging it in the server or by qlog
[...]

PR #1644, opened 2023-11-07
Update README.md to mention that pacing is required for BBR (e.g., on Linux use fq instead of fq_codel)

joergdeutschmann-i7:
It is mentioned in #1180 but could be stated explicitly in my opinion. Also see #1513.

Or maybe add a note to the [argument parser](https://github.com/cloudflare/quiche/blob/c8d10d26db8994b96b475d85dfd557dddb3d7c6b/apps/src/args.rs#L278) but this seems wrong because it's not limited to that specific app.

Checking if `SO_TXTIME` is provided by the socket would be nice but I don't know if this is possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ the network.
 Applications can use these hints by artificially delaying the sending of
 packets through platform-specific mechanisms (such as the [`SO_TXTIME`]
 socket option on Linux), or custom methods (for example by using user-space
-timers).
+timers). Pacing is required for [BBR](quiche/src/recovery) (e.g., on Linux use `fq` instead of `fq_codel`).
 
 [pace]: https://datatracker.ietf.org/doc/html/rfc9002#section-7.7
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,7 +220,7 @@ the network.
 Applications can use these hints by artificially delaying the sending of
 packets through platform-specific mechanisms (such as the [`SO_TXTIME`]
 socket option on Linux), or custom methods (for example by using user-space
-timers).
+timers). Pacing is required for [BBR](quiche/src/recovery) (e.g., on Linux use `fq` instead of `fq_codel`).
```
LPardue, line 223:
```suggestion
timers). Pacing is required for [BBR](quiche/src/recovery) (e.g., on Linux
use `fq` instead of `fq_codel`).
```
nit: wrapping

LPardue (approved):

PR #1654, opened 2023-11-15
backstage: update ctaalog entry

ghedo:
The system key is not mandatory, and add BoringSSL to dependencies while at it.


LPardue:
Typo in commit message but otherwise LGTM

LPardue (approved):

PR #1655, opened 2023-11-15
Rename methods to be more consistent and deprecate the old ones

normanmaurer:
Motivation:

Some of the methods used *source_cid* and *destination_cid* in the name while most others uses *scid* and *dcid*. Naming of methods should be consistent to make it easier to discover things.

Modifications:

Add new methods with consistent naming and deprecate the old ones

Result:

Cleanup of public API

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -512,13 +512,9 @@ fn main() {
             }
 
             // Provides as many CIDs as possible.
-            while client.conn.source_cids_left() > 0 {
+            while client.conn.scids_left() > 0 {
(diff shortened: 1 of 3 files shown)


normanmaurer:
@ghedo @LPardue let me know what you think .... I noticed this while working on exposing a few methods via ffi that I need ...

normanmaurer:
@ghedo up to you ... I only kept them to no introduce breaking changes . If you think we should remove these let me
Know and I will update the PR

LPardue:
+1 to just breaking things now rather than later :D

LPardue (approved):

PR #1656, opened 2023-11-16
Add C api for methods that are required to handle connection migratio…

normanmaurer:
…n on the server side.

Motivation:

We need to expose some methods via the C api to be able to over the remote peer connection ids that it can be used for migration

Modifications:

Add missing ffi

Result:

Be able to handle connection migrations

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -678,6 +678,30 @@ ssize_t quiche_conn_send_ack_eliciting_on_path(quiche_conn *conn,
                            const struct sockaddr *local, socklen_t local_len,
                            const struct sockaddr *peer, socklen_t peer_len);
 
+// Returns true if there are retired source connection ids and fill the parameters
+bool quiche_conn_retired_scid_next(const quiche_conn *conn, const uint8_t **out, size_t *out_len);
+
+// Returns the number of source Connection IDs that are retired.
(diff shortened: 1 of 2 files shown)


normanmaurer:
Depending on if https://github.com/cloudflare/quiche/pull/1655 is accepted and merged we might want to rename some c functions here

quiche/include/quiche.h
```diff
@@ -678,6 +678,30 @@ ssize_t quiche_conn_send_ack_eliciting_on_path(quiche_conn *conn,
                            const struct sockaddr *local, size_t local_len,
                            const struct sockaddr *peer, size_t peer_len);
 
+// Returns true if there are retired source connection ids and fill the the parameters
```
FranzBusch, line 681:
```suggestion
// Returns true if there are retired source connection ids and fill the parameters
```

normanmaurer:
Seems like the test failure is not related... @LPardue @ghedo thoughts ?

ghedo:
The clippy error should be fixed by https://github.com/cloudflare/quiche/pull/1659

normanmaurer:
rebased and pushed

LPardue (dismissed):

normanmaurer:
@ghedo rebased

ghedo (approved):

PR #1660, opened 2023-11-20
Expose C api to handle PathEvent

normanmaurer:
Motivation:

There was no C api exposed to allow handling PathEvent.

Modifications:

Expose c functions to handle PathEvent

Result:

Be able to handle and consume PathEvent

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -701,6 +701,53 @@ size_t quiche_conn_active_scids(quiche_conn *conn);
 uint64_t quiche_conn_new_scid(quiche_conn *conn,
                            const uint8_t *scid, size_t scid_len,
                            const uint8_t *reset_token, bool retire_if_needed);
+enum quiche_path_event_type {
+    QUICHE_PATH_EVENT_NEW,
+    QUICHE_PATH_EVENT_VALIDATED,
+    QUICHE_PATH_EVENT_FAILED_VALIDATION,
+    QUICHE_PATH_EVENT_CLOSED,
+    QUICHE_PATH_EVENT_REUSED_SOURCE_CONNECTION_ID,
+    QUICHE_PATH_EVENT_PEER_MIGRATED,
+};
+
+typedef struct quiche_path_event quiche_path_event;
+
+// Retrieves the next event. Returns NULL if there is no event to process.
+const quiche_path_event *quiche_conn_path_event_next(quiche_conn *conn);
+
+// Returns the type of the event.
+enum quiche_path_event_type quiche_path_event_type(quiche_path_event *ev);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_NEW.
+void quiche_path_event_new(quiche_path_event *ev,
+                           struct sockaddr_storage *local, socklen_t *local_len, struct sockaddr_storage *peer, socklen_t *peer_len);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_VALIDATED.
+void quiche_path_event_validated(quiche_path_event *ev,
+                           struct sockaddr_storage *local, socklen_t *local_len, struct sockaddr_storage *peer, socklen_t *peer_len);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_FAILED_VALIDATION.
+void quiche_path_event_failed_validated(quiche_path_event *ev,
+                           struct sockaddr_storage *local, socklen_t *local_len, struct sockaddr_storage *peer, socklen_t *peer_len);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_CLOSED.
+void quiche_path_event_closed(quiche_path_event *ev,
+                           struct sockaddr_storage *local, socklen_t *local_len, struct sockaddr_storage *peer, socklen_t *peer_len);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_REUSED_SOURCE_CONNECTION_ID.
+void quiche_path_event_reused_source_connection_id(quiche_path_event *ev, uint64_t *id,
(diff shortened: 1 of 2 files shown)


normanmaurer:
I still need to test this a bit more but wanted to leave this here to understand if the API is ok or not /cc @ghedo @LPardue

quiche/include/quiche.h
```diff
@@ -678,6 +678,54 @@ ssize_t quiche_conn_send_ack_eliciting_on_path(quiche_conn *conn,
...
+    QUICHE_PATH_EVENT_FAILED_VALIDATION,
+    QUICHE_PATH_EVENT_CLOSED,
+    QUICHE_PATH_EVENT_REUSED_SOURCE_CONNECTION_ID,
+    QUICHE_PATH_EVENT_PEER_MIGRATED,
+};
+
+typedef struct quiche_path_event quiche_path_event;
+
+// Retrieves the next event. Returns true if there was an event and ev is et, false otherwise.
+bool quiche_conn_path_event_next(quiche_conn *conn, quiche_path_event **ev);
```
normanmaurer, line 693: @ghedo I wonder if it would be better to change it to:

```
quiche_path_event* quiche_conn_path_event_next(quiche_conn *conn);
```

The caller should then just check if it returned `NULL` or not.

quiche/include/quiche.h
```diff
@@ -678,6 +678,54 @@ ssize_t quiche_conn_send_ack_eliciting_on_path(quiche_conn *conn,
...
+
+// Returns the type of the event.
+enum quiche_path_event_type quiche_path_event_type(quiche_path_event *ev);
+
+// Should be called if the quiche_path_event_type(...) returns QUICHE_PATH_EVENT_NEW.
+void quiche_path_event_new(quiche_path_event *ev,
+                           struct sockaddr_storage *local, size_t *local_len, struct sockaddr_storage *peer, size_t *peer_len);
```
normanmaurer, line 700: I made these `size_t` as we do the same for other methods. That said I wonder if it would be better to make it `socklen_t` . If we do this we should also fix the already existing methods. WDYT @LPardue @ghedo  ?

normanmaurer:
So this seems to work... just tested it with my integration code

quiche/include/quiche.h
normanmaurer: actually let me just do this.

quiche/src/ffi.rs
```diff
@@ -1351,6 +1351,140 @@ pub extern fn quiche_conn_send_quantum(conn: &Connection) -> size_t {
     conn.send_quantum() as size_t
 }
 
+#[no_mangle]
+pub extern fn quiche_conn_path_event_next(
+    conn: &mut Connection,
+) -> *mut PathEvent {
```
LPardue, line 1357: does return type need to be mut?

quiche/src/ffi.rs
```diff
@@ -1351,6 +1351,140 @@ pub extern fn quiche_conn_send_quantum(conn: &Connection) -> size_t {
...
+            *peer_addr_len = std_addr_to_c(peer, peer_addr)
+        },
+
+        _ => unreachable!(),
+    }
+}
+
+#[no_mangle]
+pub extern fn quiche_path_event_reused_source_connection_id(
+    ev: &PathEvent, connection_id: &mut u64,
```
LPardue, line 1447:
```suggestion
    ev: &PathEvent, cid_sequence_number: &mut u64,
```

or some other name to make it clear this is a sequence number not an actual CID

quiche/src/ffi.rs
normanmaurer: Good catch ... I don't think so

quiche/src/ffi.rs
normanmaurer: 👍

normanmaurer:
@LPardue all addressed

quiche/src/ffi.rs
```diff
@@ -1351,6 +1351,140 @@ pub extern fn quiche_conn_send_quantum(conn: &Connection) -> size_t {
     conn.send_quantum() as size_t
 }
 
+#[no_mangle]
+pub extern fn quiche_conn_path_event_next(
+    conn: &mut Connection,
+) -> *const PathEvent {
+    match conn.path_event_next() {
+        Some(v) => Box::into_raw(Box::new(v)),
+        None => ptr::null_mut(),
```
LPardue, line 1360: can this be just `null()`?

LPardue:
There's a lot of passing pairs of

```
foo: &mut sockaddr_storage,
foo_len: &mut socklen_t
```

wondering if maybe we should define a struct...

normanmaurer:
> There's a lot of passing pairs of
> 
> ```
> foo: &mut sockaddr_storage,
> foo_len: &mut socklen_t
> ```
> 
> wondering if maybe we should define a struct...

I was thinking about it but decided against it as other functions that take those pairs dont use a struct as well. That said I am fine with adding a struct as well... Maybe lets wait for @ghedo to chime in

normanmaurer:
rebased

quiche/src/ffi.rs
```diff
@@ -1382,7 +1382,6 @@ pub extern fn quiche_conn_new_scid(
         Err(e) => e.to_c() as u64,
     }
 }
-
 #[no_mangle]
```
ghedo, line 1385:
```suggestion

#[no_mangle]
```

ghedo (commented):
Added a nit, I'll apply it myself when merging.

quiche/include/quiche.h
```diff
@@ -701,6 +701,53 @@ size_t quiche_conn_active_scids(quiche_conn *conn);
 uint64_t quiche_conn_new_scid(quiche_conn *conn,
                            const uint8_t *scid, size_t scid_len,
                            const uint8_t *reset_token, bool retire_if_needed);
+enum quiche_path_event_type {
```
ghedo, line 704:
```suggestion

enum quiche_path_event_type {
```

ghedo (approved):

PR #1673, opened 2023-11-30
Vlad/custom buffer send

vkrasnov:
In my test setup using zero copy send with a buffer pool improves send performance by over 5%.


vkrasnov:
@ehaydenr FYI

PR #1675, opened 2023-12-02
qlog: extend serializer to support arbitrary types of event

LPardue:
Previously, the QlogStream API only supported concrete qlog::events::Event
objects. This limited the ability to stream ad-hoc types, which could be useful
for various use cases such as implementing extension Events or augmenting a
standard event with additional custom date.

This change modifies the API by replacing the Event type parameter with traits.
The new Eventable trait provides basic features need for qlog, and it is paired
with serde::Serialize to support convention.

A new JsonEvent is defined to leverage the new API. The data field is a
serde_json::Value, allow callers to construct arbitrary structured data that
can then be logged in an ad-hoc manner. This approach ensures that the QLOG-SEQ
serialization is a valid format that can also be parsed back in a structured
manner.

Example usage:

```
let data = json!({"foo": "Bar", "hello": 123});
let event = events::JsonEvent {
    time: 0.0,
    importantance: events::EventImportance::Core,
    name: "jsonevent:sample".into(),
    data,
};

// assuming a QlogStreamer object has been constructed
streamer.add_event(event);
```

### DIFF
--- a/qlog/src/events/mod.rs
+++ b/qlog/src/events/mod.rs
@@ -109,9 +109,15 @@ impl Event {
             ty,
         }
     }
+}
 
-    pub fn importance(&self) -> EventImportance {
+impl Eventable for Event {
+    fn importance(&self) -> EventImportance {
         self.ty.into()
+    }
+
+    fn set_time(&mut self, time: f32) {
+        self.time = time;
     }
 }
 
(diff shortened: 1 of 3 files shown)


ehaydenr (approved):
cool!

qlog/src/events/mod.rs
```diff
@@ -126,8 +132,30 @@ impl PartialEq for Event {
     }
 }
 
-#[derive(Clone)]
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct JsonEvent {
+    pub time: f32,
+
+    #[serde(skip)]
+    pub importantance: EventImportance,
```
fisherdarling, line 140:
```suggestion
    pub importance: EventImportance,
```

ghedo (approved):

ghedo (approved):

PR #1681, opened 2023-12-07
ffi: Fix const-correctness for quiche_h3_send_* functions

rgacogne:
Hi!

Unless I'm mistaken, these parameters are not mutable in the FFI Rust code and thus can be made `const` in the C/C++ header.


ghedo (approved):

ghedo:
Merged, thanks @rgacogne!

PR #1691, opened 2023-12-14
FFI: Expose the peer cert chain

bwoebi:
Just wanted to use Connection::peer_cert_chain() and realized it's not actually exposed via FFI.

The vector cannot be trivially returned by FFI, except by translating the vector to a `&[Certificate]` with a `struct Certificate { *mut u8; size_t; }`, and then adding a free-function for this.

I think directly fetching the values one by one is a superior approach for FFI purposes in terms of usability and complexity, hence this PR proposes this.


bwoebi:
@LPardue Sorry that I ping you here after 2 months now, but is there a chance that this will be reviewed and merged? Thanks!

catid (commented):
LGTM

PR #1692, opened 2023-12-14
qlog: add ex_data to Event

LPardue:
This change allows additional JSON fields to be contained in qlog events
without needing to change the core definitions.

### DIFF
--- a/qlog/src/streamer.rs
+++ b/qlog/src/streamer.rs
@@ -174,13 +174,11 @@ impl QlogStreamer {
     /// Writes an [Event] based on the provided [EventData] to a JSON-SEQ record
     /// at time [std::time::Instant::now()].
     pub fn add_event_data_now(&mut self, event_data: EventData) -> Result<()> {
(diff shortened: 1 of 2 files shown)


qlog/src/streamer.rs
```diff
@@ -178,6 +179,16 @@ impl QlogStreamer {
         self.add_event_data_with_instant(event_data, now)
     }
 
+    /// Writes an [Event] based on the provided [EventData] to a JSON-SEQ record
+    /// at time [std::time::Instant::now()].
+    pub fn add_event_data_ex_now(
```
fisherdarling, line 184: Maybe a builder pattern would be useful here?

```
streamer.add_event(EventBuilder::now().with_data(event_data).with_ext(ex_data));
or maybe
streamer.add_event(Event::at(instant).data(event_data).ext(ex_data));
```

LPardue: Nice idea but we'll save that for another day.

PR #1694, opened 2023-12-15
qlog: support reading JsonEvent too

LPardue:


qlog/src/reader.rs
```diff
@@ -24,9 +24,15 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use crate::events::Event;
 use crate::QlogSeq;
 
+/// Represents the format of the read event, either standard qlog::events::Event
```
ghedo, line 29: Seems better to remove the "either ..." part, and add something to each enum case instead.

qlog/src/reader.rs
```diff
@@ -24,9 +24,15 @@
...
+/// or non-standard qlog::events::JsonEvent.
+pub enum Event {
+    Qlog(Box<crate::events::Event>),
```
ghedo, line 32: Why `Box<>`?

LPardue: Clippy complains about https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant

But actually I doubt it makes much difference here, so we could just squelch the lint

PR #1701, opened 2024-01-02
lib: expose machinery for async TLS callbacks

evanrittenhouse:
This PR makes it possible for an application to use async TLS callbacks. If that is the case, the application should drive the callback forward by repeatedly calling `quiche::Connection.do_handshake()`.

This PR also introduces a new `quiche::Error` variant, `TlsRetry`. This differs from `Error::Done` in that `Done` represents the state where there is no more work to do, and thus the application should wait for some response from the remote. `Done` lacks requisite detail for async callbacks; the application using said callbacks must be able to poll a future and block while the callback completes, but not block when it has to receive data from the remote.

`TlsRetry` essentially represents the case where we _currently_ have no more work to do, as we're waiting for some TLS operation to complete, but we know that we must complete the operation before we can receive more data from the remote. This error state essentially models the following case from the [Boring docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_ERROR_PENDING_CERTIFICATE): 
>  SSL_ERROR_[error] indicates the operation failed because the [callback] was incomplete. The caller may retry the operation when [callback] has completed.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -567,6 +567,11 @@ pub enum Error {
 
     /// Error in key update.
     KeyUpdate,
+
+    /// Need to call a TLS operation again from the local side, without waiting
+    /// on data from the peer. Helpful when using asynchronous SSL
+    /// callbacks.
+    TlsRetry,
 }
 
 impl Error {
@@ -606,6 +611,7 @@ impl Error {
             Error::IdLimit => -17,
             Error::OutOfIdentifiers => -18,
             Error::KeyUpdate => -19,
+            Error::TlsRetry => -20,
         }
     }
 }
@@ -2185,6 +2191,8 @@ impl Connection {
 
                     left
                 },
+
+                Err(Error::TlsRetry) => left,
 
                 Err(e) => {
                     // In case of error processing the incoming packet, close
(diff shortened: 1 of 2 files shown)


rushilmehra:
Lgtm but will defer to @LPardue and @ghedo

LPardue:
Why is this needed?

rushilmehra:
> Why is this needed?
For async callbacks we need some way to progress the callback future by invoking SSL_do_handshake outside of a recv call. This lets us progress the future until SSL_do_handshake stops returning Error::Done

LPardue:
so why not expose quiche's do_handshake method rather than TLS one?

evanrittenhouse:
Happy to do either, wanted to expose the smallest surface but I haven't gotten to test it yet. May as well leave this WIP until we do

LPardue:
Maybe i'm overlooking something but AFAICT exposing the Handshake object means letting apps get access to internal state that they never had before. That exposes the entirety of all public methods/fields on that object, vastly increasing the current surface. Not only that, direct access means skipping all the additional state that is maintained in a quiche connection related to a handshaking, which seems pretty error prone.

The current design calls do_handshake when packets are sent or received. Under what circumstances would an app be able to progress that handshake state for any other reason?

evanrittenhouse:
We need to call `SSL_do_handshake` repeatedly to drive the Boring callback forward. We can use Quiche's `Error::Done` code as a proxy for "a `Pending` Future", but that requires that we be able to call `SSL_do_handshake` repeatedly in a separate function.

PR #1705, opened 2024-01-09
lib: minimise calls to SendBuf off_front() during emit()

LPardue:
lib: minimise calls to SendBuf off_front()

There were several places where off_front() would be called in rapid succession.
off_front() potentially has to iterate a large queue to find the "front" value.
In rapid succession, that value is identical, leading to wasted work.

This change removes the ready() function (which called off_front) and refactors
code to minimise off _front() calls.

### DIFF
--- a/quiche/src/stream/send_buf.rs
+++ b/quiche/src/stream/send_buf.rs
@@ -175,7 +175,7 @@ impl SendBuf {
         while out_len > 0 {
             let off_front = self.off_front();
 
-            if self.data.is_empty() ||
+            if self.is_empty() ||
                 off_front >= self.off ||
                 off_front != next_off ||
                 off_front >= self.max_data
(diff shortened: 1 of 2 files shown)


LPardue:
FWIW Stream::is_writable() also ends up calling off_front() twice in quick succession but I chose to leave that out.

An alternative to the approach in this PR would be to store and maintain the value of off_front in SendBuf to avoid having to recalculate it all the time. Furthermore, perhaps things are already magically optimised and implicitly cached already, making this change unnecessary. But the flamegraphs I've seen don't indicate that.

quiche/src/stream/send_buf.rs
```diff
@@ -172,11 +172,17 @@ impl SendBuf {
...
 
-        while out_len > 0 &&
-            self.ready() &&
-            self.off_front() == next_off &&
-            self.off_front() < self.max_data
-        {
+        while out_len > 0 {
+            let off_front = self.off_front();
+
+            if self.data.is_empty() ||
```
ghedo, line 178: Nit:
```suggestion
            if self.is_empty() ||
```

ghedo (approved):

PR #1711, opened 2024-01-15
frame: reject NEW_CONNECTION_ID frames with invalid CIDs

### LINKED ISSUE #1710: CIDs > 20 bytes should be rejected
quiche accepts CIDs > 20 bytes (I tried 200 bytes) in NEW_CONNECTION_ID frames, and uses them on outgoing packets.

ghedo:
Fixes #1710.

### DIFF
--- a/quiche/src/frame.rs
+++ b/quiche/src/frame.rs
@@ -261,15 +261,25 @@ impl Frame {
                 limit: b.get_varint()?,
             },
 
-            0x18 => Frame::NewConnectionId {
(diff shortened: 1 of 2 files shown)


marten-seemann (changes requested):
Not very familiar with Rust, but afaict this doesn't reject zero-length CIDs, does it?

ghedo:
Ugh, yes you are right. Fixed now, thanks.

LPardue (approved):

PR #1712, opened 2024-01-15
lib: stricter NEW_TOKEN frame checks

LPardue:
RFC 9000 has some low-level requirements for who is allowed to send
a NEW_TOKEN frame, and what the token value can be.

This change adds stricter NEW_TOKEN frame checks. It leaves any
more advanced handling of the Retry mechanism as continuing TODO.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -567,6 +567,9 @@ pub enum Error {
 
     /// Error in key update.
     KeyUpdate,
+
+    /// Protocol error.
+    ProtocolViolation,
 }
 
 impl Error {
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -6815,8 +6820,17 @@ impl Connection {
...
             frame::Frame::CryptoHeader { .. } => unreachable!(),
 
-            // TODO: implement stateless retry
-            frame::Frame::NewToken { .. } => (),
+            frame::Frame::NewToken { token } => {
+                if self.is_server {
+                    return Err(Error::ProtocolViolation);
+                }
+
+                if token.is_empty() {
```
ghedo, line 6828: This should go in the frame parsing code I think.

ghedo (approved):

PR #1713, opened 2024-01-15
remove ProtocolViolation error

ghedo:
Follow-up to 1684118a979276428ce9f994b028463b26465698.
    
This error was added in the previous commit, but we can use a more specific error instead, also to be more consistent with existing code (e.g. see HANDSHAKE_DONE case).


LPardue (approved):
Wfm

PR #1729, opened 2024-02-25
lib: extract lost frame processing to own method

LPardue:


rushilmehra (approved):

ghedo (changes requested):
Needs to be updated with PMTUD changes.

PR #1742, opened 2024-03-12
Reimplement recovery and congestion control

vkrasnov:
This change replaces the CC and pacing algorithms using the C++ google/quiche implementations rewritten in Rust.
In addition it improves on the general recovery as well.

This is essentially the same as #1490 only split into more incremental commits for easier review.


qlog/src/events/quic.rs
```diff
@@ -743,8 +743,6 @@ pub struct MetricsUpdated {
     pub congestion_window: Option<u64>,
     pub bytes_in_flight: Option<u64>,
 
-    pub ssthresh: Option<u64>,
```
LPardue, line 746: This field is still defined in the qlog RFC, so its unclear why its removed from here; see https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-metrics_updated

vkrasnov: Good point

vkrasnov: I guess for BBR that would always be 0?

LPardue: None will do

vkrasnov: Fixed

LPardue:
needs a rebase

quiche/src/recovery/cubic.rs
```diff
@@ -787,319 +764,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn cubic_hystart_css_to_ss() {
```
LPardue, line 789: What does deleting these tests really mean?

quiche/src/recovery/mod.rs
```diff
@@ -91,13 +92,100 @@ struct RecoveryEpoch {
...
+    ) -> AckedDetectionResult {
+        newly_acked.clear();
+        let mut acked_bytes = 0;
+        let mut spurious_losses = 0;
+        let mut spurious_pkt_thresh = None;
+        let mut has_ack_eliciting = false;
+        let largest_acked = self.largest_acked_packet.unwrap();
+
+        for ack in acked.iter() {
+            // Because packets always have incermenting numbers, they are always
```
LPardue, line 150: s/incermenting/incrementing

quiche/src/recovery/bbr/mod.rs
```diff
@@ -287,24 +287,22 @@ fn reset(r: &mut Recovery) {
     init::bbr_init(r);
 }
 
-fn on_packet_sent(r: &mut Recovery, sent_bytes: usize, _now: Instant) {
+fn on_packet_sent(r: &mut Recovery, _sent_bytes: usize, _now: Instant) {
     per_transmit::bbr_on_transmit(r);
-
-    r.bytes_in_flight += sent_bytes;
```
ghedo, line 293: This seems to change the behaviour of the code, rather than just being a refactoring?

`bbr_handle_restart_from_idle()` (called by `bbr_on_transmit()`) checks if `bytes_in_flight` is `0`, but if `bytes_in_fligh` is incremented before that call, then the condition will always be false after this change?

quiche/src/recovery/rtt.rs
```diff
@@ -1,94 +1,87 @@
...
 
 pub(crate) const INITIAL_RTT: Duration = Duration::from_millis(333);
 
 pub(crate) const RTT_WINDOW: Duration = Duration::from_secs(300);
+pub struct RttStats {
+    pub(super) latest_rtt: Duration,
+    pub(super) min_rtt: Minmax<Duration>,
+    pub(super) smoothed_rtt: Duration,
+    pub(super) rttvar: Duration,
+    pub(super) first_rtt_sample: Option<Instant>,
```
LPardue, line 43: AFAICT the value of first_rtt_samle is never used anywhere, so maybe just needs to be a bool?

quiche/src/recovery/cubic.rs
```diff
@@ -226,21 +227,8 @@ fn on_packet_acked(r: &mut Recovery, packet: &Acked, now: Instant) {
...
-        if new_lost < rollback_threshold {
-            let did_rollback = rollback(r);
-
-            if did_rollback {
-                return;
-            }
-        }
-    }
+    //
+    // Spurios loss detection is handled in [`Recovery`]
```
LPardue: If I recall correctly the prior behavior was added to address cubic over-reacting to spurious loss.

It's confusing the have the comment above that mentions ROLLBACK_THRESHOLD_PERCENT when this commit removes that constant entirely.

We need to discuss if this refactor commit should be changing the functionality of cubic like this.

quiche/src/recovery/mod.rs
```diff
@@ -696,10 +511,6 @@ impl Recovery {
                 self.pkt_thresh.max(thresh.min(MAX_PACKET_THRESHOLD));
         }
 
-        if newly_acked.is_empty() {
-            return Ok((0, 0));
-        }
```
LPardue: What is the effect of removing this early return?

quiche/src/recovery/bbr/mod.rs
ghedo: Actually, disregard this, I misread the change.

quiche/src/lib.rs
```diff
@@ -1286,6 +1285,9 @@ pub struct Connection {
     /// Packet number spaces.
     pkt_num_spaces: [packet::PktNumSpace; packet::Epoch::count()],
 
+    /// Next packet number
+    next_pkt_num: u64,
```
ghedo, line 1289: It looks like this merges the packet numbers from all packet number spaces into the same sequence, which is a significant functional changes, see https://datatracker.ietf.org/doc/html/rfc9000#section-12.3-8

quiche/src/recovery/congestion/reno.rs
```diff
@@ -0,0 +1,398 @@
...
+fn has_custom_pacing() -> bool {
+    false
+}
+
+fn debug_fmt(_r: &Congestion, _f: &mut std::fmt::Formatter) -> std::fmt::Result {
+    Ok(())
+}
+
+#[cfg(test)]
+pub(crate) mod test_sender {
```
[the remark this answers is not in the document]
vkrasnov: I agree, but then again all other modules use Reno at this point

quiche/src/recovery/rtt.rs
```diff
@@ -1,94 +1,87 @@
...
             rttvar: INITIAL_RTT / 2,
+            first_rtt_sample: None,
+            max_ack_delay,
         }
     }
 
     pub(crate) fn update_rtt(
-        &mut self, latest_rtt: Duration, ack_delay: Duration, now: Instant,
+        &mut self, latest_rtt: Duration, mut ack_delay: Duration, now: Instant,
+        handshake_confirmed: bool,
```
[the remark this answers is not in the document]
vkrasnov: It just does what the RFC says, and adjusts for `max_ack_delay` only after handshake is complete, and not unconditionally

quiche/src/recovery/cubic.rs
vkrasnov: The tests were not testing an actual functionality as hystart is in fact called for every sent packet and not as performed in those tests at an arbitrary point in time. Because the new code folds hystart into cubic it is impossible to fake this non-existent behavior. It is easy to change the test to match the actual behavior, but then it would be pretty moot.

quiche/src/recovery/rtt.rs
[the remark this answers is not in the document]
LPardue: If we believe there is a functional bug in the current implementation, then tracking and fixing that independent of refactor is a lot clearer.

quiche/src/recovery/mod.rs
```diff
@@ -555,11 +683,7 @@ impl Recovery {
             let latest_rtt =
                 now.saturating_duration_since(largest_newly_acked_sent_time);
 
-            let ack_delay = if epoch == packet::Epoch::Application {
-                Duration::from_micros(ack_delay)
-            } else {
-                Duration::from_micros(0)
-            };
+            let ack_delay = Duration::from_micros(ack_delay);
```
LPardue: This also seems like a functional change since it will affect how `update_rtt` works, albeit related to https://github.com/cloudflare/quiche/pull/1742/commits/a79dbfa65401f5e6497b2984988ea5572fb33bce#r1555902791

vkrasnov:
> needs a rebase

rebased, and fixed a few issues in the process

quiche/src/recovery/cubic.rs
vkrasnov: I brought the tests back.

quiche/src/recovery/rtt.rs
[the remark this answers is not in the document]
vkrasnov: Up to you if you think it is a bug or not

quiche/src/recovery/cubic.rs
vkrasnov: i brought it back

quiche/src/recovery/mod.rs
```diff
@@ -664,89 +778,80 @@ impl Recovery {
...
         self.set_loss_detection_timer(handshake_status, now);
     }
 
     pub fn on_path_change(
-        &mut self, epoch: packet::Epoch, now: Instant, trace_id: &str,
+        &mut self, epoch: packet::Epoch, now: Instant, _trace_id: &str,
     ) -> (usize, usize) {
-        self.detect_lost_packets(epoch, now, trace_id)
+        let (lost_bytes, lost_packets) =
+            self.detect_and_remove_lost_packets(epoch, now);
```
antoniovicente, line 807: I see no call to `self.pacer.on_congestion_event` after `detect_and_remove_lost_packets` in this branch.  Is this intentional?

vkrasnov: Yes, I think it is, path change is not a true congestion event, in fact maybe cc state needs to be reset completely in this case? But I would avoid big changes to working code, before it is merged, better to improve after

nmicht:
Closing this PR now.
Due to rebase problems, https://github.com/cloudflare/quiche/pull/1983 was created with the changes from this PR and then split into smaller PRs.

PR #1743, opened 2024-03-13
fix packet encoding/decoding which seems to break after 2 billions packets sent

frochet:
Hello,

Long story short, I've been using the encoding/decoding logic of packet number for another purpose than encoding pn, and observed that it would start decoding incorrect values when reaching 2^31.

I dived into RFC9000, on the Appendix for the algorithmic details of encoding/decoding, and observed that the encoding algorithm in quiche does not match the encoding algorithm suggested by the RFC. I suppose there is a reason for this, but it eludes me.

So, this pull request does an implementation close to what is suggested (hopefully more efficient), and fixes my problem, and the bug with packet numbers too. I've also added test cases for packet encoding/decoding using example values from the RFC.

This passes all unit tests, and a  simple integration test in which I downloaded a 5 GiB file through the H3 module using quiche-[client/server]. However, I am unsure about:

1) Why the largest acked value on a path can be bigger than the current packet number on that path.
2) Why largest_acked values were initialized to UINT64::MAX; I changed this to 0, and removed a condition checking for UIN64::MAX. I had to do this, because pkt_num_len() is now using information from the largest_acked value.
3) Whether the original pkt encoding is a bug, a forgotten todo, or something else. The fact that there were no test cases suggests a forgotten todo?


quiche/src/packet.rs
```diff
@@ -555,20 +555,12 @@ impl<'a> std::fmt::Debug for Header<'a> {
...
-        return Err(Error::InvalidPacket);
-    };
-
-    Ok(len)
+pub fn pkt_num_len(pn: u64, largest_acked: u64) -> Result<usize> {
+    let num_unacked: u64 = pn.checked_sub(largest_acked).map_or(0, |v| v) + 1;
+    // computes ceil of num_unacked.log2()
+    let min_bits = u64::BITS - num_unacked.leading_zeros();
+    // get the num len in bytes
+    Ok((min_bits as f32 / 8.).ceil() as usize)
```
vanc, line 563: `Ok((num_bits + 7) / 8 as usize)` should be faster.

frochet: Looks right! Good catch. The Result is not useful anymore either, and I removed it.

frochet:
@LPardue @ghedo I've other pull requests with potential quality of live improvements (nothing big) from my ongoing project using quiche. Are those welcome? If external contributions are not, I prefer not to spam :-)

(e.g., support of recvmmsg() for apps/src/client.rs used in HTTP download)

ghedo:
@frochet also, specifically about recvmmsg() support, @evanrittenhouse has been pulling all that UDP socket logic into its own crate https://github.com/cloudflare/quiche/pull/1764 so that is probably a better place to look into for that change.

frochet:
I've asked for a quicdev slack invite. Thank you for the guidance :-)

Rebase done! It looks like some changes in Master addressed the first two questions in the first post as well, so great!

ghedo (approved):

ghedo:
Merged, thanks @frochet! And sorry it took so long :disappointed:

ghedo:
And also thanks @vanc for the review and suggestions!

PR #1745, opened 2024-03-20
qlog: add length and payload_length fields to QUCI frames

LPardue:
The qlog spec includes these fields and we didn't; see https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-quic-frames

### DIFF
--- a/qlog/src/events/quic.rs
+++ b/qlog/src/events/quic.rs
@@ -384,30 +384,43 @@ pub enum QuicFrameTypeName {
 // ensure it goes out on the wire. This means that deserialization of frames
 // also works automatically.
 pub enum QuicFrame {
-    Padding,
+    Padding {
+        length: Option<u32>,
+        payload_length: u32,
+    },
 
-    Ping,
+    Ping {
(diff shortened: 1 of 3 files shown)


qlog/src/events/quic.rs
```diff
@@ -384,30 +384,43 @@ pub enum QuicFrameTypeName {
 // ensure it goes out on the wire. This means that deserialization of frames
 // also works automatically.
 pub enum QuicFrame {
-    Padding,
+    Padding {
+        length: Option<u32>,
+        payload_length: u32,
```
larseggert, line 389:
```suggestion
        payload_length: Option<u32>,
```
Maybe? (Haven't looked at the I-D, this is just based on the other additions of `payload_length`.

LPardue: Yeah I did the same but it really is not optional in https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-8.13.1

ghedo (approved):

PR #1749, opened 2024-03-26
Clippy Fixes

PSUdaemon:


PSUdaemon:
This seems to be failing on nightly builds for dead code on a public member of a public struct. I can mark it with `#[allow(dead_code)]` but that feels wrong.

PSUdaemon:
Upon deeper inspection it seems that there is no corresponding [set](https://github.com/cloudflare/quiche/blob/098664eb001c38a8370a09a4b0d99c8126320e95/quiche/src/tls/mod.rs#L758-L761) in [set_write_secret](https://github.com/cloudflare/quiche/blob/098664eb001c38a8370a09a4b0d99c8126320e95/quiche/src/tls/mod.rs#L769) and just trivially adding it there causes tests to panic on unwraps. It's not easily clear what the fix is as there are some assumptions in the code that I do not fully understand.

@ghedo can you take a look? Also, if you want me to add the `#[allow(dead_code)]` for now I am happy to.

ghedo:
Oh, interesting this never came up before... I think that `crypto_0rtt_seal` field can be removed actually, since it's not required to send 0-RTT. It was originally added in https://github.com/cloudflare/quiche/commit/746cd0077c11996a40f1a675cb0b23e03bc51e43  before we supported sending 0-RTT packets, and then I guess I forgot to remove it in https://github.com/cloudflare/quiche/commit/3174cda88120515cf624f679597c3fc1b77ca5ad ...

PSUdaemon:
> Oh, interesting this never came up before... I think that `crypto_0rtt_seal` field can be removed actually, since it's not required to send 0-RTT. It was originally added in [746cd00](https://github.com/cloudflare/quiche/commit/746cd0077c11996a40f1a675cb0b23e03bc51e43) before we supported sending 0-RTT packets, and then I guess I forgot to remove it in [3174cda](https://github.com/cloudflare/quiche/commit/3174cda88120515cf624f679597c3fc1b77ca5ad) ...

Ok, I will remove it in this PR then. Thanks.

ghedo (approved):

ghedo:
Great, thanks @PSUdaemon, it's merged now!

PR #1762, opened 2024-04-16
Add ability to convert errors to wire_error_code via C API

normanmaurer:
Motivation:

https://github.com/cloudflare/quiche/commit/1d00ee1bb2256dfd99ba0cb2dfac72fe1e59407f added the ability to retrieve the wire error code for a given error but did not add the same functionallity to the C API.

Modifications:

Expose functions to convert errors to wire_error_code via C

Result:

More complete C API

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -132,6 +132,10 @@ enum quiche_error {
 // Returns a human readable string with the quiche version number.
 const char *quiche_version(void);
 
+// Returns the wire error code for the given error as defined in RFC9000 Section 20. Error Codes.
+// The given error must be a valid quiche_error.
+uint64_t quiche_error_to_wire_error_code(ssize_t error);
+
 // Enables logging. |cb| will be called with log messages
 int quiche_enable_debug_logging(void (*cb)(const char *line, void *argp),
                                 void *argp);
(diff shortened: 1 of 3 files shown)


quiche/include/quiche.h
```diff
@@ -132,6 +132,10 @@ enum quiche_error {
 // Returns a human readable string with the quiche version number.
 const char *quiche_version(void);
 
+// Returns the wire error code for the given error as defined in RFC9000 Section 20. Error Codes.
+// The given error must be a valid quiche_error.
+uint64_t quiche_error_to_wire_error_code(ssize_t error);
```
normanmaurer, line 137: @ghedo @LPardue I only added the code for the quic errors for now but if you agree with the approach let me know and I will add the same for h3. 

One thing I was not sure about was if using a `ssize_t` is good or if we should use `quiche_error`  and expose it via FFI as enum.

ghedo: TBH it's not really clear why an application would want this? The https://github.com/cloudflare/quiche/commit/1d00ee1bb2256dfd99ba0cb2dfac72fe1e59407f change simply added public enums for the wire error codes defined in RFCs, the conversion from `Error` is not actually public, so this FFI change would add an API that is not actually available in Rust.

normanmaurer: Because an app might want to hide the fact that they use quiche underneath and just return the "RFC" error codes. That said it's easy for me to workaround it.

PR #1769, opened 2024-04-30
Add soname to libquiche.so

### LINKED ISSUE #165: soname versioning

### LINKED ISSUE #598: How to add soname into libquiche.so

junhochoi:
Using cdylib-link-lines, add soname to the generated shlibs.

It will add the following to libquiche.so:

```
% readelf -d target/debug/libquiche.so
...
 0x000000000000000e (SONAME)             Library soname: [libquiche.so.0]
```

For linux, *bsd, android, ios and macos.

Fixes #598
Fixes #165

### DIFF
--- a/quiche/Cargo.toml
+++ b/quiche/Cargo.toml
@@ -46,7 +46,7 @@ pkg-config-meta = []
 fuzzing = []
 
 # Build and expose the FFI API.
-ffi = []
+ffi = ["dep:cdylib-link-lines"]
 
 # Exposes internal APIs that have no stability guarantees across versions.
 internal = []
@@ -59,6 +59,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [build-dependencies]
 cmake = "0.1"
 pkg-config = { version = "0.3", optional = true }
+cdylib-link-lines = { version = "0.1", optional = true }
 
 [dependencies]
 either = { version = "1.8", default-features = false }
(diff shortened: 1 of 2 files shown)


quiche/Cargo.toml
```diff
@@ -57,6 +57,7 @@ features = ["boringssl-boring-crate", "qlog"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [build-dependencies]
+cdylib-link-lines = "0.1"
```
ghedo, line 60: I think we probably want to limit this dependency to when the `ffi` feature is enabled, it's not very useful otherwise (we do still build the library in that case, but there's no way around that IIRC).

junhochoi: Ok. I updated that it depends on `ffi` feature.

ghedo (approved):

PR #1778, opened 2024-05-02
(CC refactor 8) recovery: move all congestion control related fields to a new struct

ghedo:


quiche/src/recovery/cubic.rs
```diff
@@ -679,9 +689,6 @@ mod tests {
         // Exit from the recovery.
         now += rtt;
 
-        // To avoid rollback
-        r.lost_count += MIN_ROLLBACK_THRESHOLD;
```
LPardue, line 683: This PR is pretty obtuse to review, so I just wanted to check, is the removal here was covered somewhere else?

quiche/src/recovery/cubic.rs
```diff
@@ -1254,9 +1273,6 @@ mod tests {
...
```
LPardue, line 1258: Same comment as above regarding what this removal means

quiche/src/recovery/cubic.rs
ghedo: AFAICT this was never actually necessary, because the rollback doesn't happen either way.

quiche/src/recovery/cubic.rs
ghedo: Same answer :)

LPardue (approved):

ghedo:
Merged as f73f86937da6fd44468fb0fe73758b6fb0470c10.

PR #1779, opened 2024-05-02
lib: fix resetting of streams with buffered but unsent data

LPardue:
An application can buffer stream data but not have the opportunity to prepare
the packets that would send that data. Within this window of time, the stream
might get reset either in reaction to receiving a STOP_SENDING frame, or due to
the application resetting the stream themselves via stream_shutdown(). Once this
occurs, no stream data will be sent. Instead, a RESET_STREAM will be emitted
when packets are next prepared that contains a final size of whatever bytes
_had_ been sent.

Quiche needs to keep a reference to such a stream for as long as it is "active".
For a bidi stream, that means that the receive side has been finished, and the
send side closure (RESET_STREAM) has been acknowledged and validated. Only when
both these conditions are true will the stream be collected. Stream collection
is used in stream credit accounting, which is used to for determining when to
emit a MAX_STREAMS frame.

Prior to this change, resetting a stream with buffered but unsent data would
lead to some internal checks failing when processing RESET_STREAM
acknowledgement. That led to not collecting the stream and not updating
MAX_STREAMS when it was expected.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -11060,11 +11060,7 @@ mod tests {
         assert_eq!(r.next(), Some(0));
         assert_eq!(r.next(), None);
 
-        loop {
-            if pipe.server.stream_send(0, b"world", false) == Err(Error::Done) {
-                break;
-            }
-
+        while pipe.server.stream_send(0, b"world", false) != Err(Error::Done) {
             assert_eq!(pipe.advance(), Ok(()));
         }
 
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -12117,6 +12117,161 @@ mod tests {
...
+        let mut r = pipe.server.readable();
+        assert_eq!(r.next(), None);
+
+        // Server sends data, until blocked.
+        let mut r = pipe.server.writable();
+        assert_eq!(r.next(), Some(0));
+        assert_eq!(r.next(), None);
+
+        loop {
+            if pipe.server.stream_send(0, b"world", false) == Err(Error::Done) {
```
ghedo, line 12150: You could simplify this whole loop with something like:
```
while pipe.server.stream_send(0, b"world", false) != Err(Error::Done) {}
```

ghedo (dismissed):
Added minor nit, but lgtm otherwise.

quiche/src/lib.rs
LPardue: That pattern was used in a few places so I converted them all for consistency.

PR #1780, opened 2024-05-08
(CC refactor 9) recovery: move congestion into module

ghedo:


quiche/src/recovery/congestion/bbr/mod.rs
```diff
@@ -34,13 +34,13 @@ use crate::recovery::*;
...
-pub static BBR: CongestionControlOps = CongestionControlOps {
+use super::CongestionControlOps;
+
+pub(crate) static BBR: CongestionControlOps = CongestionControlOps {
     on_init,
-    reset,
     on_packet_sent,
     on_packets_acked,
     congestion_event,
-    collapse_cwnd,
```
junhochoi, line 43: Why reset() and collape_cwnd() is removed? (I understand it's not used, but no plan to implement persistent congestion in the future?)

ghedo: We'll probably want to do it at some point... but that's been the case for years, so for now removing them simplifies the refactoring a bit. We can add them back when needed.

LPardue (approved):

ghedo:
Merged as 5bccde6ead15688326e05364abab51a910242423.

PR #1783, opened 2024-05-09
ci: deprecate action-rs and other ci fix

junhochoi:
- Deprecate use of action-rs (no longer maintained)
- Update macos runner type (x86_64 and M1)
- Upgrade checkout from v3 (Node.js version warning)
- Upgrade ghaction-chocolatey to v3 (Node.js version warning)


ghedo (approved):
Ah, nice, thanks @junhochoi! I was trying to do something similar in https://github.com/cloudflare/quiche/compare/master...ci-actions-rs but couldn't get it to work.

PR #1796, opened 2024-06-06
H3: if readable after Finished, call recv again

vkrasnov:
If an H3 stream gets into a Data state that is not cleared for the stream (for example because downward backpressure), and a ResetStream arrives, the poll will transition the stream to finished and never clear the readable state of the stream, it will forever stuck in the readable state. This will mean that poll will return Finished and not Reset for the stream, and will have a somewhat significant performance impact, as for long lived connections it will take longer and longer to poll before the iterator reaches a stream that is actually readable.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -1731,6 +1731,15 @@ impl Connection {
         // events are returned when receiving empty stream frames with the fin
         // flag set.
         if let Some(finished) = self.finished_streams.pop_front() {
+            if conn.stream_readable(finished) {
+                // The connection is finished, but is still readable, it may
+                // indicate that there is a pending error, such as reset.
+                if let Err(crate::Error::StreamReset(e)) =
+                    conn.stream_recv(finished, &mut [])
+                {
+                    return Ok((finished, Event::Reset(e)));
+                }
+            }
             return Ok((finished, Event::Finished));
         }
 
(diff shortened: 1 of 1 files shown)


quiche/src/h3/mod.rs
```diff
@@ -1731,6 +1731,15 @@ impl Connection {
         // events are returned when receiving empty stream frames with the fin
         // flag set.
         if let Some(finished) = self.finished_streams.pop_front() {
+            if conn.stream_readable(finished) {
+                // The connection is finished, but is still readable, it may
```
ghedo, line 1735:
```suggestion
                // The stream is finished, but is still readable, it may
```

quiche/src/h3/mod.rs
```diff
@@ -6402,6 +6411,43 @@ mod tests {
         assert_eq!(s.poll_server(), Err(Error::Done));
     }
 
+    #[test]
+    fn reset_finished_at_server_with_data_pending() {
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        // Client sends HEADERS and doesn't fin
```
ghedo, line 6419:
```suggestion
        // Client sends HEADERS and doesn't fin.
```

quiche/src/h3/mod.rs
```diff
@@ -6402,6 +6411,43 @@ mod tests {
...
+        assert!(s.send_body_client(stream, false).is_ok());
+
+        assert_eq!(s.pipe.advance(), Ok(()));
+
+        let ev_headers = Event::Headers {
+            list: req,
+            has_body: true,
+        };
+
+        // Server receives headers and data
```
ghedo, line 6431:
```suggestion
        // Server receives headers and data...
```

quiche/src/h3/mod.rs
```diff
@@ -6402,6 +6411,43 @@ mod tests {
...
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Ok((stream, Event::Data)));
+
+        // ..then Client sends RESET_STREAM
```
ghedo, line 6435:
```suggestion
        // ..then Client sends RESET_STREAM.
```

quiche/src/h3/mod.rs
```diff
@@ -6402,6 +6411,43 @@ mod tests {
...
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Write, 0),
+            Ok(())
+        );
+
+        assert_eq!(s.pipe.advance(), Ok(()));
+
+        // Server receives the reset and there are no more readable streams
```
ghedo, line 6445:
```suggestion
        // Server receives the reset and there are no more readable streams.
```

ghedo (changes requested):

ghedo (approved):
LGTM, only applied some minor fixes to comments.

ghedo:
Also, waiting for @LPardue to look at this too before merging.

LPardue (approved):

PR #1798, opened 2024-06-11
apps: retry mio `poll()` call when it fails in quiche-server

evanrittenhouse:
Attempting to attach to the quiche-server process fails due to an interrupted syscall when mio attempts to `poll` for new events. This PR will retry the `poll` call until it succeeds.

## Testing
Terminal 1: `RUST_LOG=trace cargo run --bin quiche-server -- --cert ./apps/src/bin/cert.crt --key ./apps/src/bin/cert.key --listen 127.0.0.1:8085`
Terminal 2: `strace -p $(pgrep quiche-server)`

Before this patch, the quiche-server process fails due to the panic described above. With this patch, it succeeds and you can hit it as normal. We also see the `trace!`d log line

### DIFF
--- a/apps/src/bin/quiche-server.rs
+++ b/apps/src/bin/quiche-server.rs
@@ -187,7 +187,15 @@ fn main() {
             false => clients.values().filter_map(|c| c.conn.timeout()).min(),
         };
 
-        poll.poll(&mut events, timeout).unwrap();
+        let mut poll_res = poll.poll(&mut events, timeout);
+        while let Err(e) = poll_res.as_ref() {
+            if e.kind() == std::io::ErrorKind::Interrupted {
+                trace!("mio poll() call failed, retrying: {:?}", e);
+                poll_res = poll.poll(&mut events, timeout);
+            } else {
+                panic!("mio poll() call failed fatally: {:?}", e);
+            }
+        }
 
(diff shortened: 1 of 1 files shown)


apps/src/bin/quiche-server.rs
```diff
@@ -187,7 +187,15 @@ fn main() {
...
         };
 
-        poll.poll(&mut events, timeout).unwrap();
+        let mut poll_res = poll.poll(&mut events, timeout);
+        while let Err(e) = poll_res.as_ref() {
+            if e.kind() == std::io::ErrorKind::Interrupted {
+                trace!("mio poll() call failed, retrying: {:?}", e);
+                poll_res = poll.poll(&mut events, timeout);
+            } else {
+                panic!("mio poll() call failed fatally");
```
ghedo, line 196:
```suggestion
                panic!("mio poll() call failed fatally: {:?}", e);
```

ghedo (approved):
lgtm, minus minor suggestion.

LPardue:
Oh nice, did you know that resizing the terminal also causes panic on main? Does this fix that too?

evanrittenhouse:
> Oh nice, did you know that resizing the terminal also causes panic on main? Does this fix that too?

Yeah, this patch will fix that issue as well

ghedo (approved):

PR #1805, opened 2024-06-20
testing: make example cert/key files relative to Cargo.toml

evanrittenhouse:
Even though the `Session` struct is technically `hidden`, sometimes it's useful for testing. Currently, it's impossible to use `new()` from a different location because the `examples/cert` files aren't findable. 

Also a stray format change from running `cargo +nightly fmt`


ghedo:
Looks like this needs to be rebased. Also, the formatting is actually supposed to be the way it is, I think you need to update your rustfmt nightly.

PR #1811, opened 2024-07-02
Bump rust version to 1.67

junhochoi:
Handling the following error:
```
error: package `serde_with v3.8.2` cannot be built because it requires
rustc 1.67 or newer
```


ghedo (approved):

ghedo:
Merged, thanks @junhochoi!

PR #1815, opened 2024-07-12
ci: add cargo-machete in the workflow

junhochoi:
Run cargo-machete for unused dependency check, only on
a default ssl target. Also removed unused dependencies 
recommended by an initial run of cargo-machete.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -30,6 +30,10 @@ jobs:
           toolchain: ${{ env.RUSTTOOLCHAIN }}
           components: clippy
 
+      - name: Unused dependency check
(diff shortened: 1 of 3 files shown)


rushilmehra (commented):
Nice! Why only run on `matrix.tls-feature == ''`

junhochoi:
> Nice! Why only run on `matrix.tls-feature == ''`

Because it seems redundant. Also installing cargo-machete takes some time. We can try cargo-quickinstall instead
to save an installation time, but it will produce the same result regardless of --features.

rushilmehra (approved):

ghedo:
Merged, thanks @junhochoi! (and sorry for the delay)

PR #1816, opened 2024-07-12
ci: use ubuntu-24.04

junhochoi:
Trying ubuntu-24.04 runner.

During initial testing I met https://github.com/actions/runner-images/issues/9977, so disable i386-unknown-linux-gnu for now. Hopefully they will fix the kernel before it becomes ubuntu-latest

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   docs:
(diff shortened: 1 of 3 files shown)


junhochoi:
So simply replacing to `ubuntu-24.04` works ok. No need to change anything if `ubuntu-latest` switched to 24.04. Closing

PR #1820, opened 2024-07-21
h3i: initial version

LPardue:
See README.md


rushilmehra:
🎉

h3i/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "h3i"
+version = "0.1.0"
+edition = "2018"
+
```
ghedo, line 5: Need to add more metadata here to be able to publish to crates.io, e.g. see https://github.com/cloudflare/quiche/blob/master/quiche/Cargo.toml and https://github.com/cloudflare/quiche/blob/master/quiche/Cargo.toml

ghedo (changes requested):
Added some minor comments.

Looking at the generated docs, it looks like there are a whole bunch of public types and modules that I'm not entirely sure should actually be public as they don't seem useful to applications using h3i. Might want to go through the docs and check that the all the things that are public should be.

h3i/src/actions/h3.rs
```diff
@@ -0,0 +1,209 @@
...
+use serde_with::serde_as;
+
+use crate::encode_header_block;
+
+pub const HTTP3_CONTROL_STREAM_TYPE_ID: u64 = 0x0;
+pub const HTTP3_PUSH_STREAM_TYPE_ID: u64 = 0x1;
+pub const QPACK_ENCODER_STREAM_TYPE_ID: u64 = 0x2;
+pub const QPACK_DECODER_STREAM_TYPE_ID: u64 = 0x3;
+
+pub enum QlogEvent {
```
ghedo, line 53: Also, this could use some docs, maybe just a description of what it is.

evanrittenhouse:
All of my comments are resolved by https://github.com/cloudflare/quiche/pull/1822

h3i/src/actions/h3.rs
LPardue: actually these public types are a hangover, so I'll remove them altogether

evanrittenhouse: Oh, I found a place we use them internally. We probably want them back

LPardue: We don't need to shadow the record replay type, can you come up with a solution to use just one?

evanrittenhouse: Yeah, all good

h3i/src/prompts/h3/errors.rs
```diff
@@ -0,0 +1,253 @@
+use inquire::error::InquireResult;
```
ghedo, line 1: Nit: needs copytight/license header.

ghedo (approved):

PR #1824, opened 2024-07-26
Make host:port argument optional

evanrittenhouse:
`host:port` is now optional, though `--connect-to` is required if it's missing to prevent useless no-ops. Tested with:
```shell
cargo run --bin h3i -- --connect-to 127.0.0.1:8081
```

And observing a handshake failure as expected.


LPardue:
I don't think it makes sense to omit this from the CLI. Most people will expect to connect to a server without having g to specify the IP. If you want to omit the SNI from the handshake, that should be an explicit option

PR #1825, opened 2024-07-26
Expose the `HTTP3_CONTROL_STREAM_TYPE_ID` constant, make SNI optional

evanrittenhouse:


h3i/src/config.rs
```diff
@@ -137,7 +139,9 @@ impl Config {
     }
 
     pub fn build(self) -> Result<Self, io::Error> {
-        if self.host_port.is_empty() {
+        let no_host_port = self.host_port.is_none() ||
+            self.host_port.as_ref().filter(|s| *s == "").is_some();
+        if no_host_port && self.connect_to.is_none() {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
                 "Must provide a <host:port> to connect".to_string(),
```
LPardue, line 147: need to update the error message

h3i/src/main.rs
```diff
@@ -93,14 +93,15 @@ fn config_from_clap() -> std::result::Result<Config, String> {
         .arg(
             Arg::with_name("host:port")
                 .help("Hostname and port of the HTTP/3 server")
-                .required(true)
+                .required(false)
```
LPardue, line 96: This should stay required. If we want to omit the SNI, then add a new `--no-sni` option, which means you can avoid changing the prompter code and then just pass None to `quiche::connect` if we don't want one

PR #1827, opened 2024-07-30
Maintain DATAGRAM sent/received statistics

hendrikcech:
This PR tracks the number of sent and received DATAGRAM frames on the connection-level and on individual paths. It provides more visibility into what has been transferred over the connection. Currently only the number of sent and received packets is tracked.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1411,6 +1411,12 @@ pub struct Connection {
     /// Total number of packets sent with data retransmitted.
     retrans_count: usize,
 
+    /// Total number of sent DATAGRAM frames.
+    dgram_sent_count: usize,
+
+    /// Total number of received DATAGRAM frames.
+    dgram_recv_count: usize,
+
     /// Total number of bytes received from the peer.
     rx_data: u64,
 
@@ -1908,6 +1914,8 @@ impl Connection {
             sent_count: 0,
             lost_count: 0,
             retrans_count: 0,
+            dgram_sent_count: 0,
+            dgram_recv_count: 0,
             sent_bytes: 0,
             recv_bytes: 0,
             acked_bytes: 0,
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -1408,6 +1408,12 @@ pub struct Connection {
     /// Total number of packets sent with data retransmitted.
     retrans_count: usize,
 
+    /// Total number of sent DATAGRAM frames.
+    dgram_sent_count: usize,
+
+    /// Total number of received DATAGRAM frames.
+    dgram_recv_count: usize,
```
LPardue, line 1415: FWIW, in theory its possible for a minimally encoded length-prefixed DATAGRAM to take about 2 bytes and therefore a single packet to contain a lot of them. For a standard MTU, this would be like 600-700 but larger MTUs are possible. Therefore, there's a risk that the limit here could be breached orders of magnitude before the QUIC connection hit its packet count limit. 

Not sure if that warrants a saturating_add on the counting but wanted to flag before I forgot.

hendrikcech: Good point. I changed `+= 1` to `saturating_add(1)` and rebased the commit on top of master.

ghedo (approved):

ghedo:
Merged, thanks @hendrikcech!

quiche/src/lib.rs
```diff
@@ -7418,6 +7432,13 @@ impl Connection {
                 }
 
                 self.dgram_recv_queue.push(data)?;
+
+                let _ = self.dgram_recv_count.saturating_add(1);
```
antoniovicente, line 7436: I think this code doesn't do anything.

This is roughly equal to
`let _ = self.dgram_recv_count + 1;`
which doesn't mutate self.dgram_recv_count

I think that what you meant is closer to
`self.dgram_recv_count += 1;` or `self.dgram_recv_count.fetch_add(1);`


Consider changing the previous line to:
`self.dgram_recv_count = self.dgram_recv_count.saturating_add(1);`

Same to all of ther saturating_add calls in this PR.

antoniovicente: https://github.com/cloudflare/quiche/pull/2268

PR #1830, opened 2024-08-05
Fix largest performance issue in profiler

catid:
I'm not great at writing Rust code, but this was an easy fix.  In callgrind it was using 38% of the CPU time of the application, and afterwards 0.


LPardue:
@catid Thanks for the PR. I'm not quite sure how you were profiling but it doesn't seem like there should really be an issue from the allocation alone. Instead, there's work that can be skipped entirely if there is nothing to read on the QPACK stream . I made https://github.com/cloudflare/quiche/pull/1841 to address that, maybe you'd like to recheck your numbers with that change?

PR #1835, opened 2024-08-26
Add `CONNECTION_CLOSE` frame to h3i actions 

evanrittenhouse:


h3i/src/prompts/h3/mod.rs
```diff
@@ -423,6 +429,23 @@ fn prompt_extension() -> InquireResult<Action> {
     Ok(action)
 }
 
+pub fn prompt_connection_close() -> InquireResult<Action> {
+    let is_app = prompt_yes_no("application close:")?;
+    let error_code = prompt_varint("error code:")?;
+    let reason = Text::new("reason:")
+        .with_placeholder("empty")
```
LPardue, line 436: The field should be `reason phrase:` and the placeholder can say something like <optional reason phrase>.

h3i/src/prompts/h3/mod.rs
```diff
@@ -423,6 +429,23 @@ fn prompt_extension() -> InquireResult<Action> {
     Ok(action)
 }
 
+pub fn prompt_connection_close() -> InquireResult<Action> {
+    let is_app = prompt_yes_no("application close:")?;
```
LPardue, line 433: there's already a `prompt_transport_or_app()` to reuse

h3i/src/prompts/h3/mod.rs
```diff
@@ -423,6 +429,23 @@ fn prompt_extension() -> InquireResult<Action> {
...
+    let error_code = prompt_varint("error code:")?;
```
LPardue, line 434: There's already a transport error code suggester in `prompt_close_stream` that we should reuse.

LPardue (changes requested):

h3i/src/prompts/h3/mod.rs
```diff
@@ -423,6 +429,23 @@ fn prompt_extension() -> InquireResult<Action> {
...
+        .prompt()
+        .unwrap_or(String::new());
+
+    Ok(Action::ConnectionClose {
+        error: ConnectionError {
+            is_app,
+            error_code,
+            reason: reason.as_bytes().to_vec(),
+        },
+    })
```
LPardue, line 446: Once a connection is closed, many other actions are no longer possible. It would be nice to find a way for the prompter to reflect that to the user. This might require customizing the top-level suggestions returned in the loop in `prompt()`

h3i/src/actions/h3.rs
```diff
@@ -91,6 +92,11 @@ pub enum Action {
         error_code: u64,
     },
 
+    /// Send a CONNECTION_CLOSE frame witht the given [`ConnectionError`].
```
LPardue, line 95:
```suggestion
    /// Send a CONNECTION_CLOSE frame with the given [`ConnectionError`].
```

h3i/src/recordreplay/qlog.rs
```diff
@@ -254,7 +255,39 @@ impl From<&Action> for QlogEvents {
...
+                } else {
+                    Some(String::from_utf8(error.reason.clone()).unwrap())
+                };
+
+                let ev = fake_packet_sent(Some(smallvec![
+                    QuicFrame::ConnectionClose {
+                        error_space: Some(error_space),
+                        error_code: Some(error.error_code),
+                        // https://github.com/cloudflare/quiche/issues/1731
+                        error_code_value: Some(0),
```
LPardue, line 276: set to None

h3i/src/recordreplay/qlog.rs
```diff
@@ -336,6 +369,33 @@ impl From<&PacketSent> for H3Actions {
                         error_code: *error_code,
                     }),
 
+                    QuicFrame::ConnectionClose {
+                        error_space,
+                        error_code,
+                        reason,
+                        ..
+                    } => {
+                        let is_app = error_space
```
LPardue, line 378: qlog allows any of these fields to be `None`, so the code should be more defensive. If there is no error_space or error_code, then there is no reasonable default to assume and we should log it and either bomb out, or skip the action.

h3i/src/prompts/h3/mod.rs
evanrittenhouse: Total brain fart, nice catch

h3i/src/prompts/h3/mod.rs
evanrittenhouse: I wonder if we should refactor all of that out into a separate `errors.rs` file

h3i/src/recordreplay/qlog.rs
evanrittenhouse: Changed to `expect` the error space and added a more descriptive `expect` on the error code, rather than unwrap

h3i/src/prompts/h3/mod.rs
evanrittenhouse: After messing around with this, I think it's a larger change that warrants a separate PR. `prompt_action` will need to know about connection state at that point

h3i/src/prompts/h3/mod.rs
```diff
@@ -423,6 +430,22 @@ fn prompt_extension() -> InquireResult<Action> {
     Ok(action)
 }
 
+pub fn prompt_connection_close() -> InquireResult<Action> {
+    let (error_space, error_code) = errors::prompt_transport_or_app_error()?;
+    let reason = Text::new("reason phrase:")
+        .with_placeholder(".")
```
LPardue, line 436: can this say something useful like "optional freeform data" or something?

h3i/src/prompts/h3/mod.rs
LPardue: all good

LPardue (approved):
LGTM module one nit about placeholder text

PR #1836, opened 2024-08-27
Force Handshake packet epoch if handshake hasn't been confirmed

evanrittenhouse:
The RFC states that an endpoint can immediately close a connection by
sending a CONNECTION_CLOSE frame during the handshake:
> However, prior to confirming the handshake, it is possible that more
  advanced packet protection keys are not available to the peer, so
  another CONNECTION_CLOSE frame MAY be sent in a packet that uses a lower
  packet protection level.

Currently, when we decide the packet epoch when a CC frame is present,
we always choose Application. This commit downgrades the epoch to
Handshake if the handshake wasn't confirmed.

The test added in this PR fails without this patch - since the packet is
sent in the Application space without keys, the server can't decrypt it,
so its `peer_error` is None.

- [x] Fix CI
- [x] Squash

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -6823,7 +6823,7 @@ impl Connection {
                 crypto::Level::OneRTT => packet::Epoch::Application,
             };
 
-            if !self.is_established() {
+            if !self.handshake_confirmed {
                 match epoch {
                     // Downgrade the epoch to Handshake as the handshake is not
                     // completed yet.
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -6816,11 +6816,26 @@ impl Connection {
...
+                        packet::Epoch::Application
+                    } else {
+                        // If the handshake hasn't been confirmed, we can't send
+                        // 1-RTT packets since the peer
+                        // may not have the correct keys. Downgrade the status to
+                        // Handshake - if required, we'll downgrade further to
+                        // Initial later on.
+                        //
+                        // https://datatracker.ietf.org/doc/html/rfc9000#name-immediate-close-during-the-
+                        force_downgrade = true;
```
evanrittenhouse, line 6835: The problem with checking `handshake_confirmed` naively is that the client drops the Initial epoch state upon sending a Handshake packet. That means that downgrading all `Handshake` epoch packets naively could cause the client to drop the newly-downgraded packet

This boolean is super ugly, so I'm very open to changing it if that's preferre. I see two other alternatives:
1. Refactor such that we can trigger another CC frame getting flushed in a separate packet space. Unless I'm missing something (very possible), this seems like it'd be a lot of work. However, it's closer what the RFC recommends: 
> Prior to confirming the handshake, a peer might be unable to process 1-RTT packets, so an endpoint SHOULD send a CONNECTION_CLOSE frame in both Handshake and 1-RTT packets. A server SHOULD also send a CONNECTION_CLOSE frame in an Initial packet.
2. Change the client side so that we only drop the keys for the Initial space when the handshake gets confirmed, since that would force the client and server to be in sync.

ghedo (approved):

PR #1852, opened 2024-09-26
h3: earlier check for control and QPACK stream termination

LPardue:
h3: earlier checks for control and QPACK stream termination

A peer can FIN or reset a control stream at any point during the stream
lifetime. While doing so is an error in HTTP/3, our handling before this change
was a bit odd.

Previously, depending on the sequencing, an unexpected termination of such a
stream could trigger an InvalidStreamState or StreamReset error during poll().
This would bubble up to the app, which is weird and unexpected because the
quiche h3 layer is supposed to deal with all the control stream details. While
both those errors contain a stream ID, the app isn't given any info that the ID
is a control stream.

While any subsequent call to poll() would then cause the stream_finished
detection to kick in and enforce the RFC rules to close the connection with
CloseCriticialStream, that's a bit late and clunky.

This change adds additional checks for the control and QPACK stream terminations
in order to make sure we close connections at the earliest time and prevent
useless errors getting bubbled up to apps.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -931,6 +931,26 @@ pub struct Stats {
     pub qpack_decoder_stream_recv_bytes: u64,
 }
 
+fn close_conn_critical_stream(conn: &mut super::Connection) -> Result<()> {
+    conn.close(
+        true,
+        Error::ClosedCriticalStream.to_wire(),
+        b"Critical stream closed.",
+    )?;
+
+    Err(Error::ClosedCriticalStream)
+}
+
+fn close_conn_if_critical_stream_finished(
+    conn: &mut super::Connection, stream_id: u64,
+) -> Result<()> {
+    if conn.stream_finished(stream_id) {
+        close_conn_critical_stream(conn)?;
+    }
+
+    Ok(())
+}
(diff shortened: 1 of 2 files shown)


quiche/src/h3/mod.rs
```diff
@@ -931,6 +931,26 @@ pub struct Stats {
...
+    conn.close(
+        true,
+        Error::ClosedCriticalStream.to_wire(),
+        b"Critical stream closed.",
+    )?;
+
+    Err(Error::ClosedCriticalStream)
+}
+
+fn check_closed_critical_stream(
```
ghedo, line 944:
```suggestion
fn close_finished_critical_stream_with_error(
```

Bit of a mouthful, but seems more accurate?

ghedo: Or, idk, `close_critical_stream_if_finished(`

LPardue: I went with `close_conn_if_critical_stream_finished()`

ghedo: Oh lol, yeah, that makes way more sense.

ghedo (approved):

PR #1855, opened 2024-09-28
apps: Implement mutual tls with optional strict trust

nnathan:
Hi,

I'm looking at using QUIC for doing mutual TLS. It took me awhile to figure out even though it's rather easy to implement. I first reached for the ssl context builder toolbox because I thought that might be required, but I eventually landed on a few simple additions to the client & server.

Includes everything from #1854 .


Daniel-dev22:
Any chance this will be merged at some point? This would be a great feature.

PR #1860, opened 2024-10-14
support en/decoding the accurate ECN transport param

hawkinsw:
Properly encode/decode the accurate ECN transport param (I-D.seemann-quic-accurate-ack-ecn,
https://datatracker.ietf.org/doc/html/draft-seemann-quic-accurate-ack-ecn-01).

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -8023,6 +8023,8 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
+    /// Whether accurate ECN support is enabled.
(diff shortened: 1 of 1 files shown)


hawkinsw:
I added this small feature so that I could detect whether any implementations are currently advertising support for accurate ECN. I thought that working with quiche was an awesome experience and really appreciate all the work that you have done on the implementation. I know that the work is only an I-D at this point, but I thought I would offer it back regardless.

Thanks again for the great work on quiche!

LPardue:
Can we retire this one since the motivation is subsumed by https://github.com/cloudflare/quiche/pull/1861?

LPardue:
> Can we retire this one since the motivation is subsumed by #1861?

ping @hawkinsw

hawkinsw:
Absolutely!!

PR #1861, opened 2024-10-16
lib, qlog: Track and Log Unknown Transport Parameters

hawkinsw:
Track (and make available to qlog for logging) transport parameters and their values that this implementation does not specifically recognize.

### DIFF
--- a/qlog/src/events/quic.rs
+++ b/qlog/src/events/quic.rs
@@ -571,6 +571,14 @@ pub struct TransportParametersSet {
     pub initial_max_streams_uni: Option<u64>,
 
     pub preferred_address: Option<PreferredAddress>,
+
+    pub unknown_parameters: Vec<UnknownTransportParameter>,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct UnknownTransportParameter {
+    pub id: u64,
+    pub value: Bytes,
 }
 
 #[serde_with::skip_serializing_none]
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -382,6 +382,7 @@
 #[macro_use]
 extern crate log;
 
+use octets::BufferTooShortError;
 #[cfg(feature = "qlog")]
 use qlog::events::connectivity::ConnectivityEventType;
 #[cfg(feature = "qlog")]
@@ -813,6 +814,8 @@ pub struct Config {
     max_amplification_factor: usize,
 
     disable_dcid_reuse: bool,
+
+    track_unknown_transport_params: Option<usize>,
 }
 
 // See https://quicwg.org/base-drafts/rfc9000.html#section-15
@@ -881,6 +884,8 @@ impl Config {
             max_amplification_factor: MAX_AMPLIFICATION_FACTOR,
 
             disable_dcid_reuse: false,
+
+            track_unknown_transport_params: None,
         })
     }
 
@@ -1350,6 +1355,18 @@ impl Config {
     pub fn set_disable_dcid_reuse(&mut self, v: bool) {
         self.disable_dcid_reuse = v;
     }
+
+    /// Enables tracking unknown transport parameters.
+    ///
+    /// Specify the maximum number of bytes used to track unknown transport
+    /// parameters. The size includes the identifier and its value. If storing a
+    /// transport parameter would cause the limit to be exceeded, it is quietly
+    /// dropped.
+    ///
+    /// The default is that the feature is disabled.
+    pub fn enable_track_unknown_transport_parameters(&mut self, size: usize) {
+        self.track_unknown_transport_params = Some(size);
+    }
 }
 
 /// A QUIC connection.
@@ -1371,6 +1388,10 @@ pub struct Connection {
 
     /// Peer's transport parameters.
     peer_transport_params: TransportParams,
+
+    /// If tracking unknown transport parameters from a peer, how much space to
+    /// use in bytes.
+    peer_transport_params_track_unknown: Option<usize>,
 
     /// Local transport parameters.
     local_transport_params: TransportParams,
@@ -1895,6 +1916,9 @@ impl Connection {
 
             peer_transport_params: TransportParams::default(),
 
+            peer_transport_params_track_unknown: config
+                .track_unknown_transport_params,
+
             local_transport_params: config.local_transport_params.clone(),
 
             handshake: tls,
@@ -2198,8 +2222,11 @@ impl Connection {
         let raw_params_len = b.get_u64()? as usize;
         let raw_params_bytes = b.get_bytes(raw_params_len)?;
 
-        let peer_params =
-            TransportParams::decode(raw_params_bytes.as_ref(), self.is_server)?;
+        let peer_params = TransportParams::decode(
+            raw_params_bytes.as_ref(),
+            self.is_server,
+            self.peer_transport_params_track_unknown,
+        )?;
 
         self.process_peer_transport_params(peer_params)?;
 
@@ -6788,8 +6815,11 @@ impl Connection {
                 let raw_params = self.handshake.quic_transport_params();
 
                 if !self.parsed_peer_transport_params && !raw_params.is_empty() {
-                    let peer_params =
-                        TransportParams::decode(raw_params, self.is_server)?;
+                    let peer_params = TransportParams::decode(
+                        raw_params,
+                        self.is_server,
+                        self.peer_transport_params_track_unknown,
+                    )?;
 
                     self.parse_peer_transport_params(peer_params)?;
                 }
@@ -6807,8 +6837,11 @@ impl Connection {
         let raw_params = self.handshake.quic_transport_params();
 
         if !self.parsed_peer_transport_params && !raw_params.is_empty() {
-            let peer_params =
-                TransportParams::decode(raw_params, self.is_server)?;
+            let peer_params = TransportParams::decode(
+                raw_params,
+                self.is_server,
+                self.peer_transport_params_track_unknown,
+            )?;
 
             self.parse_peer_transport_params(peer_params)?;
         }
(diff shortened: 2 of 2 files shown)


hawkinsw:
@LPardue Let me know what you think about the _direction_ of this PR. If you think that it is on the right track, I will gladly do whatever you say so that it meets the high standards of the code that you have already written in this library.

quiche/src/lib.rs
```diff
@@ -8196,8 +8222,13 @@ impl TransportParams {
                     tp.max_datagram_frame_size = Some(val.get_varint()?);
                 },
 
-                // Ignore unknown parameters.
-                _ => (),
+                // Track unknown transport parameters specially.
+                unknown_tp_id => {
+                    tp.unknown_params.push(UnknownTransportParameter {
```
LPardue, line 8227: we probably want to put an upper bound on the number of these that we would store, because I don't think there's limit in TLS (other than the full length of the QUIC crypto stream).

I don't know what the bounds should be right now, but as inspiration we have something alongthe same lines for SETTINGS - https://github.com/cloudflare/quiche/blob/9ea2152571bbe48f0a819ebb545f62dcf7d1f738/quiche/src/h3/frame.rs#L50. I don't think it's 100% the same but might be a useful reference point.

quiche/src/lib.rs
```diff
@@ -8952,6 +8989,7 @@ mod tests {
             initial_source_connection_id: Some(b"woot woot".to_vec().into()),
             retry_source_connection_id: Some(b"retry".to_vec().into()),
             max_datagram_frame_size: Some(32),
+            unknown_params: vec![],
```
LPardue, line 8992: Its sad we can't test this, but that's because we don't allow apps to send unknown values via the quiche API. I was partly suprised to see we don't send grease TPS (https://datatracker.ietf.org/doc/html/rfc9000#name-reserved-transport-paramete) so maybe that's something we could consider. Although maybe doing it separately might be wise (and the tests could be updated if/when that happens)

hawkinsw: I believe that I have found a way to do at least a little testing on the new feature. Let me know what you think!

quiche/src/lib.rs
hawkinsw: Let me know if you like the latest revision. I decided against using `smallvec` because I don't think that users will see unknown parameters often enough to warrant pre-allocating space for any entries.

quiche/src/lib.rs
```diff
@@ -8023,7 +8113,9 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
-    // pub preferred_address: ...,
+    /// Unknown peer transport parameters and values, if any.
+    pub unknown_params: UnknownTransportParameters, /* pub preferred_address:
+                                                     * ..., */
```
LPardue, line 8118: I think cargo fmt might have tried to be clever here with the `preferred_address` . Can we move it to its own line underneath please?

quiche/src/lib.rs
```diff
@@ -7984,6 +7985,95 @@ impl std::fmt::Debug for Stats {
...
+    fn default() -> Self {
+        Self {
+            capacity: Self::MAX_UNKNOWN_TRANSPORT_PARAM_SIZE,
+            parameters: vec![],
+        }
+    }
+}
+
+impl UnknownTransportParameters {
+    const MAX_UNKNOWN_TRANSPORT_PARAM_SIZE: usize = 256;
```
LPardue, line 8033: I'm wondering if we make this configurable, and the default be 0. That way, an app can explicitly opt-in to TP storage. One reason this might be useful is that the server has to allocate for unauthenticated connections, and even allocatiing 256 bytes is something that would be nice to avoid when we don't want to.

If you agree, there's a similar example for `enable_dgram` https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L1294

quiche/src/lib.rs
```diff
@@ -7984,6 +7985,95 @@ impl std::fmt::Debug for Stats {
...
+
+    /// Push an unknown transport parameter into storage if there is space
+    /// remaining.
+    pub fn safe_push(&mut self, new: UnknownTransportParameter) -> Result<()> {
+        let new_unknown_tp_size = new.value.len();
```
LPardue, line 8038: If I read this correct. it might allow an "infinite" amount of zero-length TPs. So the capacity limit should probably factor in the size of the TP identifier two (8 bytes). Probably easiest to add a `len()` function to UnknownTransportParameter type.

LPardue (changes requested):
Overall looks like its headed in the right direction. Left a few comments to help tighten up the defensiveness of the code.

quiche/src/lib.rs
hawkinsw: You are absolutely correct -- good eyes! I have fixed that in the most recent version!

quiche/src/lib.rs
hawkinsw: I agree and have implemented it accordingly (in the latest version of the PR).

quiche/src/lib.rs
```diff
@@ -1350,6 +1355,20 @@ impl Config {
     pub fn set_disable_dcid_reuse(&mut self, v: bool) {
         self.disable_dcid_reuse = v;
     }
+
+    /// Configures tracking unknown transport params.
```
LPardue, line 1359:
```suggestion
    /// Configures tracking unknown transport parameters.
```

quiche/src/lib.rs
```diff
@@ -1350,6 +1355,20 @@ impl Config {
...
+    ///
+    /// To enable, pass `Some` `usize` with the maximum
+    /// amount of space to expend to track the ids and
+    /// values of unknown transport parameters. To disable,
+    /// pass `None`.
+    ///
+    /// The default is that the feature is disabled.
+    pub fn enable_track_unknown_transport_parameters(
+        &mut self, size: Option<usize>,
```
LPardue, line 1368:
```suggestion
        &mut self, size: usize,
```

Since the default is off,  I don't see much value in allowing a caller to disable it. Therefore I think we can just make this a value without wrapping in an option, then mark it as Some(size) inside the function. This also makes it easier to deal with in the FFI, which is something this PR should also do (see quiche.g and ffi.rs).

quiche/src/lib.rs
```diff
@@ -1350,6 +1355,20 @@ impl Config {
     pub fn set_disable_dcid_reuse(&mut self, v: bool) {
         self.disable_dcid_reuse = v;
     }
+
+    /// Configures tracking unknown transport params.
+    ///
+    /// To enable, pass `Some` `usize` with the maximum
+    /// amount of space to expend to track the ids and
+    /// values of unknown transport parameters. To disable,
+    /// pass `None`.
```
LPardue, line 1364: See comment below on the signature but it probably wants to be tweaked to something like 
```suggestion
    /// Enables tracking unknown transport parameters. 
    /// The provided `size` is the maximum size in bytes that
    /// will be allocated to storing transport parameter IDs
    /// and their values.
```

qlog/src/events/quic.rs
```diff
@@ -571,6 +571,15 @@ pub struct TransportParametersSet {
...
     pub preferred_address: Option<PreferredAddress>,
+
+    pub unknown_parameters: Vec<UnknownTransportParameter>,
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct UnknownTransportParameter {
+    pub id: u64,
+    pub value: Vec<u8>,
```
LPardue, line 582:
```suggestion
    pub value: Bytes,
```

Strictly this is a `hexstring` in qlog; see https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-packet_sent and https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#section-1.2. Basically just a string with some character constraints

We don't do a good job of dealing with that consistently through the codebase (and its a future tasks to do it at somepoint, so not asking you to do it in this PR).  

However, this qlog event should be a stringy type (Bytes is a badly names alias for String in this crate).

When it comes to serializing the actual bytes for the qlog event, we have the `qlog::HexSlice::maybe_string` util function. However, I'm not 100% sure that'll work for your impl of `From`

quiche/src/lib.rs
```diff
@@ -8196,8 +8323,18 @@ impl TransportParams {
...
+                // Track unknown transport parameters specially.
+                unknown_tp_id =>
+                    if track_unknown_parameters {
+                        // It is _not_ an error not to have space enough to track
+                        // an unknown parameter.
+                        let _track_unknown_parameter_result = tp
+                            .unknown_params
+                            .safe_push(UnknownTransportParameter {
+                                id: unknown_tp_id,
+                                value: val.to_vec(),
```
LPardue, line 8335: IIUC this will still require an allocation even if there is no capacity. Yes, there wouldn't be unbound memory growth, but it could still consume needless cycles. 

An alternative would be to change the signature of safe_push to be `id: u64, value: &[u8]`, then inside it you can create UnknownTransportParameter with a vec copy, only if you know there's capacity. You can use `val.as_ref()` to get the ref to pass

LPardue (commented):
Looking good, just a few minor things to harden it up a bit more.

quiche/src/lib.rs
hawkinsw: I just want to confirm what I think you are saying: The `to_vec()` is going to consume cycles and (potentially) cause an allocation just to pass it to the `safe_push` member function and, even after all that work, it may not even use it!

If I am hearing what you are saying, then I absolutely agree and will make the change!

LPardue: Yep!

hawkinsw: 10-4 -- I am on it!

hawkinsw: I have a pretty "cool" implementation of this. I want to look over it a little more before I send you another revision of this PR. Sorry for the delay and thank you for the feedback!

LPardue (commented):
a few little nits but otherwise very close

quiche/src/lib.rs
```diff
@@ -1350,6 +1355,16 @@ impl Config {
     pub fn set_disable_dcid_reuse(&mut self, v: bool) {
         self.disable_dcid_reuse = v;
     }
+
+    /// Enablestracking unknown transport parameters.
```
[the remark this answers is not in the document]
hawkinsw: I have no idea how _that_ slipped through! Sorry!

quiche/src/lib.rs
```diff
@@ -7984,6 +8013,102 @@ impl std::fmt::Debug for Stats {
...
+    }
+}
+
+impl From<UnknownTransportParameter<&[u8]>>
+    for UnknownTransportParameter<Vec<u8>>
+{
+    // When an instance of an UnknownTransportParameter is actually
+    // stored in UnknownTransportParameters, then we make a copy
+    // of the bytes if the source is an instance of an UnknownTransportParameter
+    // that whose value is not owned.
```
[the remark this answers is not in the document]
hawkinsw: Yeesh. Sorry again!

qlog/src/events/quic.rs
```diff
@@ -571,6 +571,15 @@ pub struct TransportParametersSet {
     pub initial_max_streams_uni: Option<u64>,
 
     pub preferred_address: Option<PreferredAddress>,
+
+    pub unknown_parameters: Vec<UnknownTransportParameter>,
+}
+
+#[serde_with::skip_serializing_none]
```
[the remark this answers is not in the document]
hawkinsw: I believe that I addressed this (Sorry for note _resolving_ it earlier)!

LPardue (approved):

LPardue:
LGTM thanks @hawkinsw . 

I'd like @ghedo to take a final look over before merging it though

hawkinsw:
Hello @LPardue! I noticed that we missed an opportunity to add a member function to test whether an unknown transport parameter's id is in the reserved space. I have updated the PR! Let me know what you think!

(Also, I rebased to make it easier to review).

hawkinsw:
I rebased against the updated master to resolve build errors.

quiche/src/lib.rs
```diff
@@ -1888,7 +1909,8 @@ impl Connection {
             next_pkt_num: 0,
 
             peer_transport_params: TransportParams::default(),
-
+            peer_transport_params_track_unknown: config
```
ghedo, line 1912:
```suggestion

            peer_transport_params_track_unknown: config
```

quiche/src/lib.rs
```diff
@@ -1888,7 +1909,8 @@ impl Connection {
...
+                .track_unknown_transport_params,
```
ghedo, line 1913:
```suggestion
                .track_unknown_transport_params,

```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
     }
 }
 
+/// QUIC Unknown Transport Parameter
```
ghedo, line 8037:
```suggestion
/// QUIC Unknown Transport Parameter.
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+///
+/// A QUIC transport parameter that is not specifically recognized
+/// by this implementation.
+#[derive(Clone, Debug, PartialEq)]
+pub struct UnknownTransportParameter<T> {
+    /// ID of an unknown transport parameter
```
ghedo, line 8043:
```suggestion
    /// The ID of the unknown transport parameter.
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+    pub id: u64,
+    /// Original data representing the value of an unknown transport parameter
```
ghedo, line 8045:
```suggestion

    /// Original data representing the value of the unknown transport parameter.
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+    pub value: T,
+}
+
+impl<T> UnknownTransportParameter<T> {
+    /// Check whether an unknown Transport Parameter's ID is in the reserved
```
ghedo, line 8050:
```suggestion
    /// Checks whether an unknown Transport Parameter's ID is in the reserved
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+            id: value.id,
+            value: value.value.to_vec(),
+        }
+    }
+}
+
+/// Track unknown transport parameters, up to a limit.
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct UnknownTransportParameters {
+    // The maximum number of bytes available to store the values of unknown
```
ghedo, line 8091:
```suggestion
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+    // transport parameters.
```
ghedo, line 8092:
```suggestion
```

quiche/src/lib.rs
```diff
@@ -8003,6 +8034,113 @@ impl std::fmt::Debug for Stats {
...
+    /// The space remaining for storing unknown transport parameters.
+    pub capacity: usize,
+    /// The unknown transport parameters.
+    pub parameters: Vec<UnknownTransportParameter<Vec<u8>>>,
+}
+
+impl UnknownTransportParameters {
+    /// Push an unknown transport parameter into storage if there is space
```
ghedo, line 8100:
```suggestion
    /// Pushes an unknown transport parameter into storage if there is space
```

quiche/src/lib.rs
```diff
@@ -8042,6 +8180,8 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
+    /// Unknown peer transport parameters and values, if any.
+    pub unknown_params: UnknownTransportParameters,
```
ghedo, line 8184: Why not just make this an option? It would also simplify the decode implementation a bit, so instead of:
```
if track_unknown_parameters {
```
you could just do
```
if tp.unknown_params.is_some()
```

ghedo (changes requested):
Added some nits and minor comments, but looks good overall.

quiche/src/lib.rs
hawkinsw: Thank you! I _think_ that I properly incorporated all these suggestions! I will close them just to save you the extra clicking, but if I missed anything, _please_ let me know!

quiche/src/lib.rs
```diff
@@ -8042,6 +8181,8 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
+    /// Unknown peer transport parameters and values, if any.
+    pub unknown_params: Option<UnknownTransportParameters>,
```
hawkinsw, line 8185: Let me know, @ghedo , if this was the type of change that you were suggesting. I _do_ think that it makes the code easier to read (and removes the need for that extra boolean). I sometimes don't like the ergonomics of working with `Option`s in Rust (when borrows are involved), but I think that your suggestion really made a big improvement. Again, please let me know if I captured your suggestion properly.

hawkinsw (commented):
Thank you again, @ghedo , for the review! I hope that I properly addressed your feedback!

ghedo (approved):

ghedo:
Added a couple additional minor tweaks (e.g. renamed safe_push -> push) and merged, thanks @hawkinsw!

PR #1867, opened 2024-11-15
h3i: add --replay-host-override option

LPardue:
One use case for replaying a recorded session with the h3i CLI is to execture
the same sequence of actions to different target servers. Many servers will
validate that there is a provided "host" or ":authority" header field, and that
it matches the certificate SAN.

This change adds the "--replay-host-override" option that can be used in
combination with the --qlog-input option. It replaces any "host" or ":authority"
header in a HEADERS frame, with the provided value.

note: the diff looks huge but most of what its doing is swapping a From impl to a function so the host_override value can be passed to the correct place

### DIFF
--- a/h3i/src/actions/h3.rs
+++ b/h3i/src/actions/h3.rs
@@ -49,7 +49,7 @@ use crate::encode_header_block;
 /// sequentially. Note that packets will be flushed when said iteration has
 /// completed, regardless of if an [`Action::FlushPackets`] was the terminal
 /// action.
-#[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


evanrittenhouse:
One thing we could do to test h3i is to add sample qlogs as test data and run the parser against them to ensure they generate the correct actions, and vice versa. Might be a good first one to try with the new option

LPardue:
Good idea

LPardue:
I added some simple tests for parsing a qlog with http headers with and without a host_override. Other parts of parsing are already covered in the qlog QlogSeqReader  itself

evanrittenhouse (approved):

PR #1868, opened 2024-11-16
h3i: extract connection creation to its own function

evanrittenhouse:
Most of the diff is just copy-pasting the function from one place to another. This will deduplicate a bunch of code between the sync and async versions

### DIFF
--- a/h3i/src/client/mod.rs
+++ b/h3i/src/client/mod.rs
@@ -38,11 +38,13 @@ use qlog::events::h3::HttpHeader;
 use quiche::ConnectionError;
 
 use std::collections::HashMap;
+use std::net::SocketAddr;
 use std::time::Instant;
 
 use crate::actions::h3::Action;
 use crate::actions::h3::StreamEvent;
 use crate::actions::h3::StreamEventType;
+use crate::config::Config;
 use crate::frame::H3iFrame;
 use crate::frame::ResetStream;
 use crate::frame_parser::FrameParseResult;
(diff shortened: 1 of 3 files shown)


h3i/src/client/mod.rs
```diff
@@ -55,10 +58,102 @@ use qlog::events::h3::Http3Frame;
...
+        url::Url::parse(&x)
+            .unwrap()
+            .to_socket_addrs()
+            .unwrap()
+            .next()
+            .unwrap()
+    };
+
+    // Create the configuration for the QUIC connection.
+    let mut config = quiche::Config::new(1).unwrap();
```
LPardue, line 93: Lets make `1` a `const QUIC_VERISION: u32 = 1`

h3i/src/client/mod.rs
```diff
@@ -55,10 +58,102 @@ use qlog::events::h3::Http3Frame;
...
 use quiche::h3::Error;
 use quiche::h3::NameValue;
+use quiche::Connection;
+use quiche::{
+    self,
+};
+
+const MAX_DATAGRAM_SIZE: usize = 1350;
+
+fn build_quiche_connection(args: Config, local_addr: SocketAddr) -> Connection {
```
LPardue, line 71: Should we wrap this in a result to avoid needing the unwraps?

h3i/src/client/sync_client.rs
```diff
@@ -66,13 +66,6 @@ pub fn connect(
...
 
-    // We'll only connect to one server.
-    let connect_url = if !args.omit_sni {
-        args.host_port.split(':').next()
-    } else {
-        None
-    };
-
     // Resolve server address.
     let peer_addr = if let Some(addr) = &args.connect_to {
```
LPardue, line 70: This code is now duplicated in `build_quiche_connection()`, so we should find a way to avoid that

LPardue (changes requested):
Looks good overall, needs a few nits addressing though

h3i/src/client/sync_client.rs
evanrittenhouse: Ah missed that one, thanks

LPardue (approved):

PR #1872, opened 2024-11-24
h3i: remove usage of http crate

evanrittenhouse:


h3i/src/frame.rs
```diff
@@ -113,12 +109,15 @@ impl From<Vec<Header>> for H3iFrame {
...
+pub type HeaderMap = HashMap<Vec<u8>, Vec<Vec<u8>>>;
+
 /// An HTTP/3 HEADERS frame with decoded headers and a [HeaderMap].
 #[derive(Clone, PartialEq, Eq)]
 pub struct EnrichedHeaders {
     header_block: Vec<u8>,
     headers: Vec<Header>,
-    header_map: HeaderMap<Vec<u8>>,
+    /// A multi-map of raw header names to values, similar to http's HeaderMap.
+    header_map: HeaderMap,
```
LPardue, line 120: This isn't a multi-map from a writing perspective, since its just backed by a raw HashMap. Yes, the parsing code emulates some multimap like behaviour but anyone naively using this type will be surprised. Consider instead using https://docs.rs/multimap/0.10.0/multimap/

evanrittenhouse: Nice

h3i/src/frame.rs
```diff
@@ -427,26 +431,3 @@ impl<'a> Serialize for SerializableQFrame<'a> {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn test_status_code() {
```
schien, line 436: should we update the test case instead of removing it?

ghedo: There's not much remaining to test, but moar tests is better than less tests I guess :shrug:

ghedo (approved):

h3i/src/frame.rs
schien: Just noticed that doc test is executed in quiche repo. I'm ok with it.
https://github.com/cloudflare/quiche/blob/master/.github/workflows/nightly.yml#L36

PR #1876, opened 2024-12-02
Replace winapi with windows-sys

DoumanAsh:
The major difference is use of unions and better compile times


ghedo (approved):

ghedo:
Merged, thanks @DoumanAsh!

PR #1878, opened 2024-12-02
h3i: add example

evanrittenhouse:
The example can be run with `cargo r --example content_length_mismatch`. TLS keys will be generated at `h3i/h3i-example`, which can be used to decrypt a corresponding PCAP if one is captured separately.


h3i/README.md
```diff
@@ -89,6 +89,19 @@ h3i is also provided as a library, which allows programmatic control over HTTP/3
...
+
+```shell
+cargo run --example content_length_mismatch
+```
+
+If you'd like to decrypt the example's output, you can do so by setting the `SSLKEYLOGFILE` environment variable:
+
+```shell
+SSLKEYLOGFILE="h3i-example" cargo run --example content_length_mismatch
+```
```
LPardue, line 103:
```suggestion
It can be useful to observe the packets on the wire that are exchanged when h3i
is run. Since QUIC is an encrypted transport protocol, tools such as Wireshark
need access to the session keys in order to dissect the packets. A common
approach is to log these keys in response to the `SSLKEYLOGFILE` environment
variable and configure Wireshark to use them; [see
more](https://wiki.wireshark.org/TLS#using-the-pre-master-secret). For example,
to log to the file `h3i-example.keys`:

```shell
SSLKEYLOGFILE="h3i-example.keys" cargo run --example content_length_mismatch
```
```

File: h3i/examples/content_length_mismatch.rs [hunk 2]
```diff
@@ -0,0 +1,74 @@
+use h3i::actions::h3::Action;
+use h3i::client::sync_client;
+use h3i::config::Config;
+use quiche::h3::frame::Frame;
+use quiche::h3::Header;
+use quiche::h3::NameValue;
+
+fn main() {
```
LPardue, line 8: Let's add a short 1-liner to explain what this example does?

h3i/examples/content_length_mismatch.rs
```diff
@@ -0,0 +1,74 @@
...
+    let config = Config::new()
+        .with_host_port("blog.cloudflare.com".to_string())
+        .with_idle_timeout(2000)
+        .build()
+        .unwrap();
+
+    let headers = vec![
+        Header::new(b":method", b"POST"),
+        Header::new(b":scheme", b"https"),
+        Header::new(b":authority", "blog.cloudflare.com".as_bytes()),
```
LPardue, line 18: does this work?
```suggestion
        Header::new(b":authority", b"blog.cloudflare.com"),
```

h3i/examples/content_length_mismatch.rs
```diff
@@ -0,0 +1,74 @@
...
+        Header::new(b":path", b"/"),
+        // We say that we're going to send a body with 5 bytes...
+        Header::new(b"content-length", b"5"),
+    ];
+
+    let header_block = encode_header_block(&headers).unwrap();
```
LPardue, line 24: Let's explain what this does
```suggestion
    // HTTP/3 uses QPACK header compression. Our HEADERS frame carries the encoded
    // QPACK bytes.
    let header_block = encode_header_block(&headers).unwrap();
```

h3i/examples/content_length_mismatch.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let actions = vec![
+        Action::SendHeadersFrame {
+            stream_id: 0,
```
LPardue, line 28: This zero looks magic, so should stick it in a variable with a comment describing briefly how request stream IDs work (mostly just a link to the RFC).

LPardue:
Overall looks good. Thoughts on adding the same example into the crate docs too?

evanrittenhouse:
Good idea, done

h3i/examples/content_length_mismatch.rs
```diff
@@ -0,0 +1,81 @@
+use h3i::actions::h3::Action;
+use h3i::client::sync_client;
+use h3i::config::Config;
+use quiche::h3::frame::Frame;
+use quiche::h3::Header;
+use quiche::h3::NameValue;
+
+/// The QUIC stream to send the frames on. See
+/// https://datatracker.ietf.org/doc/html/rfc9000#name-streams for more.
```
LPardue, line 9: lets also link https://datatracker.ietf.org/doc/html/rfc9114#request-streams

LPardue (approved):

PR #1879, opened 2024-12-04
h3i: close content_length_mismatch example after receiving a response

evanrittenhouse:


LPardue:
Don't forget the doc example.

This looks good if it works

evanrittenhouse:
Yeah it sends the CC after receiving the 400 and takes a tiny amount of time, so I think it works, let me update the doctest (thanks for the reminder)

LPardue (approved):

PR #1890, opened 2024-12-19
h3i: implement close trigger frames

evanrittenhouse:
- [x] Squash commits

Close trigger frames allow the user to specify a list of frames that the h3i client expects to receive over a given connection.

If h3i sees all of the close trigger frames over the course of the connection, it will pre-emptively close the connection with a CONNECTION_CLOSE frame. If h3i does _not_ see all of the trigger frames, the resulting ConnectionSummary will contain a list of the missing target frames for future inspection.

This gives users a way to close tests out without waiting for the idle timeout, or adding Wait/ConnectionClose actions to the end of each test. This should vastly speed up test suites that have a large number of h3i tests.

In the absence of full-fledged integration tests, I modified the `content_length_mismatch` example to expect a `400`, and remove the existing `Wait/CC` actions:
```
-        Action::Wait {
-            wait_type: WaitType::StreamEvent(StreamEvent {
-                stream_id: STREAM_ID,
-                event_type: StreamEventType::Headers,
-            }),
-        },
-        Action::ConnectionClose {
-            error: quiche::ConnectionError {
-                is_app: true,
-                error_code: quiche::h3::WireErrorCode::NoError as u64,
-                reason: vec![],
-            },
-        },
     ];
 
-    let summary =
-        sync_client::connect(config, &actions, None).expect("connection failed");
+    let summary = sync_client::connect(
+        config,
+        &actions,
+        Some(vec![ExpectedFrame::new(
+            0,
+            vec![quiche::h3::Header::new(b":status", b"400")].into(),
+        )]),
+    )
+    .expect("connection failed");
```

Taking a PCAP:
Screenshot 2024-12-19 at 10 27 50 AM

You can see that the client instantly sends the CC frame upon receiving the headers frame, without us needing to specify the action.

### DIFF
--- a/h3i/src/client/connection_summary.rs
+++ b/h3i/src/client/connection_summary.rs
@@ -38,6 +38,7 @@ use std::cmp;
 use std::collections::HashMap;
 use std::iter::FromIterator;
 
+use crate::frame::CloseTriggerFrame;
 use crate::frame::EnrichedHeaders;
 use crate::frame::H3iFrame;
 
@@ -74,22 +75,36 @@ impl Serialize for ConnectionSummary {
             self.path_stats.iter().map(SerializablePathStats).collect();
         state.serialize_field("path_stats", &p)?;
         state.serialize_field("error", &self.conn_close_details)?;
+        state.serialize_field(
+            "missed_close_trigger_frames",
+            &self.stream_map.missing_close_trigger_frames(),
+        )?;
         state.end()
     }
 }
 
 /// A read-only aggregation of frames received over a connection, mapped to the
 /// stream ID over which they were received.
+///
+/// [`StreamMap`] also contains the [`CloseTriggerFrames`] for the connection so
+/// that its state can be updated as new frames are received.
 #[derive(Clone, Debug, Default, Serialize)]
-pub struct StreamMap(HashMap<u64, Vec<H3iFrame>>);
+pub struct StreamMap {
+    stream_frame_map: HashMap<u64, Vec<H3iFrame>>,
+    close_trigger_frames: Option<CloseTriggerFrames>,
+}
 
 impl<T> From<T> for StreamMap
 where
     T: IntoIterator<Item = (u64, Vec<H3iFrame>)>,
 {
     fn from(value: T) -> Self {
-        let map = HashMap::from_iter(value);
-        Self(map)
+        let stream_frame_map = HashMap::from_iter(value);
+
+        Self {
+            stream_frame_map,
+            close_trigger_frames: None,
+        }
     }
 }
 
@@ -113,7 +128,7 @@ impl StreamMap {
     /// assert_eq!(stream_map.all_frames(), vec![headers]);
     /// ```
     pub fn all_frames(&self) -> Vec<H3iFrame> {
-        self.0
+        self.stream_frame_map
             .values()
             .flatten()
             .map(Clone::clone)
@@ -140,7 +155,10 @@ impl StreamMap {
     /// assert_eq!(stream_map.stream(0), vec![headers]);
     /// ```
     pub fn stream(&self, stream_id: u64) -> Vec<H3iFrame> {
-        self.0.get(&stream_id).cloned().unwrap_or_default()
+        self.stream_frame_map
+            .get(&stream_id)
+            .cloned()
+            .unwrap_or_default()
     }
 
     /// Check if a provided [`H3iFrame`] was received, regardless of what stream
@@ -154,8 +172,6 @@ impl StreamMap {
     /// use h3i::frame::H3iFrame;
     /// use quiche::h3::Header;
     /// use std::iter::FromIterator;
-    ///
-    /// let mut stream_map = StreamMap::default();
     ///
     /// let h = Header::new(b"hello", b"world");
     /// let headers = H3iFrame::Headers(EnrichedHeaders::from(vec![h]));
@@ -178,8 +194,6 @@ impl StreamMap {
     /// use quiche::h3::Header;
     /// use std::iter::FromIterator;
     ///
-    /// let mut stream_map = StreamMap::default();
-    ///
     /// let h = Header::new(b"hello", b"world");
     /// let headers = H3iFrame::Headers(EnrichedHeaders::from(vec![h]));
     ///
@@ -189,7 +203,10 @@ impl StreamMap {
     pub fn received_frame_on_stream(
         &self, stream: u64, frame: &H3iFrame,
     ) -> bool {
-        self.0.get(&stream).map(|v| v.contains(frame)).is_some()
+        self.stream_frame_map
+            .get(&stream)
+            .map(|v| v.contains(frame))
+            .is_some()
     }
 
     /// Check if the stream map is empty, e.g., no frames were received.
@@ -213,7 +230,7 @@ impl StreamMap {
     /// assert!(!stream_map.is_empty());
     /// ```
     pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
+        self.stream_frame_map.is_empty()
     }
 
     /// See all HEADERS received on a given stream.
@@ -226,8 +243,6 @@ impl StreamMap {
     /// use h3i::frame::H3iFrame;
     /// use quiche::h3::Header;
     /// use std::iter::FromIterator;
-    ///
-    /// let mut stream_map = StreamMap::default();
     ///
     /// let h = Header::new(b"hello", b"world");
     /// let enriched = EnrichedHeaders::from(vec![h]);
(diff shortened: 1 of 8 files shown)


h3i/src/client/connection_summary.rs
```diff
@@ -106,21 +115,105 @@ impl StreamMap {
     /// use quiche::h3::Header;
     /// use std::iter::FromIterator;
     ///
+    /// let mut stream_map = StreamMap::default();
```
LPardue, line 118: huh?

evanrittenhouse: Ah missed that in the rebase, thanks. I cleaned up another occurrence as well

h3i/src/client/connection_summary.rs
```diff
@@ -106,21 +115,105 @@ impl StreamMap {
...
+    /// ```
+    /// use h3i::client::connection_summary::StreamMap;
+    /// use h3i::frame::EnrichedHeaders;
+    /// use h3i::frame::H3iFrame;
+    /// use quiche::h3::frame::Frame;
+    /// use quiche::h3::Header;
+    /// use std::iter::FromIterator;
+    ///
+    /// let h = Header::new(b"hello", b"world");
+    /// let correct_headers = H3iFrame::Headers(EnrichedHeaders::from(vec![h]));
```
LPardue, line 143: what does correct mean here?

h3i/src/client/connection_summary.rs
```diff
@@ -106,21 +115,105 @@ impl StreamMap {
...
+        F: Fn(&H3iFrame) -> bool,
+    {
+        self.stream(stream_id)
+            .into_iter()
+            .filter(|frame| f(frame))
+            .collect()
+    }
+
+    /// Collect all frames that match the predicate. The ordering is
+    /// non-deterministic.
```
LPardue, line 174: The non-deterministic bit seems annoying. If we swap the underlying collection for a BTreeMap, they would at least be consistently in increasing stream ID order.   WDYT?

h3i/src/client/connection_summary.rs
```diff
@@ -106,21 +115,105 @@ impl StreamMap {
...
             .flatten()
+            .filter(|frame| f(frame))
             .map(Clone::clone)
-            .collect::<Vec<H3iFrame>>()
+            .collect()
     }
 
-    /// Get all frames on a given `stream_id`.
+    /// Flatten all received frames into a single vector. The ordering is
+    /// non-deterministic.
```
LPardue, line 216: similar comment to above

h3i/src/client/connection_summary.rs
```diff
@@ -213,41 +300,110 @@ impl StreamMap {
...
+    pub(crate) fn new(expected_frames: Option<ExpectedFrames>) -> Self {
+        Self {
+            expected_frames,
+            ..Default::default()
+        }
     }
 
     pub(crate) fn insert(&mut self, stream_id: u64, frame: H3iFrame) {
-        self.0.entry(stream_id).or_default().push(frame);
+        if let Some(expected) = self.expected_frames.as_mut() {
```
LPardue, line 314: I'm not following why this inserts a frame into the expected_frames struct by default

h3i/src/client/connection_summary.rs
```diff
@@ -213,41 +300,110 @@ impl StreamMap {
...
+            expected.receive_frame(stream_id, &frame);
+        }
+
+        self.map.entry(stream_id).or_default().push(frame);
+    }
+
+    pub(crate) fn check_expected_frames(&self) -> bool {
```
LPardue, line 321: Lets document this, since its not immediately obvious what it does. Consdier even returning a enum instead of Bool, since that's harder to misinterpret

h3i/src/client/connection_summary.rs
evanrittenhouse: The idea is that if the stream map sees a frame, and was configured with expected frames, we'll always want to update the expected frame state at the same time that we update the stream map state. If no expected frames were configured, this specific block will be skipped

h3i/src/client/connection_summary.rs
evanrittenhouse: That works. It's also worth noting (and my fault for not being clear in the docs) that the per-stream frame ordering is deterministic - the order in which we iterate over the streams themselves is non-deterministic

h3i/src/client/connection_summary.rs
evanrittenhouse: I renamed it to `saw_expected_frames`, similar to the `ExpectedFrames` naming

h3i/src/client/connection_summary.rs
LPardue: Yeah sorry I misread what receive_frame does, so consider documenting that function but nothing here needs change

h3i/src/frame.rs
```diff
@@ -485,25 +555,118 @@ impl ExpectedFrame {
...
+        assert!(this.is_equivalent(&other));
+
+        this = ExpectedFrame::new(0, other);
+        other =
+            vec![Header::new(b"hello", b"world"), Header::new(b"go", b"jets")]
+                .into();
+        // `other` does not contain the `go: devils` header, so it's not
+        // equivalent to `this.
+        assert!(!this.is_equivalent(&other));
+    }
```
LPardue, line 612: I'm finding it hard to follow this example. Can we just split into 2 tests or does the sequencing matter?

evanrittenhouse: Split it into 2 tests

h3i/src/client/connection_summary.rs
```diff
@@ -246,8 +251,116 @@ impl StreamMap {
...
+            is_app,
+            error_code,
+            reason,
+        }) = self.expected_frames.as_ref().map(|ef| &ef.close_with)
+        else {
+            return;
+        };
+
+        let _ = qconn.close(*is_app, *error_code, reason);
+    }
```
LPardue, line 300: Can this just be rewritten like

```
if let Some(ef) = &self.expected_frames {
  qconn.close(*ef.is_app, *ef.error_code, reason);
}
```

?

h3i/src/client/sync_client.rs
```diff
@@ -57,6 +58,15 @@ struct SyncClient {
     stream_parsers: StreamParserMap,
 }
 
+impl SyncClient {
+    fn new(expected_frames: Option<ExpectedFrames>) -> Self {
```
LPardue, line 62: Should this be part of the `Client` trait or no?

h3i/src/lib.rs
```diff
@@ -112,7 +112,7 @@
 //!    ];
 //!
 //!    let summary =
-//!        sync_client::connect(config, &actions).expect("connection failed");
+//!        sync_client::connect(config, &actions, None).expect("connection failed");
```
LPardue, line 115: Some thing here with `None` not being very helpful to readers.

h3i/src/main.rs
```diff
@@ -298,7 +298,8 @@ fn config_from_clap() -> std::result::Result<Config, String> {
 fn sync_client(
     config: Config, actions: &[Action],
 ) -> Result<ConnectionSummary, ClientError> {
-    h3i::client::sync_client::connect(config.library_config, actions)
+    // TODO: CLI doesn't support passing expected frames at the moment
```
LPardue, line 301: Guessing the qlog input format doesn't either?

LPardue (commented):
Mostly just nits

h3i/src/client/sync_client.rs
evanrittenhouse: Unfortunately not, the async client also implements `Client` but its `new` function returns an `AsyncClient`

evanrittenhouse: Well, we could probably make `Client<T>` but that will be messier without real benefit IMO

LPardue: No biggie

h3i/src/client/connection_summary.rs
```diff
@@ -74,22 +75,32 @@ impl Serialize for ConnectionSummary {
...
         state.end()
     }
 }
 
 /// A read-only aggregation of frames received over a connection, mapped to the
 /// stream ID over which they were received.
 #[derive(Clone, Debug, Default, Serialize)]
-pub struct StreamMap(HashMap<u64, Vec<H3iFrame>>);
+pub struct StreamMap {
+    map: HashMap<u64, Vec<H3iFrame>>,
```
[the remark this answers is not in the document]
evanrittenhouse: Changed to `stream_frame_map`

h3i/src/client/connection_summary.rs
```diff
@@ -74,22 +75,32 @@ impl Serialize for ConnectionSummary {
...
+        state.serialize_field(
+            "missed_expected_frames",
+            &self.stream_map.missing_expected_frames(),
+        )?;
         state.end()
     }
 }
 
 /// A read-only aggregation of frames received over a connection, mapped to the
 /// stream ID over which they were received.
```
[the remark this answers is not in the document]
evanrittenhouse: Added a note about them being there so that we can process them as the new `H3iFrame`s come in

LPardue (approved):
LGTM

PR #1893, opened 2025-01-06
Remove NGINX patch as QUIC is supported since NGINX v1.25.0

dzdidi:
See https://quic.nginx.org/ for guides, docs and interop status


LPardue (approved):

ghedo:
Merged, thanks @dzdidi! Admittedly this hasn't been updated in a while, so makes sense to remove it.

PR #1900, opened 2025-01-13
lib: support idle timeout reconfiguration

LPardue:


quiche/src/lib.rs
```diff
@@ -2198,6 +2198,24 @@ impl Connection {
...
+    /// Sets the `max_idle_timeout` transport parameter, in milliseconds.
+    ///
+    /// This must only be called immediately after creating a connection, that
+    /// is, before any packet is sent or received. Otherwise Error::Done is
+    /// returned.
+    ///
+    /// The default value is infinite, that is, no timeout is used unless
+    /// already configured when creating the connection.
+    pub fn set_max_idle_timeout(&mut self, v: u64) -> Result<()> {
+        if self.is_established() || self.is_in_early_data() {
```
ghedo, line 2210: Even if the connection is not yet established it might be too late to change the TPs. We should really check for whether TPs have already been sent, though I don't think we have that currently... so it might be fine to just not check this at all.

e.g. `set_session()` also doesn't really check if it's being called early enough.

LPardue: This issue is that if we don't check, we can cause the connection to get desynced from what the local peer thinks is an idle timeout and what the remote peer thinks. That seems quite racy and annoying. I'm inclined to add a `sent_tp` tracking variable and update the check here

LPardue: Actually, the details of TP transmission are buried deep in the TLS lib, not quiche itself, so that approach is not easy

ghedo (approved):

PR #1904, opened 2025-01-14
 packet: replace ring AEAD API for retry

ghedo:
The ring AEAD API is currently used for authenticating Retry packets, which, again, doesn't seem like something worth pulling a whole dependency for. This may or may not fall under FIPS scope though.

---

The PR also includes some refactoring commits that are needed for the actual Retry packet change, but they seem to confuse GitHub into generating a larger diff than it actually is in practice as some code was moved with no changes... sorry about that...


quiche/src/packet.rs
```diff
@@ -789,8 +787,9 @@ pub fn verify_retry_integrity(
 
 fn compute_retry_integrity_tag(
     b: &octets::OctetsMut, odcid: &[u8], version: u32,
-) -> Result<aead::Tag> {
+) -> Result<Vec<u8>> {
     const KEY_LEN: usize = RETRY_AEAD_ALG.key_len();
+    const TAG_LEN: usize = RETRY_AEAD_ALG.key_len();
```
LPardue, line 792: Should this be ` RETRY_AEAD_ALG.tag_len();`?

ghedo: Ooops, yeah, fixed now. It just so happen that tag and key len are the same for AES128 GCM...

LPardue (dismissed):

PR #1908, opened 2025-01-14
ffi: make `quiche_conn_path_event_next` return `mut` pointer (fix #1905)

### LINKED ISSUE #1905: ffi: `const quiche_path_event*` expected to be passed to `quiche_path_event*` parameters
The only way to get a `quiche_path_event` is from the return value of `quiche_conn_path_event_next`. It returns a `const quiche_path_event*`. However, all the functions that operate on a `quiche_path_event` take a mutable `quiche_path_event*` in quiche.h, meaning you have to cast away `const` on the pointer to use it anywhere or else you get an error. This seems to be a mistake, because in ffi.rs none of those function parameters are `mut` (with the exception of `quiche_path_event_free`). Perhaps `quiche_conn_path_event_next` should just return a mutable pointer?

ichordev:
Changes `quiche_conn_path_event_next` so that it returns a mutable pointer on both the C and Rust side. This fixes issue #1905.
Also added `const` to the `quiche_path_event*` parameters in the following functions from  `quiche.h` in order to match their mutability in `ffi.rs`:
- `quiche_path_event_type`
- `quiche_path_event_new`
- `quiche_path_event_validated`
- `quiche_path_event_failed_validation`
- `quiche_path_event_closed`
- `quiche_path_event_reused_source_connection_id`
- `quiche_path_event_peer_migrated`

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -797,36 +797,36 @@ enum quiche_path_event_type {
 typedef struct quiche_path_event quiche_path_event;
 
 // Retrieves the next event. Returns NULL if there is no event to process.
-const quiche_path_event *quiche_conn_path_event_next(quiche_conn *conn);
+quiche_path_event *quiche_conn_path_event_next(quiche_conn *conn);
 
 // Returns the type of the event.
-enum quiche_path_event_type quiche_path_event_type(quiche_path_event *ev);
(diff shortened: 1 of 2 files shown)


ghedo (approved):

ghedo:
Merged, thanks @ichordev!

PR #1913, opened 2025-01-17
ffi: add functions to deal with sliced strings (fix #1909)

### LINKED ISSUE #1909: ffi: `const char*` parameters have no specified length (null-terminated)
I come from another language with slices, so it's rather vexing that quiche's FFI doesn't expose string parameters as slices (pointer + length), but rather as null-terminated arrays (pointer + memory leak). This means that before sending a string slice to quiche I have to null-terminate it and pass only its pointer to quiche, only for quiche to find the null-terminator and re-slice it anyway. In other words:

> safe bounds-checked slice ⇒ unsafe null-terminated string ⇒ safe bounds-checked slice

I would prefer if there were alternative functions that take a pointer & length so that I don't have to do this dance. Many C libraries (including portions of the C standard library) also expect `const char*`s to be passed alongside a length, so I don't think it's unreasonable to expect it here.

ichordev:
Adds `*_sliced` variants of the following functions that accept an additional length parameter for strings instead of using null-termination:
- `quiche_version_sliced`
- `quiche_enable_debug_logging_sliced`
- `quiche_config_load_cert_chain_from_pem_file_sliced`
- `quiche_config_load_priv_key_from_pem_file_sliced`
- `quiche_config_load_verify_locations_from_file_sliced`
- `quiche_config_load_verify_locations_from_directory_sliced`
- `quiche_config_set_cc_algorithm_name_sliced`
- `quiche_connect_sliced`
- `quiche_conn_set_keylog_path_sliced`
- `quiche_conn_set_qlog_path_sliced`
- `quiche_conn_set_qlog_fd_sliced`

This fixes issue #1909.
I'm not very experienced with Rust, so please double-check that I didn't do anything wrong in `ffi.rs`. I'm using `std::str::from_utf8_unchecked` to create the `str&` in Rust so that there's no runtime overhead for checking UTF-8 validity.


ichordev:
Is this check failing because of mis-formatting?

ichordev:
Ah, it finally passed! 🥹

PR #1921, opened 2025-01-23
introduce tx_cap factor

ghedo:
Currently we buffer outgoing data up to `min(cwnd, flow_control)`, but this can be lead to starving a connection while waiting for additional data from the application.

This change adds a configuration option to change the send capacity "factor", that is multiplied to the existing send capacity calculation (so e.g. instead of `min(cwnd, flow_control)` once could change that to `2 * min(cwnd, flow_control)`).

Making this configurable rather than a fixed value lets application experiment with different values depending on their needs.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -499,6 +499,9 @@ const DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS: usize = 10;
 // The maximum data offset that can be stored in a crypto stream.
 const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 
+// The send capacity factor.
+const TX_CAP_FACTOR: f64 = 1.0;
+
 /// A specialized [`Result`] type for quiche operations.
 ///
 /// This type is used throughout quiche's public API for any operation that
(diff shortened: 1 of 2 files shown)


jrouviere (approved):

quiche/src/lib.rs
```diff
@@ -495,6 +495,9 @@ const DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS: usize = 10;
 // The maximum data offset that can be stored in a crypto stream.
 const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 
+// The send capacity factor.
+const TX_CAP_FACTOR: usize = 1;
```
antoniovicente, line 499: It is strange that this is an integer.

quiche/src/lib.rs
```diff
@@ -14190,6 +14236,58 @@ mod tests {
             .app_limited());
     }
 
+    #[test]
+    fn tx_cap_factor() {
```
toidiu, line 14240: Is it worth testing the different cc algorithms?

quiche/src/lib.rs
ghedo: Changed to `f64` though TBH I think we'd still want something like 2, but at least now we can experiment with more values.

quiche/src/lib.rs
ghedo: The send capacity is applied independently of CC implementation, so I don't think it would provide much added coverage TBH.

LPardue (approved):

evanrittenhouse (approved):

rushilmehra (approved):

PR #1931, opened 2025-01-31
chore: Derive `default` for a bunch of types

larseggert:
Which either (IMO) have a natural default, or that only consist of `Option`s.

This makes initialization of these types less verbose for the user.

### DIFF
--- a/qlog/src/events/quic.rs
+++ b/qlog/src/events/quic.rs
@@ -37,7 +37,7 @@ use super::Token;
 use crate::HexSlice;
 use crate::StatelessResetToken;
 
-#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Default)]
 #[serde(rename_all = "snake_case")]
 pub enum PacketType {
     Initial,
@@ -51,6 +51,7 @@ pub enum PacketType {
 
     Retry,
     VersionNegotiation,
+    #[default]
     Unknown,
 }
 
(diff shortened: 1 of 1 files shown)


qlog/src/events/quic.rs
```diff
@@ -63,7 +64,7 @@ pub enum PacketNumberSpace {
 }
 
 #[serde_with::skip_serializing_none]
-#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug, Default)]
```
LPardue, line 67: Open question. For your use case, does default obviate the need for `new()`?

https://github.com/cloudflare/quiche/pull/1931/files#diff-6674de9cf1901c352c7eda222350674ca2c177b89a965d0299d7b5bef8449150R88

larseggert:
Don't think so? I mostly want to shorten tings like this:
https://github.com/mozilla/neqo/blob/1090de386550ca69a8b0a1fd530e0bcfe5f79cfd/neqo-transport/src/qlog.rs#L230-L253

LPardue:
I interpret that as yes lol. If you're not using new() or with_type() and the `Default` impl can make more concise code in apps, perhaps we should just remove those 2 functions as well. It doesn't need to be in this PR but could be a followup the does some quiche library internal tidyup too

larseggert:
I think `new` wants all the info, too, and `with_type` doesn't have all the args for info we want.

LPardue (approved):

LPardue:
Merged, thanks.

PR #1932, opened 2025-02-03
ci: add ubuntu-24.04-arm

junhochoi:
Ubuntu ARM runner is available now: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


junhochoi:
It just works ok but we might need to remove some duplicate steps (e.g. does clippy need to run every OS?) to reduce the build time.

PR #1933, opened 2025-02-03
chore: use qlog PacketSent default inside quiche

LPardue:
In followup to https://github.com/cloudflare/quiche/pull/1931, we can cut down on boilerplate internally


ghedo:
AFAICT there are a few more `EventData` cases in `lib.rs`, `h3/mod.rs` and `recovery/mod.rs` that could also be converted, though not quite as satisfying as these changes, do we want to convert those too while at it?

LPardue:
commit count :chart_with_upwards_trend: 

(or more like lets squash these then merge"

PR #1934, opened 2025-02-06
lib: Fix overflow when detecting reserved transport params

hawkinsw:
In cases where a QUIC connection's transport parameter has an ID of less than 27, checking whether that transport parameter is reserved will result in a panic caused by an attempt to store a negative value into a variable with an unsigned type.

Handling transport parameters with ID of less than 27 specially will prevent this from happening.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -8068,8 +8068,12 @@ impl<T> UnknownTransportParameter<T> {
     ///
     /// See Section 18.1 in [RFC9000](https://datatracker.ietf.org/doc/html/rfc9000#name-reserved-transport-paramete).
     pub fn is_reserved(&self) -> bool {
-        let n = (self.id - 27) / 31;
-        self.id == 31 * n + 27
+        if self.id < 27 {
+            false
+        } else {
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -8068,8 +8068,12 @@ impl<T> UnknownTransportParameter<T> {
     ///
     /// See Section 18.1 in [RFC9000](https://datatracker.ietf.org/doc/html/rfc9000#name-reserved-transport-paramete).
     pub fn is_reserved(&self) -> bool {
-        let n = (self.id - 27) / 31;
-        self.id == 31 * n + 27
+        if self.id < 27 {
+            false
+        } else {
+            let n = (self.id - 27) / 31;
```
ghedo, line 8074: A simpler fix would be to have something like:
```suggestion
            let n = self.id.saturating_sub(27) / 31;
```
to avoid the if / else.

ghedo: Might be worth adding explicit test cases for `0` and `27` TPs, just to make sure we cover the edge cases.

PR #1935, opened 2025-02-11
Open source tokio-quiche utils

mendess:


ghedo (changes requested):
Mostly just looked at the general setup, and I don't think these should be in their own workspace.

I suspect that once moved they will need to be reformatted as we use nightly rustfmt for the rest of the repo.

tokio-quiche-utils/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[workspace]
+members = ["./crates/*"]
+resolver = "2"
+
+[workspace.package]
+version = "0.1.0"
+edition = "2021"
+repository = "https://github.com/cloudflare/net-foundations"
```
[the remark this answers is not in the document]
ghedo: I think `description` is also required to publish.

tokio-quiche-utils/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[workspace]
+members = ["./crates/*"]
+resolver = "2"
+
+[workspace.package]
+version = "0.1.0"
```
ghedo, line 6: For other crates in the repo we don't use a single version, to avoid having to publish all crates every time we make a release. So, IMo each crate should have its own version so they can be released independently.

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = [ "apps", "h3i", "octets", "qlog", "quiche" ]
+members = [ "apps", "h3i", "octets", "qlog", "quiche", "./tokio-quiche-utils/buffer-pool", "./tokio-quiche-utils/datagram-socket", "./tokio-quiche-utils/task-killswitch" ]
```
ghedo, line 2: Do these need to be on a subdirectory? Given all other crates are on the root I kind of prefer to move these there as well, particularly given they could be useful outside of tokio-quiche itself.

ghedo (changes requested):
Added one more comment, but otherwise lgtm.

Cargo.toml
mendess: Sure!

ghedo (approved):

PR #1942, opened 2025-02-20
h3i: more workspace dependencies

ghedo:
This bumps serde_with and url to match the versions specified in the workspace.

It also appears that the `serde_with` dependency was entirely unnecessary, as the `serde_as` use in the code never actually worked properly and it triggers a warning with newer versions (https://github.com/jonasbb/serde_with/pull/502) so I just removed it.

---

This depends on https://github.com/cloudflare/quiche/pull/1941.

### DIFF
--- a/h3i/Cargo.toml
+++ b/h3i/Cargo.toml
@@ -25,8 +25,7 @@ octets = { workspace = true }
 qlog = { workspace = true }
 quiche = { workspace = true, features = ["internal", "qlog"] }
 rand = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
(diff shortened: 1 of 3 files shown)


evanrittenhouse (dismissed):

ghedo:
FTR, I had originally misunderstood the issue with `serde_as`... in practice the way we were using it never actually worked, as that use case (tagging enum variants) was never actually supported (https://github.com/jonasbb/serde_with/issues/499#issuecomment-1214394728), and newer versions will warn about it.

In the end I just completely removed the `serde_with` dependency as it wasn't really needed.

LPardue (approved):

PR #1946, opened 2025-02-21
h3i: encode WaitType(Duration) as number

ghedo:
Currently the Duration type is serialized as a struct. This changes it to a simple number of milliseconds.

### DIFF
--- a/h3i/src/actions/h3.rs
+++ b/h3i/src/actions/h3.rs
@@ -39,6 +39,7 @@ use quiche::h3::Header;
 use quiche::ConnectionError;
 use serde::Deserialize;
 use serde::Serialize;
+use serde_with::serde_as;
 
 use crate::encode_header_block;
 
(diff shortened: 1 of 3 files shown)


h3i/src/recordreplay/qlog.rs
```diff
@@ -599,14 +599,13 @@ mod tests {
...
-                Duration::from_millis(0),
+                Duration::from_millis(12345),
             ))
             .unwrap(),
         };
         let serialized = serde_json::to_string(&ev);
 
-        let expected =
-            r#"{"time":123.0,"name":"h3i:wait","data":{"secs":0,"nanos":0}}"#;
+        let expected = r#"{"time":123.0,"name":"h3i:wait","data":12345}"#;
```
LPardue, line 608: In qlog, the `data` field is always a structure. So this should be something like 

`r#"{"time":123.0,"name":"h3i:wait","data":{"duration":12345}}"#;`

i.e. the `WaitDuration` enum variant should either contain a structure, or the ser/der handles tagging it

h3i/src/actions/h3.rs
```diff
@@ -105,11 +106,12 @@ pub enum Action {
 }
 
 /// Configure the wait behavior for a connection.
+#[serde_as]
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum WaitType {
     /// Wait for a time before firing the next action
-    WaitDuration(Duration),
+    WaitDuration(#[serde_as(as = "serde_with::DurationMilliSeconds")] Duration),
```
LPardue, line 114: In order to maintain consistency with other qlog timestamp fields, should this be https://docs.rs/serde_with/latest/serde_with/struct.DurationMilliSecondsWithFrac.html ?

ghedo: idk, you tell me :laughing: I changed it to `DurationMilliSecondsWithFrac<f64>` which incidentally also makes the post-fmt code more readable.

h3i/src/recordreplay/qlog.rs
ghedo: Ok, fixed.

PR #1954, opened 2025-02-24
tokio-quiche: add support for sending additional headers

ghedo:
Since quiche will reject multiple calls to `send_response*()` on the same stream, we need to call `send_additional_headers()` after the initial headers are sent.

### DIFF
--- a/tokio-quiche/src/http3/driver/streams.rs
+++ b/tokio-quiche/src/http3/driver/streams.rs
@@ -50,6 +50,8 @@ pub(crate) struct StreamCtx {
     /// This is used as temporary storage when waiting for `recv`.
     pub(crate) queued_frame: Option<OutboundFrame>,
     pub(crate) audit_stats: Arc<H3AuditStats>,
+    /// Indicates the stream sent initial headers.
+    pub(crate) initial_headers_sent: bool,
     /// Indicates the stream received fin. No more data will be received.
     pub(crate) fin_recv: bool,
     /// Indicates the stream sent fin. No more data will be sent.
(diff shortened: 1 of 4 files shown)


tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -74,6 +76,8 @@ impl StreamCtx {
             queued_frame: None,
             audit_stats: Arc::new(H3AuditStats::new(stream_id)),
 
+            headers_sent: false,
```
LPardue, line 79: The name is sort of bugging me. It's only need to track the whether the very first headers frame was sent or not. But like, I also won't die inside this bikeshed from paint fumes.

LPardue (approved):

tokio-quiche/tests/integration_tests/headers.rs
```diff
@@ -0,0 +1,95 @@
...
+                        let IncomingH3Headers { mut send, .. } = headers;
+
+                        // Send initial headers.
+                        send.send(OutboundFrame::Headers(vec![Header::new(
+                            b":status", b"103",
+                        )]))
+                        .await
+                        .unwrap();
+
+                        tokio::task::yield_now().await;
```
evanrittenhouse, line 62: small nit: may be worth leaving a comment as to why the yield is necessary

evanrittenhouse (approved):

tokio-quiche/tests/integration_tests/headers.rs
ghedo: It's not really necessary TBH and I just copied it from the other test we have, but it seems maybe better to avoid sending the frames back to back, just to make sure it still works with a bit of delay.

tokio-quiche/src/http3/driver/streams.rs
ghedo: Changed to `initial_headers_sent`, but now you need to approve again :P

LPardue (approved):

PR #1957, opened 2025-02-27
Bump rust version in Dockerfile

LPardue:
The Docker builds are failing due to litemap and zerofrom crates
requiring rust 1.81.


rushilmehra (dismissed):

ghedo:
Updated to also bump the MSRV of quiche so it's actually tested in CI.

ghedo (approved):

PR #1960, opened 2025-03-01
Implement MOQT qlog events

LPardue:
Implements https://lpardue.github.io/draft-pardue-moq-qlog-moq-events/#go.draft-pardue-moq-qlog-moq-events.html

### DIFF
--- a/qlog-dancer/src/reports/events.rs
+++ b/qlog-dancer/src/reports/events.rs
@@ -25,6 +25,9 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use qlog::events::http3::Http3Frame;
+use qlog::events::moqt::MOQTByteString;
+use qlog::events::moqt::MOQTControlMessage;
+use qlog::events::moqt::MOQTParameter;
 use qlog::events::quic::AckedRanges;
 use qlog::events::quic::QuicFrame;
 use qlog::events::EventData;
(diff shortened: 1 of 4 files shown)


qlog/src/events/moqt.rs
```diff
@@ -0,0 +1,627 @@
...
+
+#[serde_with::skip_serializing_none]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Default)]
+#[serde(tag = "type")]
+#[serde(rename_all = "snake_case")]
+// Strictly, the qlog spec says that all these control messages have a type
+// field. But instead of making that a rust object property, just use serde to
+// ensure it goes out on the wire. This means that deserialization of control
+// messages also works automatically.
+pub enum MOQTControlMessage {
```
antoniovicente, line 263: Missing doc comments — The project enforces #![warn(missing_docs)]. The new public types in moqt.rs (e.g., MOQTControlMessage, MOQTParameter, MOQTEventType, etc.) lack doc comments, which will trigger warnings in CI if the feature is enabled.

qlog-dancer/src/wirefilter.rs
```diff
@@ -150,6 +279,25 @@ pub fn filter_sqlog_events(mut events: Vec<Event>, filter: &str) -> Vec<Event> {
...
+                    scheme.get_field("moqt.object_id").unwrap(),
+                    moqt_filter_fields.object_ids,
+                )
+                .unwrap();
+
+                ctx.set_field_value(
+                    scheme.get_field("moqt.object_payload_length").unwrap(),
+                    moqt_filter_fields.object_payload_length,
+                )
+                .unwrap();
```
antoniovicente, line 299: Duplicate ctx.set_field_value blocks — The three moqt.* field assignments appear twice in filter_sqlog_events: once in the Event::Qlog branch and once in the Event::Json branch (~24 lines duplicated). This could be refactored — e.g., set all fields unconditionally before the match, since MOQTFilterFields already handles the Json case gracefully (returning empty arrays).

qlog-dancer/src/reports/events.rs
```diff
@@ -235,6 +238,429 @@ fn http_frame_to_string(frame: &Http3Frame) -> String {
...
+            track_namespace_prefix,
+            number_of_parameters,
+            parameters,
+        } => {
+            s += "PUBLISH_NAMESPACE {";
+            s += "track_namespace_prefix=[";
+            for tn in track_namespace_prefix {
+                printyo!("value", &tn.value, s);
+                printyo!("value_bytes", &tn.value_bytes, s);
+            }
```
antoniovicente, line 574: moqt_track_namespace_to_string inlined for prefix/suffix variants — Three message arms (SubscribeNamespace, PublishNamespace, PublishNamespaceCancel, Namespace) inline the same "iterate Vec<MOQTByteString>" loop instead of calling moqt_track_namespace_to_string. The function already exists and should be reused.

antoniovicente (approved):

qlog/src/events/moqt.rs
LPardue: This seems fine for now, all the other qlog events omit public docs. Once the RFC is published, we can consider pub docs that link to a relevant section but doing so now is likely to just cause more churn and maintenance overhead

qlog-dancer/src/wirefilter.rs
LPardue: oh nice, yeah the code here can be simplified, even for the old code (I'll do that in a separate PR)

qlog-dancer/src/reports/events.rs
LPardue: This is not quite right, the `moqt_track_namespace_to_string` function is specialized for printing a track namespace, not a namespace prefix. Blindly using the function would cause a semantic issue. 

The solutin however, is to make the functin more generic so it can handle printing a `&[MOQTByteString]` with an paramaterized name. That's easy enough to do

antoniovicente: Sounds good.  Happy with this going in as is.

PR #1970, opened 2025-03-19
tls: don't log nullbytes from log_ssl_error

dextero:
The function was always logging a 1024 byte long buffer padded with nullbytes. This doesn't look intentional, and may cause trouble for some logger implementations that don't expect nulls in &str.

See also: rust-mobile/android_logger-rs#90


ghedo (approved):

ghedo:
Merged, thanks @dextero!

PR #1972, opened 2025-03-20
Avoid fragmentation when using gso with feature = "gcongestion"

antoniovicente:
Skipping the call to tune_max_send_size will result in small packets when using GSO when attempting to burst more than 64k bytes worth of packets in one go.

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -355,14 +355,13 @@ where
             }
 
             #[cfg(not(feature = "gcongestion"))]
-            let max_send_size = tune_max_send_size(
-                segment_size,
-                qconn.send_quantum(),
-                send_buf.len(),
-            );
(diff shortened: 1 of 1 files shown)


vkrasnov:
That is what GSO_THRESHOLD is for it will not GSO for less than 1000B

antoniovicente:
> That is what GSO_THRESHOLD is for it will not GSO for less than 1000B

My read is that tune_max_send_size will set buffer_is_full to true when the following condition is true:
`send_buf.len() - self.write_state.bytes_written < segment_size`

So we won't generate a packet smaller than the max packet size when writing a series of packets with the max size.

Looking more closely at the code...  I wonder how often the send_quantum is smaller send_buf.len().  It is possible that we can simplify the code by removing the send_quantum argument to tune_max_send_size.

antoniovicente:
Closing for now, may revisit later.

PR #1974, opened 2025-03-25
Switch from feature to cfg flag for tokio-quiche's capture_keylogs

TheJokr:
Whether to capture keylogs (i.e., secrets) should only be decided by the final binary that incorporates tokio-quiche somewhere in its dependency graph. Using a feature allows any dependency to enable keylog capturing, whereas a cfg flag must be explicitly set in RUSTFLAGS or similar.

I've opted to keep the feature flag around for now. We can remove it on the next major bump. This makes the migration a bit easier.


evanrittenhouse (approved):

evanrittenhouse:
cc @ghedo since we spoke about this feature a couple of days ago

PR #1975, opened 2025-03-25
Fix now_or_never() spuriously returning None

gregor-cf:
`now_or_never()` can spuriously return None even if the underlying is Future is ready.  This is due to tokio's cooperative scheduler returning Pending if a task's coop budget is exhausted. We can avoid this by wrapping the future in a `tokio::task::unconstrained()`


ghedo (approved):
Note that we use nightly rustfmt in this repo (hence the "spurious" formatting changes). I reformatted now, and lgtm.

PR #1980, opened 2025-03-28
add mechanism to change connection settings from BoringSSL handshake callbacks

ghedo:
It can be useful to tweak a specific connection's configuration based on information like SNI. However this is only available after the client's ClientHello is fully processed, which is not entirely apparent from an application PoV (e.g. ClientHello can be split across multiple Initial packets).

BoringSSL provides a number of callbacks called during the TLS handshake and sometimes even before any packet has been sent (e.g. `select_cert_cb`) which is useful to tweak congestion control settings or transport parameters,  that provide the required information, but the underlying quiche connection is not exposed to application code from inside these callbacks currently.

This change adds a mechanism to allow applications to change specific connection settings from inside these BoringSSL callbacks.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1237,6 +1237,13 @@ impl Config {
         self.local_transport_params.disable_active_migration = v;
     }
 
+    /// Sets the congestion control algorithm used.
+    ///
+    /// The default value is `CongestionControlAlgorithm::CUBIC`.
+    pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
+        self.cc_algorithm = algo;
+    }
+
     /// Sets the congestion control algorithm used by string.
     ///
     /// The default value is `cubic`. On error `Error::CongestionControl`
@@ -1260,13 +1267,6 @@ impl Config {
     /// The default value is 10.
     pub fn set_initial_congestion_window_packets(&mut self, packets: usize) {
         self.initial_congestion_window_packets = packets;
-    }
-
-    /// Sets the congestion control algorithm used.
-    ///
-    /// The default value is `CongestionControlAlgorithm::CUBIC`.
-    pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
-        self.cc_algorithm = algo;
     }
 
     /// Configures whether to enable HyStart++.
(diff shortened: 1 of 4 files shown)


quiche/src/lib.rs
```diff
@@ -2247,6 +2247,150 @@ impl Connection {
         self.encode_transport_params()
     }
 
+    /// Sets the congestion control algorithm used.
```
antoniovicente, line 2250: I would consider a `set_recovery_config_in_handshake` method that takes a `RecoveryConfig` as a way to minimize future plumbing if additional fields are added to `RecoveryConfig`.  But this works.

quiche/src/lib.rs
```diff
@@ -1237,6 +1237,13 @@ impl Config {
         self.local_transport_params.disable_active_migration = v;
     }
 
+    /// Sets the congestion control algorithm used.
```
antoniovicente, line 1240: Not sure why you moved this method up.  I think that we should prefer set_cc_algorithm_name over set_cc_algorithm

The reason is related to us introducing additional algorithm names as I am prototyping in https://github.com/cloudflare/quiche/commit/d4899cea6c5304703371b8df1308c73ec51203f8.  See also the comment below about `set_recovery_config_in_handshake`

quiche/src/lib.rs
```diff
@@ -15633,32 +15790,80 @@ mod tests {
...
+        let mut pipe = testing::Pipe::with_client_and_server_config(
+            &mut client_config,
+            &mut server_config,
+        )?;
+
+        // Client sends initial flight.
+        let (len, _) = pipe.client.send(&mut buf).unwrap();
+        pipe.server_recv(&mut buf[..len]).unwrap();
+
+        assert_eq!(pipe.server.tx_cap, 36000);
```
antoniovicente, line 15862: Consider changing to:

`assert_eq!(pipe.server.tx_cap, icwnd * 1200);`

and using icwnd in the set_select_certificate_callback.

Also, it is possible to query `pipe.server.tx_cap` prior to `pipe.client.send`?  Wondering if we can `assert_eq!(pipe.server.tx_cap, default_icwnd * 1200);`

antoniovicente (commented):
Looks fantastic.  Thanks for looking into this.

quiche/src/lib.rs
ghedo: The problem is that not all fields in RecoveryConfig can be changed with this method, e.g. `max_ack_delay` is negotiated via transport parameters. And even if that was supported then we'd need to make sure any future config option is also compatible, which seems kind of limiting.

quiche/src/lib.rs
ghedo: I moved it just for grouping the `set_cc_algorithm*` methods together. Given that `_name()` is implemented via converting string to the enum it just kinda made sense to put th enum method first.

As for the gcongestion stuff, why not just add new enum values rather than reuse the existing ones?

quiche/src/lib.rs
ghedo: I changed the check to use a constant for the custom initial window, and also added a check for tx_cap before the update, but note that before receiving the client's initial tx_cap is always 0 (as that's the default value for the flow control window).

PR #1982, opened 2025-03-28
datagram-socket: add min_rtt_us, rtt_var_us, bytes_unsent and delivery_rate to SocketStats

jrouviere:


tokio-quiche/src/quic/connection/mod.rs
```diff
@@ -120,6 +130,12 @@ impl AsSocketStats for QuicConnectionStats {
             bytes_recvd: self.stats.recv_bytes,
             bytes_lost: self.stats.lost_bytes,
             bytes_retrans: self.stats.stream_retrans_bytes,
+            bytes_unsent: 0,
```
antoniovicente, line 133: Is it correct to hard code this to 0?

jrouviere: This was implemented like that in the legacy codebase. I should add a comment.
AFAIK we don't have this value in quiche, but we have it for TCP sockets, so it's worth adding to `SocketStats` for compatibility.

antoniovicente (dismissed):

tokio-quiche/src/quic/connection/mod.rs
antoniovicente: I would add a comment to the struct or code for posterity.

rushilmehra (dismissed):

datagram-socket/src/socket_stats.rs
```diff
@@ -44,6 +44,8 @@ pub trait AsSocketStats {
 pub struct SocketStats {
     pub pmtu: u16,
     pub rtt_us: i64,
+    pub min_rtt_us: i64,
```
antoniovicente, line 47: Do we need to update the package version in order to provide access to these new fields?

jrouviere: I think we generally create a separate release commit like this: https://github.com/cloudflare/quiche/commit/603b77b129816fad8211f189f694aa831b1673fc

PR #1984, opened 2025-04-01
tokio-quiche: remove unconstrained() call

ghedo:
This partially reverts 35130251 as the change was found to cause regressions on some internal deployments.


jrouviere (approved):
LGTM, I verified that fixes the stalling issue in my example

PR #1983, opened 2025-03-31
[gcongestion] Pull in BBR and BBRv2 gcongestion implementations

antoniovicente:
Pull in BBR and BBRv2 implementations from the gcongestion branch in use by mallard.  Also, pull in recent fixes to the gcongestion implementation from the quiche-mallard repo.

This version of the change differs from https://github.com/cloudflare/quiche/pull/1742 in that it attempts to preserve the current congestion control implementations without any functional change, while adding the gcongestion versions to a separate directory.  I decided to only include the BBR and BBRv2 implementations in order to reduce the size of the change; the CUBIC and Reno implementations could be added in a followup change.

Contents of the quiche/src/recovery/gcongestion directory and changes to quiche/src/lib.rs are based on https://github.com/cloudflare/quiche/pull/1742

Changes to tokio-quiche should behave similar to the gcongestion feature when using the gcongestion algorithms

Changes up to 96d3b4fc87dcc50f09a2c1c824e9c18668083ff2 should be no-op refactors to make room for the gcongestion implementations.

PR #1986, opened 2025-04-01
Move the contents of recovery/mod.rs to recovery/congestion/recovery.rs

antoniovicente:
To make room for changes in https://github.com/cloudflare/quiche/pull/1983

No behavior changes expected.


antoniovicente:
I think we should squash these changes when merging.

quiche/src/recovery/mod.rs
```diff
@@ -24,31 +24,22 @@
...
 
 #[cfg(feature = "qlog")]
 use qlog::events::EventData;
 
 use smallvec::SmallVec;
 
-use self::congestion::pacer;
-use self::congestion::Congestion;
-use self::rtt::RttStats;
+mod congestion;
```
ghedo, line 39: Nit: general convention in quiche is to put the `mod`s at the end of the file.

quiche/src/recovery/mod.rs
```diff
@@ -318,39 +74,22 @@ impl LossDetectionTimer {
...
-    /// A resusable list of acks.
-    newly_acked: Vec<Acked>,
+impl std::fmt::Debug for LossDetectionTimer {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self.time {
+            Some(v) => {
+                let now = Instant::now();
+                if v > now {
+                    let d = v.duration_since(now);
+                    write!(f, "timer={d:?} ")
```
ghedo, line 85: Was the idea here to use this as part of `Recovery`'s `Debug` impl? If so, I think the `timer=` prefix can be removed (below as well), so it can be used as `write!("timer={:?}", value)?;` in the other `Debug`.

ghedo (dismissed):
Added a couple of minor comments, but lgtm otherwise.

quiche/src/recovery/mod.rs
antoniovicente: Yes, that was the idea.  The gcongestion implementation depends on Debug for LossDetectionTimer

It makes sense to use it also for congestion after the suggested cleanups.

quiche/src/recovery/mod.rs

quiche/src/recovery/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub tx_in_flight: usize,
+
+    pub lost: u64,
+
+    pub has_data: bool,
+
+    pub pmtud: bool,
+}
+
+impl std::fmt::Debug for Sent {
```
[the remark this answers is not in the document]
toidiu: This doesn't need to be blocking since this was the impl in the original version.

Frame does implement [Debug](https://github.com/cloudflare/quiche/blob/master/quiche/src/frame.rs#L1040) so it should be possible.

However I do see some fields missing from the custom impl (e.g. `ack_eliciting`), which could be intentional or a mistake that the custom Debug impl leads to since you forget to update both places when adding new fields.

toidiu (approved):

quiche/src/recovery/mod.rs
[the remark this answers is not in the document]
antoniovicente: I'll see if we can improve on this in a followup PR.

PR #1987, opened 2025-04-01
Introduce interface to integrate with gcongestion

antoniovicente:
Followup to https://github.com/cloudflare/quiche/pull/1986, introduce trait that will be used to integrate the gcongestion implementation into quiche recovery.

See https://github.com/cloudflare/quiche/pull/1983 for followup changes.

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -29,6 +29,8 @@ use std::time::Duration;
 use std::time::Instant;
 
 use crate::frame;
+use crate::packet;
+use crate::ranges::RangeSet;
 use crate::Config;
 
 #[cfg(feature = "qlog")]
@@ -36,7 +38,7 @@ use qlog::events::EventData;
 
 use smallvec::SmallVec;
 
-pub use self::congestion::recovery::Recovery;
+use self::congestion::recovery::LegacyRecovery;
 
 // Loss Recovery
 const INITIAL_PACKET_THRESHOLD: u64 = 3;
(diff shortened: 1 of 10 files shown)


quiche/src/recovery/mod.rs
```diff
@@ -379,609 +120,159 @@ impl RecoveryConfig {
...
+    Legacy(LegacyRecovery),
+}
 
-    #[cfg(test)]
-    pub fn new(config: &Config) -> Self {
-        Self::new_with_config(&RecoveryConfig::from_config(config))
-    }
+#[enum_dispatch::enum_dispatch]
+/// Api for the Recovery implementation
+pub trait RecoveryApi {
```
ghedo, line 132: `RecoveryOps` maybe?

quiche/src/recovery/mod.rs
```diff
@@ -118,6 +120,96 @@ impl RecoveryConfig {
     }
 }
 
+#[enum_dispatch::enum_dispatch(RecoveryApi)]
```
toidiu, line 123: I was able to compile this without enum_dispatch and wondering if we really need it.

The below replacement code can be annoying to write but I have a few concerns which might not be a big deal.

My concerns:
- I havnt read thru and understood what enum_dispatch is doing. the [docs](https://docs.rs/enum_dispatch/latest/enum_dispatch/) talk about dynamic dispatch, which I dont think we care about (atleast in this PR) and make me less sure why we need it.
- enum_dispatch is `0.3`
- all that goes with additional dependencies
  - need to understand yet another codebase
  - for a library all consumers will need to adopt it also, etc

```
  impl RecoveryApi for Recovery {
      fn lost_count(&self) -> usize {
          match self {
              Recovery::Legacy(legacy_recovery) => legacy_recovery.lost_count(),
          }
      }

    fn bytes_lost(&self) -> u64 {
    ...
```

toidiu:
Besides my concern with `enum_dispatch` this looks good to me.

quiche/src/recovery/mod.rs

toidiu (dismissed):

toidiu (dismissed):

toidiu (approved):

PR #1990, opened 2025-04-03
Add get_next_release_time and on_app_limited methods to RecoveryOps

antoniovicente:
Depends on https://github.com/cloudflare/quiche/pull/1987

PR #1991, opened 2025-04-03
BBRv2 gcongestion implementation, not hooked into the recovery module yet.

antoniovicente:
Depends on https://github.com/cloudflare/quiche/pull/1990


antoniovicente:
See https://github.com/cloudflare/quiche/pull/1993 instead

PR #1993, opened 2025-04-03
[gcongestion] BBRv2 gcongestion implementation, not hooked into the recovery module yet.

antoniovicente:
See https://github.com/cloudflare/quiche/pull/1983 for the full chain of commits.

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -365,7 +365,7 @@ struct QlogMetrics {
     rttvar: Duration,
     cwnd: u64,
     bytes_in_flight: u64,
-    ssthresh: u64,
+    ssthresh: Option<u64>,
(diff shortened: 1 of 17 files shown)


quiche/src/recovery/mod.rs
```diff
@@ -1510,4 +1510,6 @@ mod tests {
 }
 
 mod congestion;
+#[allow(dead_code)]
+mod gcongestion;
```
toidiu, line 1514: Verified that code still compiles when we removing this; making this PR essentially a Noop

toidiu (approved):

antoniovicente:
Added two more commits which should have been part of this PR in the first place.  PTAL at 917f193 and f4e1068

toidiu (approved):

PR #1995, opened 2025-04-03
[gcongestion] Apply fixes to gcongestion implementation from the cloudflare/quiche-mallard repo.

antoniovicente:
After this change most of the functionality from quiche-mallard should be available from mainline quiche.  The exception being the "zero-copy" feature.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4799,7 +4799,7 @@ impl Connection {
             is_app_limited: false,
             tx_in_flight: 0,
             lost: 0,
-            has_data,
+            has_data: has_data | dgram_emitted,
             pmtud: pmtud_probe,
         };
 
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4784,6 +4784,12 @@ impl Connection {
             aead,
         )?;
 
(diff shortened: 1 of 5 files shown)


quiche/src/lib.rs
```diff
@@ -4799,7 +4799,7 @@ impl Connection {
             is_app_limited: false,
             tx_in_flight: 0,
             lost: 0,
-            has_data,
+            has_data: has_data | dgram_emitted,
```
antoniovicente, line 4802: May affect behavior without gcongestion

quiche/src/lib.rs
```diff
@@ -4799,7 +4799,8 @@ impl Connection {
             is_app_limited: false,
             tx_in_flight: 0,
             lost: 0,
-            has_data,
+            has_data: has_data |
+                (dgram_emitted && path.recovery.gcongestion_enabled()),
```
toidiu, line 4803: Can you pull this out into a separate variable to make it easier to read.

```
let has_data = if path.recovery.gcongestion_enabled() {
    dgram_emitted || has_data
} else {
    has_data
};
```

toidiu (approved):

PR #1998, opened 2025-04-04
Add support for literal headers for h3i

josja-cf:
This PR adds support for literal headers, such as header names with upper-case letters, to h3i.
It does not affect the existing functions, instead introducing a new one (`send_headers_frame_literal`) that can be used when testing specifically for these invalid headers.

I made sure that the internal quiche functions and constants that needed to be exposed to h3i are hidden behind the `internal` feature.

### DIFF
--- a/h3i/src/actions/h3.rs
+++ b/h3i/src/actions/h3.rs
@@ -42,6 +42,7 @@ use serde::Serialize;
 use serde_with::serde_as;
 
 use crate::encode_header_block;
+use crate::encode_header_block_literal;
 
 /// An action which the HTTP/3 client should take.
 ///
@@ -62,6 +63,7 @@ pub enum Action {
     SendHeadersFrame {
         stream_id: u64,
         fin_stream: bool,
+        literal_headers: bool,
         headers: Vec<Header>,
         frame: Frame,
     },
@@ -201,6 +203,25 @@ pub fn send_headers_frame(
         stream_id,
         fin_stream,
         headers,
+        literal_headers: false,
+        frame: Frame::Headers { header_block },
+    }
+}
+
+/// Convenience to convert between header-related data and a
+/// [Action::SendHeadersFrame]. Unlike [`send_headers_frame`],
+/// this version encodes the headers literally as they are provided,
+/// not converting the header names to lower-case.
+pub fn send_headers_frame_literal(
+    stream_id: u64, fin_stream: bool, headers: Vec<Header>,
+) -> Action {
+    let header_block = encode_header_block_literal(&headers).unwrap();
+
+    Action::SendHeadersFrame {
+        stream_id,
+        fin_stream,
+        headers,
+        literal_headers: true,
         frame: Frame::Headers { header_block },
     }
 }
--- a/h3i/src/recordreplay/qlog.rs
+++ b/h3i/src/recordreplay/qlog.rs
@@ -50,6 +50,7 @@ use smallvec::smallvec;
 use crate::actions::h3::Action;
 use crate::actions::h3::WaitType;
 use crate::encode_header_block;
+use crate::encode_header_block_literal;
 use crate::fake_packet_sent;
 use crate::HTTP3_CONTROL_STREAM_TYPE_ID;
 use crate::HTTP3_PUSH_STREAM_TYPE_ID;
@@ -108,6 +109,7 @@ impl From<&Action> for QlogEvents {
                 stream_id,
                 fin_stream,
                 headers,
+                literal_headers,
                 ..
             } => {
                 let qlog_headers = headers
@@ -132,6 +134,10 @@ impl From<&Action> for QlogEvents {
 
                 if *fin_stream {
                     ex.insert("fin_stream".to_string(), json!(true));
+                }
+
+                if *literal_headers {
+                    ex.insert("literal_headers".to_string(), json!(true));
                 }
 
                 vec![QlogEvent::Event {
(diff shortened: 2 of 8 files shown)


LPardue:
This looks great.

It would be nice to add this as another option to the CLI too. We already have a `headers` and `headers_no_pseudo` option, see https://github.com/cloudflare/quiche/blob/cb93c9991041698cd3a14a040e335583df361604/h3i/src/prompts/h3/mod.rs#L141

How about we add `headers_literal` and plumb the option to use your new functions if selected?

josja-cf:
I've added the actions (one for with and one for without pseudo headers). 
One thing that still needs to be done is properly replaying these from the qlog. Currently, when the qlog is read, all the headers are encoded normally (not literally). I tried to see if this could be changed to literal, but the headers are not converted to lowercase before being saved in the qlog. That means qlogs generated with non-literal headers might still have capital letters in the header names, which would then be parsed differently. 

Any suggestions on how to approach this? 
I am not very familiar with the qlog format, so not sure what the best way to mark these headers as literal is.

LPardue:
So the core qlog specs can't help us here but we have prior art for encoding additional information that is useful for h3i. We embed the `fin_stream`  prompt option in `Action::SendHeadersFrame` (https://github.com/cloudflare/quiche/blob/828f030312ace4219345b0d108909c08e66b8d51/h3i/src/prompts/h3/headers.rs#L80-L87) and later serialize that into "extra data" at https://github.com/cloudflare/quiche/blob/828f030312ace4219345b0d108909c08e66b8d51/h3i/src/recordreplay/qlog.rs#L98

The seralized qlog event is then

```
{"time":0.099764995,"name":"http:frame_created","data":{"stream_id":0,"frame":{"frame_type":"headers","headers":[{"name":":method","value":"GET"},{"name":":path","value":"/delay/1"},{"name":":scheme","value":"https"},]}},"fin_stream":true}
```

We round-trip this back in while parsing here: https://github.com/cloudflare/quiche/blob/master/h3i/src/recordreplay/qlog.rs#L448

My suggestion is to keep your current code for encoding in `send_headers_frame_literal()`. In addition add a `literal_headers` bool field to `Action::SendHeadersFrame` and set that to true in this case. Then serialize that into extra data . And finally, when parsing the code use the extra data to decide the encoding here: https://github.com/cloudflare/quiche/blob/master/h3i/src/recordreplay/qlog.rs#L507

josja-cf:
Thank you for the input, that worked well. This is a breaking change in the API (the `Action::SendHeadersFrame` has an additional field), so we do need to account for that.

I've implemented your suggestion and tested it locally + with one added unit test.

h3i/src/actions/h3.rs
```diff
@@ -201,6 +203,25 @@ pub fn send_headers_frame(
...
+        literal_headers: false,
         frame: Frame::Headers { header_block },
     }
 }
+
+/// Convenience to convert between header-related data and a
+/// [Action::SendHeadersFrame]. Unlike [`send_headers_frame`],
+/// this version encodes the headers literally as they are provided,
+/// not converting the header names to lower-case.
+pub fn send_headers_frame_literal(
```
evanrittenhouse, line 215: Can we add a comment to `send_headers_frame` that any HEADERS sent with that function will be lowercased due to QPACK?

h3i/src/recordreplay/qlog.rs
```diff
@@ -108,8 +109,11 @@ impl From<&Action> for QlogEvents {
                 stream_id,
                 fin_stream,
                 headers,
+                literal_headers,
                 ..
             } => {
+
+                dbg!(&headers);
```
evanrittenhouse, line 116: Is this necessary?

evanrittenhouse (commented):
Overall LGTM, left a couple of comments. 

As far as the breaking change goes, I think it should be fine to add here without breaking any internal users. From an internal scan, I don't think there are any users that use the action directly - they all use `send_headers_frame()` so as to avoid needing to encode a header block directly.

quiche/src/h3/qpack/encoder.rs
```diff
@@ -117,7 +117,7 @@ fn lookup_static<T: NameValue>(h: &T) -> Option<(u64, bool)> {
     None
 }
 
-fn encode_int(
```
LPardue, line 120: should making these pub be behind the `internal` config too?

h3i/src/recordreplay/qlog.rs
josja-cf: It is not, removed.

h3i/src/actions/h3.rs
josja-cf: done

quiche/src/h3/qpack/encoder.rs
josja-cf: This should be hidden behind `internal` already. `qpack` is only `pub use` when the `internal` feature is enabled ([see here](https://github.com/cloudflare/quiche/pull/1998/files#diff-851097865419440f06ef5f0c6fcbca011863b552fe556fdcded5c368fdeb9715R7194)).
It is not possible to put the `pub` keyword itself behind a feature (according to my googling), so this was the next best option.

LPardue (dismissed):
LGTM

There's a few minor build failures and it needs a rebase but then its good to go

quiche/src/h3/mod.rs
```diff
@@ -7191,4 +7191,6 @@ pub mod frame;
 mod frame;
 #[doc(hidden)]
 pub mod qpack;
+#[cfg(feature = "internal")]
+pub use qpack::LITERAL;
```
ghedo, line 7195: Why is this needed? AFAICT we already expose all `qpack::constants::*` as well, so this seems superfluous.

josja-cf: I did that because `internal != hidden`, so I assumed that semver restrictions still held for that module. 
Those constants were not public before, and I didn't want to make anything `pub` that was not already.
If we are good to make all the constants part of the public API, we can get rid of this.

josja-cf: Done

PR #1999, opened 2025-04-04
bbr: account for bytes buffered at lower layers

ghedo:
For example, due to pacing offload some packets might be buffered in the fq qdisc queue and not yet sent on the network.

This is more or less meant to be equivalent to Linux' `bbr_packets_in_net_at_edt()` function.

---

This is an alternative to https://github.com/cloudflare/quiche/pull/1636 and instead of walking the list of sent packets which is quite expensive, we just sort of estimate the amount of bytes by looking at BW and pacing rate.

### DIFF
--- a/quiche/src/recovery/congestion/bbr/per_ack.rs
+++ b/quiche/src/recovery/congestion/bbr/per_ack.rs
@@ -258,11 +258,35 @@ fn bbr_check_drain(r: &mut Congestion, bytes_in_flight: usize, now: Instant) {
     }
 
     if r.bbr_state.state == BBRStateMachine::Drain &&
-        bytes_in_flight <= bbr_inflight(r, 1.0)
+        bbr_bytes_in_net(r, bytes_in_flight, now) <= bbr_inflight(r, 1.0)
     {
         // we estimate queue is drained
         bbr_enter_probe_bw(r, now);
     }
+}
+
+/// Estimates the number of bytes "in the network" based on pacing.
+///
+/// When pacing is implemented at lower layers (e.g. when TX_TIME and the sch_fq
+/// qdisc are used), the in-flight data can be higher than the amount data
+/// actually sent on the network.
+///
+/// This is largely based on the [`bbr_packets_in_net_at_edt()`] function from
+/// Linux' BBR implementation.
+///
+/// `bbr_packets_in_net_at_edt()`: https://elixir.bootlin.com/linux/v6.13.7/source/net/ipv4/tcp_bbr.c#L437
+fn bbr_bytes_in_net(r: &Congestion, in_flight: usize, now: Instant) -> usize {
+    let edt = r.pacer.next_time().max(now);
+    let interval = edt.saturating_duration_since(now);
+    let interval_delivered = interval.as_secs_f64() * r.bbr_state.btlbw as f64;
+
+    let mut in_flight_at_edt = in_flight;
+
+    if r.bbr_state.pacing_gain > 1.0 {
+        in_flight_at_edt += r.send_quantum;
+    }
+
+    in_flight_at_edt.saturating_sub(interval_delivered as usize)
 }
 
 // 4.3.4.3.  Gain Cycling Algorithm
@@ -305,9 +329,11 @@ fn bbr_is_next_cycle_phase(r: &mut Congestion, now: Instant) -> bool {
     let bbr = &mut r.bbr_state;
     let lost_bytes = bbr.newly_lost_bytes;
     let pacing_gain = bbr.pacing_gain;
-    let prior_in_flight = bbr.prior_bytes_in_flight;
 
     let is_full_length = (now - bbr.cycle_stamp) > bbr.rtprop;
+
+    let prior_in_flight = bbr.prior_bytes_in_flight;
+    let prior_in_flight = bbr_bytes_in_net(r, prior_in_flight, now);
 
     // pacing_gain == 1.0
     if (pacing_gain - 1.0).abs() < f64::EPSILON {
--- a/quiche/src/recovery/congestion/bbr/mod.rs
+++ b/quiche/src/recovery/congestion/bbr/mod.rs
@@ -537,8 +537,7 @@ mod tests {
             );
         }
 
-        // Stop at right before filled_pipe=true.
-        for _ in 0..5 {
+        for _ in 0..7 {
             let pkt = Sent {
                 pkt_num: pn,
                 frames: smallvec![],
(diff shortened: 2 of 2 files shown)


quiche/src/recovery/congestion/bbr/per_ack.rs
```diff
@@ -258,13 +258,28 @@ fn bbr_check_drain(r: &mut Congestion, bytes_in_flight: usize, now: Instant) {
...
     if r.bbr_state.state == BBRStateMachine::Drain &&
-        bytes_in_flight <= bbr_inflight(r, 1.0)
+        bbr_bytes_in_net(r, bytes_in_flight, now) <= bbr_inflight(r, 1.0)
     {
         // we estimate queue is drained
         bbr_enter_probe_bw(r, now);
     }
 }
 
+fn bbr_bytes_in_net(r: &Congestion, in_flight: usize, now: Instant) -> usize {
```
toidiu, line 268: Could you link to the [source](https://elixir.bootlin.com/linux/v6.13.7/source/net/ipv4/tcp_bbr.c#L437) and add a comment explaining the context

quiche/src/recovery/congestion/bbr/per_ack.rs
```diff
@@ -258,13 +258,28 @@ fn bbr_check_drain(r: &mut Congestion, bytes_in_flight: usize, now: Instant) {
...
+    let edt = r.pacer.next_time().max(now);
+    let interval = edt.saturating_duration_since(now);
+    let interval_delivered = interval.as_secs_f64() * r.bbr_state.btlbw as f64;
+
+    let mut in_flight_at_edt = in_flight;
+
+    // pacing_gain > 1.0
+    if (r.bbr_state.pacing_gain - 1.0).abs() > f64::EPSILON {
+        in_flight_at_edt += r.send_quantum;
+    }
```
toidiu, line 278: - Are we missing these [lines](https://elixir.bootlin.com/linux/v6.13.7/source/net/ipv4/tcp_bbr.c#L451-L452) and do we need them?  `if (interval_delivered >= inflight_at_edt) return 0;`

If I am understanding correctly,  `r.send_quantum` is the approximation of what we approximate is buffered. Thats more than I would have expected.

quiche/src/recovery/congestion/bbr/per_ack.rs
```diff
@@ -258,13 +258,28 @@ fn bbr_check_drain(r: &mut Congestion, bytes_in_flight: usize, now: Instant) {
...
     {
         // we estimate queue is drained
         bbr_enter_probe_bw(r, now);
     }
 }
 
+fn bbr_bytes_in_net(r: &Congestion, in_flight: usize, now: Instant) -> usize {
+    let edt = r.pacer.next_time().max(now);
+    let interval = edt.saturating_duration_since(now);
+    let interval_delivered = interval.as_secs_f64() * r.bbr_state.btlbw as f64;
```
toidiu, line 271: I cant tell but is `r.bbr_state.btlbw ` equivalent to `bbr_bw`. If its different should we call out the difference in a comment?

antoniovicente: I think that r.bbr_state.btlbw is equivalent to [lt_bw](https://elixir.bootlin.com/linux/v6.13.7/C/ident/lt_bw) in the kernel code, so bbr_bw would match btlbw when [lt_use_bw](https://elixir.bootlin.com/linux/v6.13.7/C/ident/lt_use_bw) is true.

I assume that we have a valid btlbw at this point, but it wouldn't hurt to double check (btlbw starts as 0)

quiche/src/recovery/congestion/bbr/per_ack.rs
antoniovicente: The `saturating_sub` in the next line is equivalent to the `if (interval_delivered >= inflight_at_edt) return 0;` expression.

I think that the `send_quantum` computation in the previous lines corresponds to https://elixir.bootlin.com/linux/v6.13.7/source/net/ipv4/tcp_bbr.c#L449-L450

quiche/src/recovery/congestion/bbr/per_ack.rs
```diff
@@ -258,13 +258,28 @@ fn bbr_check_drain(r: &mut Congestion, bytes_in_flight: usize, now: Instant) {
...
+
+    let mut in_flight_at_edt = in_flight;
+
+    // pacing_gain > 1.0
+    if (r.bbr_state.pacing_gain - 1.0).abs() > f64::EPSILON {
```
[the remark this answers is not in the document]
antoniovicente: See existing impl of `pacing_gain == 1.0`

https://github.com/cloudflare/quiche/blob/c8cabed1220e0390abc1d7cdac3ae5cd7fb59e24/quiche/src/recovery/congestion/bbr/per_ack.rs#L312

quiche/src/recovery/congestion/bbr/per_ack.rs
ghedo: Yeah, good point. Added a comment.

quiche/src/recovery/congestion/bbr/per_ack.rs
```diff
@@ -305,10 +320,12 @@ fn bbr_is_next_cycle_phase(r: &mut Congestion, now: Instant) -> bool {
     let bbr = &mut r.bbr_state;
     let lost_bytes = bbr.newly_lost_bytes;
     let pacing_gain = bbr.pacing_gain;
-    let prior_in_flight = bbr.prior_bytes_in_flight;
 
     let is_full_length = (now - bbr.cycle_stamp) > bbr.rtprop;
 
+    let prior_in_flight = bbr.prior_bytes_in_flight;
+    let prior_in_flight = bbr_bytes_in_net(r, prior_in_flight, now);
```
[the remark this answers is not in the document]
ghedo: The previous implementation was quite expensive which IIRC is why it wasn't used here, so technically that PR was wrong.

quiche/src/recovery/congestion/bbr/per_ack.rs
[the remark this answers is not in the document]
antoniovicente: Comparing against `> epsilon` or `> 1.0 + epsilon` is more correct, but it I don't know how much error we can end up accumulating here.  Is gain of 1.0 something that can only happen when we assign 1.0, or can it also happen after `(pacing_gain * N)` followed by `* (1/N)` or `pacing_gain + N` followed by `- N`?

antoniovicente (approved):

quiche/src/recovery/congestion/bbr/per_ack.rs
toidiu: From what I can tell `btlbw` is the closest thing to `bw` in [BBR1 state](https://github.com/cloudflare/quiche/blob/951e1fce3267c62177a5f308f0cc1f9b286d6600/quiche/src/recovery/congestion/bbr/mod.rs#L103).

The RFC only seems to mention [BBR.BtlBw](https://datatracker.ietf.org/doc/html/draft-cardwell-iccrg-bbr-congestion-control-00#section-4.1.1).

toidiu (approved):

PR #2002, opened 2025-04-06
Add bounds check in un-safe code in octets macros

gregor-cf:
The `put_u` and `peek_u` macros in `octets` contains unsafe blocks. This commit adds a safety check to make sure the length we are being passed is within the proper bounds. Without this check it's possible to access invalid memory.

Note that the macro is not exported and all current uses of it are safe. So this commits just prevents future issues.

### DIFF
--- a/octets/Cargo.toml
+++ b/octets/Cargo.toml
@@ -8,3 +8,8 @@ repository = { workspace = true }
 license = { workspace = true }
 keywords = { workspace = true }
 categories = { workspace = true }
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(test_invalid_len_compilation_fail)',
+] }
--- a/octets/src/lib.rs
+++ b/octets/src/lib.rs
@@ -52,6 +52,14 @@ impl std::error::Error for BufferTooShortError {
     }
 }
 
+/// Helper macro that asserts at compile time. It requires that
+/// `cond` is a const expression.
+macro_rules! static_assert {
+    ($cond:expr) => {{
+        const _: () = assert!($cond);
+    }};
+}
+
 macro_rules! peek_u {
     ($b:expr, $ty:ty, $len:expr) => {{
         let len = $len;
@@ -61,6 +69,7 @@ macro_rules! peek_u {
             return Err(BufferTooShortError);
         }
 
+        static_assert!($len <= mem::size_of::<$ty>());
         let mut out: $ty = 0;
         unsafe {
             let dst = &mut out as *mut $ty as *mut u8;
@@ -95,6 +104,7 @@ macro_rules! put_u {
 
         let dst = &mut $b.buf[$b.off..($b.off + len)];
 
+        static_assert!($len <= mem::size_of::<$ty>());
         unsafe {
             let src = &<$ty>::to_be(v) as *const $ty as *const u8;
             let off = (mem::size_of::<$ty>() - len) as isize;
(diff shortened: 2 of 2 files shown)


gregor-cf:
> It it possible to make this a compile-time check?

Probably yes. Let me try. (Should be able to either use the `static_assert` crate or just implement `static_assert` directly).

gregor-cf:
Asserting a compile-time now:
```
$ cargo rustc  -- --cfg test_invalid_len_compilation_fail 
   Compiling octets v0.3.0 (/home/gregor/r/tq-perf/quiche/octets)
error[E0080]: evaluation of constant value failed
   --> octets/src/lib.rs:59:23
    |
59  |         const _: () = assert!($cond);
    |                       ^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: 2 <= mem::size_of::<u8>()', octets/src/lib.rs:713:9
...
713 |         peek_u!(b, u8, 2)
    |         ----------------- in this macro invocation
    |
    = note: this error originates in the macro `assert` which comes from the expansion of the macro `peek_u` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0080]: evaluation of constant value failed
   --> octets/src/lib.rs:59:23
    |
59  |         const _: () = assert!($cond);
    |                       ^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: 2 <= mem::size_of::<u8>()', octets/src/lib.rs:719:9
...
719 |         put_u!(b, u8, v, 2)
    |         ------------------- in this macro invocation
    |
    = note: this error originates in the macro `assert` which comes from the expansion of the macro `put_u` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0080`.
error: could not compile `octets` (lib) due to 2 previous errors
```

ghedo:
This looks good, but somewhat unfortunate that the tests don't run all the time. I wonder if they could be turned into doc tests on `peek_u`/`put_u` with the `compile_fail` attribute instead? https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html?highlight=compile_fail#attributes

gregor-cf:
> This looks good, but somewhat unfortunate that the tests don't run all the time. I wonder if they could be turned into doc tests on `peek_u`/`put_u` with the `compile_fail` attribute instead? https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html?highlight=compile_fail#attributes

Yes and no. I could use a doc-test, but doc-tests are outside the crate so I would then need to make the macros public with `#[macro_export]`.

octets/Cargo.toml
```diff
@@ -8,3 +8,8 @@ repository = { workspace = true }
 license = { workspace = true }
 keywords = { workspace = true }
 categories = { workspace = true }
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
```
evanrittenhouse, line 13: nit: should we make this an `error` level warning?

gregor-cf: That's actually unrelated to the error the compiler produces if the invariant is violated. 

For some reason that I don't fully understand (and didn't feel like investigating)  my "test" that's gated by `#[cfg(test_invalid_len_compilation_fail)]`  produces a lint warning about unexpected and I just copy-pasted the suggested remediation.... Let me add a comment here to clarify what this does. 
Screenshot 2025-04-09 at 10 59 21

ghedo (approved):

PR #2003, opened 2025-04-06
tokio-quiche: Handle zero-length HTTP bodies without FIN

gregor-cf:
Change the return value of h3::Connection::send_body() to return Ok(0) if a zero-length body *without* FIN is sent (from Err(Done)).

Two reasons why:
* It eliminiates a possible infinte loop in tokio-quiche. If TQ gets a Err(Done) it will retry the send, which will never succeed. But if it gets an Ok(n) TQ will move on if `n == body.len()`.
* It brings the behavior in line with what send_stream() does, which also returns Ok(0) if a zero-length write w/o fin is attempted.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -650,6 +650,11 @@ impl<H: DriverHooks> H3Driver<H> {
 
             OutboundFrame::Body(body, fin) => {
                 let len = body.as_ref().len();
+                if len == 0 && !*fin {
+                    // quiche doesn't allow sending an empty body when the fin
+                    // flag is not set
+                    return Ok(());
+                }
                 if *fin {
                     // If this is the last body frame, close the receiver in the
                     // stream map to signal that we shouldn't


gregor-cf:
> This sounds more like a bug in TQ than anything. TQ should deal with the result of send_body rather than bubble it up.
> 
> QUIC and HTTP/3 are different - you can send a 0-length QUIC STREAM frame without consuming flow control but HTTP frames always consume flow control.
> 
> I'm hesitant to make any changes here due to the impact that this might have on apps' control loop that expect the current behaviour.

Fair enough. I fully understand not wanting to change the current behavior.
That being said: I don't think applications would rely on that behavior today anyways. And this change will make applications more resilient if they try to send an empty body frame by accident. 
Today, an application either never tries to send a zero-length body without `fin` (by checking before calling `send_body()`) or they would do the same as TQ and retry any `Err(Done)` and have an infinite loop. 

If an application would rely on the current behavior, they'd have to do something like the following which wouldn't make a lot of sense I think:
```
match conn.send_body(..., &buf, fin) {
     Ok(n) => { ... },
     Err(Done) => { 
          if buf.is_empty() && !fin {
               // nevermind. Should have done this check earlier and not even tried to send this
          } else {
               // stream is blocked. Retry later.
               return Err(Done);
          }
     },
}
```

All that said: I'm more than happy to instead add a check to TQ (probably a good idea in any case). 
I think this PR is still useful, but if you disagree I'm happy close it

LPardue:
Sending a 0-length DATA frame without fin doesn't make much logical sense to me. Its unfortunate we overload using Error::Done for different conditions. Perhaps the clearer thing to do it return a different error code to make it clear this is something that 0-length DATA without fin should not be retried

gregor-cf:
That makes sense. I've update the PR and 
a) changed TQ to not send empty bodies w/o fin 
b) changed the error that `send_body()` return to `Error::FrameUnexpected`. I thought that was the best "fit". Let me know what you think

rushilmehra (approved):

LPardue (changes requested):
I think the quiche lib change needs some more consideration before it's ready to merge.

Perhaps we can do the TQ change separate since that seems like a net improvement no matter what

rushilmehra (approved):

gregor-cf:
> I think the quiche lib change needs some more consideration before it's ready to merge.
> 
> Perhaps we can do the TQ change separate since that seems like a net improvement no matter what

Good point. Update this PR to only include TQ changes.

gregor-cf:
> I think the quiche lib change needs some more consideration before it's ready to merge.

For the quiche changes: Let me know if you think Error::UnexpectedFrame would make sense or a different error variant. Or maybe we'll just leave it as-is (Error::Done) and update the doc for `send_body()` to document the behavior

evanrittenhouse (approved):

LPardue (approved):

PR #2012, opened 2025-04-10
Parameterize recovery module tests so they run against all congestion control algorithms.

antoniovicente:
As a pre-requisite, add Instant::now as an argument to get_packet_send_time so it can be driven by the test.

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -556,6 +556,8 @@ impl ReleaseDecision {
 
 #[cfg(test)]
 mod tests {
+    use rstest::rstest;
+
     use super::*;
     use crate::packet;
     use crate::ranges;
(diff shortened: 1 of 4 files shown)


quiche/src/recovery/mod.rs
```diff
@@ -1301,7 +1324,7 @@ mod tests {
             is_app_limited: false,
             tx_in_flight: 0,
             lost: 0,
-            has_data: false,
+            has_data: true,
```
antoniovicente, line 1327: bbr2_gcongestion ignores sends with has_data set to false.

quiche/src/recovery/mod.rs
```diff
@@ -1431,9 +1452,16 @@ mod tests {
...
+            // For bbr2_gcongestion, send time is almost 13000 / pacing_rate.
+            // Don't know where 13000 comes from.
+            1.08333332
+        } else {
+            1.0
+        };
         assert_eq!(
             r.get_packet_send_time(now) - now,
-            Duration::from_secs_f64(12000.0 / pacing_rate as f64)
+            Duration::from_secs_f64(scale_factor * 12000.0 / pacing_rate as f64)
```
antoniovicente, line 1464: The pacing rate for bbr2_gcongestion may be too conservative; it is sending 2.5x slower than all the other congestion control algorithms

ghedo (approved):

PR #2018, opened 2025-04-11
tokio-quiche: expose quiche settings

evanrittenhouse:
I opted to not change to a builder pattern for API consistency's sake. If we include one on `QuicSettings`, we should also include one on `Http3Settings`, which would significantly increase the scope of this PR.

If we really want to do that, I'm fine doing it as a separate commit/follow-up


antoniovicente (approved):

antoniovicente:
Looks great.  Thanks for adding these params.

PR #2022, opened 2025-04-11
Drop dependency on quiche-mallard

antoniovicente:
Quiche has all the required functionality, so mallard can switch over to mainline quiche.


ghedo:
Rebased, I also changed the quiche dependency in tokio-quiche to use the workspace version again.

PR #2023, opened 2025-04-15
refactor: pass bbr2 parameters to components

toidiu:
Stop accessing global params struct from the gcongestion implementation so we can override these params per connection

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -52,7 +52,7 @@ use super::RttStats;
 
 const MAX_MODE_CHANGES_PER_CONGESTION_EVENT: usize = 4;
 
-#[derive(PartialEq)]
+#[derive(Debug, PartialEq)]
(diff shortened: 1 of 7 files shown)


quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -168,7 +169,7 @@ struct Params {
     bw_lo_mode: BwLoMode,
 }
 
-const PARAMS: Params = Params {
+const DEFAULT_PARAMS: Params = Params {
```
antoniovicente, line 172: Any way to make these params private?  It seems like PARAMS was accessed within this crate before your refactor.

antoniovicente (approved):
Looks good.

quiche/src/recovery/gcongestion/bbr2.rs
toidiu: They are already private to the bbr2 module. Passing a single instance is a good practice we have to enforce.

PR #2024, opened 2025-04-15
[gcongestion] implement initial_congestion_window_packets config for bbr2_gcongestion algorithm

antoniovicente:


antoniovicente:
I see some 32-bit test failures that are already fixed in https://github.com/cloudflare/quiche/pull/2013

So this depends on that other PR going in first.

PR #2026, opened 2025-04-15
quiche: release 0.24.0

evanrittenhouse:


antoniovicente (approved):
Looks good, but I think you need to push via CLI to avoid changing hash.

PR #2030, opened 2025-04-21
Open source async h3i 

evanrittenhouse:
Tests for internal services pass when using this commit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ smallvec = { version = "1.10", default-features = false }
 task-killswitch = { version = "0.1.0", path = "./task-killswitch" }
 thiserror = { version = "1" }
 tokio = { version = "1.44", default-features = false }
+tokio-quiche = { version = "0.5.0", path = "./tokio-quiche" }
 tokio-stream = { version = "0.1" }
 tokio-util = { version = "0.7.13" }
 triomphe = { version = "0.1" }
--- a/h3i/src/client/async_client.rs
+++ b/h3i/src/client/async_client.rs
@@ -29,6 +29,7 @@
 
 use buffer_pool::ConsumeBuffer;
 use buffer_pool::Pooled;
+use log;
 use quiche::PathStats;
 use quiche::Stats;
 use std::future::Future;
@@ -41,14 +42,10 @@ use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 use tokio::time::sleep_until;
 use tokio::time::Instant;
-
-use log;
-
 use tokio_quiche::buf_factory::BufFactory;
 use tokio_quiche::metrics::Metrics;
 use tokio_quiche::quic::HandshakeInfo;
 use tokio_quiche::quic::QuicheConnection;
-use tokio_quiche::quiche;
 use tokio_quiche::settings::Hooks;
 use tokio_quiche::settings::QuicSettings;
 use tokio_quiche::socket::Socket;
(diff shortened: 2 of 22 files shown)


Cargo.toml
```diff
@@ -63,6 +63,7 @@ smallvec = { version = "1.10", default-features = false }
 task-killswitch = { version = "0.1.0", path = "./task-killswitch" }
 thiserror = { version = "1" }
 tokio = { version = "1.44", default-features = false }
+tokio-quiche = { version = "0.5.0", path = "./tokio-quiche" }
```
LPardue, line 66: Does this need to be a workspace dep?

h3i/src/client/mod.rs
```diff
@@ -161,7 +84,7 @@ fn handle_qlog(
 }
 
 #[derive(Debug, Serialize)]
-/// Represents different errors that can occur when [sync_client] runs.
+/// Represents different errors that can occur when h3i runs.
```
LPardue: `h3i` sounds too generic, lets be clear this is a client role

h3i/src/client/sync_client.rs
```diff
@@ -87,51 +117,71 @@ impl Client for SyncClient {
...
     // Create the UDP socket backing the QUIC connection, and register it with
     // the event loop.
-    let mut socket =
-        mio::net::UdpSocket::bind(bind_addr.parse().unwrap()).unwrap();
+    let mut socket = mio::net::UdpSocket::bind(bind_addr).unwrap();
     poll.registry()
         .register(&mut socket, mio::Token(0), mio::Interest::READABLE)
         .unwrap();
 
+    let mut config = create_config(&args);
```
LPardue: How about we make add a param for the keylogging to create_config and then we don't need to make the config mut here

h3i/src/client/sync_client.rs
```diff
@@ -87,51 +117,71 @@ impl Client for SyncClient {
 ///
 /// Returns a [ConnectionSummary] on success, [ClientError] on failure.
 pub fn connect(
-    args: Config, actions: &[Action],
+    args: Config, actions: Vec<Action>,
```
LPardue, line 124: Why change to Vec?

h3i/src/client/async_client.rs
```diff
@@ -0,0 +1,437 @@
+// Copyright (C) 2024, Cloudflare, Inc.
```
LPardue:
```suggestion
// Copyright (C) 2025, Cloudflare, Inc.
```

:stuck_out_tongue:

h3i/src/client/sync_client.rs
evanrittenhouse: The async `H3iDriver` takes ownership of the actions, so it's easier to use a `Vec`. I changed this one for simplicity to unify the async/sync `connect()` signatures, but I can leave them different instead?

Cargo.toml
evanrittenhouse: Async h3i now depends on TQ, so I think it makes sense to have it depend on the workspace version

h3i/src/client/sync_client.rs
evanrittenhouse: I put the `config.log_keys()` call into `create_config`, though `config` has to stay `mut` because `quiche::connect()` requires an `&mut Config`

h3i/src/client/sync_client.rs
LPardue: nah Vec is fine for now, these are not in the hot path

Cargo.toml
LPardue: Adding it here makes it consistent with other crates in the repo, so this seems fine AFAICT

LPardue (approved):

PR #2043, opened 2025-05-01
Update delivery_rate units for gcongestion

toidiu:
This PR updates the gcongestion's delivery_rate so the units match the congestion implementation.

**Reasoning**
`recovery.delivery_rate()` is only [used](https://github.com/cloudflare/quiche/commit/e7dba7c39109e6e5d71f0d91418af2905188e9f1) for PathStats and qlog. Therefore as noted on [PathStats](https://github.com/cloudflare/quiche/blob/master/quiche/src/path.rs#L910), the units should be bytes/s

### DIFF
--- a/quiche/src/recovery/congestion/delivery_rate.rs
+++ b/quiche/src/recovery/congestion/delivery_rate.rs
@@ -191,6 +191,7 @@ impl Rate {
 
 #[derive(Default, Debug)]
 struct RateSample {
+    // The sample delivery_rate in bytes/sec
     delivery_rate: u64,
 
     is_app_limited: bool,
--- a/quiche/src/recovery/congestion/delivery_rate.rs
+++ b/quiche/src/recovery/congestion/delivery_rate.rs
@@ -32,6 +32,8 @@
 use std::time::Duration;
 use std::time::Instant;
 
+use crate::recovery::bandwidth::Bandwidth;
+
 use super::Acked;
 use super::Sent;
 
@@ -73,7 +75,7 @@ impl Default for Rate {
 
             largest_acked: 0,
 
-            rate_sample: RateSample::default(),
+            rate_sample: RateSample::new(),
         }
     }
 }
@@ -149,9 +151,11 @@ impl Rate {
 
             if !interval.is_zero() {
                 // Fill in rate_sample with a rate sample.
-                self.rate_sample.delivery_rate =
-                    (self.rate_sample.delivered as f64 / interval.as_secs_f64())
-                        as u64;
+                let bytes_per_second = (self.rate_sample.delivered as f64 /
+                    interval.as_secs_f64())
+                    as u64;
+                self.rate_sample.bandwidth =
+                    Bandwidth::from_bytes_per_second(bytes_per_second);
             }
         }
     }
@@ -168,8 +172,8 @@ impl Rate {
         self.delivered
     }
 
-    pub fn sample_delivery_rate(&self) -> u64 {
-        self.rate_sample.delivery_rate
+    pub fn sample_delivery_rate(&self) -> Bandwidth {
+        self.rate_sample.bandwidth
     }
 
     pub fn sample_rtt(&self) -> Duration {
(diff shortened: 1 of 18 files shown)


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -856,10 +856,12 @@ impl RecoveryOps for GRecovery {
         r.rtt() + (r.rttvar() * 4).max(GRANULARITY)
     }
 
+    /// The most recent data delivery rate estimate in bytes/s.
     fn delivery_rate(&self) -> u64 {
         self.pacer
             .bandwidth_estimate(&self.rtt_stats)
-            .to_bits_per_second()
+            .to_bits_per_second() /
+            8
```
antoniovicente, line 864: Please look at the implementation of `maybe_qlog` also.  It seems like the two congestion control implementations put different info in that field.  Don't know how to best reconcile the two.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -856,10 +856,12 @@ impl RecoveryOps for GRecovery {
         r.rtt() + (r.rttvar() * 4).max(GRANULARITY)
     }
 
+    /// The most recent data delivery rate estimate in bytes/s.
```
antoniovicente, line 859: Any tests that we can run against both implementations in order to validate that they mostly compute the same thing?

antoniovicente (commented):
Thanks for the fix.  Seems more or less correct; the gcongestion implementation of `self.delivery_rate()` seems to only be used in `maybe_qlog` and `path::stats`

quiche/src/recovery/gcongestion/recovery.rs
toidiu: I added a test for congestion https://github.com/cloudflare/quiche/pull/2043/commits/68b1bc6c493dfc7b30bf23a393092563cdb18e85

gcongestion operates very differently and I wasn't able to easily add a test that increases the delivery_rate. I am in the process of reverse engineering how gcongestion operates so I can control it.

quiche/src/recovery/gcongestion/recovery.rs
toidiu: Yea not sure who is using qlog currently. I'll track this separately for now.

quiche/src/recovery/mod.rs
```diff
@@ -185,6 +185,7 @@ pub trait RecoveryOps {
 
     fn pto(&self) -> Duration;
 
+    /// The most recent data delivery rate estimate in bytes/s.
     fn delivery_rate(&self) -> u64;
```
antoniovicente, line 189: Would be good to change this function to return a Bandwidth

fn delivery_rate(&self) -> Bandwidth;

toidiu: Address in https://github.com/cloudflare/quiche/pull/2043/commits/656aa246cb503d227f127332f143f4525352a935

quiche/src/recovery/congestion/delivery_rate.rs
```diff
@@ -210,6 +215,22 @@ struct RateSample {
...
+            interval: Duration::ZERO,
+            delivered: 0,
+            prior_delivered: 0,
+            prior_time: None,
+            send_elapsed: Duration::ZERO,
+            ack_elapsed: Duration::ZERO,
+            rtt: Duration::ZERO,
+        }
+    }
+}
```
toidiu, line 232: `RateSample` was auto-deriving `Default` before, now it will be via `new()`

I specifically didn't want to auto-devire `Default` for `Bandwidth` since that could lead to subtle errors. Instead Bandwidth should be created via explicit methods such as `zero`, `infinity`...

quiche/src/recovery/congestion/delivery_rate.rs
```diff
@@ -168,8 +172,8 @@ impl Rate {
         self.delivered
     }
 
-    pub fn sample_delivery_rate(&self) -> u64 {
-        self.rate_sample.delivery_rate
```
antoniovicente, line 172: So apparently delivery rate and bandwidth are slightly different concepts...

The name delivery_rate seems more descriptive name that refers to something in the BBR draft.  Bandwidth may be an ok name for the struct, but it may make sense to use delivery rate for many of the function and data member names because it tells us more about where the bandwidth measurement came from.

toidiu: Yep that makes sense.

antoniovicente (approved):
thanks

PR #2044, opened 2025-05-02
tokio-quiche: add support for setting per-response h3 priorities

ghedo:
Currently tokio-quiche always sets the same default ptiority for all h3 responses. This change adds support for applications to specify an optional priority on a response-by-response basis instead, while still keeping the current default priority for backwards compatibility.

This is done by adding a new `Option<Priority>` field to `OutboundFrame::Headers()`. Applications that don't want to provide any priority can simply set this to `None`, in which case the existing default priority logic applies.

In order to allow setting priorities for additional headers as well, a new method `send_additional_headers_with_priority()` is also added. This is required because the `Priority` fields are private, which means it would not be possible to call the lower level `stream_priority()` method with their values.

Finally, `Clone` and `Copy` traits are derived for `Priority` as well to simplify use in applications.

---

Testing the changes is a bit chllenging unfortunately, as there is no clearly visible signal for priorities, though in practice all this change amounts to calling `send_response_with_priority()` and `send_additional_headers_with_priority()` with a different value priority value, so in practice it shouldn't affect functionality much.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -811,7 +811,7 @@ pub enum Event {
 /// Structured Fields Dictionary field value. I.e, use `TryFrom` to parse the
 /// value of a Priority header field or a PRIORITY_UPDATE frame. Using this
 /// trait requires the `sfv` feature to be enabled.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[repr(C)]
 pub struct Priority {
     urgency: u8,
@@ -1203,10 +1203,6 @@ impl Connection {
     /// Sends an HTTP/3 response on the specified stream with specified
     /// priority.
     ///
-    /// The `priority` parameter represents [Extensible Priority]
-    /// parameters. If the urgency is outside the range 0-7, it will be clamped
-    /// to 7.
-    ///
     /// This method sends the provided `headers` as a single initial response
     /// without a body.
     ///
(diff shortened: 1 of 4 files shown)


LPardue:
> In order to allow setting priorities for additional headers as well, a new method send_additional_headers_with_priority() is also added. This is required because the Priority fields are private, which means it would not be possible to call the lower level stream_priority() method with their values.

I'm fine adding another function of _with_priority. However, I'm not following the justification. The quiche and h3 modules view of priority is similar but different. Any direct call to stream_priority() would require doing an urgency value remapping anyway. The method takes primitives, so those seem easy enough for an app to manage. We could make lives easier if we made the fields public, or added getters/setters, and possibly even exposed some conversion methods between the two spaces. 

There are use cases where we want to change the stream priorities outside of HEADERS sending events. It would be nice to also make it ergonomic to support that at some point.

ghedo:
@LPardue what I mean is that given a `Priority` value (like what we get here from `OutboundFrame::Headers`), it's not possible to use it to call `stream_priority()` from outside quiche itself (e.g. in tokio-quiche), as the fields of `Priority` are private. So the choice was to either add the `_with_priority()` variant or make those fields public like you said, but adding the new method seemed better as it siplifies the application code (only one method to call instead of 2), and doesn't require exposing the mapping logic.

> There are use cases where we want to change the stream priorities outside of HEADERS sending events. It would be nice to also make it ergonomic to support that at some point.

Yeah, I think that can also be implemented by adding a new command that can be sent by applications synchronously, though it isn't necessary for our immediate use so I didn't implement it yet. It's a fairly different change from this PR anyway (wouldn't really touch the h3 code I think), so it can be implemented independently at a later time.

LPardue:
Yeah all sounds good to me, incremental improvements (pun intended)

LPardue (approved):

Issue #1259, opened 2022-06-22
rustls crypto backend


tbu-:
Duplicate of #1116 which was closed by its author.

Would a rustls backend be considered if it were pull-requested? How would it look like? Probably enabled using feature flags?

tbu-:
This would probably fix #1148.

dignifiedquire:
Would a PR providing this support be considered for merge?

dzmitry-lahoda:
on of reasons we do not use quiche - to have simplest possible static builds. so we full rust, so unfortunately cannot use quiche to do sans-io (have to use quinn with async noise)

dzmitry-lahoda:
boringssl and openssl abstracted https://github.com/cloudflare/quiche/blob/b17904e522777746b2a6b12667d85283e1df368c/quiche/src/crypto/mod.rs#L621 with some tests, and there is example of using rustls in quinn.  so it should be easy to pr (and may be add more test if needed). 

so need approve from @quiche that this PR at all will be considered

dzmitry-lahoda:
cc @ghedo

ghedo:
Closing as duplicate of #1116 (also closed). Quiche currently has no plans to add a rustls backend; BoringSSL via `boring-sys` remains the sole supported TLS backend.

PR #2045, opened 2025-05-06
Add rustls compile time support

hargut:
This PR adds [rustls](https://rustls.dev/) compile time support to quiche. #1259 

The implementation intended to change as little as possible within the existing quiche library. As rustls is in some ways different to the already integrated TLS providers some changes had been necessary.

Additionally it was intended to stick to the existing configuration defaults which required to use some `danger` areas of rustls (related to changing the certificate validation defaults).

Any feedback is highly appreciated. 

Have a great time. :smiley:


toidiu:
Hi @hargut thank you for the PR and nice job.

Can you help me understand if there is something that rustls provides that you can't get from boringSSL (our default TLS provider)? Another TLS backend means more code that we need to test and maintain and adds to the complexity.

hargut:
Hi @toidiu,

the reasons for rustls are that it is a TLS implementation that is written in Rust (asides cryptography) and the performance it can deliver.  It can also provide a solution to use quiche for any projects that contain an OpenSSL dependency (avoiding conflicts during linking).

As within QUIC TLS is part of the protocol core it is not possible to use quiche with another external TLS provider (like it was e.g. possible on HTTP1 & HTTP2).

I understand that adding rustls as TLS provider adds to complexity and maintenance efforts but think that it can be a valuable addition for upstream projects that already use/offer rustls as TLS layer and would like to add QUIC/HTTP3 functionality through quiche.

Kind regards,
Harald

toidiu:
As you mentioned the complexity and maintenance effort is the primary blocker for us.

We have limited resources and maintain quiche primary to match our use case at Cloudflare. The added complexity/maintenance introduced in this PR would be more than we can take on at this time.

I want to thank you for this PR, it seems like you put a lot of effort into it. I also opened an issue to add a [CONTRIBUTING.md](https://github.com/cloudflare/quiche/issues/2056) guide to avoid surprises in the future.

hargut:
Hi @toidiu,

given the comments within related issues on the subject I'm not surprised or disappointed by the outcome. The decision is understandable and fully respected.

As interest was expressed by others within #1259 I thought it can be worthwhile to share the work. For now I'll keep the forked repository up for further reference.

This is an interesting topic for me and provided an opportunity to dive deeper into the TLS and encryption mechanisms within the Quic protocol.

Thank you for taking the time to look into it. :+1: 

Kind regards,
Harald

PR #2046, opened 2025-05-06
tokio-quiche: Add input path to file errors

TheJokr:
BoringSSL does this internally when we ask it to load a PEM file, but for RPK we call `fs::read` ourselves. The error message is not very helpful in those cases, as we simply bubble up an io::Error without any context.

`anyhow` is already an implicit dependency via foundations, so I figured using it was fine.

### DIFF
--- a/tokio-quiche/src/settings/config.rs
+++ b/tokio-quiche/src/settings/config.rs
@@ -223,10 +223,10 @@ fn quiche_config_with_tls(
         #[cfg(feature = "rpk")]
         CertificateKind::RawPublicKey => {
             let mut ssl_ctx_builder = boring::ssl::SslContextBuilder::new_rpk()?;
-            let raw_public_key = std::fs::read(tls.cert)?;
+            let raw_public_key = read_file(tls.cert)?;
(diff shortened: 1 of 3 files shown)


tokio-quiche/src/settings/config.rs
```diff
@@ -245,3 +245,11 @@ fn quiche_config_with_tls(
         },
     }
 }
+
+#[cfg_attr(not(feature = "rpk"), expect(dead_code))]
+fn read_file(path: &str) -> QuicResult<Vec<u8>> {
+    use anyhow::Context as _;
```
rushilmehra, line 251: what does the `as _` do?

rushilmehra (approved):
Just a few questions, lgtm otherwise

tokio-quiche/src/settings/config.rs
TheJokr: Imports it without adding it to the scope, which makes the trait's methods available without conflicting with e.g. a struct named `Context`.

See https://rust-lang.github.io/rust-clippy/master/index.html#unused_trait_names

rushilmehra (approved):

PR #2047, opened 2025-05-06
Cleanup the gcongestion implementation

toidiu:
The current implementation assumes being generic over cc algorithms (bbr and bbr2). However, we currently only have the bbr2 implementaion. The generic implementation makes the code hard to reason (e.g. why does the Pacer implement the `CongestionControl` trait) about and the use of enum_dispatch makes code navigation more cumbersome.

This PR cleanups up gcongestion with the assumption that we wont be porting over the bbr implementation.

**Review recommendation**
It going to be easiest to review the code one commit at a time since each commit is an atomic change.


toidiu:
> This PR cleanups up gcongestion with the assumption that we wont be porting over the bbr implementation.

Is this assumption correct?

I would argue, that even if we decide to port the bbr as implementation at some point in the future, it should not prevent us from improving code quality now.

PR #2048, opened 2025-05-07
Add Stat for spurious lost packets

antoniovicente:


quiche/src/recovery/mod.rs
```diff
@@ -130,6 +130,13 @@ pub(crate) enum Recovery {
     GCongestion(GRecovery),
 }
 
+#[derive(Debug, Default, PartialEq)]
+pub struct OnAckReceivedResult {
+    pub lost_packets: usize,
+    pub lost_bytes: usize,
+    pub acked_bytes: usize,
+}
```
toidiu: In Rust, "Result" implies an <Ok, Err> enum by convention and makes it easy to tell if a function can fail. Can we rename this to something like `OnAckReceivedOutcome`

antoniovicente: Thanks for catching that.  Fixed.

tokio-quiche/src/quic/connection/mod.rs
```diff
@@ -125,6 +125,7 @@ impl AsSocketStats for QuicConnectionStats {
             packets_sent: self.stats.sent as u64,
             packets_recvd: self.stats.recv as u64,
             packets_lost: self.stats.lost as u64,
+            packets_lost_spurious: self.stats.spurious_lost as u64,
```
antoniovicente, line 128: I forgot to add it to SocketStats PTAL

toidiu (approved):

PR #2049, opened 2025-05-07
Add max RTT to QUIC connection stats

antoniovicente:


quiche/src/recovery/rtt.rs
```diff
@@ -124,7 +132,19 @@ impl RttStats {
     }
 
     pub(crate) fn min_rtt(&self) -> Option<Duration> {
-        self.min_rtt.ne(&Duration::ZERO).then_some(*self.min_rtt)
```
antoniovicente, line 127: We are no longer initializing min_rtt to ZERO, so this check was not producing the intended result.  Looking at has_first_rtt_sample is more correct.

toidiu: verified that min_rtt updated here: https://github.com/cloudflare/quiche/commit/11e4e65041b533530cef5bc22a272f068be7fe78.

Ideally this would be a separate PR but separate commit works.

quiche/src/recovery/rtt.rs
```diff
@@ -77,17 +80,22 @@ impl RttStats {
...
+            self.has_first_rtt_sample = true;
             return;
         }
 
         // min_rtt ignores acknowledgment delay.
         self.min_rtt.running_min(RTT_WINDOW, now, latest_rtt);
 
+        if latest_rtt > self.max_rtt {
+            self.max_rtt = latest_rtt;
+        }
```
toidiu, line 97:
```suggestion
self.max_rtt = self.max_rtt.max(latest_rtt);
```

quiche/src/recovery/rtt.rs
```diff
@@ -37,6 +37,8 @@ pub(crate) const RTT_WINDOW: Duration = Duration::from_secs(300);
 pub struct RttStats {
     pub(super) latest_rtt: Duration,
 
+    pub(super) max_rtt: Duration,
```
toidiu, line 40: nit: would love to avoid `pub` if possible. Do we access this outside this file?

quiche/src/recovery/rtt.rs
antoniovicente: Is it truly a readability improvement?

quiche/src/recovery/rtt.rs
antoniovicente: Done.

quiche/src/recovery/rtt.rs
toidiu: I think so since it avoids an if condition and captures the intent of the if condition more precisely (set max_rtt to the max of the two value).

antoniovicente: done

toidiu (approved):

toidiu:
Looks great and nice catching the zero bug.

PR #2050, opened 2025-05-08
PMTUD Tests and Updates

cf-rhett:
In attempts to use PMTUD I found a number of issues with the existing implementation. This PR attempts to address the following

1. PMTUD probes were only sent when handshake_done_sent was true. As this will only be true for the server implementation, I've changed this to handshake_completed which I believe conforms to original intent
2. If I do not limit the number of probes being sent at once, then they are not consistently acked/dropped so PMTUD cannot function. Additionally, the algorithm did not take into account the sized of the failed probes so many 100s/1000s of probes would be sent at the initial probe size and if the PMTU was less than the probe size this would cause the PMTU to be calculated to always be 1200 (minimum supported MTU)
3. PMTUD was stopped at the first acked probe. e.g if the the PMTU was 1349 the prior implementation would calculate the PMTU to be 1275 with minimum MTU of 1200 and maximum of 1350. 

I've additionally modified some of the function names to be more readable

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3283,17 +3283,14 @@ impl<F: BufFactory> Connection<F> {
                     frame::Frame::Ping {
                         mtu_probe: Some(mtu_probe),
                     } => {
-                        let pmtud_next = p.pmtud.get_current();
-                        p.pmtud.set_current(cmp::max(pmtud_next, mtu_probe));
-
-                        // Stop sending path MTU probes after successful probe.
-                        p.pmtud.should_probe(false);
+                        p.pmtud.set_estimated_pmtu(mtu_probe);
+                        p.pmtud.successful_probe(mtu_probe);
                         pmtud_probe = true;
 
                         trace!(
                             "{} pmtud acked; pmtu size {:?}",
                             self.trace_id,
-                            p.pmtud.get_current()
+                            p.pmtud.get_estimated_pmtu()
                         );
                     },
 
@@ -3388,8 +3385,8 @@ impl<F: BufFactory> Connection<F> {
             if pmtud_probe {
                 trace!(
                     "{} updating pmtu {:?}",
-                    p.pmtud.get_current(),
-                    self.trace_id
+                    self.trace_id,
+                    p.pmtud.get_estimated_pmtu()
                 );
 
                 qlog_with_type!(
@@ -3402,7 +3399,7 @@ impl<F: BufFactory> Connection<F> {
                         let pmtu_data = EventData::MtuUpdated(
                             qlog::events::connectivity::MtuUpdated {
                                 old: Some(p.recovery.max_datagram_size() as u16),
-                                new: p.pmtud.get_current() as u16,
+                                new: p.pmtud.get_estimated_pmtu() as u16,
                                 done: Some(pmtud_probe),
                             },
                         );
@@ -3412,7 +3409,7 @@ impl<F: BufFactory> Connection<F> {
                 );
 
                 p.recovery
-                    .pmtud_update_max_datagram_size(p.pmtud.get_current());
+                    .pmtud_update_max_datagram_size(p.pmtud.get_estimated_pmtu());
             }
         }
 
@@ -3701,11 +3698,11 @@ impl<F: BufFactory> Connection<F> {
         let send_path = self.paths.get_mut(send_pid)?;
 
         // Update max datagram size to allow path MTU discovery probe to be sent.
-        if send_path.pmtud.get_probe_status() {
+        if send_path.pmtud.should_probe() {
             let size = if self.handshake_confirmed || self.handshake_done_sent {
                 send_path.pmtud.get_probe_size()
             } else {
-                send_path.pmtud.get_current()
+                send_path.pmtud.get_estimated_pmtu()
             };
 
             send_path.recovery.pmtud_update_max_datagram_size(size);
@@ -3912,8 +3909,11 @@ impl<F: BufFactory> Connection<F> {
                         self.ids.mark_retire_dcid_seq(seq_num, true)?;
                     },
 
-                    frame::Frame::Ping { mtu_probe } if mtu_probe.is_some() => {
-                        p.pmtud.pmtu_probe_lost();
+                    frame::Frame::Ping {
+                        mtu_probe: Some(failed_probe),
+                    } => {
+                        trace!("pmtud probe dropped: {}", failed_probe);
+                        p.pmtud.failed_probe(failed_probe);
                     },
 
                     _ => (),
@@ -3930,7 +3930,7 @@ impl<F: BufFactory> Connection<F> {
 
         let mut left = if path.pmtud.is_enabled() {
             // Limit output buffer size by estimated path MTU.
-            cmp::min(path.pmtud.get_current(), b.cap())
+            cmp::min(path.pmtud.get_estimated_pmtu(), b.cap())
         } else {
             b.cap()
         };
@@ -4131,25 +4131,22 @@ impl<F: BufFactory> Connection<F> {
 
             let pmtu_probe = active_path.should_send_pmtu_probe(
                 self.handshake_confirmed,
-                self.handshake_done_sent,
+                self.handshake_completed,
                 out_len,
                 is_closing,
                 frames.is_empty(),
             );
 
-            trace!("{} pmtud probe status {} hs_con={} hs_sent={} cwnd_avail={} out_len={} left={}", self.trace_id, pmtu_probe, self.handshake_confirmed, self.handshake_done_sent,
-                    active_path.recovery.cwnd_available(), out_len, left);
-
             if pmtu_probe {
+                let probe_size = active_path.pmtud.get_probe_size();
                 trace!(
-                    "{} sending pmtud probe pmtu_probe={} next_size={} pmtu={}",
+                    "{} sending pmtud probe pmtu_probe={} estimated_pmtu={}",
                     self.trace_id,
-                    active_path.pmtud.get_probe_size(),
-                    active_path.pmtud.get_probe_status(),
-                    active_path.pmtud.get_current(),
+                    probe_size,
+                    active_path.pmtud.get_estimated_pmtu(),
                 );
 
-                left = active_path.pmtud.get_probe_size();
+                left = probe_size;
 
                 match left.checked_sub(overhead) {
                     Some(v) => left = v,
@@ -4169,12 +4166,12 @@ impl<F: BufFactory> Connection<F> {
                 }
 
                 let frame = frame::Frame::Padding {
-                    len: active_path.pmtud.get_probe_size() - overhead - 1,
+                    len: probe_size - overhead - 1,
                 };
 
                 if push_frame_to_pkt!(b, frames, frame, left) {
                     let frame = frame::Frame::Ping {
-                        mtu_probe: Some(active_path.pmtud.get_probe_size()),
+                        mtu_probe: Some(probe_size),
                     };
 
                     if push_frame_to_pkt!(b, frames, frame, left) {
@@ -4183,6 +4180,7 @@ impl<F: BufFactory> Connection<F> {
                     }
                 }
 
+                active_path.pmtud.set_inflight(true);
                 pmtud_probe = true;
             }
 
@@ -4999,7 +4997,7 @@ impl<F: BufFactory> Connection<F> {
         if active_path.pmtud.is_enabled() {
             active_path
                 .recovery
-                .pmtud_update_max_datagram_size(active_path.pmtud.get_current());
+                .pmtud_update_max_datagram_size(active_path.pmtud.get_estimated_pmtu());
         }
 
         Ok((pkt_type, written))
@@ -6883,6 +6881,16 @@ impl<F: BufFactory> Connection<F> {
         ConnectionId::from_ref(e.cid.as_ref())
     }
 
+    /// Returns the PMTU for the active path if it exists
+    #[inline]
+    pub fn pmtu(&self) -> Option<usize> {
+        if let Ok(path) = self.paths.get_active() {
+            path.pmtud.get_pmtu()
+        } else {
+            None
+        }
+    }
+
     /// Returns true if the connection handshake is complete.
     #[inline]
     pub fn is_established(&self) -> bool {
@@ -7120,7 +7128,7 @@ impl<F: BufFactory> Connection<F> {
 
         active_path.recovery.update_max_ack_delay(max_ack_delay);
 
-        if active_path.pmtud.get_probe_status() {
+        if active_path.pmtud.should_probe() {
             active_path.recovery.pmtud_update_max_datagram_size(
                 active_path
                     .pmtud
@@ -7341,7 +7349,7 @@ impl<F: BufFactory> Connection<F> {
                 self.streams.has_stopped() ||
                 self.ids.has_new_scids() ||
                 self.ids.has_retire_dcids() ||
-                send_path.pmtud.get_probe_status() ||
+                send_path.pmtud.should_probe() ||
                 send_path.needs_ack_eliciting ||
                 send_path.probing_required())
         {
@@ -18766,13 +18774,13 @@ mod tests {
 
         // Check that PMTU params are configured correctly
         let pmtu_param = &mut pipe.server.paths.get_mut(pid_1).unwrap().pmtud;
-        assert!(pmtu_param.get_probe_status());
+        assert!(pmtu_param.should_probe());
         assert_eq!(pmtu_param.get_probe_size(), 1350);
         assert_eq!(pipe.advance(), Ok(()));
 
         for (_, p) in pipe.server.paths.iter_mut() {
-            assert_eq!(p.pmtud.get_current(), 1350);
-            assert!(!p.pmtud.get_probe_status());
+            assert_eq!(p.pmtud.get_estimated_pmtu(), 1350);
+            assert!(!p.pmtud.should_probe());
         }
     }
 
@@ -18815,7 +18823,7 @@ mod tests {
 
         // Check that PMTU params are configured correctly
         let pmtu_param = &mut pipe.server.paths.get_mut(pid_1).unwrap().pmtud;
-        assert!(pmtu_param.get_probe_status());
+        assert!(pmtu_param.should_probe());
         assert_eq!(pmtu_param.get_probe_size(), 1350);
         std::thread::sleep(
             pipe.server.paths.get_mut(pid_1).unwrap().recovery.rtt() +
@@ -18826,10 +18834,10 @@ mod tests {
         let pmtu_param = &mut active_server_path.pmtud;
 
         // PMTU not updated since probe is not ACKed
-        assert_eq!(pmtu_param.get_current(), 1200);
+        assert_eq!(pmtu_param.get_estimated_pmtu(), 1200);
 
         // Continue searching for PMTU
-        assert!(pmtu_param.get_probe_status());
+        assert!(pmtu_param.should_probe());
     }
 }
 
--- a/quiche/src/path.rs
+++ b/quiche/src/path.rs
@@ -346,10 +346,11 @@ impl Path {
         is_closing: bool, frames_empty: bool,
     ) -> bool {
         (hs_confirmed && hs_done) &&
-            self.pmtud.get_probe_size() > self.pmtud.get_current() &&
+            self.pmtud.get_probe_size() > self.pmtud.get_estimated_pmtu() &&
             self.recovery.cwnd_available() > self.pmtud.get_probe_size() &&
             out_len >= self.pmtud.get_probe_size() &&
-            self.pmtud.get_probe_status() &&
+            self.pmtud.should_probe() &&
+            !self.pmtud.get_inflight() &&
             !is_closing &&
             frames_empty
     }
@@ -578,7 +579,6 @@ impl PathMap {
         // Enable path MTU Discovery and start probing with the largest datagram
         // size.
         if enable_pmtud {
-            initial_path.pmtud.should_probe(enable_pmtud);
             initial_path.pmtud.set_probe_size(max_send_udp_payload_size);
             initial_path.pmtud.enable(enable_pmtud);
         }
--- a/quiche/src/pmtud.rs
+++ b/quiche/src/pmtud.rs
@@ -1,92 +1,289 @@
+/// This file contains the logic to implement PMTUD. Given a minimum supported MTU and an initial
+/// probe size that marks the maximum supported MTU, finds the PMTU between the two.
+
 #[derive(Default)]
 pub struct Pmtud {
-    /// The current path MTU estimate.
-    cur_size: usize,
+    /// The current PMTU estimate.
+    estimated_pmtu: usize,
 
-    /// The last MTU probe size that was attempted.
-    probe: usize,
+    /// The PMTU after the completion of PMTUD.
+    /// Will be [`None`] if the PMTU is less than the minimum supported MTU.
+    pmtu: Option<usize>,
 
-    /// Indicated if Path MTU probe needs to be generated.
-    next_size: bool,
+    /// The current PMTUD probe size. The initial value is the largest supported MTU.
+    probe_size: usize,
 
-    /// Check config for PMTU variable.
-    enable: bool,
+    /// The minimum supported MTU.
+    minimum_supported_mtu: usize,
+
+    /// The size of the smallest failed probe.
+    last_failed_probe_size: Option<usize>,
+
+    /// The size of the largest successful probe.
+    last_successful_probe_size: Option<usize>,
+
+    /// Indicates if PMTUD requires continued probing.
+    should_probe: bool,
+
+    /// Is PMTUD enabled.
+    enabled: bool,
+
+    /// Indicates if a PMTUD probe is inflight.
+    inflight: bool,
 }
 
 impl Pmtud {
     /// Creates new PMTUD instance.
-    pub fn new(cur_size: usize) -> Self {
+    pub fn new(minimum_supported_mtu: usize) -> Self {
         Self {
-            cur_size,
+            estimated_pmtu: minimum_supported_mtu,
+            minimum_supported_mtu,
             ..Default::default()
         }
     }
 
-    /// Enables Path MTU Discovery for the connection.
+    /// Enables PMTUD for the connection.
     pub fn enable(&mut self, enable: bool) {
-        self.enable = enable;
+        self.enabled = enable;
     }
 
-    /// Returns enable status for Path MTU Discovery for the connection.
+    /// Returns enabled status for PMTUD for the connection.
     pub fn is_enabled(&mut self) -> bool {
-        self.enable
+        self.enabled
     }
 
-    /// Specifies whether Path MTU Discovery should be performed at the next
-    /// opportunity, i.e., when the next packet is sent out if possible.
-    ///
-    /// Once Path MTU has been discovered, this maybe set to false.
-    pub fn should_probe(&mut self, pmtu_next: bool) {
-        self.next_size = pmtu_next;
+    /// Indicates whether probing should continue on the connection.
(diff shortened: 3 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -4131,25 +4131,22 @@ impl<F: BufFactory> Connection<F> {
...
             let pmtu_probe = active_path.should_send_pmtu_probe(
                 self.handshake_confirmed,
-                self.handshake_done_sent,
+                self.handshake_completed,
                 out_len,
                 is_closing,
                 frames.is_empty(),
             );
 
-            trace!("{} pmtud probe status {} hs_con={} hs_sent={} cwnd_avail={} out_len={} left={}", self.trace_id, pmtu_probe, self.handshake_confirmed, self.handshake_done_sent,
```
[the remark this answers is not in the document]
antoniovicente: Seems like too much logging.  Thanks for cleaning up.

quiche/src/lib.rs
```diff
@@ -6888,6 +6886,34 @@ impl<F: BufFactory> Connection<F> {
         ConnectionId::from_ref(e.cid.as_ref())
     }
 
+    /// Returns the PMTU for the active path if it exists
```
antoniovicente, line 6889: I don't really follow what the following 3 methods do.  Would be good to expand descriptions to describe what recalculate and validate imply.  Do they involve local actions only or need to transmit packets over the network?

quiche/src/lib.rs
```diff
@@ -6883,6 +6881,16 @@ impl<F: BufFactory> Connection<F> {
...
+    #[inline]
+    pub fn pmtu(&self) -> Option<usize> {
```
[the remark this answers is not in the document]
antoniovicente: Is there a planned tokio-quiche change in a followup PR to this repo?

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
...
-    cur_size: usize,
+    /// The current PMTU estimate.
+    estimated_pmtu: usize,
+
+    /// The PMTU after the completion of PMTUD.
+    /// Will be [`None`] if the PMTU is less than the minimum supported MTU.
+    pmtu: Option<usize>,
+
+    /// The current PMTUD probe size. The initial value is the largest supported MTU.
+    probe_size: usize,
```
antoniovicente, line 14: Should this be Option<usize> ?  I assume that most of the time there are 0 active mtu probes.

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
+/// This file contains the logic to implement PMTUD. Given a minimum supported MTU and an initial
+/// probe size that marks the maximum supported MTU, finds the PMTU between the two.
+
 #[derive(Default)]
 pub struct Pmtud {
-    /// The current path MTU estimate.
-    cur_size: usize,
+    /// The current PMTU estimate.
+    estimated_pmtu: usize,
```
antoniovicente, line 7: What is the role of estimated_pmtu vs pmtu?

Is estimated_pmtu eventually promoted to pmtu once based on a combination of successful probes and unsuccessful probes?

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
...
+    pmtu: Option<usize>,
+
+    /// The current PMTUD probe size. The initial value is the largest supported MTU.
+    probe_size: usize,
+
+    /// The minimum supported MTU.
+    minimum_supported_mtu: usize,
+
+    /// The maximum supported MTU.
+    maximum_supported_mtu: usize,
```
antoniovicente, line 20: Seems like maximum_supported_mtu is not always set.  Consider changing the type to Option<usize>

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
...
+        self.smallest_failed_probe_size = None;
+        self.largest_successful_probe_size = None;
+        self.estimated_pmtu = self.minimum_supported_mtu;
+        self.pmtu = None;
+    }
+
+    // Checks that a probe of PMTU size can be ack'd by enabling
+    // a probe on the next opportunity. If this probe is dropped
+    // PMTUD will restart from a fresh state
+    pub fn validate_pmtu(&mut self) {
```
antoniovicente, line 187: suggestion: `revalidate_pmtu`

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
...
+            self.smallest_failed_probe_size = Some(probe_size);
+        }
+
+        self.update_probe_size();
+        self.inflight = false;
+    }
+
+    // Resets PMTUD internals such that PMTUD will be recalculated
+    // on the next opportunity
+    pub fn recalculate_pmtu(&mut self) {
```
antoniovicente, line 176: I wonder if a method like `restart_pmtud` or `reset_pmtud` or `schedule_pmtu_recalculation` would better hint at the behavior triggered by this method.

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,423 @@
...
     /// Based on the Optimistic Binary algorithm defined in:
     /// Ref: <https://www.hb.fh-muenster.de/opus4/frontdoor/deliver/index/docId/14965/file/dplpmtudQuicPaper.pdf>
-    pub fn update_probe_size(&mut self) {
-        self.probe = self.cur_size + ((self.probe - self.cur_size) / 2);
+    fn update_probe_size(&mut self) {
+        match (
+            self.smallest_failed_probe_size,
+            self.largest_successful_probe_size,
+        ) {
+            // Binary search between successful and failed probes
```
antoniovicente, line 102: Sounds like the pmtud process will take around `log_2(mtu - 1200)` RTTs.  That seems like a long time with only 1 probe possible per RTT.  And it seems like we can't take advantage of any MTU increases until after the whole discovery process is complete?

Should the byte threshold for exiting PMTUD be configurable?  Targetting the right multiple of 10 bytes would speed up the discovery process by about 3 RTTs.

quiche/src/lib.rs
```diff
@@ -3417,7 +3414,7 @@ impl<F: BufFactory> Connection<F> {
                 );
 
                 p.recovery
-                    .pmtud_update_max_datagram_size(p.pmtud.get_current());
+                    .pmtud_update_max_datagram_size(p.pmtud.get_estimated_pmtu());
```
antoniovicente, line 3417: update_mss in `bbr2.rs` seems to be doing some odd updates to cwnd when pmtu is updated.  Don't know if the roundoff errors as the packet size is updated slightly would lead to problems eventually.

The updates in `congestion/recovery.rs` seem easier to follow

quiche/src/lib.rs
```diff
@@ -4136,25 +4136,22 @@ impl<F: BufFactory> Connection<F> {
...
                     self.trace_id,
-                    active_path.pmtud.get_probe_size(),
-                    active_path.pmtud.get_probe_status(),
-                    active_path.pmtud.get_current(),
+                    probe_size,
+                    active_path.pmtud.get_estimated_pmtu(),
                 );
 
-                left = active_path.pmtud.get_probe_size();
+                left = probe_size;
```
antoniovicente, line 4154: When using GSO the packets written to the output buffer must always be written in decreasing order.  I don't see an argument to Connection::send indicating if it is OK to add a large packet to the output buffer.  Perhaps we need to add that signal.

quiche/src/lib.rs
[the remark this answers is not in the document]
cf-rhett: No this is accessible already doing something like

```
async fn pmtu(&self) -> Option<usize> {
        let (tx, rx) = oneshot::channel();

        if let Err(err) = self.cmd_sender.send(QuicCommand::Custom(Box::new(|conn| {
            let pmtu = conn.pmtu();
            let _ = tx.send(pmtu);
        }))) {
            tracing::error!(
                ?err,
                "Failed to send QUIC command. PMTU cannot be established"
            );

            return None;
        };

        match tokio::time::timeout(QUIC_COMMAND_TIMEOUT, rx).await {
            Ok(Ok(pmtu)) => pmtu,
            Ok(Err(err)) => {
                tracing::error!(?err, "Failed to retrieve PMTU due receive error");
                return None;
            }
            Err(err) => {
                tracing::error!(?err, "Failed to retrieve PMTU in the allotted time");
                return None;
            }
        }
    }
 ```

File: quiche/src/path.rs [hunk 12]
```diff
@@ -346,10 +346,11 @@ impl Path {
         is_closing: bool, frames_empty: bool,
     ) -> bool {
         (hs_confirmed && hs_done) &&
-            self.pmtud.get_probe_size() > self.pmtud.get_current() &&
+            self.pmtud.get_probe_size() > self.pmtud.get_estimated_pmtu() &&
```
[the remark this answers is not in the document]
cf-rhett: I'd just tested macOS and Windows. Linux was next so I'll rebase on this before my testing

quiche/src/pmtud.rs
cf-rhett: estimated_pmtu is a relic of the initial work here. In the code its 1200 and then is set to the largest successful probe iteratively. Internal state is modified quite a bit based on the estimated pmtu and I didn't want to change the current behavior too much as I didn't have the context for a lot of the interactions. Let's talk about this in a 1-1 because I'd be happy to simplify this where I can.

pmtu is only set when we believe that the actual PMTU has been discovered. When its set it acts both as a signal that PMTUD has completed and a vehicle to deliver the value.

quiche/src/pmtud.rs
cf-rhett: It can be optional in the code but it can't be optional for PMTUD to work. I didn't make it optional for that reason. If I were to redesign this completely the constructor would take the max/min MTU but the PMTUD object is created far away from where the probe size is set.

quiche/src/pmtud.rs
cf-rhett: Sure, I can do `restart_pmtud`

quiche/src/pmtud.rs
cf-rhett: Sure -- I can add some logic for this

quiche/src/lib.rs
cf-rhett: I really only renamed the functionality here. I was hesitant to change anything functionally without context but happy to do so based on the advice on someone who works in the code

quiche/src/lib.rs
cf-rhett: I'll add some more details here

quiche/src/lib.rs
cf-rhett: let's talk about this

antoniovicente: I'm looking into the sendmsg issue.  Will report back when complete, you can proceed without worrying about this for now.

quiche/src/lib.rs
antoniovicente: Yeah, just brainstorming about interactions with the lower levels.  I know this change is a no-op rename.

quiche/src/lib.rs
```diff
@@ -1958,10 +1958,16 @@ impl<F: BufFactory> Connection<F> {
...
             &recovery_config,
             config.path_challenge_recv_max_queue_len,
-            MIN_CLIENT_INITIAL_LEN,
             true,
         );
 
+        if config.pmtud {
+            path.pmtud = Some(pmtud::Pmtud::new(
+                MIN_CLIENT_INITIAL_LEN,
+                config.max_send_udp_payload_size,
```
antoniovicente, line 1967: There's something odd here.  I thought that max_send_udp_payload_size was the size that the quiche implementation uses when sending out packets.

I would expect the pmtud probe size to be a larger configured value.  It can't be the same as config.max_send_udp_payload_size

antoniovicente: The source of my confusion is https://github.com/cloudflare/quiche/blob/d45788b/quiche/src/lib.rs#L5008

config.max_send_udp_payload_size acts as the send packet size when pmtud is disabled.  When pmtud is enabled, recovery.max_send_udp_payload_size is set to 1200 bytes by this code and then pmtud can try to raise it.

quiche/src/lib.rs
```diff
@@ -3707,16 +3712,18 @@ impl<F: BufFactory> Connection<F> {
...
+        if let Some(pmtud) = send_path.pmtud.as_mut() {
+            if pmtud.should_probe() {
+                let size = if self.handshake_confirmed || self.handshake_done_sent {
+                    pmtud.get_probe_size()
+                } else {
+                    pmtud.get_largest_succesful_probe()
+                };
 
-            send_path.recovery.pmtud_update_max_datagram_size(size);
+                send_path.recovery.pmtud_update_max_datagram_size(size);
```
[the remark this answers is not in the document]
cf-rhett: > Or something needs to call send_path.recovery.pmtud_update_max_datagram_size(pmtud.get_largest_succesful_probe()) after the probe packet is generated.

Doesn't the following do this?
https://github.com/cloudflare/quiche/pull/2050/files#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R3423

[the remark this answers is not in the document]
cf-rhett: > I think this will result in all packets generated by the recovery module to be of pmtud.get_probe_size() size from then on.

Ahh so this is a good catch. Since I'm working from the client side I never enter 

```
let size = if self.handshake_confirmed || self.handshake_done_sent {
                    pmtud.get_probe_size()
}
```

As the client will never have `self.handshake_done_sent` as the client doesn't send the handshake. I'll test this with `self.handshake_confirmed || self.handshake_completed` as I think this is the more appropriate logic

quiche/src/lib.rs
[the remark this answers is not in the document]
ghedo: Why not just get this from the socket stats in tokio-quiche?

quiche/src/lib.rs
```diff
@@ -6892,6 +6904,40 @@ impl<F: BufFactory> Connection<F> {
...
+            path.pmtud.as_ref().and_then(|pmtud| pmtud.get_pmtu())
+        } else {
+            None
+        }
+    }
+
+    /// Restarts the PMTUD process for the active path. This will send a series
+    /// of packets to probe for the PMTU within the supported MTU range.
+    #[inline]
+    pub fn restart_pmtud(&mut self) {
```
ghedo: What is the use case for this new API? Same for `revalidate_pmtu()`.

cf-rhett: `revalidate_pmtu` is used to periodically/conditionally validate that PMTU is still valid. e.g we have a medium sized cell service provider that has different PMTUs between different towers and if the PMTU decreases such that connection with the previous settings is dropping too much we need to adjust.

`restart_pmtud` is just future proofing this a bit as right now we're only functionally concerned about the pmtu dropping but eventually we'd like to be able to take of path changes that allow us to increase the size. I was originally intending on using just this but revalidate_pmtu is more lightweight as most of the time it only sends one probe whereas this sends 6-7 on average. I can remove this for now if we want.

quiche/src/lib.rs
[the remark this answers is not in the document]
cf-rhett: it doesn't allow me to recognize when pmtud is finished as on socket stats pmtu is set in the following fashion `pmtu: self.recovery.max_datagram_size()` so this will change for each successful probe. PMTUD is relatively quick so I'd prefer to only react to the changes once PMTUD completes. I could just keep checking the socket stats periodically but didn't think this added too much overhead

cf-rhett:
I'm currently investigating an issue on Windows where if PMTUD finds the PMTU to be <= 1325 everything breaks. This can be recreated without PMTUD by setting `max_send_udp_payload_size` on the QUIC Settings that manage the initial connection to the same range

cf-rhett:
See #2074 for a related fix on handling low MTU environments on macOS

quiche/src/lib.rs
```diff
@@ -19475,30 +19559,45 @@ mod tests {
...
+            .paths
+            .get_active_mut()
+            .unwrap()
+            .pmtud
+            .as_mut()
+            .unwrap();
+
         // MTU should finally update
-        assert_eq!(active_path.pmtud.get_current_mtu(), 1250);
+        assert_eq!(pmtud.get_current_mtu(), 1299);
```
cf-rhett: Since we've simulated a probe failing at 1300 but the actual MTU is greater than this the PMTUD probes will continue til they find a PMTU of 1299

quiche/src/pmtud.rs
```diff
@@ -1,96 +1,184 @@
...
+    maximum_supported_mtu: usize,
+
+    /// The size of the smallest failed probe.
+    smallest_failed_probe_size: Option<usize>,
+
+    /// The size of the largest successful probe.
+    largest_successful_probe_size: Option<usize>,
+
+    /// Indicates if a PMTUD probe is inflight. Used to limit probes to 1/RTT.
+    inflight: bool,
```
LPardue: nit: it should be two words "in flight", so in rust style `in_flight` here and elsewhere

quiche/src/lib.rs
LPardue: I'm not opposed to the idea of explicitly revalidating the currently chosen MTU. However, I wonder how an application would know to trigger this because it doesn't have the precise visibility into packet sizes, packet numbers, and loss. Maybe it would help for there to be an internal counter of "lost packets between 1200 and the current MTU" which could be used as a stronger indicator that the path MTU might have shrunk.

quiche/src/lib.rs
```diff
@@ -1972,10 +1972,16 @@ impl<F: BufFactory> Connection<F> {
             peer,
             &recovery_config,
             config.path_challenge_recv_max_queue_len,
-            MIN_CLIENT_INITIAL_LEN,
             true,
         );
 
+        if config.pmtud {
```
rushilmehra: We can probably just pass `config.pmtud` into the `Path` constructor and initialize `Path::pmtud` there

rushilmehra: Or just `&Config`, so you can access both `config.pmtud` and `config.max_send_udp_payload_size`

quiche/src/lib.rs
```diff
@@ -3391,7 +3395,7 @@ impl<F: BufFactory> Connection<F> {
         }
 
         // Following flag used to upgrade datagram size, if probe is successful.
-        let mut successful_pmtud_probe = false;
+        let mut successful_pmtud_probe: Option<usize> = None;
```
rushilmehra: While we're here, I think it'd be nice to put L3507-L3505 into a dedicated method on `quiche::Connection`, and then call that method inline on L3410. I'm not a huge fan of stashing the result in a temporary variable and then updating PMTU inline a bit later down in the function

quiche/src/pmtud.rs
cf-rhett: done

quiche/src/lib.rs
cf-rhett: Ended up passing Option<config> as not all path creation locations have a config but done

quiche/src/lib.rs
cf-rhett: I can't do it in a function as it requires a double borrow of self -- I moved this inline as there's not too much logic/LoC. Possible there's some Rust that lets you extract this cleanly but I wasn't able to do so

quiche/src/lib.rs
cf-rhett: I've removed it as I have no use for it currently -- if I find a use for it I'll come back and add it with reasons

quiche/src/lib.rs
```diff
@@ -7277,10 +7298,17 @@ impl<F: BufFactory> Connection<F> {
...
 
-        if active_path.pmtud.get_should_probe() {
+        if active_path
+            .pmtud
+            .as_ref()
+            .map(|pmtud| pmtud.should_probe())
+            .unwrap_or(false)
+        {
             active_path.recovery.pmtud_update_max_datagram_size(
                 active_path
```
rushilmehra, line 7309: This makes me wonder if we should respect the peer's `max_udp_payload_size` in our initial PMTUD probe size. The spec doesn't mandate anything here, but we may have a better chance of settling on the optimal MTU earlier if we respect this parameter. This isn't something you need to tackle in this PR, just thinking out loud

quiche/src/pmtud.rs
```diff
@@ -1,96 +1,181 @@
...
     }
 
-    /// Specifies whether PMTUD should be performed at the next opportunity,
-    /// i.e., when the next packet is sent out if possible.
-    ///
-    /// Once Path MTU has been discovered, this may be set to false.
-    pub fn set_should_probe(&mut self, should_probe: bool) {
-        self.should_probe = should_probe;
+    /// Returns the PMTUD probe size.
+    pub fn get_probe_size(&mut self) -> usize {
```
rushilmehra:
```suggestion
    pub fn get_probe_size(&self) -> usize {
```

quiche/src/pmtud.rs
```diff
@@ -1,96 +1,181 @@
...
+        self.probe_size
     }
 
-    /// Returns the value of the Path MTU Discovery flag.
-    pub fn get_should_probe(&self) -> bool {
-        self.should_probe
+    /// Returns the largest successful PMTUD probe size if one exists, otherwise
+    /// returns the minimum supported MTU.
+    pub fn get_current_mtu(&mut self) -> usize {
```
rushilmehra:
```suggestion
    pub fn get_current_mtu(&self) -> usize {
```

quiche/src/pmtud.rs
```diff
@@ -1,96 +1,181 @@
...
 
-    /// The last PMTU probe size that was attempted.
+    /// The current PMTUD probe size. Set to maximum_supported_mtu at
+    /// initialization.
     probe_size: usize,
 
-    /// Whether or not a PMTU probe needs to be generated.
-    should_probe: bool,
+    /// The minimum supported MTU.
+    minimum_supported_mtu: usize,
```
rushilmehra: What's the use case for making this configurable? QUIC mandates a minimum of 1200 bytes, and it seems sensible to always support that as the minimum

quiche/src/pmtud.rs
```diff
@@ -1,96 +1,181 @@
...
+            self.pmtu.is_none() &&
+            self.smallest_failed_probe_size != Some(self.minimum_supported_mtu)
     }
 
-    /// Whether or not PMTUD is enabled for the connection.
-    pub fn is_enabled(&mut self) -> bool {
-        self.enabled
+    /// Sets the PMTUD probe size.
+    pub fn set_probe_size(&mut self, probe_size: usize) {
+        self.probe_size = std::cmp::min(probe_size, self.maximum_supported_mtu);
```
rushilmehra, line 50: Should this scream a bit louder when `probe_size > self.maximum_supported_mtu`? We may want to go a step further and just make this API private

quiche/src/lib.rs
cf-rhett: Hmm yeah should the initial probe_size/maximum supported MTU be min(max_udp_payload_size, max_send_udp_payload_size) or prefer one of max_udp_payload_size or max_send_udp_payload_size

quiche/src/pmtud.rs
cf-rhett: We have some clients that operate in a pure IPv6 environment so I was thinking about the cases where I set the minimum to ~1280. As there's no way to pipe this setting right now and functionally it just uses 1200 everywhere I'll just go ahead and default to 1200 here instead of making this configurable

quiche/src/pmtud.rs
cf-rhett: I'll make this private.

quiche/src/pmtud.rs
cf-rhett: With the redesign this cannot be optional.

quiche/src/pmtud.rs
cf-rhett: There won't always be an active probe but this will always be set to the size of the next probe that should be sent until we converge on the pmtu.

quiche/src/pmtud.rs
cf-rhett: Discussed offline and suggested finding the exact value was sufficient for now

quiche/src/pmtud.rs
cf-rhett: done

quiche/src/pmtud.rs
cf-rhett: done

quiche/src/pmtud.rs
cf-rhett: Also per this comment to make this configurable https://github.com/cloudflare/quiche/pull/2050#discussion_r2103654309 -- I think this is fine to do in a separate PR as it will need some settings additions to provide a field to make this configurable but just as a note

quiche/src/lib.rs
LPardue: This is basically described in the DPLPMTUD spec https://www.rfc-editor.org/rfc/rfc8899#section-5.3.1

> The method discovers the search range by confirming the minimum PLPMTU and then using the probe method to select a PROBED_SIZE less than or equal to MAX_PLPMTU. MAX_PLPMTU is the minimum of the local MTU and EMTU_R (when this is learned from the remote endpoint). The MAX_PLPMTU MAY be reduced by an application that sets a maximum to the size of datagrams it will send.

in other words starting probe being min(max_udp_payload_size, max_send_udp_payload_size)  sounds right ot me

cf-rhett: I've made the change to use min(max_udp_payload_size, max_send_udp_payload_size) here -- let me know what you think

quiche/src/lib.rs
```diff
@@ -3390,30 +3388,48 @@ impl<F: BufFactory> Connection<F> {
...
+                                }
+                            );
 
-                        trace!(
-                            "{} pmtud acked; current mtu: {:?}",
-                            self.trace_id,
-                            p.pmtud.get_current_mtu()
-                        );
-                    },
+                            p.recovery.pmtud_update_max_datagram_size(mtu_probe);
```
rushilmehra: `pmtud.successful_probe()` ends up respecting `std::cmp::max(mtu_probe, self.largest_successful_probe_size)`, but we always call `p.recovery.pmtud_update_max_datagram_size()` with `mtu_probe`. I think this is meant to be resilient against out-of-order packets / retransmissions in ACKs from the peer, but in those cases we may inadvertently lower MTU

rushilmehra: Ditto for the qlog event

cf-rhett: Ahh good catch -- updated so this always sets this to largest successful pmtud probe

quiche/src/lib.rs
```diff
@@ -3388,30 +3386,56 @@ impl<F: BufFactory> Connection<F> {
...
 
-                        trace!(
-                            "{} pmtud acked; current mtu: {:?}",
-                            self.trace_id,
-                            p.pmtud.get_current_mtu()
-                        );
-                    },
+                            // Ensure the probe is within the supported MTU range
+                            // before updating the max datagram size
+                            if let Some(successful_probe) =
```
rushilmehra: nit
```suggestion
                            if let Some(current_mtu) =
```

quiche/src/lib.rs
```diff
@@ -3388,30 +3386,56 @@ impl<F: BufFactory> Connection<F> {
...
+
+                                        q.add_event_data_with_instant(
+                                            pmtu_data, now,
+                                        )
+                                        .ok();
+                                    }
+                                );
+
+                                p.recovery.pmtud_update_max_datagram_size(
+                                    successful_probe,
```
rushilmehra: nit
```suggestion
                                    current_mtu,
```

quiche/src/lib.rs
cf-rhett: done

quiche/src/lib.rs
cf-rhett: done

rushilmehra (approved):

rushilmehra:
Thanks for your patience on this one @cf-rhett, merged

PR #2053, opened 2025-05-12
Make initial RTT configurable

aochagavia:
Before this change, the initial RTT was hardcoded to `333`. With this change, it's now possible for the user to configure that value through `Config::set_initial_rtt`. This is important for high-latency networks.

Context: I'm collaborating with the [tiptop IETF working group](https://datatracker.ietf.org/doc/charter-ietf-tiptop/) to test QUIC in a deep-space setting.

cc @marcblanchet

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -438,6 +438,9 @@ pub const MAX_CONN_ID_LEN: usize = crate::packet::MAX_CID_LEN as usize;
 /// The minimum length of Initial packets sent by a client.
 pub const MIN_CLIENT_INITIAL_LEN: usize = 1200;
 
+/// The default initial RTT.
+const DEFAULT_INITIAL_RTT: Duration = Duration::from_millis(333);
+
 #[cfg(not(feature = "fuzzing"))]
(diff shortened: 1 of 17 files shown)


quiche/src/lib.rs
```diff
@@ -8583,6 +8593,8 @@ impl<'a> Iterator for UnknownTransportParameterIterator<'a> {
 pub struct TransportParams {
     /// Value of Destination CID field from first Initial packet sent by client
     pub original_destination_connection_id: Option<ConnectionId<'static>>,
+    /// The initial RTT
+    pub initial_rtt: Duration,
```
ghedo: This isn't really a QUIC transport parameter, so I think it'd make more sense to add it directly to the `Config` struct.

aochagavia: That makes sense, just rebased and pushed a new version of the PR that has the parameter in `Config`

ghedo:
Merged, thanks @aochagavia!

aochagavia:
It's a bit late to mention, but I just wanted to thank you for finding the time to review and merge this. Very appreciated!

PR #2057, opened 2025-05-15
stats: expose the startup stats for connection

toidiu:
Based on https://github.com/cloudflare/quiche/pull/2061

BBR can exit the statup mode either due to excessive loss or bandwidth plateau. Exposing this reason can help in understanding CCA behavior and network characteristics.

### DIFF
--- a/datagram-socket/src/socket_stats.rs
+++ b/datagram-socket/src/socket_stats.rs
@@ -59,6 +59,26 @@ pub struct SocketStats {
     pub bytes_retrans: u64,
     pub bytes_unsent: u64,
     pub delivery_rate: u64,
+    pub startup_exit_reason: Option<StartupExit>,
+}
+
+/// The reason a CCA exited the startup phase.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct StartupExit {
+    pub cwnd: usize,
+    pub reason: StartupExitReason,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum StartupExitReason {
+    /// Exit startup due to excessive loss
+    Loss,
+
+    /// Exit startup due to bandwidth plateau.
+    BandwidthPlateau,
+
+    /// Exit startup due to persistent queue.
+    PersistentQueue,
 }
 
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -403,6 +403,7 @@ use qlog::events::EventImportance;
 use qlog::events::EventType;
 #[cfg(feature = "qlog")]
 use qlog::events::RawInfo;
+use recovery::OnLossDetectionTimeoutOutcome;
 use stream::StreamPriorityKey;
 
 use std::cmp;
@@ -1647,6 +1648,9 @@ where
 
     /// The anti-amplification limit factor.
     max_amplification_factor: usize,
+
+    /// The reason a CCA exited the startup phase.
+    startup_exit_reason: Option<StartupExit>,
 }
 
 /// Creates a new server-side connection.
@@ -2137,6 +2141,7 @@ impl<F: BufFactory> Connection<F> {
             stopped_stream_remote_count: 0,
 
             max_amplification_factor: config.max_amplification_factor,
+            startup_exit_reason: None,
         };
 
         if let Some(odcid) = odcid {
@@ -6323,12 +6328,20 @@ impl<F: BufFactory> Connection<F> {
                 if timer <= now {
                     trace!("{} loss detection timeout expired", self.trace_id);
 
-                    let (lost_packets, lost_bytes) = p.on_loss_detection_timeout(
+                    let OnLossDetectionTimeoutOutcome {
+                        lost_packets,
+                        lost_bytes,
+                        startup_exit_reason,
+                    } = p.on_loss_detection_timeout(
                         handshake_status,
                         now,
                         self.is_server,
                         &self.trace_id,
                     );
+                    // Record the first startup exit reason.
+                    if self.startup_exit_reason.is_none() {
+                        self.startup_exit_reason = startup_exit_reason;
+                    }
 
                     self.lost_count += lost_packets;
                     self.lost_bytes += lost_bytes as u64;
@@ -7013,6 +7026,7 @@ impl<F: BufFactory> Connection<F> {
             reset_stream_count_remote: self.reset_stream_remote_count,
             stopped_stream_count_remote: self.stopped_stream_remote_count,
             path_challenge_rx_count: self.path_challenge_rx_count,
+            startup_exit_reason: self.startup_exit_reason,
         }
     }
 
@@ -7418,6 +7432,7 @@ impl<F: BufFactory> Connection<F> {
                         lost_bytes,
                         acked_bytes,
                         spurious_losses,
+                        startup_exit_reason,
                     } = p.recovery.on_ack_received(
                         &ranges,
                         ack_delay,
@@ -7426,6 +7441,11 @@ impl<F: BufFactory> Connection<F> {
                         now,
                         &self.trace_id,
                     );
+
+                    // Record the first startup exit reason.
+                    if self.startup_exit_reason.is_none() {
+                        self.startup_exit_reason = startup_exit_reason;
+                    }
 
                     self.lost_count += lost_packets;
                     self.lost_bytes += lost_bytes as u64;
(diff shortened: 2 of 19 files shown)


quiche/Cargo.toml
```diff
@@ -70,6 +70,7 @@ cdylib-link-lines = { version = "0.1", optional = true }
 
 [dependencies]
 boring = { workspace = true, optional = true }
+datagram-socket = { workspace = true }
```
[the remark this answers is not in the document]
toidiu: Yea didnt see that, its a no go.

I dont know the purpose of datagram-socket, but can it depend on quiche?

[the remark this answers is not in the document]
toidiu: Avoided the dep in https://github.com/cloudflare/quiche/pull/2057/commits/ab6412bf4d1dec847eec5c6bd8a8910ded5571b5

Long term I would love something like `quiche_core` which could house reusable logic like stats.

datagram-socket/src/socket_stats.rs
```diff
@@ -59,6 +59,23 @@ pub struct SocketStats {
...
+    Loss {
+        /// The cwnd at startup exit
+        cwnd: usize,
+    },
+
+    /// Bandwidth (BBR)
+    BandwidthPlateau {
+        /// The cwnd at startup exit
+        cwnd: usize,
+    },
```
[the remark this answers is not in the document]
antoniovicente: The cwnd appearing in both of the enums seems a bit odd.

Would we consider a struct for StartupExitReason with a field for cwnd, a reason enum?

It would make extracing cwnd out somewhat easier.

datagram-socket/src/socket_stats.rs
```diff
@@ -59,6 +59,23 @@ pub struct SocketStats {
     pub bytes_retrans: u64,
     pub bytes_unsent: u64,
     pub delivery_rate: u64,
+    pub startup_exit_reason: Option<StartupExitReason>,
+}
+
+/// The reason a CCA exited the startup phase.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum StartupExitReason {
```
antoniovicente, line 67: Should we consider also recording the bandwidth estimate, number of packets in the most recent RTT and number of losses/retransmits in the most recent RTT?

This info is relevant regardless of the exit reason being Loss vs BandwidthPlateau.

quiche/src/lib.rs
```diff
@@ -7418,6 +7423,7 @@ impl<F: BufFactory> Connection<F> {
                         lost_bytes,
                         acked_bytes,
                         spurious_losses,
+                        startup_exit_reason,
```
antoniovicente, line 7426: I would consider just keeping this state inside recovery rather than returning it from on_ack_received.

Also, I think we could exit slow start during an RTO instead of ack processing instead of ack processing.

quiche/src/lib.rs
```diff
@@ -7013,6 +7017,7 @@ impl<F: BufFactory> Connection<F> {
             reset_stream_count_remote: self.reset_stream_remote_count,
```
antoniovicente, line 7017: Odd that Stats does not contain RTT information or other information stored in the recovery object.

quiche/src/recovery/gcongestion/bbr2/network_model.rs
```diff
@@ -630,21 +630,30 @@ impl BBRv2NetworkModel {
...
-            self.full_bandwidth_reached = true;
+            if self.rounds_with_queueing >= params.max_startup_queue_rounds {
+                self.full_bandwidth_reached = true;
+                outcome.full_bandwidth_reached = true;
+            }
         }
+
+        outcome.rounds_with_queueing = self.rounds_with_queueing;
+
+        outcome
```
antoniovicente, line 656: Refactors in this code make me uneasy given the testing state of the code in the gcongestion directory.

At the very least no-op refactors should be in separate commits which can be inspected independent of code changes that introduce new behaviors.

I don't think that this refactor is related to the introduction of startup related stats.  I strongly suggest we undo all changes to this function.

quiche/src/recovery/gcongestion/bbr2/network_model.rs
```diff
@@ -555,7 +555,7 @@ impl BBRv2NetworkModel {
         true
     }
 
-    pub(super) fn is_inflight_too_high(
+    pub(super) fn is_loss_too_high(
```
antoniovicente, line 558: Please undo rename.  My understanding is that these names were chosen to be consistent with a C++ codebase that implements similar functionality.

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -537,6 +538,7 @@ impl CongestionControl for BBRv2 {
         &mut self, _rtt_updated: bool, prior_in_flight: usize,
         _bytes_in_flight: usize, event_time: Instant, acked_packets: &[Acked],
         lost_packets: &[Lost], least_unacked: u64, _rtt_stats: &RttStats,
+        recovery_stats: &mut RecoveryStats,
```
antoniovicente, line 541: Could the Recovery module provide an accessor for RecoveryStats instead of having it be an output argument for on_congestion_event?

quiche/src/lib.rs
toidiu: Yea good point about other times we can exit startup and your suggestion makes sense.

I dont love keeping pure stat state inside recovery because it adds bloat (eventually I would like to make this a user opt-in functionality).

toidiu: > I would consider just keeping this state inside recovery rather than returning it from on_ack_received.

I think the return pattern is more type safe because it makes consume side-effects obvious and makes sure we consume them.

Return value:
"on_ack_received returns on_startup_exit reason. It might have recorded a value so make sure to consume it."

vs

Explicit API:
"The on_ack_received might have modified the on_ack_received value. You have to explicitly know that on_ack_received potentially records a new value and call a separate API `recovery.get_startup_exit_reason()`."

quiche/src/lib.rs
toidiu: Some rtt stats seem to be exposed [here](https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/quic/connection/mod.rs#L84) as part of [PathStats](https://github.com/cloudflare/quiche/blob/master/quiche/src/path.rs#L883-L894)

quiche/src/recovery/gcongestion/bbr2.rs
toidiu: Lets discuss in the above comment: https://github.com/cloudflare/quiche/pull/2057#discussion_r2096491363

quiche/src/recovery/gcongestion/bbr2/startup.rs
```diff
@@ -64,26 +69,50 @@ impl ModeImpl for Startup {
...
+        if exit_due_to_bw_plateau {
+            recovery_stats.startup_exit_reason =
+                Some(StartupExit::new(cwnd, StartupExitReason::BandwidthPlateau));
+        } else if exit_due_to_excessive_loss {
+            recovery_stats.startup_exit_reason =
+                Some(StartupExit::new(cwnd, StartupExitReason::Loss));
         }
 
-        if self.model.full_bandwidth_reached() {
+        if exit_due_to_bw_plateau || exit_due_to_excessive_loss {
```
antoniovicente, line 115: Possible way to simplify: just use exit_due_to_excessive_loss

```
if self.model.full_bandwidth_reached() {
  if exit_due_to_excessive_loss {
    recovery_stats.startup_exit_reason = Some(StartupExitReason::Loss)
  } else {
    recovery_stats.startup_exit_reason = Some(StartupExitReason::BandwidthPlateau)
  }
  self.into_drain(event_time, Some(congestion_event), params)
} else {
...
```

quiche/src/lib.rs
```diff
@@ -8453,6 +8473,9 @@ pub struct Stats {
 
     /// The total number of PATH_CHALLENGE frames that were received.
     pub path_challenge_rx_count: u64,
+
+    /// The reason a CCA exited the startup phase.
+    pub startup_exit_reason: Option<StartupExit>,
```
ghedo, line 8478: This should be in `PathStats` I think, since CC state is per-path.

datagram-socket/src/socket_stats.rs
antoniovicente: This can be a followup.  Just things to keep in mind when thinking about the data structures used to represent the results.

quiche/src/lib.rs
antoniovicente: Consider the following:
* Slow start is done on each path, so the right place to expose this is from `PathStats` as Alessandro points out.
* You are forced to remember to look at this output argument in both on_ack_received and on_loss_detection_timeout, not just 1 place
* The startup_exit_reason remains constant past the point at which it changes
* Significant prior art of getting info from the recovery module in `PathStats` accessor https://github.com/cloudflare/quiche/blob/master/quiche/src/path.rs#L504

toidiu:
replaced with https://github.com/cloudflare/quiche/pull/2064

PR #2059, opened 2025-05-16
Add toolchain file for nightly formatting

rushilmehra:
This will help instruct auto formatters like rust analyzer to use the
nightly toolchain.


evanrittenhouse (approved):

rushilmehra:
Don't think this is possible unfortunately, this overrides the toolchain for everything and it doesn't seem like this is possible to restrict to just rustfmt

PR #2060, opened 2025-05-16
tokio-quiche: always set UDP_SEGMENT cmsg when calling sendmsg()

rushilmehra:
Current logic avoids setting the UDP_SEGMENT cmsg on sendmsg() calls when sending a single packet over the wire. However, there are niche cases where it's desirable to update the segment size after a socket has been created. If the updated segment size is greater than what the socket was initialized with, and we omit the UDP_SEGMENT cmsg, then the kernel will fall back to the segment size that is configured on the socket and erroneously truncate the packet on the way out.

We now set the UDP_SEGMENT cmsg on each packet no matter what.

To help avoid this footgun, we now initialize GSO on the socket with the maximum possible segment size.

### DIFF
--- a/tokio-quiche/src/lib.rs
+++ b/tokio-quiche/src/lib.rs
@@ -194,9 +194,7 @@ where
             #[cfg_attr(not(target_os = "linux"), expect(unused_mut))]
             let mut socket = s.try_into()?;
             #[cfg(target_os = "linux")]
-            socket.apply_max_capabilities(
-                params.settings.max_send_udp_payload_size,
-            );
+            socket.apply_max_capabilities();
             Ok(socket)
         })
         .collect::<io::Result<_>>()?;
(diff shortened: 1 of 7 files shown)


tokio-quiche/src/lib.rs
```diff
@@ -194,9 +194,7 @@ where
             #[cfg_attr(not(target_os = "linux"), expect(unused_mut))]
             let mut socket = s.try_into()?;
             #[cfg(target_os = "linux")]
-            socket.apply_max_capabilities(
-                params.settings.max_send_udp_payload_size,
```
antoniovicente, line 198: params argument seems unused.  Could it be removed from listen()?

rushilmehra: `params` is piped down into `listen_with_capabilities`

antoniovicente: You may need to rename to _params

rushilmehra: Piping it to `listen_with_capabilities` is still considered as using the parameter

antoniovicente: I missed that line, sorry :)

antoniovicente (dismissed):

antoniovicente (approved):

PR #2062, opened 2025-05-20
gcongestion: don't consider app limited rounds for bw growth

ghedo:
Currently rounds without growth that are app limited are counted towards the threshold for number of rounds without growth.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2/network_model.rs
+++ b/quiche/src/recovery/gcongestion/bbr2/network_model.rs
@@ -182,8 +182,12 @@ pub(super) struct BBRv2NetworkModel {
     full_bandwidth_baseline: Bandwidth,
     rounds_without_bandwidth_growth: usize,
 
-    // Used by STARTUP and PROBE_UP to decide when to exit.
+    /// Used by STARTUP and PROBE_UP to decide when to exit.
     rounds_with_queueing: usize,
+
+    /// Determines whether app limited rounds with no bandwidth growth count
+    /// towards the rounds threshold to exit startup.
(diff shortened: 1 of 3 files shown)


antoniovicente:
Code change looks good.  Could we protected with a parameter in bbr2.rs Params and add an override to BbrParams?

https://github.com/cloudflare/quiche/blob/master/quiche/src/recovery/gcongestion/mod.rs#L168

ghedo:
Added BBR parameter to enable the new behaviour (and disabled by default).

quiche/src/recovery/gcongestion/bbr2/network_model.rs
```diff
@@ -611,12 +618,15 @@ impl BBRv2NetworkModel {
...
+
+        if !ignore_round {
+            self.rounds_without_bandwidth_growth += 1;
+        }
 
         // full_bandwidth_reached is only set to true when not app-limited
-        if self.rounds_without_bandwidth_growth >= params.startup_full_bw_rounds &&
-            !congestion_event.last_packet_send_state.is_app_limited
-        {
+        if self.rounds_without_bandwidth_growth >= params.startup_full_bw_rounds {
```
antoniovicente, line 629: But alert: we lost the `&& !congestion_event.last_packet_send_state.is_app_limited` in the next line.

That was fine when `self.rounds_without_bandwidth_growth` was only incremented when not appl limited, but that's no longer the case with the introduction of config.

antoniovicente (approved):
Thanks

rushilmehra (approved):

PR #2064, opened 2025-05-20
Emit statistics from when a CCA exits Startup

toidiu:
This PR emits stats from when a CCA exits the startup phase. Its currently only implemented for gcongestion (BBR).

**How to review**
Its probably best to review this code one commit at a time. The second commit modified the `startup.rs` and the most important piece. The other two commits are essentially the glue.

**Callouts**
This replaces https://github.com/cloudflare/quiche/pull/2057 and https://github.com/cloudflare/quiche/pull/2061. Those were trying to refactor the code, which has a lot of side effects and were introducing bugs in the process. This PR avoids refactoring the code to avoid that pitfall.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2/startup.rs
+++ b/quiche/src/recovery/gcongestion/bbr2/startup.rs
@@ -31,6 +31,9 @@
 use std::time::Instant;
 
 use crate::recovery::gcongestion::bbr2::Params;
+use crate::recovery::RecoveryStats;
+use crate::recovery::StartupExit;
+use crate::recovery::StartupExitReason;
 
 use super::mode::Mode;
 use super::mode::ModeImpl;
@@ -55,6 +58,7 @@ impl ModeImpl for Startup {
         _acked_packets: &[Acked], _lost_packets: &[Lost],
         congestion_event: &mut BBRv2CongestionEvent,
         _target_bytes_inflight: usize, params: &Params,
+        recovery_stats: &mut RecoveryStats, cwnd: usize,
     ) -> Mode {
         if self.model.full_bandwidth_reached() {
             return self.into_drain(event_time, Some(congestion_event), params);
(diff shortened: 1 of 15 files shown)


quiche/src/recovery/gcongestion/bbr2/startup.rs
```diff
@@ -70,21 +70,45 @@ impl ModeImpl for Startup {
...
+
+            if self.model.full_bandwidth_reached() {
+                recovery_stats.set_startup_exit(StartupExit::new(
+                    cwnd,
+                    StartupExitReason::Loss,
+                ));
+            }
         }
 
         if self.model.full_bandwidth_reached() {
```
antoniovicente, line 114: I would consider calling `recovery_stats.set_startup_exit` in the next line before the call to `self.into_drain` with an exit reason computed from the branches above.  The reason being it providing a single place where this stat is initialized, so it would be easier to add extra info to StartupExit in the future.  But your version is fine for now, let's go with it.

quiche/src/recovery/mod.rs
```diff
@@ -572,7 +572,16 @@ impl ReleaseDecision {
...
 #[derive(Default, Debug)]
 pub struct RecoveryStats {
-    pub startup_exit: Option<StartupExit>,
+    startup_exit: Option<StartupExit>,
+}
+
+impl RecoveryStats {
+    // Record statistics when a CCA first exits startup.
+    pub fn set_startup_exit(&mut self, startup_exit: StartupExit) {
+        if let None = self.startup_exit {
```
antoniovicente, line 581: Possible nit, this is equivalent to: `if self.startup_exit.is_none() {`

Clippy seems to have a similar opinion:
```
error: redundant pattern matching, consider using `is_none()`
   --> quiche/src/recovery/mod.rs:581:16
    |
581 |         if let None = self.startup_exit {
    |         -------^^^^-------------------- help: try: `if self.startup_exit.is_none()`
    |
    = help: for further information visit [https://rust-lang.github.io/rust-](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching)
```

quiche/src/recovery/gcongestion/bbr2/startup.rs
```diff
@@ -31,6 +31,9 @@
 use std::time::Instant;
 
 use crate::recovery::gcongestion::bbr2::Params;
+use crate::recovery::RecoveryStats;
+use crate::recovery::StartupExit;
+use crate::recovery::StartupExitReason;
```
antoniovicente, line 36: Some of these includes seem unused in this commit, but will become used when looking at the full PR.  I recommend squashing on merge since CI / clippy does not run cleanly on each commit.

I would consider squashing commits now.

antoniovicente (commented):
Looks good otherwise.

quiche/src/recovery/gcongestion/bbr2/startup.rs
toidiu: Yep I want to squash commit.. wish that was the default. I can do that since you are done with the review.

Breaking into separate commits was done by eye (for the ease of review) so I am not surprised I missed that.

quiche/src/recovery/mod.rs
```diff
@@ -566,6 +569,50 @@ impl ReleaseDecision {
...
+pub struct StartupExit {
+    /// The congestion_window recorded at Startup exit.
+    pub cwnd: usize,
+
+    /// The reason a CCA exited the startup phase.
+    pub reason: StartupExitReason,
+}
+
+impl StartupExit {
+    fn new(cwnd: usize, reason: StartupExitReason) -> Self {
```
antoniovicente, line 598: I think this function is now unused after your refactor to [startup.rs](https://github.com/cloudflare/quiche/pull/2064/files#diff-0d91abc11d379b124d2a6a20e11d1e04bc52581bbedffcc9a3cad414ec45b5e8)

quiche/src/recovery/gcongestion/bbr2/startup.rs
```diff
@@ -64,26 +68,47 @@ impl ModeImpl for Startup {
...
-        #[allow(clippy::absurd_extreme_comparisons)]
-        if params.max_startup_queue_rounds > 0 && !has_bandwidth_growth {
+        let check_persisten_queue =
+            params.max_startup_queue_rounds > 0 && !has_bandwidth_growth;
+        if check_persisten_queue {
             // 1.75 is less than the 2x CWND gain, but substantially more than
             // 1.25x, the minimum bandwidth increase expected during
             // STARTUP.
             self.model.check_persistent_queue(1.75, params);
+            if self.model.full_bandwidth_reached() {
```
antoniovicente, line 86: `if startup_exit_reason.is_none() && self.model.full_bandwidth_reached() {`

toidiu: The none check is now part of the `set_startup_exit` function.

antoniovicente (approved):
Looks great.  Thanks!

PR #2065, opened 2025-05-23
Allow configuring pmtud during the TLS handshake

rushilmehra:


quiche/src/lib.rs
```diff
@@ -7197,6 +7220,18 @@ impl<F: BufFactory> Connection<F> {
...
                     }
+
+                    if let Some(discover) = ex_data.pmtud {
+                        let path = self.paths.get_active_mut()?;
+
+                        path.pmtud.enable(discover);
+                        // It's harmless to set probe size even if we're disabling
+                        // pmtud.
+                        path.pmtud.set_probe_size(
+                            self.recovery_config.max_send_udp_payload_size,
```
antoniovicente, line 7231: probe size has to be larger than max_send_udp_payload_size for pmtud to do anything.

We need a new parameter that can be set per experiment.

quiche/src/lib.rs
```diff
@@ -7197,6 +7220,18 @@ impl<F: BufFactory> Connection<F> {
                     if ex_data.tx_cap_factor != self.tx_cap_factor {
                         self.tx_cap_factor = ex_data.tx_cap_factor;
                     }
+
+                    if let Some(discover) = ex_data.pmtud {
+                        let path = self.paths.get_active_mut()?;
+
+                        path.pmtud.enable(discover);
```
antoniovicente, line 7227: I think that path mtu discovery needs to be done by path.  We're overriding here based on crypto handshake happening...  I assume that crypto handshake doesn't happen per path, how do we ensure that experiments that enable pmtud use it on every path, not just the first one created?

rushilmehra: Updated to enable pmtud on all existing paths for the connection

quiche/src/path.rs
```diff
@@ -843,6 +843,19 @@ impl PathMap {
 
         Ok(())
     }
+
+    /// Configures path MTU discovery on all existing paths.
+    pub fn set_discover_pmtu_on_existing_paths(
+        &mut self, discover: bool, max_send_udp_payload_size: usize,
+    ) {
+        for (_, path) in self.paths.iter_mut() {
+            path.pmtud.enable(discover);
```
antoniovicente, line 852: What calls path.pmtud.enable on new path creation?

I think we're not modifying `Config {pmtud: bool}` in `lib.rs`.  See call to `path::PathMap::new`

https://github.com/cloudflare/quiche/blob/5a463e767b3c8999a5ce80097879f809ad973957/quiche/src/lib.rs#L1975

rushilmehra: > What calls path.pmtud.enable on new path creation?
> 
> I think we're not modifying `Config {pmtud: bool}` in `lib.rs`. See call to `path::PathMap::new`
> 
> https://github.com/cloudflare/quiche/blob/5a463e767b3c8999a5ce80097879f809ad973957/quiche/src/lib.rs#L1975

We can't modify `Config {pmtud: bool}` in the middle of the handshake, and I think only one `PathMap` is constructed per connection even if you're intending to utilize multiple `Path`s (they just get inserted into that `PathMap`. Also `config.pmtud` is only used for enabling PMTUD on the very initial path. Right now we don't support connection migration, so the initial path is the only one we use, but in the future we'll need to ensure that we enable pmtud for all `Path`s that need it. Out of scope for this PR though imo

antoniovicente (approved):
Thanks for adding tests.

antoniovicente (approved):

PR #2066, opened 2025-05-27
support changing transport params from BoringSSL handshake callbacks

ghedo:
This adds support for changing some transport parameters during the QUIC hanshake from inside BoringSSL callbacks (e.g. `select_cert_cb`). Currently only `max_idle_timeout` and `initial_max_streams_bidi` are supported, but more can be added later if needed.

Each parameter is configured with its own method rather than having a single method taking a new `TransportParameters` value to allow applications to overwrite specific parameters rather than having to provide an entire new set of parameters.

In the future we might consider adding a new `TransportParameters`-like type with optional fields, but that didn't seem necessary here as we are only interested in a few TPs at this point.

Because the server's TPs are sent in the first flight inside TLS messages generated by BoringSSL itself, they need to be re-encoded during the early handshake callback before control is returned to quiche, so this is done directly inside each `*_in_handshake()` method rather than in a single place (e.g. `do_handshake()`).

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -2552,6 +2552,81 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Sets the `max_idle_timeout` transport parameter, in milliseconds.
+    ///
+    /// This function can only be called inside one of BoringSSL's handshake
+    /// callbacks, before any packet has been sent. Calling this function any
+    /// other time will have no effect.
+    ///
+    /// See [`Config::set_max_idle_timeout()`].
+    ///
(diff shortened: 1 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -7227,6 +7297,24 @@ impl<F: BufFactory> Connection<F> {
...
                         );
                     }
+
+                    if ex_data.local_transport_params !=
+                        self.local_transport_params
+                    {
+                        if ex_data.local_transport_params.initial_max_streams_bidi !=
+                            self.local_transport_params
+                                .initial_max_streams_bidi
+                        {
```
toidiu, line 7307: do we really need this if statement? Can we always just call `set_max_streams_bidi` and if they are the same its a noop.

toidiu (approved):

PR #2069, opened 2025-05-28
Expose WouldBlock errors on sendmsg

rushilmehra:


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -528,6 +528,7 @@ where
                     current_send_buf,
                     self.write_state.segment_size,
                     self.write_state.tx_time,
+                    self.metrics.clone(),
```
rushilmehra, line 531: Not sure if this clone is going to add too much overhead. It's possible to rework this to avoid passing in the metrics struct to `send_to`, but it's a bit tedious

evanrittenhouse: Agreed, IMO we should rework it

rushilmehra: Yeah that would work, I was hesitant since it seemed a bit messy but it avoids the clone issue, so might as well. I pushed it up

PR #2070, opened 2025-05-29
Annotate qlog on congestion control state transitions.

antoniovicente:
Based on https://github.com/cloudflare/quiche/pull/1357, but extended to cover BBR2 and HyStart++


quiche/src/recovery/mod.rs
```diff
@@ -270,6 +277,27 @@ impl Recovery {
...
+        &mut self, qlog: &mut qlog::streamer::QlogStreamer, now: Instant,
+    ) {
+        if let Some(ev_data) = self.get_updated_qlog_event_data() {
+            qlog.add_event_data_with_instant(ev_data, now).ok();
+        }
+
+        if let Some(cc_state) = self.get_updated_qlog_cc_state(now) {
+            let ev_data = EventData::CongestionStateUpdated(
+                qlog::events::quic::CongestionStateUpdated {
+                    old: None,
```
evanrittenhouse, line 291: Given we already have the old state, is it worth attaching here for completeness? `self.get_updated_qlog_cc_state` could have return `(curr, prev)` states for use here, for example. Feel free to resolve if it wouldn't give much benefit though

antoniovicente: I'ld need an API to return both old and new.  Visualizations just based on the new state work pretty well.  Is old really needed?

LPardue (dismissed):

LPardue:
Looks good. The only suggestion I had was Maybe we should make the text more consistent and specific. E.g I algorith. Name and version in every variant, so there's no ambiguity . The counterargument is that we could log the algorithm once in a different event, or determine it in some other way

antoniovicente:
> Looks good. The only suggestion I had was Maybe we should make the text more consistent and specific. E.g I algorith. Name and version in every variant, so there's no ambiguity . The counterargument is that we could log the algorithm once in a different event, or determine it in some other way

I could see value in having a qlog message that describes the client and server configs, which would include info on congestion control algorithm name and any tuning parameters that they would like to share.

Having more generic names for the events may allow for use of the same name in multiple contexts, but as we have seen so far the names of the phases of Cubic+hystart differ significantly from the BBR1/BBR2 phases.  There are even some differences in names of the phases between BBR1 and BBR2 since BBR2 breaks down the probe BW phase into multiple sub-phases.

I don't have a strong opinion about adding an alg name prefix to the congestion control annotations; let me know if I should try to do that:
* add cubic/reno prefix to slow_start
* change bbr_ prefix to bbr1_, bbr2, bbr3_

antoniovicente:
> I think adding bbr version is a no brainer, it's just one character.
> 
> For reno/cubic, it's don't have a strong opinion, will leave it to your discretion

Changed the prefix to "bbr2_" for the BBR2 phases.

quiche/src/lib.rs
```diff
@@ -3347,9 +3347,7 @@ impl<F: BufFactory> Connection<F> {
 
         qlog_with_type!(QLOG_PACKET_RX, self.qlog, q, {
```
LPardue, line 3364: So i think there's a latent bug with this line, since the event is a metrics one, it should probably have alway been QLOG_METRICS. However, this makes me realise a broader issue with the change. The way this design is supposed to work is that we only do work when all the stars align on the config of quiche. buried in the qlog_with_type! metric is a check of qlog importance and annoyingly the metrics_updated has `core` importance while congestion_state_updated has `base` importance. 

This means if quiche is configured to log only `core` events, the PR will also log congestion_state_updated which is unexpected. Its not a disaster but ideally we'd be able to avoid this problem when adding the new event. 

The easiest thing to do for now might be to put the code block behind an `if cfg qlog` check and move the qlog_with_type check into your modified `maybe_qlog` function; internally it can then call qlog_with_type for the two different event types it wants to log. This approach, of pushing recovery qlogging deeper into the recovery code seems like a cleaner design anyway. 

WDYT?

quiche/src/recovery/congestion/bbr/mod.rs
```diff
@@ -337,6 +349,10 @@ fn has_custom_pacing() -> bool {
     true
 }
 
+fn state_str(r: &Congestion, _now: Instant) -> &'static str {
```
LPardue, line 354: CI is complaining about "method `state_str` is never used", so I wonder if this function needs to go behind `#[cfg(feature = "qlog")]`

quiche/src/lib.rs
antoniovicente: You're right.  I started looking at a fix, but wondering if it is worth doing as a followup PR since there are other qlog changes blocked behind this one.

LPardue: lets take it as follow up work since it was already like this

LPardue (approved):

PR #2073, opened 2025-05-31
Fix inconsistency in BBR2 A0 point selection logic between cloudflare/quiche and google/quiche

antoniovicente:
The A0 point is used in bandwidth computations when BW over-estimation protections are enabled (the default).

Without this change, BW is under estimated in some edge cases.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
+++ b/quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
@@ -455,7 +455,9 @@ impl RecentAckPoints {
     }
 
     fn less_recent_point(&self) -> Option<AckPoint> {
-        self.ack_points[0].or(self.ack_points[1])
+        self.ack_points[0]
+            .filter(|ack_point| ack_point.total_bytes_acked > 0)
+            .or(self.ack_points[1])
     }
 }
 
@@ -789,7 +791,7 @@ impl BandwidthSampler {
 
         while let Some(candidate) = a0_candidates.get(1) {
             if candidate.total_bytes_acked > total_bytes_acked {
-                return Some(*candidate);
+                break;
             }
             a0_candidates.pop_front();
         }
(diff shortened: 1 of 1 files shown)


quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
```diff
@@ -789,7 +791,7 @@ impl BandwidthSampler {
 
         while let Some(candidate) = a0_candidates.get(1) {
             if candidate.total_bytes_acked > total_bytes_acked {
-                return Some(*candidate);
+                break;
```
antoniovicente, line 794: Note that we were returning the `i` th element, instead of `i-1`

https://github.com/google/quiche/blob/66dc426/quiche/quic/core/congestion_control/bandwidth_sampler.cc#L524

quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
```diff
@@ -915,11 +917,11 @@ mod bandwidth_sampler_tests {
...
             );
 
-            let max_bandwidth =
-                self.max_bandwidth.max(sample.sample_max_bandwidth.unwrap());
+            let sample_max_bandwidth = sample.sample_max_bandwidth.unwrap();
+            self.max_bandwidth = self.max_bandwidth.max(sample_max_bandwidth);
 
             let bandwidth_sample = BandwidthSample {
-                bandwidth: max_bandwidth,
+                bandwidth: sample_max_bandwidth,
```
antoniovicente, line 924: The original test does not do `self.max_bandwidth.max` on the bw sample returned by this function in the test harness.  This fixes the `app_limited` test at idx `61`.

https://github.com/google/quiche/blob/66dc426afd733258dd95b6e3307a5833a48e5df4/quiche/quic/core/congestion_control/bandwidth_sampler_test.cc#L102

quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
```diff
@@ -1386,19 +1388,13 @@ mod bandwidth_sampler_tests {
...
-                assert!(
-                    sample.bandwidth < expected_bandwidth * 0.7,
-                    "{} {:?} vs {:?}",
-                    i,
-                    sample.bandwidth,
-                    expected_bandwidth * 0.7
-                );
-            } else {
-                // Needs further investigation: when using overestimate_avoidance,
-                // sample.bandwidth increases 1 packet earlier than expected.
```
antoniovicente, line 1399: The fix to choose_a0_point makes this test work without the hacks added in da5379760

PR #2074, opened 2025-06-02
Handle low PMTU on macOS

cf-rhett:
On macOS with a quiche client in low MTU situations the incoming packet router should be able to gracefully handle packets marked with fragmentation needed

### DIFF
--- a/tokio-quiche/Cargo.toml
+++ b/tokio-quiche/Cargo.toml
@@ -50,7 +50,6 @@ foundations = { workspace = true, default-features = false, features = [
 futures = { workspace = true }
 futures-util = { workspace = true }
 ipnetwork = { workspace = true }
-libc = { workspace = true }
 log = { workspace = true }
 octets = { workspace = true }
 pin-project = { workspace = true }
@@ -81,4 +80,5 @@ serde_json = { workspace = true }
 tokio = { workspace = true, features = ["time", "test-util"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
+libc = { workspace = true }
 nix = { workspace = true, features = ["signal"] }
--- a/tokio-quiche/Cargo.toml
+++ b/tokio-quiche/Cargo.toml
@@ -50,6 +50,7 @@ foundations = { workspace = true, default-features = false, features = [
 futures = { workspace = true }
 futures-util = { workspace = true }
 ipnetwork = { workspace = true }
+libc = { workspace = true }
 log = { workspace = true }
 octets = { workspace = true }
 pin-project = { workspace = true }
@@ -80,5 +81,4 @@ serde_json = { workspace = true }
 tokio = { workspace = true, features = ["time", "test-util"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
-libc = { workspace = true }
 nix = { workspace = true, features = ["signal"] }
--- a/tokio-quiche/src/quic/router/mod.rs
+++ b/tokio-quiche/src/quic/router/mod.rs
@@ -694,7 +694,7 @@ where
 
                 Poll::Ready(Err(e)) => {
                     log::error!("Incoming packet router encountered recvmsg error"; "error" => e);
-                        continue;
+                    continue;
                 },
 
                 Poll::Pending => {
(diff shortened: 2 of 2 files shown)


cf-rhett:
During the process of implementing PMTUD (#2050) we found in low MTU environments on macOS PMTUD cannot complete as the incoming packet router returns early and infinitely restarts PMTUD on each start.

I've limited the change here to be a specific as possible and to otherwise retain the same behavior

tokio-quiche/Cargo.toml
```diff
@@ -50,6 +50,7 @@ foundations = { workspace = true, default-features = false, features = [
 futures = { workspace = true }
 futures-util = { workspace = true }
 ipnetwork = { workspace = true }
+libc = { workspace = true }
```
cf-rhett: I can do some more intense cfg'ing if we don't want to include this for Windows but this allows for the cleanest code (imo)

rushilmehra:
Did you end up determining that this is specific to Mac OS?

LPardue:
I find it suprising, albeit plausible, that an OS would inform an application that there was a datagram larger than the locally configured UDP MTU. I'm more suspicious that this is caused by some bug in the size of buffers we provide to recvmsg causing those buffers to be too small (and therefore something we should deal with). I am happy to be proven wrong though.

This sort of thing would benefit from test cases so that we can gain confidence we understand the root cause, cover OS differences, and can prevent any regressions should something change in the future.

cf-rhett:
Strange that github doesn't allow me to respond in-thread to messages that I don't initiate but anyways.

> I find it surprising, albeit plausible, that an OS would inform an application that there was a datagram larger than the locally configured UDP MTU. 

Specifically this only occurs if its not the locally configured MTU. If I set the MTU on the machine itself, macOS seems to handle this but if I have a router upstream with a lower MTU than I do start to see this.

> I'm more suspicious that this is caused by some bug in the size of buffers we provide to recvmsg causing those buffers to be too small (and therefore something we should deal with). I am happy to be proven wrong though.

I logged the buffers and they seemed to be plenty large enough. I'll get some numbers and report back though.

> This sort of thing would benefit from test cases so that we can gain confidence we understand the root cause, cover OS differences, and can prevent any regressions should something change in the future.

I can add some tests. I wasn't sure yall would take this so wanted to have some discussions before diving in

cf-rhett:
> Did you end up determining that this is specific to Mac OS?

It doesn't occur on Windows but Toks is testing Linux today and I'll come back with an update

LPardue:
Thanks for the clarification. Can you explain more about the case  

> Specifically this only occurs if its not the locally configured MTU. If I set the MTU on the machine itself, macOS seems to handle this but if I have a router upstream with a lower MTU than I do start to see this.

QUIC [requires](https://datatracker.ietf.org/doc/html/rfc9000#section-14-7) that UDP datagrams are not fragmented 

> UDP datagrams MUST NOT be fragmented at the IP layer. In IPv4 [[IPv4](https://datatracker.ietf.org/doc/html/rfc791)], the Don't Fragment (DF) bit MUST be set if possible, to prevent fragmentation on the path.

So I'm not sure what your router is doing and how that relates to the recvmsg call. If DF is properly set, then the router should be dropping it, not passing it through.

cf-rhett:
>So I'm not sure what your router is doing and how that relates to the recvmsg call. If DF is properly set, then the router should be dropping it, not passing it through.

The router is dropping the UDP packet but from my understanding a router that receives a network packet larger than the next hop's MTU with the Don't Fragment (DF) flag bit set in the packet's header should send an ICMP message which indicates the condition Fragmentation Needed (Type 3, Code 4) back to the device.

[wiki-src](https://en.wikipedia.org/wiki/IP_fragmentation#:~:text=drop%20the%20packet%20if%20the%20Don%27t%20Fragment%20%28DF%29%20flag%20bit%20is%20set%20in%20the%20packet%27s%20header%20and%20send%20an%20Internet%20Control%20Message%20Protocol%20%28ICMP%29%20message%20which%20indicates%20the%20condition%20Fragmentation%20Needed%20%28Type%203%2C%20Code%204%29)

So this is more handling this message than QUIC itself although I don't know why this behavior appears to only be on macOS

rushilmehra (approved):

PR #2075, opened 2025-06-03
Calculate the bytes_in_flight active time for QUIC connections

antoniovicente:


quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,144 @@
+use std::time::Duration;
+use std::time::Instant;
+
+#[derive(Default)]
+pub struct BytesInFlight {
+    // Current bytes in flight.
+    bytes_in_flight: usize,
+
+    // Instant at which bytes_in_flight transitioned from 0 to >0.
+    // Set if if bytes_in_flight is currently >0 which indicates that
```
LPardue:
```suggestion
    // Set if bytes_in_flight is currently >0 which indicates that
```

quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,144 @@
...
+    // Sum of closed interval durations seen so far.
+    closed_interval_time: Duration,
+}
+
+impl BytesInFlight {
+    // Add to bytes in flight.  Record the start time when
+    // bytes_in_flight was 0 at the beginning of the function.
+    pub(crate) fn add(&mut self, delta: usize, now: Instant) {
+        self.bytes_in_flight += delta;
+        if delta > 0 {
```
LPardue: nit: the function only does something with an effect if `delta` in non-zero, so consider instead changing this to return early `if delta == 0` which removes a layer of code nesting

quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,144 @@
...
+                self.update_in_flight_times(now);
+            } else {
+                self.bytes_in_flight_interval_start = Some(now);
+            }
+        }
+    }
+
+    // Substract from bytes in flight.  If bytes_in_flight drops to 0,
+    // end the current bytes_in_flight >0 interval.
+    pub(crate) fn subtract(&mut self, delta: usize, now: Instant) {
```
LPardue: Not clear to me why we need this variant when there is saturating_subtract too.

quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.bytes_in_flight = self.bytes_in_flight.saturating_sub(delta);
+        self.update_in_flight_times(now);
+    }
+
+    pub(crate) fn get(&self) -> usize {
+        self.bytes_in_flight
+    }
+
+    // Total time during which bytes_in_flight was > 0.
+    pub(crate) fn get_time(&self) -> Duration {
```
LPardue: nit: since time might be conflated with a timestamp

```suggestion
    pub(crate) fn get_duration(&self) -> Duration {
```

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Possibly to catch code errors, but it is not clear why we needed a saturating version.  Happy to have just a saturating version.

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Done.

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Done.

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Done.  Also changed the stats field from `bytes_in_flight_time_us` to `bytes_in_flight_duration_us`

quiche/src/recovery/bytes_in_flight.rs
LPardue: in my experience, we put in the saturating call if we experienced an underflow panic when running in more exotic  environments. I'm not sure we have a consistent philosophy about our approach, so I'm not going to ligitate the matter if we want to stick with how things currently are. We can always change this later if we wanted.

antoniovicente: Made a commit that replaces `substract` with `saturating_subtract`

quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,140 @@
+use std::time::Duration;
```
LPardue, line 27: nit: needs the copyright boilerplate

quiche/src/recovery/bytes_in_flight.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.bytes_in_flight += delta;
+
+        if self.bytes_in_flight_interval_start.is_some() {
+            self.update_in_flight_duration(now);
+        } else {
+            self.bytes_in_flight_interval_start = Some(now);
+        }
+    }
+
+    // Substract from bytes in flight.  If bytes_in_flight drops to 0,
```
LPardue: nit: I think you need `///` for public comments

LPardue (dismissed):
LGTM modulo some nits

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Done.

quiche/src/recovery/bytes_in_flight.rs
antoniovicente: Done.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -477,7 +479,7 @@ impl GRecovery {
         let mut duration = self.pto() * (1 << self.pto_count);
 
         // Arm PTO from now when there are no inflight packets.
-        if self.bytes_in_flight == 0 {
+        if self.bytes_in_flight.get() == 0 {
```
jrouviere: super nit: maybe a `is_zero()` would be useful here

quiche/src/lib.rs
```diff
@@ -8570,6 +8577,9 @@ pub struct Stats {
 
     /// The total number of PATH_CHALLENGE frames that were received.
     pub path_challenge_rx_count: u64,
+
+    /// Total time during which bytes_in_flight > 0.
```
jrouviere: nit: maybe that's obvious but I would specify in the doc that the "bytes in flight" are bytes we are sending. So I guess "unacked sent bytes" is another way to describe it.

quiche/src/lib.rs
```diff
@@ -7113,9 +7113,16 @@ impl<F: BufFactory> Connection<F> {
             reset_stream_count_remote: self.reset_stream_remote_count,
             stopped_stream_count_remote: self.stopped_stream_remote_count,
             path_challenge_rx_count: self.path_challenge_rx_count,
+            bytes_in_flight_duration: self.bytes_in_flight_duration(),
         }
     }
 
+    fn bytes_in_flight_duration(&self) -> Duration {
+        self.paths.iter().fold(Duration::ZERO, |acc, (_, path)| {
+            acc + path.bytes_in_flight_duration()
```
jrouviere, line 7130: Seems like adding up the duration over multiple paths will give a confusing value here.

For instance if I have a connection that last 10 sec, and it used two paths with bytes in flight for 8 sec each. Does that mean the connection `bytes_in_flight_duration` is 16 sec?

Disclaimer: I'm not sure I understand how multipath work.

antoniovicente:
> LGTM assuming we were already incrementing/decrementing `bytes_in_flight` everywhere we should, which is hard to verify.

That was part of the reason behind me introducing a new type: by adding a new type I forced us to fine all places where the old type was updated and switched them to the new API which computes the inflight active time and the number of inflight bytes together.

quiche/src/lib.rs
antoniovicente: Added a comment with a TODO for us to revisit the computation when multiple paths are in more active use.  My understanding is that only 1 path is used right now.

quiche/src/lib.rs
antoniovicente: Attempted to clarify.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: Done.

PR #2076, opened 2025-06-04
lib: add support for qlogging packet lost events

LPardue:
To date, we've focused on reporting packet loss counts as aggregate metrics.

This change adds support for the qlog packet_lost event. We already log packet_sent events and include the frames in that event, so we can reduce overheads by omitting them from the packet_lost events.

The tricky part of this change is wiring up logging so that it can access the data it needs. Rather than modify the recovery module to change it from passing a count to a list of packets, I opted instead to conditionally store lost packets based on the qlog feature, and then provide an accessor for reading them at key points in the connection lifecycle (in the receive loop, or after a timeout). By storing the loss Instant, we ensure the event is logged with a timestamp reflective of the actual time quiche determined the loss.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1920,6 +1920,10 @@ const QLOG_METRICS: EventType =
     EventType::RecoveryEventType(RecoveryEventType::MetricsUpdated);
 
 #[cfg(feature = "qlog")]
+const QLOG_PACKET_LOST: EventType =
+    EventType::RecoveryEventType(RecoveryEventType::PacketLost);
+
+#[cfg(feature = "qlog")]
 const QLOG_CONNECTION_CLOSED: EventType =
     EventType::ConnectivityEventType(ConnectivityEventType::ConnectionClosed);
 
@@ -3364,6 +3368,20 @@ impl<F: BufFactory> Connection<F> {
         qlog_with_type!(QLOG_PACKET_RX, self.qlog, q, {
             let recv_path = self.paths.get_mut(recv_pid)?;
             recv_path.recovery.maybe_qlog(q, now);
+        });
+
+        qlog_with_type!(QLOG_PACKET_LOST, self.qlog, q, {
+            let recv_path = self.paths.get_mut(recv_pid)?;
+            for (header, time_lost) in recv_path.recovery.get_lost_pkts(epoch) {
+                let ev_data =
+                    EventData::PacketLost(qlog::events::quic::PacketLost {
+                        header: Some(header),
+                        frames: None,
+                        trigger: None,
+                    });
+
+                q.add_event_data_with_instant(ev_data, time_lost).ok();
+            }
         });
 
         if let Some(e) = frame_processing_err {
@@ -6474,6 +6492,29 @@ impl<F: BufFactory> Connection<F> {
 
                     qlog_with_type!(QLOG_METRICS, self.qlog, q, {
                         p.recovery.maybe_qlog(q, now);
+                    });
+
+                    qlog_with_type!(QLOG_PACKET_LOST, self.qlog, q, {
+                        for epoch in [
+                            packet::Epoch::Initial,
+                            packet::Epoch::Handshake,
+                            packet::Epoch::Application,
+                        ] {
+                            for (header, time_lost) in
+                                p.recovery.get_lost_pkts(epoch)
+                            {
+                                let ev_data = EventData::PacketLost(
+                                    qlog::events::quic::PacketLost {
+                                        header: Some(header),
+                                        frames: None,
+                                        trigger: None,
+                                    },
+                                );
+
+                                q.add_event_data_with_instant(ev_data, time_lost)
+                                    .ok();
+                            }
+                        }
                     });
                 }
             }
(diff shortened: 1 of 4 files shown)


quiche/src/recovery/congestion/recovery.rs
```diff
@@ -246,6 +252,28 @@ impl RecoveryEpoch {
 
                     self.in_flight_count -= 1;
 
+                    #[cfg(feature = "qlog")]
+                    {
+                        // TODO: can we get the actual real packet type?
```
LPardue, line 279: As the comment suggests, this seems kludgy. It seems that we don't story the actual packet type anywhere I could read it and I didn't want to make too large a change. But maybe its really required.

The type is a required field in the qlog PacketHeader object defined in https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-8.8

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -245,7 +251,7 @@ impl RecoveryEpoch {
 
     fn detect_and_remove_lost_packets(
         &mut self, loss_delay: Duration, pkt_thresh: u64, now: Instant,
-        newly_lost: &mut Vec<Lost>,
+        newly_lost: &mut Vec<Lost>, epoch: packet::Epoch,
```
LPardue, line 268: Note the new `epoch` field is only necessary for the qlogging, and so generates a CI error when the `qlog` feature is not enabled. There's ways to workaround that but it suggests we shouldn't be using `epoch` to determine packet type.

LPardue:
@antoniovicente this PR will end up conflicting with https://github.com/cloudflare/quiche/pull/2070 but I wanted to keep the proposal as simple as possible to highlight the issues with storing/passing lost packet details. It should be easy enough to land this after your changes once we agree the important parts.

quiche/src/lib.rs
```diff
@@ -6475,6 +6493,29 @@ impl<F: BufFactory> Connection<F> {
...
+                            packet::Epoch::Application,
+                        ] {
+                            for (header, time_lost) in
+                                p.recovery.get_lost_pkts(epoch)
+                            {
+                                let ev_data = EventData::PacketLost(
+                                    qlog::events::quic::PacketLost {
+                                        header: Some(header),
+                                        frames: None,
+                                        trigger: None,
```
antoniovicente, line 6510: It seems that we could claim: `PacketLostTrigger::PtoExpired`

quiche/src/lib.rs
```diff
@@ -3366,6 +3370,20 @@ impl<F: BufFactory> Connection<F> {
...
         });
 
+        qlog_with_type!(QLOG_PACKET_LOST, self.qlog, q, {
+            let recv_path = self.paths.get_mut(recv_pid)?;
+            for (header, time_lost) in recv_path.recovery.get_lost_pkts(epoch) {
+                let ev_data =
+                    EventData::PacketLost(qlog::events::quic::PacketLost {
+                        header: Some(header),
+                        frames: None,
+                        trigger: None,
```
antoniovicente, line 3380: Is it possible to determine the loss detection trigger?  I'm guessing this one is `ReorderingThreshold`

quiche/src/lib.rs
```diff
@@ -6475,6 +6493,29 @@ impl<F: BufFactory> Connection<F> {
...
                     });
+
+                    qlog_with_type!(QLOG_PACKET_LOST, self.qlog, q, {
+                        for epoch in [
+                            packet::Epoch::Initial,
+                            packet::Epoch::Handshake,
+                            packet::Epoch::Application,
+                        ] {
+                            for (header, time_lost) in
+                                p.recovery.get_lost_pkts(epoch)
```
antoniovicente, line 6504: One possible worry is that in cases where qlog is enabled but QLOG_PACKET_LOST isn't, get_lost_pkts won't be called, so we'll accumulate a potentially large number of lost packets in this array until `on_pkt_num_space_discarded` is called at a future date.

quiche/src/recovery/congestion/recovery.rs
antoniovicente: Makes sense.  We can always try to refine this info further in the future.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: Any way to annotate to avoid the error?  _epoch?

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -588,6 +618,13 @@ impl RecoveryOps for GRecovery {
         std::mem::take(&mut self.epochs[epoch].lost_frames)
     }
 
+    #[cfg(feature = "qlog")]
+    fn get_lost_pkts(
+        &mut self, epoch: packet::Epoch,
+    ) -> Vec<(events::quic::PacketHeader, Instant)> {
+        std::mem::take(&mut self.epochs[epoch].lost_pkts)
```
antoniovicente, line 625: No code that clears this array in cases where get_lost_pkts is never called!

antoniovicente: I think it belongs in the `fn discard` from the RecoveryEpoch impl

PR #2077, opened 2025-06-05
 Unconditionally call process_reads() in RunningApplication phase

evanrittenhouse:
If a server ACKs a packet containing a HEADERS frame in the same flight that the handshake completes, the Handshake IOW must have taken all packets out of the context's `incoming_pkt_receiver`.

That means that the first iteration of the `RunningApplication` work loop will no-op - since `did_recv` is false, that iteration won't call `h3::Connection::poll()`. We therefore unconditionally call `on_read()` before starting the `RunningApplication` work loop to prevent this case.

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -765,6 +765,19 @@ where
     pub(crate) async fn run<A: ApplicationOverQuic>(
         mut self, mut qconn: QuicheConnection, mut ctx: ConnectionStageContext<A>,
     ) -> Closing<Tx, M, A> {
+        // Perform a single call to process_reads()/process_writes(),
+        // unconditionally, to ensure that any application data (e.g.
+        // STREAM frames or datagrams) processed by the Handshake
+        // stage are properly passed to the application.
+        if let Err(e) = self.conn_stage.on_read(true, &mut qconn, &mut ctx) {
+            return Closing {
+                params: self.into(),
+                context: ctx,
+                work_loop_result: Err(e),
+                qconn,
+            };
+        };
+
         let work_loop_result = self.work_loop(&mut qconn, &mut ctx).await;
 
         Closing {


rushilmehra (approved):

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -765,6 +765,17 @@ where
     pub(crate) async fn run<A: ApplicationOverQuic>(
         mut self, mut qconn: QuicheConnection, mut ctx: ConnectionStageContext<A>,
     ) -> Closing<Tx, M, A> {
+        // Perform a single call to process_reads()/process_writes(), unconditionally, to ensure
+        // that any HTTP/3 frames processed by the Handshake stage are passed to the application.
```
LPardue: Do we need to call out HTTP/3 specifically here? Since this is `A: ApplicationOverQuic` should we say instead something like "ensure that any application data (Streams or datagrams) processed before the Handshake stage can be read by the application"

LPardue: cargo fmt is also failing on this comment so you'll need to change it either way :joy:

toidiu:
Here is a screenshot of the before and after comparing the delivery rate. Since processing was delayed until we received an ACK from the peer, the expectation is to start sending data approx 1 RTT earlier.

In both runs the handshake completes at the same time, which means both connections should have a fair chance at sending data over time. The test were run on linux, with fq enabled, rate limiting 10mbps, and 20ms RTT delay.

**before**
screen_shot_pause

**after**
screen_shot_no_pause

toidiu (approved):

tokio-quiche/src/quic/io/worker.rs
evanrittenhouse: Changed,  linters ruin all the fun /s

toidiu (approved):

PR #2078, opened 2025-06-05
Emit RTO count statistics to estimate loss events

toidiu:
Loss usually happens in burst so the number of packets lost will depend on the volume of inflight packets at the time of loss (which can be arbitrary). RTO count measures the number of loss events and provide a normalized loss metric.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -6948,7 +6948,7 @@ impl<F: BufFactory> Connection<F> {
     /// Note that the value returned can change throughout the connection's
     /// lifetime.
     #[inline]
-    pub fn source_id(&self) -> ConnectionId {
+    pub fn source_id(&self) -> ConnectionId<'_> {
         if let Ok(path) = self.paths.get_active() {
             if let Some(active_scid_seq) = path.active_scid_seq {
                 if let Ok(e) = self.ids.get_scid(active_scid_seq) {
@@ -6966,7 +6966,7 @@ impl<F: BufFactory> Connection<F> {
     /// An iterator is returned for all active IDs (i.e. ones that have not
     /// been explicitly retired yet).
     #[inline]
-    pub fn source_ids(&self) -> impl Iterator<Item = &ConnectionId> {
+    pub fn source_ids(&self) -> impl Iterator<Item = &ConnectionId<'_>> {
         self.ids.scids_iter()
     }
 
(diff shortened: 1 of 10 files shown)


quiche/src/lib.rs
```diff
@@ -6433,6 +6433,9 @@ impl<F: BufFactory> Connection<F> {
                     self.lost_count += lost_packets;
                     self.lost_bytes += lost_bytes as u64;
 
+                    // Track RTO timeout event
+                    p.rto_count += 1;
```
LPardue: would it be better to increment this inside on_loss_detection_timeout() itself?

quiche/src/lib.rs
```diff
@@ -6433,6 +6433,9 @@ impl<F: BufFactory> Connection<F> {
                     self.lost_count += lost_packets;
                     self.lost_bytes += lost_bytes as u64;
 
+                    // Track PTO timeout event
+                    p.total_pto_count += 1;
```
LPardue: I think my last comment got nuked by a change. So just in case - would it make sense to put the increment inside `p.on_loss_detection_timeout(` itself?

toidiu: Yep good suggestion.

quiche/src/path.rs
```diff
@@ -170,6 +170,14 @@ pub struct Path {
     /// Total number of packets sent with data retransmitted from this path.
     pub retrans_count: usize,
 
+    /// Total number of times PTO (probe timeout) fired.
+    ///
+    /// Loss usually happens in burst so the number of packets lost will depend
+    /// on the volume of inflight packets at the time of loss (which can be
+    /// arbitrary). PTO count measures the number of loss events and provide a
```
LPardue:
```suggestion
    /// arbitrary). PTO count measures the number of loss events and provides a
```

quiche/src/path.rs
```diff
@@ -170,6 +170,14 @@ pub struct Path {
     /// Total number of packets sent with data retransmitted from this path.
     pub retrans_count: usize,
 
+    /// Total number of times PTO (probe timeout) fired.
+    ///
+    /// Loss usually happens in burst so the number of packets lost will depend
```
LPardue:
```suggestion
    /// Loss usually happens in a burst so the number of packets lost will depend
```

or `bursts`

quiche/src/path.rs
```diff
@@ -894,6 +907,14 @@ pub struct PathStats {
     /// The number of sent QUIC packets with retransmitted data.
     pub retrans: usize,
 
+    /// The number of times PTO (probe timeout) fired.
+    ///
+    /// Loss usually happens in burst so the number of packets lost will depend
+    /// on the volume of inflight packets at the time of loss (which can be
+    /// arbitrary). PTO count measures the number of loss events and provide a
+    /// normalized loss metric.
```
LPardue: same nits apply here :smile:

LPardue (dismissed):
LGTM modulo some nits

LPardue (approved):

PR #2079, opened 2025-06-10
ci: specify nightly ci jobs

toidiu:


evanrittenhouse (approved):

.github/workflows/nightly.yml
```diff
@@ -74,7 +70,6 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUSTTOOLCHAIN }}
-          components: rustfmt
```
evanrittenhouse, line 77: Should we call the `fuzz` job `fuzz_nightly` as well?

toidiu: fuzz requires nightly to work so there is no stable equivalent. I think we avoid the suffix for fuzz since nightly is simply an implementation detail.

Btw the goal of this will to make the nightly jobs as not required for PRs since new clippy warnings can block merges

LPardue: Comedy timing https://blog.rust-lang.org/inside-rust/2025/06/21/announcing-the-clippy-feature-freeze/

PR #2081, opened 2025-06-15
Option to disable BBR code that attempts to avoid bandwidth overestimation

antoniovicente:


antoniovicente:
> Can we add some details into why this would help in the commit message?

Added an explanation.  Also note that this is just exposing a config knob that we can use to tweak the behavior of the library from the outside.

PR #2082, opened 2025-06-15
Config protect fixes to a0 choice used in BW overestimation avoidance code

antoniovicente:
This code doesn't seem to be working well, and we would like to compare results with and without the fix.  It seems that overestimation avoidance is a work in progress and we should recommend disabling the code for now.

Original change: https://github.com/cloudflare/quiche/pull/2073

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
+++ b/quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
@@ -123,6 +123,10 @@ pub struct BandwidthSampler {
     /// sampler to exit the app-limited phase.
     end_of_app_limited_phase: Option<u64>,
     overestimate_avoidance: bool,
+    // If true, apply the fix to A0 point selection logic so the
+    // implementation is consistent with the behavior of the
+    // google/quiche implementation.
+    choose_a0_point_fix: bool,
     limit_max_ack_height_tracker_by_send_rate: bool,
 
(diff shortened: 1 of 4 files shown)


antoniovicente:
In draft since this builds on top of https://github.com/cloudflare/quiche/pull/2081

quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs
```diff
@@ -784,14 +798,19 @@ impl BandwidthSampler {
...
+        choose_a0_point_fix: bool,
     ) -> Option<AckPoint> {
         if a0_candidates.is_empty() {
             return None;
         }
 
         while let Some(candidate) = a0_candidates.get(1) {
             if candidate.total_bytes_acked > total_bytes_acked {
-                break;
+                if choose_a0_point_fix {
```
evanrittenhouse, line 809: Is it worth setting two config options, one for here and one for the `> 0` filter change, or not really? Asking out of curiosity, IIRC this one was safer than the other. Not sure if it'd narrow the scope of debugging in any meaningful way

antoniovicente: That was my original plan, but then I went down the route of `enable_overestimate_avoidance = false` from https://github.com/cloudflare/quiche/pull/2081 which I expect will do better; the overestimation avoidance is kinda buggy.  When that option is false, the a0 choice code is unreachable.

PR #2083, opened 2025-06-15
Clean up PMTUD related variables + APIs, add unit tests to pmtud.rs

rushilmehra:
This makes some of the internal state variables in pmtud.rs more
intuitive, and in general makes the code easier to reason about.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3375,7 +3375,7 @@ impl<F: BufFactory> Connection<F> {
         }
 
         // Following flag used to upgrade datagram size, if probe is successful.
-        let mut pmtud_probe = false;
+        let mut successful_pmtud_probe = false;
 
         // Process acked frames. Note that several packets from several paths
         // might have been acked by the received packet.
@@ -3385,17 +3385,17 @@ impl<F: BufFactory> Connection<F> {
                     frame::Frame::Ping {
                         mtu_probe: Some(mtu_probe),
                     } => {
-                        let pmtud_next = p.pmtud.get_current();
-                        p.pmtud.set_current(cmp::max(pmtud_next, mtu_probe));
+                        let current_mtu = p.pmtud.get_current_mtu();
+                        p.pmtud.set_current_mtu(cmp::max(current_mtu, mtu_probe));
 
                         // Stop sending path MTU probes after successful probe.
-                        p.pmtud.should_probe(false);
-                        pmtud_probe = true;
+                        p.pmtud.set_should_probe(false);
+                        successful_pmtud_probe = true;
 
                         trace!(
-                            "{} pmtud acked; pmtu size {:?}",
+                            "{} pmtud acked; current mtu: {:?}",
                             self.trace_id,
-                            p.pmtud.get_current()
+                            p.pmtud.get_current_mtu()
                         );
                     },
 
@@ -3487,10 +3487,10 @@ impl<F: BufFactory> Connection<F> {
             }
 
             // Update max datagram send size with newly acked probe size.
-            if pmtud_probe {
+            if successful_pmtud_probe {
                 trace!(
-                    "{} updating pmtu {:?}",
-                    p.pmtud.get_current(),
+                    "{} updating pmtu to {:?}",
+                    p.pmtud.get_current_mtu(),
                     self.trace_id
                 );
 
@@ -3504,8 +3504,8 @@ impl<F: BufFactory> Connection<F> {
                         let pmtu_data = EventData::MtuUpdated(
                             qlog::events::connectivity::MtuUpdated {
                                 old: Some(p.recovery.max_datagram_size() as u16),
-                                new: p.pmtud.get_current() as u16,
-                                done: Some(pmtud_probe),
+                                new: p.pmtud.get_current_mtu() as u16,
+                                done: Some(true),
                             },
                         );
 
@@ -3514,7 +3514,7 @@ impl<F: BufFactory> Connection<F> {
                 );
 
                 p.recovery
-                    .pmtud_update_max_datagram_size(p.pmtud.get_current());
+                    .pmtud_update_max_datagram_size(p.pmtud.get_current_mtu());
             }
         }
 
@@ -3803,11 +3803,11 @@ impl<F: BufFactory> Connection<F> {
         let send_path = self.paths.get_mut(send_pid)?;
 
         // Update max datagram size to allow path MTU discovery probe to be sent.
-        if send_path.pmtud.get_probe_status() {
+        if send_path.pmtud.get_should_probe() {
             let size = if self.handshake_confirmed || self.handshake_done_sent {
                 send_path.pmtud.get_probe_size()
             } else {
-                send_path.pmtud.get_current()
+                send_path.pmtud.get_current_mtu()
             };
 
             send_path.recovery.pmtud_update_max_datagram_size(size);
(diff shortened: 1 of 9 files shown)


quiche/src/pmtud.rs
```diff
@@ -1,92 +1,193 @@
...
+    enabled: bool,
 }
 
 impl Pmtud {
     /// Creates new PMTUD instance.
-    pub fn new(cur_size: usize) -> Self {
+    pub fn new(initial_mtu: usize) -> Self {
         Self {
-            cur_size,
+            current_mtu: initial_mtu,
```
antoniovicente, line 25: We should think about the invariants that this code should have.  Some possible examples:

if `!enabled`, then `!should_probe`.  Also known as `assert!(enabled || !should_probe)`.

If `should_probe`, `current_mtu <= probe_size`.  Or even `current_mtu < probe_size` since probing only makes sense if the probe size is larger than the MTU in use. This one makes me think that we should initialize `probe_size = initial_mtu` or have the setter for `should_probe` take `probe_size` as an argument: `disable_probe()` and `schedule_probe(size)`

And `initial_mtu >= 1200` since that's required by QUIC.

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,193 @@
...
 
     /// Specifies whether Path MTU Discovery should be performed at the next
     /// opportunity, i.e., when the next packet is sent out if possible.
     ///
-    /// Once Path MTU has been discovered, this maybe set to false.
-    pub fn should_probe(&mut self, pmtu_next: bool) {
-        self.next_size = pmtu_next;
+    /// Once Path MTU has been discovered, this may be set to false.
+    pub fn set_should_probe(&mut self, pmtu_next: bool) {
+        self.should_probe = pmtu_next;
```
antoniovicente: `pmtu_next` argument name seems incorrect, should be `should_probe`

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,193 @@
...
+
+        // Simulate probe loss - should update to midpoint
+        pmtud.update_probe_size();
+        // Expected: 1200 + ((1500 - 1200) / 2) = 1200 + 150 = 1350
+        assert_eq!(pmtud.get_probe_size(), 1350);
+
+        // Another probe loss
+        pmtud.update_probe_size();
+        // Expected: 1200 + ((1350 - 1200) / 2) = 1200 + 75 = 1275
+        assert_eq!(pmtud.get_probe_size(), 1275);
```
antoniovicente, line 155: It would be good to cover the `1201` to `1200` transition in a test.

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,193 @@
...
+        let mut pmtud = Pmtud::new(1200);
+        pmtud.set_probe_size(1500);
+        pmtud.set_should_probe(false);
+
+        // Simulate probe loss
+        pmtud.pmtu_probe_lost();
+
+        // Should re-enable probing and adjust size
+        assert!(pmtud.get_should_probe());
+        assert_eq!(pmtud.get_probe_size(), 1350); // binary search result
```
antoniovicente, line 197: Also check current mtu

I think it will be: `assert_eq!(pmtud.get_current_mtu(), 1200);`

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,193 @@
...
+        pmtud.set_probe_size(1400);
+
+        // Simulate successful probe
+        pmtud.set_current_mtu(1400);
+
+        assert_eq!(pmtud.get_current_mtu(), 1400);
+    }
+
+    #[test]
+    fn pmtud_binary_search_convergence() {
```
antoniovicente, line 213: It would be good to write a test that finds an MTU of arbitrary size.

For example: initial 1200, probe 1500, actual MTU 1300.
Progression:
probe 1500 -> fail
probe 1350 -> fail
probe 1275 -> success
probe 1312 -> fail
probe 1293 -> success
probe 1302 -> fail
probe 1297 -> success
probe 1299 -> success
probe 1300 -> success
probe 1301 -> fail

quiche/src/lib.rs
```diff
@@ -19100,45 +19096,23 @@ mod tests {
...
-        let new_path_size = pipe
-            .server
-            .paths
-            .get_active_mut()
-            .unwrap()
-            .pmtud
-            .get_current();
-        assert_eq!(new_path_size, 1350);
+        let active_path = pipe.server.paths.get_active_mut().unwrap();
+        assert_eq!(active_path.pmtud.get_current_mtu(), 1350);
```
antoniovicente, line 19110: Consider adding {}'s around this block for consistency.

rushilmehra: Done

quiche/src/pmtud.rs
rushilmehra: Thanks, fixed

antoniovicente (approved):

quiche/src/pmtud.rs
```diff
@@ -1,92 +1,191 @@
...
-        self.probe
+        self.probe_size
     }
 
-    /// Sets the current Path MTU Discovery size after a successful probe has
-    /// been performed.
-    pub fn set_current(&mut self, pmtu: usize) {
-        self.cur_size = pmtu;
+    /// Sets the current discovered PMTU after a successful probe has
+    /// been ACKed.
```
LPardue: nit: This is a tautology, the probe is only successful when it has been acked

The old comment seems more correct to me

quiche/src/lib.rs
```diff
@@ -19100,45 +19096,25 @@ mod tests {
...
-
-        let current_path_size = pipe
-            .server
-            .paths
-            .get_active_mut()
-            .unwrap()
-            .pmtud
-            .get_current();
-        assert_eq!(current_path_size, 1200);
+        {
```
LPardue: Do we really need these scopings?

rushilmehra: Removed, thanks

quiche/src/pmtud.rs
rushilmehra: Yeah makes sense - restored the previous terminology

quiche/src/pmtud.rs
rushilmehra: Quiche currently implements unidirectional binary search for PMTU. Once we successfully probe a specific size, we consider it the best we can do and stop probing. The RFC calls for periodic attempts to raise the MTU, similar to the algorithm you're outlining. Once we support that, I'll add the test you're suggesting.

antoniovicente: I don't think I follow.  Is the intent in the current implementation that we send out several size probes at the same time and set the MTU to the highest successful one?

I think we're trying to improve how the code handles a single probe size, no binary searching.

antoniovicente (dismissed):

quiche/src/pmtud.rs
rushilmehra: I agree with all of your points, though I think those will be best suited for a followup PR. I added the assertion that `initial_mtu >= 1200` for now

quiche/src/pmtud.rs
rushilmehra: Done

quiche/src/pmtud.rs
rushilmehra: Done

antoniovicente (approved):
Thanks

quiche/src/pmtud.rs
rushilmehra: As far as I understand, the current algorithm in quiche only sends a single probe at a time, starting at some maximum value (`max_send_udp_payload_size`). As soon as a probe is successful, PMTUD terminates. If a probe is lost, we reduce probe size and try again. We never increase probe size after success

PR #2084, opened 2025-06-15
Ensure 1 probe per rtt, fix client-side PMTUD, add better integration tests

rushilmehra:
This commit fixes two issues:

1. Fixes client-side PMTUD by ensuring probes are actually enabled after the handshake. The code was using `handshake_done_sent`, which is specific to the server. Credit for this fix goes to Rhett, as he discoverd this in https://github.com/cloudflare/quiche/pull/2050, but I needed to pull it in here to write tests for the second issue.

2. Ensures 1 probe per rtt by calling `set_should_probe(false)` after a probe has been sent. Previously, we would send a probe for the same size on every single iteration of quiche::send() so long as the congestion window allowed for it. This is bad for obvious reasons since it 1) wastes bandwidth and 2) blocks application data from being sent. If the flurry of initial probes results in ACKs because the PMTU is large enough, then we will happily adjust our packet size and disable PMTUD. However, if the packets get lost because the PMTU is too small, we'll call `pmtud_probe_lost()` for every single probe that we sent, even though they were all the same size. As a result, we will call `update_probe_size()` multiple times, which in some cases effectively reduces our probe size all the way down to 1200 in a single fight. This basically renders PMTUD useless whenever the initial probe size is too high for the PMTU.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3820,7 +3820,7 @@ impl<F: BufFactory> Connection<F> {
 
         // Update max datagram size to allow path MTU discovery probe to be sent.
         if send_path.pmtud.get_should_probe() {
-            let size = if self.handshake_confirmed || self.handshake_done_sent {
+            let size = if self.handshake_confirmed || self.handshake_completed {
                 send_path.pmtud.get_probe_size()
             } else {
                 send_path.pmtud.get_current_mtu()
@@ -4253,7 +4253,7 @@ impl<F: BufFactory> Connection<F> {
             // (e.g. due to the anti-amplification limits).
             let should_probe_pmtu = active_path.should_send_pmtu_probe(
                 self.handshake_confirmed,
-                self.handshake_done_sent,
+                self.handshake_completed,
                 out_len,
                 is_closing,
                 frames.is_empty(),
@@ -4302,6 +4302,9 @@ impl<F: BufFactory> Connection<F> {
                     }
                 }
 
+                // Reset probe flag after sending to prevent duplicate probes in a
+                // single flight.
+                active_path.pmtud.set_should_probe(false);
                 is_pmtud_probe = true;
             }
 
@@ -19331,60 +19334,45 @@ mod tests {
     }
 
     #[rstest]
-    fn successful_probe_pmtud(
+    fn pmtud_probe_success(
         #[values("cubic", "bbr2", "bbr2_gcongestion")] cc_algorithm_name: &str,
     ) {
         let mut config = Config::new(crate::PROTOCOL_VERSION).unwrap();
-        assert_eq!(config.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
+        config.set_cc_algorithm_name(cc_algorithm_name).unwrap();
         config
             .load_cert_chain_from_pem_file("examples/cert.crt")
             .unwrap();
         config
             .load_priv_key_from_pem_file("examples/cert.key")
             .unwrap();
-        config
-            .set_application_protos(&[b"proto1", b"proto2"])
-            .unwrap();
+        config.set_application_protos(&[b"proto1"]).unwrap();
         config.verify_peer(false);
-        config.set_initial_max_data(100000);
-        config.set_initial_max_stream_data_bidi_local(100000);
-        config.set_initial_max_stream_data_bidi_remote(100000);
-        config.set_initial_max_streams_bidi(2);
-        config.set_active_connection_id_limit(4);
-        config.set_max_send_udp_payload_size(1350);
-        config.set_max_recv_udp_payload_size(1350);
+        config.set_max_send_udp_payload_size(1400);
         config.discover_pmtu(true);
 
-        // Perform initial handshake.
         let mut pipe = testing::Pipe::with_config(&mut config).unwrap();
         assert_eq!(pipe.handshake(), Ok(()));
 
-        let server_addr = testing::Pipe::server_addr();
-        let client_addr = testing::Pipe::client_addr();
-        let pid_1 = pipe
-            .server
-            .paths
-            .path_id_from_addrs(&(server_addr, client_addr))
-            .expect("no such path");
-
-        // Check that PMTU params are configured correctly
-        let pmtu_param = &mut pipe.server.paths.get_mut(pid_1).unwrap().pmtud;
-        assert!(pmtu_param.get_should_probe());
-        assert_eq!(pmtu_param.get_probe_size(), 1350);
+        // Send probe and let it be acknowledged
         assert_eq!(pipe.advance(), Ok(()));
 
-        for (_, p) in pipe.server.paths.iter_mut() {
-            assert_eq!(p.pmtud.get_current_mtu(), 1350);
-            assert!(!p.pmtud.get_should_probe());
-        }
+        // Verify probing is disabled after successful probe
+        let active_path = pipe.client.paths.get_active_mut().unwrap();
+        assert!(!active_path.pmtud.get_should_probe());
+
+        // Verify MTU was updated
+        let current_mtu = active_path.pmtud.get_current_mtu();
+        assert_eq!(current_mtu, 1400);
     }
 
     #[rstest]
-    fn pmtud_probe_loss(
+    /// This test verifies that multiple send() calls after handshake completion
+    /// only generate one PMTUD probe packet, not multiple identical probes.
+    fn pmtud_no_duplicate_probes(
         #[values("cubic", "bbr2", "bbr2_gcongestion")] cc_algorithm_name: &str,
     ) {
         let mut config = Config::new(crate::PROTOCOL_VERSION).unwrap();
-        assert_eq!(config.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
+        config.set_cc_algorithm_name(cc_algorithm_name).unwrap();
         config
             .load_cert_chain_from_pem_file("examples/cert.crt")
             .unwrap();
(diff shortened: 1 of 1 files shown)


rushilmehra:
FYI @cf-rhett

quiche/src/lib.rs
```diff
@@ -18941,93 +18978,145 @@ mod tests {
...
+            frames.append(
+                testing::decode_pkt(&mut pipe.server, &mut buf[..len])
+                    .unwrap()
+                    .as_mut(),
+            );
         }
+
+        assert!(matches!(frames[0], frame::Frame::ACK { .. }));
+        assert!(matches!(frames[1], frame::Frame::Padding { .. }));
+
```
LPardue, line 19414: shouldn't the MTU probe contain a PING too?

quiche/src/lib.rs
```diff
@@ -18941,93 +18978,145 @@ mod tests {
...
+        let mut buf = [0; 1400];
+        assert_eq!(pipe.client.send(&mut buf).unwrap_err(), Error::Done);
+
+        // Verify probe flag was reset after sending
+        let active_path = pipe.client.paths.get_active_mut().unwrap();
+        assert!(!active_path.pmtud.get_should_probe());
```
LPardue, line 19420: To match the logic, it seems this assert should happen after the first send, not the second

quiche/src/lib.rs
```diff
@@ -18941,93 +18978,145 @@ mod tests {
...
+
+        let mut pipe = testing::Pipe::with_config(&mut config).unwrap();
+        assert_eq!(pipe.handshake(), Ok(()));
+
+        let active_path = pipe.client.paths.get_active_mut().unwrap();
+        assert!(active_path.pmtud.get_should_probe());
+
+        // Send probe packet
+        let mut out = [0; 4096];
+        while let Ok(_) = pipe.client.send(&mut out) {}
```
LPardue: Do we really need this loop here and elsewhere? I think it would be better to keep explicit line-by-line checks on send() return value in case something else changes in future that invalidates expectations

cf-rhett:
Ahh looking at the PR and this definitely explains some of the issues I was seeing. Thanks for the update! I need to rebase my branch on this and remove the probe tracking I was doing to see if I can remove that aspect from pmtud.rs since it's now handled correctly here.

quiche/src/lib.rs
```diff
@@ -18940,93 +18977,145 @@ mod tests {
...
+            let (len, _) = pipe.client.send(&mut buf).unwrap();
+            frames.append(
+                testing::decode_pkt(&mut pipe.server, &mut buf[..len])
+                    .unwrap()
+                    .as_mut(),
+            );
         }
+
+        assert!(matches!(frames[0], frame::Frame::ACK { .. }));
+        assert!(matches!(frames[1], frame::Frame::Padding { .. }));
```
antoniovicente, line 19412: Would be good to verify the full contents of the frames array.  Either by also checking `assert_eq!(frames.len(), 2)` or doing a comparison on the full frames Vec`assert!(matches!(frames, {ACK, Padding}));`

quiche/src/lib.rs
antoniovicente: The inclusion of an ACK frame in the MTU probe seems suspicious.  The MTU probe could be lost when MTU is not big enough.  ACKs are important to moving the protocol forward, so they shouldn't be in a packet that has a higher probability of being lost.

quiche/src/lib.rs
```diff
@@ -18940,93 +18977,145 @@ mod tests {
...
+        assert!(!active_path.pmtud.get_should_probe());
+
+        // Simulate probe loss
+        active_path.pmtud.pmtu_probe_lost();
+
+        // Verify probe flag is re-enabled and size is reduced
+        assert!(active_path.pmtud.get_should_probe());
+        let second_probe_size = active_path.pmtud.get_probe_size();
+        assert!(second_probe_size < initial_probe_size);
+        assert!(second_probe_size > 1200);
```
antoniovicente: We could do an exact comparison here.  We expect the second probe to be 1300 bytes.

assert_eq!(second_probe_size, 1300);

quiche/src/lib.rs
```diff
@@ -4231,7 +4231,7 @@ impl<F: BufFactory> Connection<F> {
             // (e.g. due to the anti-amplification limits).
             let should_probe_pmtu = active_path.should_send_pmtu_probe(
                 self.handshake_confirmed,
-                self.handshake_done_sent,
+                self.handshake_completed,
```
cf-rhett, line 4256: Should https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L3807 be changed as part of this to 

```
let size = if self.handshake_confirmed || self.handshake_completed {
```

from

```
let size = if self.handshake_confirmed || self. handshake_done_sent {
```

to match the logic here

quiche/src/lib.rs
rushilmehra: The probe isn't sent until the second send, the first only contains an ACK frame

quiche/src/lib.rs
```diff
@@ -18940,93 +18977,145 @@ mod tests {
...
+        // Verify third probe is even smaller
+        assert!(active_path.pmtud.get_should_probe());
+        let third_probe_size = active_path.pmtud.get_probe_size();
+        assert!(third_probe_size < second_probe_size);
 
-        // Continue searching for PMTU
-        assert!(pmtu_param.get_should_probe());
+        // Verify binary search convergence
+        let expected_third_size = 1200 + ((second_probe_size - 1200) / 2);
+        assert_eq!(third_probe_size, expected_third_size);
```
antoniovicente: Consider instead:
```
// Third probe should be 1250 bytes which is halfway between 1200 and the second probe size=1300.
assert_eq!(third_probe_size, 1250);
```

quiche/src/lib.rs
```diff
@@ -18940,93 +18977,145 @@ mod tests {
...
+
+        // Verify probe flag was reset after sending
+        let active_path = pipe.client.paths.get_active_mut().unwrap();
+        assert!(!active_path.pmtud.get_should_probe());
     }
 
     #[rstest]
-    fn pmtud_probe_loss(
+    /// Test that probe flag is properly re-enabled after probe loss
+    fn pmtud_probe_flag_reenabled_after_loss(
```
antoniovicente: This test seems redundant with `pmtud_probe_retry_after_loss` further down.

quiche/src/lib.rs
```diff
@@ -18940,93 +18977,145 @@ mod tests {
...
-        // PMTU not updated since probe is not ACKed
-        assert_eq!(pmtu_param.get_current_mtu(), 1200);
+        // Verify third probe is even smaller
+        assert!(active_path.pmtud.get_should_probe());
+        let third_probe_size = active_path.pmtud.get_probe_size();
+        assert!(third_probe_size < second_probe_size);
 
-        // Continue searching for PMTU
-        assert!(pmtu_param.get_should_probe());
+        // Verify binary search convergence
```
antoniovicente: Worth doing these tests?
- First 2 probes fail like in this test, 3rd succeeds.  MTU updated to 1250.
- All probes fail until MTU settles at 1200.

quiche/src/lib.rs
rushilmehra: They're in two separate packets:
```
        let mut frames: Vec<frame::Frame> = Vec::new();
        for _ in 0..2 {
            let mut buf = [0; 1400];
            let (len, _) = pipe.client.send(&mut buf).unwrap();
            let mut pkt =
                testing::decode_pkt(&mut pipe.server, &mut buf[..len]).unwrap();
            dbg!(&pkt);
            frames.append(&mut pkt);
        }
```
yields

```
[quiche/src/lib.rs:19039:13] &pkt = [
    ACK delay=7 blocks=[3..4] ecn_counts=None,
]
[quiche/src/lib.rs:19039:13] &pkt = [
    PADDING len=1365,
    PING mtu_probe=None,
]
```

quiche/src/lib.rs
rushilmehra: See https://github.com/cloudflare/quiche/pull/2084#discussion_r2151011280

quiche/src/lib.rs
antoniovicente: Is `assert!(matches!(frames[2], frame::Frame::Ping { .. }));` ?

Would be good to capture how frames are mapped into packets and resulting packet sizes.

rushilmehra: Yeah, I have a diff locally that asserts on both the ping frame slotted in `frames[2]` and `assert_eq!(frames.len(), 3)`. Will push it up soon

rushilmehra: For completeness, `should_send_pmtu_probe` returns `false` when frames have already been written to the output buffer:

* https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L4237
* https://github.com/cloudflare/quiche/blob/master/quiche/src/path.rs#L355

quiche/src/lib.rs
rushilmehra: Updated to line-by-line `quiche::send()`s

quiche/src/lib.rs
rushilmehra: Yeah great catch, updated

quiche/src/lib.rs
rushilmehra: Done

quiche/src/lib.rs
rushilmehra: Done

quiche/src/lib.rs
rushilmehra: Done

quiche/src/lib.rs
rushilmehra: Indeed, I removed it

quiche/src/lib.rs
```diff
@@ -3804,7 +3804,7 @@ impl<F: BufFactory> Connection<F> {
 
         // Update max datagram size to allow path MTU discovery probe to be sent.
         if send_path.pmtud.get_should_probe() {
-            let size = if self.handshake_confirmed || self.handshake_done_sent {
+            let size = if self.handshake_confirmed || self.handshake_confirmed {
```
antoniovicente: This should be:

`let size = if self.handshake_confirmed || self.handshake_completed {`

rushilmehra: Yeah I fixed

quiche/src/lib.rs
antoniovicente: Changes look good.  At least file a ticket for a followup test that covers the case of probes stopping when MTU settles at 1200 after a series of failures.  And that no probe is send for 1200.

antoniovicente (approved):
Changes look good. At least file a ticket for a followup test that covers the case of probes stopping when MTU settles at 1200 after a series of failures. And that no probe is send for 1200.

quiche/src/lib.rs
rushilmehra: I enhanced `pmtud_probe_retry_after_loss` to make the third probe successful and ensure `current_mtu` gets updated to 1250

antoniovicente (approved):

antoniovicente (approved):

PR #2085, opened 2025-06-18
Specify nightly for cargo fmt

cf-rhett:
quiche uses the nightly cargo fmt. Specify this in code so those that switch between nightly and stable use the correct formatter


cf-rhett:
Can't do this just for fmt -- sigh

rushilmehra:
> Can't do this just for fmt -- sigh

Did the same thing last week lol

PR #2089, opened 2025-06-26
Fuzz: add targets

catenacyber:
The main goal of this PR is to add fuzz targets to fuzz connection with multiple packets (and not only the first packet as the current fuzz targets do)

There are 2 fuzz targets that are pretty similar, but one generates the QUIC handshake, to be able to fuzz HTTP/3 directly

To be able to pass the handshake, I had to make `cargo test h3::tests --features fuzzing` pass and thus the first commit is modifying the behavior with `features fuzzing` to get closer to reality

Not finding new bugs after a few hours


catenacyber:
Oops, fixing the format with fixup commits

catenacyber:
Should I rebase to get rid of unrelated clippy errors ?

ghedo:
Yeah, rebase should fix them. Can also squash the fixup commits while at it.

I haven't had a chance to look at the PR though, as I've been mostly away for the past couple weeks.

ghedo:
Hey, thank you for the PR and sorry for the delay. It looks good to me overall, but there seems to be a fair bit of duplication between the 2 fuzzers, so it'd be nice if common code could be refactored and used across them (e.g. `PktsData`, `PktIterator`, the h3 event loop, etc...).

It would also be useful to update the https://github.com/cloudflare/quiche/blob/master/tools/gen_fuzz_seeds.sh script that generates initial seeds for each fuzzer, this way we have a way of bootstrapping the test data.

> To be able to pass the handshake, I had to make cargo test h3::tests --features fuzzing pass

Not sure I understand what the relationship between the fuzzers and the tests? Are the tests used to create the initial corpus for the fuzzers?

catenacyber:
Thanks for the feedback Alessandro

> but there seems to be a fair bit of duplication between the 2 fuzzers, so it'd be nice if common code could be refactored and used across them (e.g. PktsData, PktIterator, the h3 event loop, etc...).

I will fix that.

At first, I thought the fuzz target packets_recv_server should be good enough to reach and fuzz HTTP/3 but it does not seem the case yet. I think there is some randomness/unreproducibility somewhere. I did not investigate too much as I worked around it with the other fuzz target.

> It would also be useful to update the https://github.com/cloudflare/quiche/blob/master/tools/gen_fuzz_seeds.sh script that generates initial seeds for each fuzzer, this way we have a way of bootstrapping the test data.

Cool, I am adding packets_recv_server to it
But for packets_afterhs_server, I should be able to know which is the first packet after quic handshake...
(Also, I see some discrepancies lurking in the 4th packet logged by the server (3.pkt))

> > To be able to pass the handshake, I had to make cargo test h3::tests --features fuzzing pass
> 
> Not sure I understand what the relationship between the fuzzers and the tests? Are the tests used to create the initial corpus for the fuzzers?

I wanted to fuzz HTTP/3 and try to mimic the unit tests behavior into the fuzzing.
The difference in fuzzing is that we use `#[cfg(feature = "fuzzing")]` to "bypass" encryption. This used to be done with a zero-length AEAD tag but this behavior prevented me from making a successful quic handshake. So, I switched to a normal length (16) AEAD tag full of zeroes

catenacyber:
Feel free to squash the 2 last commits as they are the changes over the last iteration (+ rebase and fix of quiche::testing became quiche::test_utils)

catenacyber:
any updates @ghedo after https://github.com/cloudflare/quiche/security/advisories/GHSA-m3hh-f9gh-74c2 ?

ghedo:
Hey @catenacyber, sorry for the delay. Still want to get this merged, I just need to find some time to look at the code again.

catenacyber:
Friendly reminder @ghedo ;-)

catenacyber:
I also rebased and force-pushed, but I need approval to run CI...

catenacyber:
I managed to get rid of the unwanted submodule diff in the commits

catenacyber:
Great, so we are waiting for the merge, then I can create a PR to oss-fuzz, right ?

PR #2092, opened 2025-07-02
tokio-quiche: add an example asynchronous HTTP/3 server

evanrittenhouse:


tokio-quiche/Cargo.toml
```diff
@@ -74,10 +74,15 @@ triomphe = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
+clap = { version = "4.5.40", features = ["derive"] }
+env_logger = { workspace = true }
 h3i = { workspace = true }
+http = "1"
+http-body = "1"
+http-body-util = "0.1.3"
```
toidiu: nit:

```suggestion
http-body-util = "0.1"
```

tokio-quiche/examples/cert.crt
toidiu, line 1: I suggest that we add a `generate_cert.sh` script which can be used to re-generates the certs/keys. Ideally we do this for all test certs in the codebase.

Test cert expiration causes a lot of pain because tests start to fail and the original test intent is lost to time.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -192,6 +191,17 @@ pub struct IncomingH3Headers {
     pub h3_audit_stats: Arc<H3AuditStats>,
 }
 
+impl fmt::Debug for IncomingH3Headers {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
toidiu, line 195: Why is this change needed?

tokio-quiche/examples/async_http3_server/main.rs
```diff
@@ -0,0 +1,99 @@
...
+        log::info!("received new connection!");
+
+        // Create an `H3Driver` to serve the connection.
+        let (driver, mut controller) =
+            ServerH3Driver::new(Http3Settings::default());
+
+        // Start the driver. This will execute the handshake under the hood, which
+        // lets us start receiving ServerH3Events without needing to do
+        // anything extra after this future resolves.
+        conn.expect("socket dropped").start(driver);
```
toidiu: Rust recommends message style for expect.

https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

tokio-quiche/examples/async_http3_server/main.rs
toidiu, line 1: Can you add a Readme.md which explains how to run this example.

- short description
- how to get it running
  - args to start server
  - args to start client

Also worth building examples in CI so we know it continues to work?

tokio-quiche/examples/async_http3_server/body.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    #[cfg(not(feature = "zero-copy"))]
+    let frame = OutboundFrame::Body(BufFactoryImpl::get_empty_buf(), true);
+    #[cfg(feature = "zero-copy")]
+    let frame = OutboundFrame::Body(
+        <BufFactoryImpl as quiche::BufFactory>::buf_from_slice(&[]),
+        true,
+    );
+
+    frame_sender.send(frame).await.ok()?;
```
toidiu: If I am understanding this correctly, we are sending an empty frame here. Why do we need that?

Also why the zero-copy (we eventually do want to support this but maybe for the first pass we dont)?

tokio-quiche/examples/async_http3_server/body.rs
```diff
@@ -0,0 +1,153 @@
...
+    B: Body<Data = Bytes> + Send + Unpin + 'static,
+    B::Error: Send + Debug,
+{
+    let mut body_stream = BodyDataStream::new(body);
+
+    while let Some(maybe_chunk) = body_stream.next().await {
+        match maybe_chunk {
+            Ok(chunk) => {
+                for chunk in chunk.chunks(BufFactoryImpl::MAX_BUF_SIZE) {
+                    // Is it too many levels of chunking?
```
toidiu: Can you elaborate on this. What is the implication of this and if there an alternative.

tokio-quiche/examples/async_http3_server/body.rs
```diff
@@ -0,0 +1,153 @@
...
+        })
+    }
+
+    fn size_hint(&self) -> SizeHint {
+        SizeHint::with_exact(self.remaining as u64)
+    }
+}
+
+/// Use the `frame_sender` to send all frames for the `Body` over to
+/// tokio-quiche. The sender is paired with the underlying `H3Driver`.
```
toidiu: nit: can you structure this doc comment to match [convention](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#summary-sentence). There are other doc comments that this also applies to.

```
/// Summary (ideally 1 line).
///
/// Additional details.
```

tokio-quiche/examples/async_http3_server/server.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// client.
+    async fn end_stream(frame_sender: &mut OutboundFrameSender) {
+        let _ = frame_sender.send(OutboundFrame::PeerStreamError).await;
+    }
+}
+
+/// Convert a list of [Header]s into a request object which can be processed by
+/// the [`Server`].
+///
+/// This isn't complete or robust by any means.
```
toidiu: I find this a bit confusing. Could you rephrase this to say what is missing or what is supported.

tokio-quiche/examples/async_http3_server/body.rs
evanrittenhouse: The empty frame sends the `fin` bit to the client thanks to passing `true`.

As for the zero-copy thing, relic of me initially putting it in `apps` and needing to please CI - I can remove it now, I think, now that it's here in `examples`

tokio-quiche/src/http3/driver/mod.rs
evanrittenhouse: Not strictly required, I just noticed that logging headers currently is messy because it prints the channel state. See https://github.com/cloudflare/quiche/pull/2092/commits/98cfcbbe3a390c546ccf0d195c3271b53ab39fc0 for a concrete example

tokio-quiche/examples/async_http3_server/main.rs
evanrittenhouse: I added a README. I'd like to add example builds as follow-up to avoid cluttering this PR more, it doesn't look like the existing workflows use `--examples`: https://github.com/cloudflare/quiche/blob/master/.github/workflows/stable.yml#L158-L200

tokio-quiche/examples/async_http3_server/body.rs
evanrittenhouse: Removed it, copy-paste relic

tokio-quiche/examples/async_http3_server/server.rs
evanrittenhouse: Added some more detail, please let me know if it still isn't clear

tokio-quiche/examples/cert.crt
evanrittenhouse: Can we do that as followup? PR is already kinda large 😅

toidiu: Yep no need to block on this.

tokio-quiche/examples/README.md
```diff
@@ -0,0 +1,94 @@
...
+Verbosities can be specified with typical [`env_logger`](https://docs.rs/env_logger/latest/env_logger/#enabling-logging) syntax.
+
+The default TLS certificate covers `test.com`. Certificates can be passed via
+the `--tls-cert-path` CLI argument, while private keys can be passed via the
+`--tls-private-key-path` argument.
+
+Once the server is up and running, you can hit it with your favorite client:
+
+```shell
+❯ curl https://test.com/ --http3 -vk --connect-to ::127.0.0.1:5757
```
toidiu: Any reason to prefer curl over `quiche/http-client`, which seems easier to build and use IMO?
```
RUST_LOG=info ./target/debug/examples/http3-client https://127.0.0.1:8000/stream-bytes/33
```

The default curl didnt support http3 on my Mac. If we want curl we should provide installation instructions.

tokio-quiche/examples/async_http3_server/body.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Use the `frame_sender` to send DATA frames to tokio-quiche.
+///
+/// The sender is paired with the underlying tokio-quiche `H3Driver`.
+pub(crate) async fn send_body<B>(
+    body: B, mut frame_sender: OutboundFrameSender,
+) -> Option<()>
+where
+    B: Body<Data = Bytes> + Send + Unpin + 'static,
+    B::Error: Send + Debug,
+{
```
toidiu:
```
impl ExampleBody {
...

  pub(crate) async fn send(
      self, mut frame_sender: OutboundFrameSender,
  ) -> Option<()> {
     ... 
  }
}
```

tokio-quiche/examples/async_http3_server/server.rs
```diff
@@ -0,0 +1,276 @@
...
+                                        * connection was terminated */
+            }
+        }
+    }
+
+    /// Handle a [`ServerH3Event`].
+    ///
+    /// For simplicity's sake, we only handle a couple of events here.
+    // TODO(evanrittenhouse): support POST requests
+    async fn handle_forwarded_event(
```
toidiu: nit: rename to `handle_h3_event"?

tokio-quiche/examples/README.md
evanrittenhouse: Updated to just regular `quiche-client`

tokio-quiche/examples/async_http3_server/body.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Use the `frame_sender` to send DATA frames to tokio-quiche.
+    ///
+    /// The sender is paired with the underlying tokio-quiche `H3Driver`.
+    pub(crate) async fn send(
+        self, mut frame_sender: OutboundFrameSender,
+    ) -> Option<()> {
+        let mut body_stream = BodyDataStream::new(self);
+
+        while let Some(maybe_chunk) = body_stream.next().await {
+            match maybe_chunk {
```
toidiu: This was confusing since the type is Result<Bytes> not Option<Bytes>.

```suggestion
        while let Some(chunk) = body_stream.next().await {
            match chunk {
```

tokio-quiche/examples/async_http3_server/server.rs
```diff
@@ -0,0 +1,273 @@
...
+        match event {
+            // Received an explicit connection level error. Not much to do here.
+            H3Event::ConnectionError(err) => QuicResult::Err(Box::new(err)),
+
+            // The connection has shutdown.
+            H3Event::ConnectionShutdown(err) => {
+                let error = err.map_or_else(
+                    || Box::new(quiche::h3::Error::Done) as BoxError,
+                    |e| Box::new(e),
+                );
```
toidiu: nit: this seems neater

```suggestion
                 let err = match err {
                     Some(err) => Box::new(err),
                     None => Box::new(quiche::h3::Error::Done) as BoxError,
                 };
```


Also trick for better readability and type-safety is to overwrite the same variable name (`err`) once you convert it into the new type (previously was called `error`).

From a readability perspective, I can easily tell that since `err` is overwritten, it wont be used again, and that it has a new value.

From a type safety perspective, if someone adds code later, its impossible for them to use the incorrect value of err/error.

toidiu (approved):

Issue #2025, opened 2025-04-15
QUIC compatibility


maxime-bruno:
Hello,

I'm currently conducting a study to figure out why QUIC outperforms TCP in some scenarios.
In addition to a theoretical study, I wanted an empirical one.

Quiche boasts about its high configurability. That's why I chose it.

Following several inconsistent results (#1988 among others), I tried with Quinn (another well-used QUIC lib).
And to my great surprise, the results are different. We find standard results.

I then chose to see if they were both able to communicate. That when I find out that:
* A quinn server and a quiche client can communicate
* A quiche server and a quinn client suffer some issues

So here is my question: Does Quiche implement a RFC compatible version of QUIC, or is the issue coming from Quinn, or my code ?

For information, here is my code:

The part to create a server
```rust
use std::collections::HashSet;
use std::fmt::Display;
use std::net::SocketAddr;
use std::sync::Arc;
use tokio::sync::RwLock as TokioRwLock;
use tokio::net::UdpSocket;

pub struct Server<'a> {
    udp_socket: Arc<UdpSocket>,
    paths: Arc<TokioRwLock<HashSet<Path>>>,
    config: Config<'a>,
}

type Path = (SocketAddr, SocketAddr);

impl Server<'_> {
    pub async fn start(host: impl Display, port: u16, config: Config<'_>) -> error::Result<Server> {
        let paths = Arc::new(TokioRwLock::new(HashSet::new()));

        let udp_socket =
            Arc::new(tokio::net::UdpSocket::bind((host.to_string().as_str(), port)).await?);

        Ok(Server {
            udp_socket,
            paths,
            config,
        })
    }

    pub async fn accept(&mut self) -> error::Result<Connection> {
        let mut interval = tokio::time::interval(TICK);
        let local = self.udp_socket.local_addr()?;

        loop {
            let peer = self.udp_socket.peek_sender().await?;
            if self.paths.read().await.contains(&(local, peer)) {
                log::trace!("Connection already exists");
                interval.tick().await;
                continue;
            }
            return Connection::accept(
                &self.config,
                Arc::clone(&self.udp_socket),
                Arc::clone(&self.paths),
                (local, peer),
            )
            .await;
        }
    }

    pub fn local_addr(&self) -> error::Result<SocketAddr> {
        Ok(self.udp_socket.local_addr()?)
    }
}
```

The part to handle the connection:
```rust
use core::pin::Pin;
use core::task::{Context, Poll};
[...]

PR #2096, opened 2025-07-06
Key update with TLS_AES_256_GCM_SHA384 trunkates bytes from the next secret. Incoming packets get discarded. 

Ruben2424:
A timeout with a [quinn](https://github.com/quinn-rs/quinn) client was reported in:
https://github.com/seanmonstar/reqwest/issues/2610 and also #2025 

While debugging i noticed, that after a attempted key update from quinn incoming packets could not be decrypted. Also the new secret was shorter than the old one.

Quinn performs the early key update by design https://github.com/quinn-rs/quinn/blob/64139181e0fcdf337737d32dac077e96391d9d52/quinn-proto/src/connection/mod.rs#L299-L305

This PR ensures, that the buffer is large enough to support the current key size.


ghedo:
Nice catch! Merged now, thanks @Ruben2424.

PR #2102, opened 2025-07-11
Add unchecked congestion window algorithm

cf-rhett:
Adds an opt-in congestion control algorithm for deployments that constrain transmission outside quiche. `congestion_window_unchecked` keeps the congestion window at `usize::MAX` while preserving ACK processing, RTT estimation, loss detection, recovery diagnostics, and shared bytes-in-flight accounting.

The earlier version copied Reno because congestion controllers previously owned bytes-in-flight accounting. That accounting moved into generic recovery in #1772, so this refresh reduces the algorithm to a small callback table. It retains saturating PRR arithmetic for the maximum-window edge case.

This is hidden behind the non-default `congestion_window_unchecked_available` feature and forwarded by tokio-quiche as `congestion_window_unchecked`. It provides no congestion-window protection and makes optimistic-ACK probes extremely infrequent, so callers must supply their own transmission constraints.

Feature-enabled tests cover send capacity, ACK and RTT accounting, loss and recovery accounting, packet-number-space discard, and PRR saturation. Stable and nightly CI exercise the feature, including 32-bit Windows in stable CI.

I will squash all commits before merge but keeping separate for reviewers for now


quiche/src/recovery/congestion/cwnd_max.rs
```diff
@@ -0,0 +1,119 @@
...
+    true
+}
+
+fn has_custom_pacing() -> bool {
+    false
+}
+
+#[cfg(feature = "qlog")]
+fn state_str(r: &Congestion, now: Instant) -> &'static str {
+    reno::state_str(r, now)
```
LPardue: Let's change this to "cwnd_max"

cf-rhett: changed to cc_disabled per offline discussion

aochagavia:
I'm a bit surprised about the amount of code in this PR. Why do we need to keep any cc logic at all inside the provided functions? To illustrate my point, consider a no-op cc algorithm I implemented in my own fork of quiche a few months ago (see [diff](https://github.com/aochagavia/quiche/commit/a7644715a296ee87ca9cc7dae5a4ff91e7efef01)), which looks much simpler. Am I missing something? I hope I'm not making a showcase of ignorance with my question...

quiche/src/recovery/congestion/cc_disabled.rs
```diff
@@ -0,0 +1,117 @@
...
+use crate::recovery;
+
+use crate::recovery::congestion::Acked;
+use crate::recovery::rtt::RttStats;
+use crate::recovery::Sent;
+
+use super::Congestion;
+use super::CongestionControlOps;
+
+pub(crate) static CC_DISABLED: CongestionControlOps = CongestionControlOps {
```
antoniovicente: @toidiu

One challenge I see with this change is the future of the code in the gcongestion directory.  Ideally this change would target the APIs in that other directory instead, since that's closer to the future.

But you have what seems to be working code, so I hesitate to force you to adjust it to use a different interface.

quiche/src/recovery/mod.rs
```diff
@@ -353,6 +356,8 @@ impl FromStr for CongestionControlAlgorithm {
             #[cfg(feature = "gcongestion")]
             "bbr2" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
             "bbr2_gcongestion" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
+            #[cfg(feature = "cc_disabled")]
+            "cc_disabled" => Ok(CongestionControlAlgorithm::CcDisabled),
```
antoniovicente: It would be good to add more tests for the cc_disabled functionality, possibly based on other tests in this file.

quiche/Cargo.toml
```diff
@@ -55,6 +55,9 @@ ffi = ["dep:cdylib-link-lines"]
 # Exposes internal APIs that have no stability guarantees across versions.
 internal = []
 
+# Disable congestion control
+cc_disabled = []
```
toidiu: Could you expand on this with warnings about why this is a terrible idea and one should only do this if they are super very sure what they are doing.

quiche/src/recovery/mod.rs
```diff
@@ -335,6 +335,9 @@ pub enum CongestionControlAlgorithm {
     /// BBRv2 congestion control algorithm implementation from gcongestion
     /// branch. `bbr2_gcongestion` in a string form.
     Bbr2Gcongestion = 4,
+    /// Disabled congestion control. `cc_disabled` in a string form.
```
toidiu: Do we need this comment?

quiche/src/recovery/congestion/recovery.rs
```diff
@@ -844,8 +844,9 @@ impl RecoveryOps for LegacyRecovery {
         }
 
         // Open more space (snd_cnt) for PRR when allowed.
-        self.cwnd().saturating_sub(self.bytes_in_flight.get()) +
-            self.congestion.prr.snd_cnt
+        self.cwnd()
+            .saturating_sub(self.bytes_in_flight.get())
+            .saturating_add(self.congestion.prr.snd_cnt)
```
toidiu, line 894: Was this change motivated by some networks failing without CC constraints? Is it possible to add tests for this.

The comment https://github.com/cloudflare/quiche/pull/2102/files#r2237321998 also suggests additional tests and it would be a good idea since our CI currently doesn't run in this mode.

For that matter please add a CI task that runs quiche with the noop CC mode enabled.

toidiu:
> I'm a bit surprised about the amount of code in this PR. Why do we need to keep any cc logic at all inside the provided functions? To illustrate my point, consider a no-op cc algorithm I implemented in my own fork of quiche a few months ago (see [diff](https://github.com/aochagavia/quiche/commit/a7644715a296ee87ca9cc7dae5a4ff91e7efef01)), which looks much simpler. Am I missing something? I hope I'm not making a showcase of ignorance with my question...

Agreed, seems like there seems to be too much implementation for a "disabled" CC... but maybe there is a good reason for that?

quiche/Cargo.toml
```diff
@@ -55,6 +55,9 @@ ffi = ["dep:cdylib-link-lines"]
 # Exposes internal APIs that have no stability guarantees across versions.
 internal = []
 
+# Disable congestion control
```
antoniovicente: The feature name and description doesn't seem accurate.  Just setting this option doesn't disable congestion control, but expose a congestion control algorithm that doesn't apply traditional congestion window based limits.  The default cc is still cubic.

quiche/src/recovery/mod.rs
antoniovicente: It is consistent with the comments associated with the rest of the enum values.  Seems OK to me.

quiche/src/recovery/congestion/cc_disabled.rs
```diff
@@ -0,0 +1,117 @@
...
+    checkpoint,
+    rollback,
+    has_custom_pacing,
+    #[cfg(feature = "qlog")]
+    state_str,
+    debug_fmt,
+};
+
+pub fn on_init(r: &mut Congestion) {
+    r.congestion_window = usize::MAX;
```
antoniovicente: We could smaller values here while still achieving the same effect and avoiding the usize overflow issues that require changing `+` to `saturating_add`

But I don't have much of a problem with the `saturating add`, so we could go either way.

quiche/src/recovery/mod.rs
cf-rhett: I've added this to the existing test suite. Let me know if there are any other items you'd like to test here

quiche/Cargo.toml
cf-rhett: expanded on -- let me know if this is sufficient for you

quiche/Cargo.toml
cf-rhett: Renamed and annoted with comments more closely describing it's usage

quiche/src/recovery/congestion/cc_disabled.rs
cf-rhett: I went to smaller values but felt the saturing_add doesn't harm anything so left this

cf-rhett:
> I'm a bit surprised about the amount of code in this PR. Why do we need to keep any cc logic at all inside the provided functions? To illustrate my point, consider a no-op cc algorithm I implemented in my own fork of quiche a few months ago (see [diff](https://github.com/aochagavia/quiche/commit/a7644715a296ee87ca9cc7dae5a4ff91e7efef01)), which looks much simpler. Am I missing something? I hope I'm not making a showcase of ignorance with my question...

This is what my team tried at first (very simple noop cc) but found that it eventually broke in production. This was before my time so I don't know if that would still be the case as this was 2.5 years ago but this version (or close to it) has been reliable in production so hesitant to change it too much

quiche/src/recovery/congestion/mod.rs
```diff
@@ -40,6 +40,16 @@ use crate::StartupExitReason;
...
 
+/// 16 GiB
+#[cfg(any(test, feature = "congestion_window_unchecked_available"))]
+#[cfg(target_pointer_width = "64")]
+pub const UNCHECKED_WINDOW: usize = 16 << 30;
+
+/// 2 GiB
+#[cfg(any(test, feature = "congestion_window_unchecked_available"))]
+#[cfg(not(target_pointer_width = "64"))]
+pub const UNCHECKED_WINDOW: usize = 2 << 30;
```
antoniovicente: I wonder if it would be preferable to use u64 instead of usize to track this so we can avoid issues related to usize width.

quiche/src/recovery/congestion/cwnd_unchecked.rs
```diff
@@ -0,0 +1,119 @@
...
+        if r.hystart.on_packet_acked(packet, rtt_stats.latest_rtt, now) {
+            // Exit to congestion avoidance if CSS ends.
+            r.ssthresh.update(r.congestion_window, true);
+        }
+    } else {
+        // Congestion avoidance.
+        r.bytes_acked_ca += packet.size;
+
+        if r.bytes_acked_ca >= r.congestion_window {
+            r.bytes_acked_ca -= r.congestion_window;
```
antoniovicente: I guess this is reachable now that `r.congestion_window` is only 2GB or 16GB.

I don't have a good intuition for why this code is needed though.  Although I'm fine with this code mostly going in as is, I'ld like that we put some effort into simplifying this stuff further and figuring out what parts are actually needed and which aren't.

marcblanchet:
any update on this? would love to get it merged, as it would have immediate use for our use case.

quiche/src/recovery/mod.rs
cf-rhett: Done. The existing PTO, timer-loss, and reordering tests now include a feature-conditional congestion_window_unchecked case. Focused tests cover the unchecked send window, ACK and RTT accounting, packet-space discard, and saturation.

quiche/src/recovery/congestion/recovery.rs
cf-rhett: No specific network failure motivated the arithmetic change. usize::MAX makes the existing PRR addition overflow unless it saturates. 

Stable and nightly CI enable the feature and run this in CI

quiche/Cargo.toml
cf-rhett: The gate is now congestion_window_unchecked_available. Enabling it only exposes the algorithm; Cubic remains the default until a caller explicitly selects it.

quiche/src/recovery/congestion/mod.rs
cf-rhett: Removed the constants here in the simplification

quiche/src/recovery/congestion/cwnd_unchecked.rs
cf-rhett: The copied Reno implementation is gone.

quiche/src/recovery/congestion/recovery.rs
```diff
@@ -1159,6 +1160,18 @@ mod tests {
     use crate::recovery::RecoveryConfig;
     use std::time::Instant;
```
csujedihy, line 1162: Can we add a test that verifies that cwnd is not changed after loss and after spurious loss?

csujedihy: and maybe cwnd never grows on ACK

cf-rhett: added/modified existing test to assert for these

csujedihy: thanks for adding them!

csujedihy (approved):

quiche/src/recovery/congestion/mod.rs
```diff
@@ -283,6 +283,24 @@ pub(crate) struct CongestionControlOps {
     ) -> std::fmt::Result,
 }
 
+#[cfg(feature = "congestion_window_unchecked_available")]
+static CONGESTION_WINDOW_UNCHECKED: CongestionControlOps = CongestionControlOps {
+    on_init: |recovery| recovery.congestion_window = usize::MAX,
+    on_packet_sent: |_, _, _, _| {},
+    on_packets_acked: |_, _, _, _, _| {},
```
csujedihy, line 290: Just a thought: I wonder if we should still set cwnd to `usize::MAX` just to be defensive, in case there are logic in the future that temporarily expands the cwnd without using `saturating_add`, which could wraps around the cwnd causing cwnd to be pinned at very small value permanently.

quiche/src/recovery/congestion/mod.rs
```diff
@@ -283,6 +283,24 @@ pub(crate) struct CongestionControlOps {
...
+    congestion_event: |recovery, _, _, largest_lost_packet, now| {
+        if !recovery.in_congestion_recovery(largest_lost_packet.time_sent) {
```
gregor-cf: do we need this? It would mean that once we set it, we never leave `congestion_recovery`. IDK off the top of my head what implications that has.

quiche/src/recovery/congestion/mod.rs
gregor-cf: +1
I think a static, large window (20--100MB) is probably not a bad idea. It also serves as an emergency backstop, since it prevents us from having and unbounded amount of packets in flight.

gregor-cf, line 290: stupid question: Loss detection and in-flight tracking is handled by `recovery` directly and still works, correct, even with `on_packet_sent` and `on_packets_acked` being no-ops, correct?

cf-rhett: That's correct from my understanding. These callbacks only control how the congestion window reacts

quiche/src/recovery/congestion/mod.rs
cf-rhett: No, I don't think we do -- removed

quiche/src/recovery/congestion/mod.rs
cf-rhett: I'd like some input from the protocols team here -- I'm not against doing this but think their insight might be valuable here.

@ghedo @antoniovicente thoughts?

quiche/src/recovery/congestion/mod.rs
csujedihy: oh good catch!

PR #2110, opened 2025-07-17
tokio-quiche: round-robin send over all network paths

ghedo:
Currently tokio-quiche depends on the `send()` method to pick the network path to send packets on. This can mean that only packets for a single path end being flushed to the network.

In addition to that, the destination address for `sendmsg()` calls is fixed to the peer's initial address, initialized when a connection is established.

These issues end up preventing connections from migrating correctly when the peer's address changes (e.g. due to NAT rebinding) because packets for any new path would have to be flushed too, and in any case the original peer address would be used to send packets to anyway, rather than the new address the peer migrated to.

This change fixes both of those issues, but note that while this allows "passive" migration of peers, it does not allow "active" migration, where a peer explicitly decides to switch network path by picking a new connection ID. Supporting that would require providing additional connection IDs on our part too, which could be implemented in the future, but is out of scope of this change.

The test added is somewhat complicated, mostly because it needs to use "plain" quiche to implement the migrating client, such that we can control which address packets are sent from, but it simply uses 2 sockets bound to different addresses to simulate the client migrating (similarly to what `quiche-client` does).

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -79,6 +79,7 @@ const GSO_THRESHOLD: usize = 1_000;
 pub struct WriterConfig {
     pub pending_cid: Option<ConnectionId<'static>>,
     pub peer_addr: SocketAddr,
+    pub local_addr: SocketAddr,
     pub with_gso: bool,
     pub pacing_offload: bool,
     pub with_pktinfo: bool,
@@ -95,9 +96,11 @@ pub(crate) struct WriteState {
     // If pacer schedules packets too far into the future, we want to pause
     // sending, until the future arrives
     next_release_time: Option<Instant>,
-    // If set, outgoing packets will be sent to the peer from the `send_from`
-    // address rather than the listening socket.
-    send_from: Option<SocketAddr>,
+    // The selected source and destination addresses for the current write
+    // cycle.
+    selected_path: Option<(SocketAddr, SocketAddr)>,
+    // Iterator over the network paths that haven't been flushed yet.
+    pending_paths: quiche::SocketAddrIter,
 }
 
 pub(crate) struct IoWorkerParams<Tx, M> {
@@ -274,13 +277,6 @@ where
         }
     }
 
-    #[inline]
-    fn gather_data_from_quiche_conn(
-        &mut self, qconn: &mut QuicheConnection, send_buf: &mut [u8],
-    ) -> QuicResult<usize> {
-        self.fill_send_buffer(qconn, send_buf)
-    }
-
     #[cfg(feature = "perf-quic-listener-metrics")]
     fn measure_complete_handshake_time(&mut self) {
         if let Some(init_rx_time) = self.init_rx_time.take() {
@@ -294,7 +290,7 @@ where
         }
     }
 
-    fn fill_send_buffer(
+    fn gather_data_from_quiche_conn(
         &mut self, qconn: &mut QuicheConnection, send_buf: &mut [u8],
     ) -> QuicResult<usize> {
         let mut segment_size = None;
@@ -302,6 +298,8 @@ where
 
         self.write_state.num_pkts = 0;
         self.write_state.bytes_written = 0;
+
+        self.write_state.selected_path = None;
 
         let now = Instant::now();
 
@@ -348,16 +346,10 @@ where
             );
 
             let packet_size = match outcome {
-                Ok(0) => {
-                    self.write_state.has_pending_data = false;
+                Ok(0) => break Ok(0),
 
-                    break Ok(0);
-                },
-                Ok(bytes_written) => {
-                    self.write_state.has_pending_data = true;
+                Ok(bytes_written) => bytes_written,
 
-                    bytes_written
-                },
                 Err(e) => break Err(e),
             };
 
@@ -458,6 +450,34 @@ where
         buffer_write_outcome
     }
 
+    /// Selects a network path, if none already selected.
+    ///
+    /// This will return the first path available in the write state's
+    /// `pending_paths` iterator. If that is empty a new iterator will be
+    /// created by querying quiche itself.
+    ///
+    /// Note that the connection's statically configured local address will be
+    /// used to query quiche for available paths, so this can't handle multiple
+    /// local addresses currently.
+    fn select_path(
+        &mut self, qconn: &QuicheConnection,
+    ) -> Option<(SocketAddr, SocketAddr)> {
+        if self.write_state.selected_path.is_some() {
+            return self.write_state.selected_path;
+        }
+
+        let from = self.cfg.local_addr;
+
+        // Initialize paths iterator.
+        if self.write_state.pending_paths.len() == 0 {
+            self.write_state.pending_paths = qconn.paths_iter(from);
+        }
+
+        let to = self.write_state.pending_paths.next()?;
+
+        Some((from, to))
+    }
+
     #[cfg(not(feature = "gcongestion"))]
     fn pacing_enabled(&self, qconn: &QuicheConnection) -> bool {
         self.cfg.pacing_offload && qconn.pacing_enabled()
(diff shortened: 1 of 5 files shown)


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -475,26 +498,56 @@ where
...
             Err(QuicheError::Done) => {
-                // Flush to network and yield when there are no
-                // more packets to write.
+                // Flush the current buffer to network. If no other path needs
+                // to be flushed to the network also yield the work loop task.
+                //
+                // Otherwise the write loop will start again and the next path
+                // will be selected.
+                let has_pending_paths = self.write_state.pending_paths.len() > 0;
+                let has_flushed_paths = self.write_state.has_flushed_paths;
```
evanrittenhouse: AFAICT we never set `self.write_state.has_flushed_paths` to `true`, I think we'd want to set it in `flush_buffer_to_socket`?

More generally though, can you elaborate on why do we need to continue writing if we've already flushed data? I must be missing something, as it seems to me that we could just loop so long as `has_pending_paths` is true if `send_on_path` returns `Error::Done`

ghedo: IIRC the reason was to ensure the path iterator is up-to-date before giving up, so you'd loop through all paths, then update the path iterator and go through the paths again to make sure there is no more pending data. Though TBH I'm not entirely sure it's needed. For the simple passive migration case I _think_ it should be fine to remove though.

ghedo: Removed.

evanrittenhouse (dismissed):

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -475,26 +495,53 @@ where
...
-            // function properly
+            // function properly.
             send_buf = &mut send_buf[..segment_size.unwrap_or(usize::MAX)];
         }
 
-        match qconn.send(send_buf) {
+        // On the first call to `select_path()` a path will be chosen based on
+        // the local address the connection initially landed on. Once a path is
+        // selected following calls to `select_path()` will return it, until it
+        // is reset at the start of the next write cycle.
```
toidiu, line 505: > Once a path is selected following calls to `select_path()` will return it, until it
> is reset at the start of the next write cycle.

How important is it that select_path return the same path? As in if it doesnt happen would it be detrimental or simply less efficient?

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -458,6 +450,34 @@ where
...
+        let from = self.cfg.local_addr;
+
+        // Initialize paths iterator.
+        if self.write_state.pending_paths.len() == 0 {
+            self.write_state.pending_paths = qconn.paths_iter(from);
+        }
+
+        let to = self.write_state.pending_paths.next()?;
+
+        Some((from, to))
```
toidiu, line 478: Can you add a `debug_assert!(self.write_state.selected_path.is_none())` to catch future mistakes.

tokio-quiche/tests/integration_tests/migration.rs
```diff
@@ -0,0 +1,231 @@
...
+///
+/// The test simply binds a UDP socket on one address which is then used to
+/// complete the handshake and send an initial HTTP/3 request, then uses a new
+/// socket bound to a different port to send an additional HTTP/3 request, if
+/// both requests complete that means that the client was successfully migrated
+/// to the new address.
+///
+/// This requires using "plain" quiche as a client to properly control when and
+/// where packets are sent to, which is not possible using h3i.
+async fn test_passive_migration() {
```
toidiu, line 48: Nice comment!

tokio-quiche/tests/integration_tests/migration.rs
```diff
@@ -0,0 +1,231 @@
...
+    let mut quic_settings = QuicSettings::default();
+    quic_settings.active_connection_id_limit = 2;
+    quic_settings.disable_active_migration = true;
```
toidiu, line 51: Would passive migration work with `active_connection_id_limit = 1`. Whats the default value for this?

tokio-quiche/tests/integration_tests/migration.rs
```diff
@@ -0,0 +1,231 @@
...
+    // client doesn't know that the path changes (e.g. due to NAT rebinding).
+    h3_conn.send_request(&mut conn, &req, true).unwrap();
+    emit_flight(&migrated_socket, &mut conn).await;
+    process_flight(&migrated_socket, client_addr, &mut conn).await;
+
+    // Client responds to PATH_CHALLENGE.
+    emit_flight(&migrated_socket, &mut conn).await;
+
+    // Client receives response for the second request.
+    process_flight(&migrated_socket, client_addr, &mut conn).await;
```
toidiu, line 154: Anyway to verify we got a challenge and are responding to it?

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -458,6 +450,34 @@ where
...
+    /// created by querying quiche itself.
+    ///
+    /// Note that the connection's statically configured local address will be
+    /// used to query quiche for available paths, so this can't handle multiple
+    /// local addresses currently.
+    fn select_path(
+        &mut self, qconn: &QuicheConnection,
+    ) -> Option<(SocketAddr, SocketAddr)> {
+        if self.write_state.selected_path.is_some() {
+            return self.write_state.selected_path;
```
antoniovicente, line 466: Is there value in having a config parameter or build time feature to disable the new functionality in case something with it goes wrong?

tokio-quiche/src/quic/io/worker.rs
ghedo: It's explained in the next paragraph. Basically returning different paths in the middle of a write cycle would break GSO.

tokio-quiche/tests/integration_tests/migration.rs
ghedo: 2 is the minimum value for `active_connection_id_limit` per RFC 9000. It's also the default value so the previous line is maybe unnecessary, but seems more clear to be explicit here.

tokio-quiche/tests/integration_tests/migration.rs
ghedo: There is a `path_challenge_rx_count` stats in quiche, though it only tells you a challenge was received. I'll add that check anyway, seems useful.

tokio-quiche/src/quic/io/worker.rs
ghedo: Generally speaking, it seems weird to have a config option that is basically "make it work how it's supposed to"... but more practically, either all the new code is wrapped in if / else which IMO would be very messy, or it seems unlikely a config would be effective. As for a build time option I don't think it would save much work compared to just reverting the change TBH.

tokio-quiche/src/quic/io/worker.rs
ghedo: Not sure I follow. The method starts by checking if `selected_path` is Some, so not sure how checking again would help?

tokio-quiche/src/quic/io/worker.rs
toidiu: So that seems like a potential sharp edge especially since the logic lives in another function `select_path()` (someone might accidentally change that fn). Maybe we already have a test to detect this, in which case its fine but GSO bugs can be hard to reason about.

The idea behind https://github.com/cloudflare/quiche/pull/2110/files#r2285821179 was that we check at the end of `select_path()` that it never returns `Some(...)` if `self.write_state.selected_path` was already set, because it should have returned early.

Would it be cleaner to have the early return logic in this function? Don't know who else calls `select_path()`.

```
if self.write_state.selected_path.is_some() {
            return self.write_state.selected_path;
```

tokio-quiche/src/quic/io/worker.rs
toidiu: Left a related [comment](https://github.com/cloudflare/quiche/pull/2110/files#r2288943988). Lmk if I am being too paranoid and/or if we already have a test to cover it.

tokio-quiche/src/quic/io/worker.rs
ghedo: I still don't understand the issue TBH... Are we trying to protect against compiler bugs? Otherwise I don't see how the scenario you are describing is possible. And if we do that here, then we should add `debug_assert!()` to every other `if` block? Like, why is this case special.

toidiu: Spoke offline. We have tests that will break if we remove the early return.

toidiu (approved):

PR #2112, opened 2025-07-21
refactor: remove impl CongestionControl for Pacer

toidiu:
The original impl had a bunch of abstraction to support bbr1/bbr2/bbr3. Making pacer implement CongestionControl, using enum_dispatch, etc were some of the abstractions. We are only using bbr3 so we need to get rid of the weird stuff like pacer implementing CongestionControl trait (which makes no sense).

### DIFF
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -41,7 +41,6 @@ use crate::Result;
 use super::pacer::Pacer;
 use super::Acked;
 use super::Congestion;
-use super::CongestionControl;
 use super::Lost;
 
 // Congestion Control
@@ -128,7 +127,7 @@ struct LossDetectionResult {
 impl RecoveryEpoch {
     /// Discard the Epoch state and return the total size of unacked packets
     /// that were discarded
-    fn discard(&mut self, cc: &mut impl CongestionControl) -> usize {
+    fn discard(&mut self, cc: &mut Pacer) -> usize {
         let unacked_bytes = self
             .sent_packets
             .drain(..)
--- a/quiche/src/recovery/gcongestion/mod.rs
+++ b/quiche/src/recovery/gcongestion/mod.rs
@@ -138,11 +138,6 @@ pub(super) trait CongestionControl: Debug {
     #[allow(dead_code)]
     fn is_cwnd_limited(&self, bytes_in_flight: usize) -> bool;
 
-    #[cfg(test)]
-    fn is_app_limited(&self, bytes_in_flight: usize) -> bool {
-        !self.is_cwnd_limited(bytes_in_flight)
-    }
-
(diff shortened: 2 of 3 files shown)


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -400,10 +399,12 @@ pub struct GRecovery {
 impl GRecovery {
     pub fn new(recovery_config: &RecoveryConfig) -> Option<Self> {
         let cc = match recovery_config.cc_algorithm {
-            CongestionControlAlgorithm::Bbr2Gcongestion => Congestion::bbrv2(
+            CongestionControlAlgorithm::Bbr2Gcongestion => BBRv2::new(
```
antoniovicente: We need a credible plan for merging the contents of the `congestion` and `gcongestion` directories in order to minimize the current massive code duplication in those directories.  We need to preserve some way of select between multiple implementations selected by cc_algorithm.

quiche/src/recovery/gcongestion/bbr2/startup.rs
```diff
@@ -85,6 +85,7 @@ impl ModeImpl for Startup {
         let check_persisten_queue =
             params.max_startup_queue_rounds > 0 && !has_bandwidth_growth;
         if check_persisten_queue {
+            // https://github.com/google/quiche/blob/27eca0257490df89d2bd2c2a8bcea15565e7831c/quiche/quic/core/congestion_control/bbr2_startup.cc#L60-L62
```
antoniovicente: What branch is 27eca0257490df89d2bd2c2a8bcea15565e7831c associated with?  I see it on github, but it is not a revision that my local git clients know about.

toidiu: Its just a permalink to google quiche to avoid code drift.

quiche/src/recovery/gcongestion/recovery.rs
toidiu: This PR is only concerned with gcongestion and work to combine recover and gcongestion can happen later.

There is another use of enum_dispatch which controls the switch between recovery vs gcongestion. Link [1](https://github.com/cloudflare/quiche/blob/678859667a1814f69d6d1a00030d21d65b6a8555/quiche/src/recovery/mod.rs#L129-L135) and [2](https://github.com/cloudflare/quiche/blob/678859667a1814f69d6d1a00030d21d65b6a8555/quiche/src/recovery/mod.rs#L284-L291)

toidiu:
As per discussion, I'll split each commit into a separate PR

quiche/src/recovery/gcongestion/recovery.rs
toidiu: I have split the PRs so lets talk about it in the follow up PR.

PR #2113, opened 2025-07-25
Fixes to BBR2 initial pacing rate for compliance with the BBR draft RFC, and introduce option to override initial pacing rate for new connections

antoniovicente:
Changes:
* Use startup pacing gain when computing initial pacing rate.  2.885 is an old version of the startup pacing gain value.
* fix argument name to BBRv2::new to initial_rtt from smoothed_rtt.  sRTT is not known this early in the connection
* Fix BBR initial pacing rate calculation for consistency with draft-ietf-ccwg-bbr
* Add option to override initial pacing rate

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -354,7 +354,7 @@ fn initial_pacing_rate(cwnd_in_bytes: usize, params: &Params) -> Bandwidth {
     }
 
     // sRTT is not known yet.  Compute the initial pacing rate by
-    // dividing cwnd by 1msec as recommented by draft-ietf-ccwg-bbr-03.
+    // dividing cwnd by 1msec as recommended by draft-ietf-ccwg-bbr-03.
     // https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-03.html#name-pacing-rate-cpacing_rate
     Bandwidth::from_bytes_and_time_delta(cwnd_in_bytes, Duration::from_millis(1)) *
         params.startup_pacing_gain
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -173,6 +173,12 @@ struct Params {
     /// Determines whether app limited rounds with no bandwidth growth count
     /// towards the rounds threshold to exit startup.
     ignore_app_limited_for_no_bandwidth_growth: bool,
+
+    /// Initial pacing rate for a new connection before an RTT
+    /// estimate is available.  This rate serves as an upper bound on
+    /// pacing rate calculated by dividing the initial cwnd by the
+    /// first RTT estimate.
+    initial_pacing_rate_bytes_per_second: Option<u64>,
 }
 
 impl Params {
@@ -181,6 +187,14 @@ impl Params {
             ($field:ident) => {
                 if let Some(custom_value) = custom_bbr_settings.$field {
                     self.$field = custom_value;
+                }
+            };
+        }
+
+        macro_rules! apply_optional_override {
+            ($field:ident) => {
+                if let Some(custom_value) = custom_bbr_settings.$field {
+                    self.$field = Some(custom_value);
                 }
             };
         }
@@ -202,6 +216,7 @@ impl Params {
         apply_override!(use_bytes_delivered_for_inflight_hi);
         apply_override!(decrease_startup_pacing_at_end_of_round);
         apply_override!(ignore_app_limited_for_no_bandwidth_growth);
+        apply_optional_override!(initial_pacing_rate_bytes_per_second);
 
         if let Some(custom_value) = custom_bbr_settings.bw_lo_reduction_strategy {
             self.bw_lo_mode = custom_value.into();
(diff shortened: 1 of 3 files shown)


quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -331,6 +348,18 @@ impl<T: Ord + Clone + Copy + From<u8>> Limits<T> {
     }
 }
 
+fn initial_pacing_rate(cwnd_in_bytes: usize, params: &Params) -> Bandwidth {
+    if let Some(pacing_rate) = params.initial_pacing_rate_bytes_per_second {
+        return Bandwidth::from_bytes_per_second(pacing_rate);
+    }
+
+    // sRTT is not known yet.  Compute the initial pacing rate by
+    // dividing cwnd by 1msec as recommented by draft-ietf-ccwg-bbr-03.
```
LPardue:
```suggestion
    // dividing cwnd by 1msec as recommended by draft-ietf-ccwg-bbr-03.
```

LPardue (dismissed):
lgtm modulo a typo

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -422,7 +422,7 @@ impl BBRv2 {
             mode: Mode::startup(BBRv2NetworkModel::new(&params, smoothed_rtt)),
             cwnd,
             pacing_rate: Bandwidth::from_bytes_and_time_delta(cwnd, smoothed_rtt) *
-                2.885,
+                params.startup_pacing_gain,
```
toidiu: This is changing the default value but quiche is used by other teams so are we sure its a safe change? Where did the previous value come from?

---
Changes to default values should be in a separate commit.

Can you add a comment [here](https://github.com/cloudflare/quiche/blob/37defda36f23a351eacd36667eda0b6edde72d44/quiche/src/recovery/gcongestion/bbr2.rs#L232C26-L232C32) specify where that value is coming from (link to kernel source or bbr spec)

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -421,8 +421,13 @@ impl BBRv2 {
...
             cwnd,
-            pacing_rate: Bandwidth::from_bytes_and_time_delta(cwnd, initial_rtt) *
-                params.startup_pacing_gain,
+            // sRTT is not known yet.  Compute the initial pacing rate by
+            // dividing cwnd by 1msec as recommented by draft-ietf-ccwg-bbr-03.
+            // https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-03.html#name-pacing-rate-cpacing_rate
+            pacing_rate: Bandwidth::from_bytes_and_time_delta(
+                cwnd,
+                Duration::from_millis(1),
+            ) * params.startup_pacing_gain,
```
toidiu: From what I can tell this is changing behavior since we currently use [INITIAL_RTT](https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L447). The RFC logic does `SRTT ? SRTT : 1ms` but if we are using INITIAL_RTT I don't expect the 1ms to be dead code.

---

Also can you add the pseudo code as a comment here if we want this behavior.
```
  BBRInitPacingRate():
    nominal_bandwidth = C.InitialCwnd / (SRTT ? SRTT : 1ms)
    C.pacing_rate =  BBR.StartupPacingGain * nominal_bandwidth
```

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -173,6 +173,12 @@ struct Params {
     /// Determines whether app limited rounds with no bandwidth growth count
     /// towards the rounds threshold to exit startup.
     ignore_app_limited_for_no_bandwidth_growth: bool,
+
+    /// Initial pacing rate for a new connection before an RTT
+    /// estimate is available.  This rate serves as an upper bound on
+    /// pacing rate calculated by dividing the initial cwnd by the
+    /// first RTT estimate.
+    initial_pacing_rate_bytes_per_second: Option<u64>,
```
toidiu, line 181: > This rate serves as an upper bound on
> pacing rate calculated by dividing the initial cwnd by the
> first RTT estimate.

Might be missing something but how is this enforcing an upper bound? It seems to override the value based on my read of the code below.

```
    if let Some(pacing_rate) = params.initial_pacing_rate_bytes_per_second {
        return Bandwidth::from_bytes_per_second(pacing_rate);
    }
```

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: There are 4 commits in this PR in an attempt to separate semantic fixes, from spec consistency and new parameters.  I split the last commit into a separate PR since it is the change I care the most about at this time and hopefully not a controversial change.  See https://github.com/cloudflare/quiche/pull/2115

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: An initial RTT estimate of 333ms is not a valid sRTT measurement.  It is the absence of a measurement, so the code should really be using 1msec when computing the initial pacing rate.

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: See https://github.com/cloudflare/quiche/pull/2115#discussion_r2237998067

antoniovicente:
Abandoning the rest of the changes for now that https://github.com/cloudflare/quiche/pull/2115 is merged.  May revisit the rest of the changes in a few weeks.

PR #2115, opened 2025-07-28
Add option to override initial pacing rate

antoniovicente:
Extracted from https://github.com/cloudflare/quiche/pull/2113, while not addressing the more controversial fixes to reduce the delta between the implementation and the BBR draft RFC.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -173,6 +173,12 @@ struct Params {
     /// Determines whether app limited rounds with no bandwidth growth count
     /// towards the rounds threshold to exit startup.
     ignore_app_limited_for_no_bandwidth_growth: bool,
+
+    /// Initial pacing rate for a new connection before an RTT
+    /// estimate is available.  This rate serves as an upper bound on
+    /// the initial pacing rate, which is calculated by dividing the
+    /// initial cwnd by the first RTT estimate.
+    initial_pacing_rate_bytes_per_second: Option<u64>,
 }
 
 impl Params {
@@ -181,6 +187,14 @@ impl Params {
             ($field:ident) => {
                 if let Some(custom_value) = custom_bbr_settings.$field {
                     self.$field = custom_value;
+                }
+            };
+        }
+
+        macro_rules! apply_optional_override {
+            ($field:ident) => {
+                if let Some(custom_value) = custom_bbr_settings.$field {
+                    self.$field = Some(custom_value);
                 }
             };
         }
@@ -202,6 +216,7 @@ impl Params {
         apply_override!(use_bytes_delivered_for_inflight_hi);
         apply_override!(decrease_startup_pacing_at_end_of_round);
         apply_override!(ignore_app_limited_for_no_bandwidth_growth);
+        apply_optional_override!(initial_pacing_rate_bytes_per_second);
 
         if let Some(custom_value) = custom_bbr_settings.bw_lo_reduction_strategy {
             self.bw_lo_mode = custom_value.into();
@@ -283,6 +298,8 @@ const DEFAULT_PARAMS: Params = Params {
     bw_lo_mode: BwLoMode::InflightReduction,
 
     ignore_app_limited_for_no_bandwidth_growth: false,
+
+    initial_pacing_rate_bytes_per_second: None,
 };
 
 #[derive(Debug, PartialEq)]
@@ -329,6 +346,16 @@ impl<T: Ord + Clone + Copy + From<u8>> Limits<T> {
             hi: val,
         }
     }
+}
+
+fn initial_pacing_rate(
+    cwnd_in_bytes: usize, smoothed_rtt: Duration, params: &Params,
+) -> Bandwidth {
+    if let Some(pacing_rate) = params.initial_pacing_rate_bytes_per_second {
+        return Bandwidth::from_bytes_per_second(pacing_rate);
+    }
+
+    Bandwidth::from_bytes_and_time_delta(cwnd_in_bytes, smoothed_rtt) * 2.885
 }
 
 #[derive(Debug)]
@@ -421,8 +448,7 @@ impl BBRv2 {
         BBRv2 {
             mode: Mode::startup(BBRv2NetworkModel::new(&params, smoothed_rtt)),
             cwnd,
-            pacing_rate: Bandwidth::from_bytes_and_time_delta(cwnd, smoothed_rtt) *
-                2.885,
+            pacing_rate: initial_pacing_rate(cwnd, smoothed_rtt, &params),
             cwnd_limits: Limits {
                 lo: initial_congestion_window * max_segment_size,
                 hi: max_congestion_window * max_segment_size,
@@ -464,6 +490,18 @@ impl BBRv2 {
                 self.cwnd,
                 self.mode.min_rtt(),
             );
+
+            if let Some(pacing_rate) =
+                self.params.initial_pacing_rate_bytes_per_second
+            {
+                // Do not allow the pacing rate calculated from the first RTT
+                // measurement to be higher than the configured initial pacing
+                // rate.
+                let initial_pacing_rate =
+                    Bandwidth::from_bytes_per_second(pacing_rate);
+                self.pacing_rate = self.pacing_rate.min(initial_pacing_rate);
+            }
+
             return;
         }
 
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -1669,6 +1669,129 @@ mod tests {
         assert_eq!(r.startup_exit(), None);
     }
 
+    fn helper_initial_pacing_rate_override_test(
+        initial_pacing_rate_hint: Bandwidth,
+        expected_pacing_with_rtt_measurement: Bandwidth,
+    ) {
+        let cc_algorithm_name = "bbr2_gcongestion";
+        let mut cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        assert_eq!(cfg.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
+        cfg.set_custom_bbr_params(BbrParams {
+            initial_pacing_rate_bytes_per_second: Some(
+                initial_pacing_rate_hint.to_bytes_per_second(),
+            ),
+            ..Default::default()
+        });
+
+        let mut r = Recovery::new(&cfg);
+
+        let mut now = Instant::now();
+
+        assert_eq!(r.sent_packets_len(packet::Epoch::Application), 0);
+
+        // send out first packet burst (a full initcwnd).
+        for i in 0..10 {
+            let p = Sent {
+                pkt_num: i,
+                frames: smallvec![],
+                time_sent: now,
+                time_acked: None,
+                time_lost: None,
+                size: 1200,
+                ack_eliciting: true,
+                in_flight: true,
+                delivered: 0,
+                delivered_time: now,
+                first_sent_time: now,
+                is_app_limited: false,
+                tx_in_flight: 0,
+                lost: 0,
+                has_data: true,
+                is_pmtud_probe: false,
+            };
+
+            r.on_packet_sent(
+                p,
+                packet::Epoch::Application,
+                HandshakeStatus::default(),
+                now,
+                "",
+            );
+        }
+
+        assert_eq!(r.sent_packets_len(packet::Epoch::Application), 10);
+        assert_eq!(r.bytes_in_flight(), 12000);
+        assert_eq!(r.bytes_in_flight_duration(), Duration::ZERO);
+
+        // Initial pacing rate matches the override value.
+        assert_eq!(
+            r.pacing_rate(),
+            initial_pacing_rate_hint.to_bytes_per_second()
+        );
+        assert_eq!(r.get_packet_send_time(now), now);
+
+        assert_eq!(r.cwnd(), 12000);
+        assert_eq!(r.cwnd_available(), 0);
+
+        // Wait 50ms for ACK.
+        now += Duration::from_millis(50);
+
+        let mut acked = RangeSet::default();
+        acked.insert(0..10);
+
+        assert_eq!(
+            r.on_ack_received(
+                &acked,
+                10,
+                packet::Epoch::Application,
+                HandshakeStatus::default(),
+                now,
+                None,
+                "",
+            )
+            .unwrap(),
+            OnAckReceivedOutcome {
+                lost_packets: 0,
+                lost_bytes: 0,
+                acked_bytes: 12000,
+                spurious_losses: 0,
+            }
+        );
+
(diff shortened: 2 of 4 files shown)


quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -464,6 +490,18 @@ impl BBRv2 {
                 self.cwnd,
                 self.mode.min_rtt(),
             );
+
+            if let Some(pacing_rate) =
+                self.params.initial_pacing_rate_bytes_per_second
+            {
+                // Do not allow the pacing rate calculated from the first RTT
+                // measurement to be higher than the configured initial pacing
+                // rate.
```
antoniovicente, line 499: Added this comment explaining the second meaning of the initial_pacing_rate_bytes_per_second parameter: serving as a max initial pacing rate

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -331,6 +348,16 @@ impl<T: Ord + Clone + Copy + From<u8>> Limits<T> {
...
 
+fn initial_pacing_rate(
+    cwnd_in_bytes: usize, smoothed_rtt: Duration, params: &Params,
+) -> Bandwidth {
+    if let Some(pacing_rate) = params.initial_pacing_rate_bytes_per_second {
+        return Bandwidth::from_bytes_per_second(pacing_rate);
+    }
+
+    Bandwidth::from_bytes_and_time_delta(cwnd_in_bytes, smoothed_rtt) * 2.885
+}
```
toidiu, line 359: nit:

```suggestion
fn initial_pacing_rate(
    cwnd_in_bytes: usize, smoothed_rtt: Duration, params: &Params,
) -> Bandwidth {
    if let Some(pacing_rate) = params.initial_pacing_rate_bytes_per_second {
        Bandwidth::from_bytes_per_second(pacing_rate)
    } else {
        Bandwidth::from_bytes_and_time_delta(cwnd_in_bytes, smoothed_rtt) * 2.885
    }
}
```

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -173,6 +173,12 @@ struct Params {
     /// Determines whether app limited rounds with no bandwidth growth count
     /// towards the rounds threshold to exit startup.
     ignore_app_limited_for_no_bandwidth_growth: bool,
+
+    /// Initial pacing rate for a new connection before an RTT
+    /// estimate is available.  This rate serves as an upper bound on
+    /// pacing rate calculated by dividing the initial cwnd by the
+    /// first RTT estimate.
```
toidiu: nit: is this correct? we are only trying to cap the initial pacing rate right?

```suggestion
    /// Initial pacing rate for a new connection before an RTT
    /// estimate is available. This rate serves as an upper bound on
    /// the initial pacing rate, which is calculated by dividing the initial cwnd by the
    /// first RTT estimate.
```

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
         assert_eq!(r.startup_exit(), None);
     }
 
+    fn initial_pacing_rate_override_test(
```
toidiu: nit: could we add a `helper` prefix and move this to the [bottom](https://github.com/cloudflare/quiche/blob/master/quiche/src/test_utils.rs#L525) of the test mod.

```suggestion
    fn helper_initial_pacing_rate_override_test(
```

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+
+        // Pacing rate is recalculated based on initial cwnd when the
+        // first RTT estimate is available.
+        assert_eq!(
+            r.pacing_rate(),
+            expected_pacing_with_rtt_measurement.to_bytes_per_second()
+        );
+    }
+
+    #[rstest]
```
toidiu, line 1773:
```suggestion
    #[test]
```

quiche/src/recovery/gcongestion/bbr2.rs
toidiu: Are we sure this is only called during the initial calculation? Its gated on `self.mode.total_bytes_acked() == bytes_acked` which could be called after also?

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: I used to advocate for code like this, but over time I have been pushed towards a coding style that discourages unnecessary nesting.  The code turns up more readable and less clustered.

The extreme version of what you're suggesting may be something along the lines of:

`params.initial_pacing_rate_bytes_per_second.map(Bandwidth::from_bytes_per_second).unwrap_or(Bandwidth::from_bytes_and_time_delta(cwnd_in_bytes, smoothed_rtt) * 2.885)`

quiche/src/recovery/mod.rs
antoniovicente: What's the benefit?  All tests in this file use rstest today.

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: I do believe this code only runs once per connection on the assumption that acks are always have a non zero number of bytes associated with them.  The comments in the code do suggest that this block only runs very early in the connection, before there are changes to the initial window used at the beginning of the connection.

quiche/src/recovery/mod.rs

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: Done.

quiche/src/recovery/mod.rs
toidiu: Not a big deal. Lets keep rstest

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: Consider that `total_bytes_acked` should be the `sum of all bytes_acked` on a QUIC connection path.

`self.mode.total_bytes_acked() == bytes_acked` should only happen when processing an ACK where the previous state of total_bytes_acked was 0.

quiche/src/recovery/gcongestion/bbr2.rs
toidiu: Primarily I dislike early `return` statements but balancing then with deep nesting should also be considered. In this case since you already have an if clause, the else seemed a bit better.

The `map` feel way more confusing to read.

But its a `nit` and i'll leave it up to you.

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+    fn initial_pacing_rate_override_decrease_when_rtt_measured() {
+        let bw = Bandwidth::from_bytes_and_time_delta(
+            12000,
+            Duration::from_millis(50),
+        );
+        helper_initial_pacing_rate_override_test(bw * 2.0, bw);
+    }
+
+    #[rstest]
```
LPardue: Do we need rstest if the test is not parameterized?

quiche/src/recovery/gcongestion/bbr2.rs
toidiu: > // After the first ACK, cwnd is still the initial congestion window.

Yep comment from [here](https://github.com/cloudflare/quiche/pull/2115/files#diff-ad4b2a983ab827f21b0de5b2449da3abde4e40ef3415342499f12322e92b2bd9R488) suggests its run only once.. just wish the code was easier to reason about :\

Thx for looking into this, lets move forward with this.

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
         assert_eq!(r.startup_exit(), None);
     }
 
+    fn helper_initial_pacing_rate_override_test(
+        initial_pacing_rate_hint: Bandwidth,
+        expected_pacing_with_rtt_measurement: Bandwidth,
+    ) {
+        let cc_algorithm_name = "bbr2_gcongestion";
+        let mut cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        assert_eq!(cfg.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
```
toidiu, line 1695:
```suggestion
        cfg.set_cc_algorithm_name(cc_algorithm_name).unwrap();
```

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+        });
+
+        let mut r = Recovery::new(&cfg);
+
+        let mut now = Instant::now();
+
+        assert_eq!(r.sent_packets_len(packet::Epoch::Application), 0);
+
+        // send out first packet burst (a full initcwnd).
+        for i in 0..10 {
```
toidiu: Can we use the named default value here and below.

```suggestion
        for i in 0..DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS {
```

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+            let p = Sent {
```
toidiu: Possible to use [helper](https://github.com/cloudflare/quiche/blob/83b1d638c3688d1f2e14e1b158c82f6d62e8b5e0/quiche/src/test_utils.rs#L525) here and improve readibility?

quiche/src/recovery/mod.rs
antoniovicente: Both worked.  Changed to `#[test]`

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+                lost_packets: 0,
+                lost_bytes: 0,
+                acked_bytes: 12000,
+                spurious_losses: 0,
+            }
+        );
+
+        assert_eq!(r.sent_packets_len(packet::Epoch::Application), 0);
+        assert_eq!(r.bytes_in_flight(), 0);
+        assert_eq!(r.bytes_in_flight_duration(), Duration::from_millis(50));
```
toidiu: `Duration::from_millis(50);` is used a few times. Can you pull it out into a variable. Also if it need to match the `Duration::from_millis(50)` in the [actual test](https://github.com/cloudflare/quiche/pull/2115/files#diff-61276adac12565ae851ac862eca24f85ae650e9461acefe87e12125a256805b9R1778), lets pass it in as a variable.

LPardue (approved):

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,129 @@ mod tests {
...
+    fn initial_pacing_rate_override_decrease_when_rtt_measured() {
+        let bw = Bandwidth::from_bytes_and_time_delta(
+            12000,
+            Duration::from_millis(50),
+        );
+        helper_initial_pacing_rate_override_test(bw * 2.0, bw);
+    }
+
+    #[test]
+    fn initial_pacing_rate_override_does_not_increase_when_rtt_measured() {
```
toidiu: It seems like the rtt stays the same here `bw * 0.5`. Also do we need 3 test cases "same", "decrease", "increase"?

quiche/src/recovery/mod.rs
antoniovicente: I could go either way on this one...

Debug output from the current version:
```
assertion `left == right` failed
  left: Err(CongestionControl)
 right: Ok(())
```

Your version:
```
called `Result::unwrap()` on an `Err` value: CongestionControl
```

I guess these two provide more or less the same information: The error was of type `CongestionControl`.


Last let's consider: `let _ = cfg.set_cc_algorithm_name(cc_algorithm_name);`

The similarity between this last one and `cfg.set_cc_algorithm_name(cc_algorithm_name).unwrap();` is what makes me hesitate against the `.unwrap()` version.  I'm honestly sad about the `let _ =` being allowed.

quiche/src/recovery/mod.rs
antoniovicente: Changed this to 2 packets since I think it provides better insight into what is happening.  In rate calculations we do not depend on the number of packets or bytes sent, we depend only on the initial window, first RTT measurement, and the initial pacing rate.

quiche/src/recovery/mod.rs
antoniovicente: Done.  I had to adjust the helper to set `has_data: true` for it to work correctly.

quiche/src/recovery/mod.rs

quiche/src/recovery/mod.rs
antoniovicente: Refactored to make this a parametrized test that covers the same, decrease and increase cases.

quiche/src/recovery/mod.rs
toidiu: Sry, I should have marked this as a `nit` since we do the `assert` pattern in other parts of the code.

The cc_algo name is not the focus of this test so I would be inclined to just do the unwrap and go with the less verbose version.

rushilmehra (dismissed):

rushilmehra (approved):

quiche/src/recovery/mod.rs
```diff
@@ -1669,6 +1669,104 @@ mod tests {
...
+    #[rstest]
+    // initial_cwnd / first_rtt == initial_pacing_rate.  Pacing is 1.0 * bw before
+    // and after.
+    #[case::bw_estimate_equal_after_first_rtt(1.0, 1.0)]
+    // initial_cwnd / first_rtt < initial_pacing_rate.  Pacing decreases from 2 *
+    // bw to 1.0 * bw.
+    #[case::bw_estimate_decrease_after_first_rtt(2.0, 1.0)]
+    // initial_cwnd / first_rtt > initial_pacing_rate from 0.5 * bw to 1.0 * bw.
+    // Initial pacing remains 0.5 * bw.
+    #[case::bw_estimate_increase_after_first_rtt(0.5, 0.5)]
```
toidiu: Seems like this is not really measuring the `>` case (i.e. 0.5 vs 0.5) because in the first round you are capped by the override delivery rate.

Is it worth sending a second round to exercise the increase in the expected value. The math does get less nice and test more ugly however.

antoniovicente: This test is verifying that `self.pacing_rate = self.pacing_rate.min(initial_pacing_rate);` does what I expect it should.

antoniovicente: Also consider that bw measurements can never be larger than `cwnd / rtt` since you can only ever send cwnd per RTT.  In order to increase bw estimates, either cwnd must increase or rtt has to decrease.

toidiu: Yep these make sense. The 0.5 vs 0.5 is confusing without this context. Could you add a comment next to the test explaining what you described here.

toidiu (approved):

PR #2121, opened 2025-08-07
add qlog event definitions for Careful Resume

Finic8:
This adds the qlog event definitions for [Careful Resume](https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-careful-resume) according to the [draft](https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-careful-resume-qlog).

### DIFF
--- a/qlog/src/events/mod.rs
+++ b/qlog/src/events/mod.rs
@@ -262,6 +262,10 @@ impl From<EventType> for EventImportance {
             EventType::QuicEventType(QuicEventType::EcnStateUpdated) =>
                 EventImportance::Extra,
 
+            EventType::QuicEventType(
+                QuicEventType::CarefulResumePhaseUpdated,
+            ) => EventImportance::Extra,
+
             EventType::Http3EventType(Http3EventType::ParametersSet) =>
                 EventImportance::Base,
             EventType::Http3EventType(Http3EventType::StreamTypeSet) =>
@@ -374,6 +378,9 @@ impl From<&EventData> for EventType {
                 EventType::QuicEventType(QuicEventType::MarkedForRetransmit),
             EventData::QuicEcnStateUpdated { .. } =>
                 EventType::QuicEventType(QuicEventType::EcnStateUpdated),
+
+            EventData::CarefulResumePhaseUpdated { .. } =>
+                EventType::QuicEventType(QuicEventType::CarefulResumePhaseUpdated),
 
             EventData::Http3ParametersSet { .. } =>
                 EventType::Http3EventType(Http3EventType::ParametersSet),
@@ -545,6 +552,9 @@ pub enum EventData {
     #[serde(rename = "quic:ecn_state_updated")]
     QuicEcnStateUpdated(quic::EcnStateUpdated),
 
+    #[serde(rename = "recovery:careful_resume_phase_updated")]
+    CarefulResumePhaseUpdated(resume::CarefulResumePhaseUpdated),
+
     // HTTP/3
     #[serde(rename = "http3:parameters_set")]
     Http3ParametersSet(http3::ParametersSet),
@@ -696,3 +706,4 @@ pub struct TupleEndpointInfo {
 
 pub mod http3;
 pub mod quic;
+pub mod resume;
--- a/qlog/src/events/mod.rs
+++ b/qlog/src/events/mod.rs
@@ -552,7 +552,7 @@ pub enum EventData {
     #[serde(rename = "quic:ecn_state_updated")]
     QuicEcnStateUpdated(quic::EcnStateUpdated),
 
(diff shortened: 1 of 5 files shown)


qlog/src/events/mod.rs
```diff
@@ -276,6 +276,10 @@ impl From<EventType> for EventImportance {
                 RecoveryEventType::MarkedForRetransmit,
             ) => EventImportance::Extra,
 
+            EventType::RecoveryEventType(
+                RecoveryEventType::CarefulResumePhaseUpdated,
```
LPardue: This isn't strictly a recovery event in qlog terms (although its somewhat complicated by the this crate lagging behind the latest specs.

On balance, I think it would be better to define a different EventType for careful resume. I.e. defining a separate enum in [qlog/src/events/quic.rs](https://github.com/cloudflare/quiche/pull/2121/files#diff-6674de9cf1901c352c7eda222350674ca2c177b89a965d0299d7b5bef8449150)

qlog/src/events/resume.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum CarefulResumeTrigger {
+    /// From [Reconnaissance][CarefulResumePhase::Reconnaissance] to
+    /// [Unvalidated][CarefulResumePhase::Unvalidated]
+    ///
+    /// When sender has confirmed the RTT, has received an ACK for the initial
+    /// data without reported congestion and has more data to send than the CWND
+    /// would allow.
+    #[serde(rename = "cwnd_limited")]
```
LPardue: This doesn't seem to match the name in the spec: https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-careful-resume-qlog#section-3.3.1

qlog/src/events/resume.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
+pub struct CarefulResumeStateParameters {
+    pub pipesize: u64,
+    pub first_unvalidated_packet: u64,
+    pub last_unvalidated_packet: u64,
+    pub congestion_window: Option<u64>,
+    pub ssthresh: Option<u64>,
```
LPardue, line 125: spec says all these are u32, not u64

LPardue:
Hey @Finic8, apologies for the delay. We've now landed a major qlog schema bump and we're ready to get careful resume qlog support landed properly. If you can find the time to rebase and resolve conflicts, we can get this looked at quickly

qlog/src/events/resume.rs
Finic8: We have changed the spec to use u64, to be consistant with [quic-qlog-events](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-quic-events-11).

qlog/src/events/resume.rs
Finic8: Fixed in lateste commits.

qlog/src/events/mod.rs
Finic8: I have changed the namespace from `recovery:careful_resume_phase_updated` to `quic:careful_resume_phase_updated` as written in the spec.

Finic8:
I've rebased the PR now onto the latest master and made sure it matches the latest [careful-resume-qlog](https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-careful-resume-qlog) spec.

qlog/src/events/resume.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub old: Option<CarefulResumePhase>,
+    pub new: CarefulResumePhase,
+    pub state_data: CarefulResumeStateParameters,
+    pub restored_data: Option<CarefulResumeRestoredParameters>,
+    pub trigger: Option<CarefulResumeTrigger>,
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum CarefulResumeTrigger {
```
LPardue, line 16: nit: can we order these the same as in https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-careful-resume-qlog#section-3.3.1 please? That makes it a bit easier to maintain

LPardue:
Thanks, overall this looks on the right track. There's one nit to address. However, more importantly, the careful resume qlog spec is missing defining a schema value, which we'd want to support in order to align with the core qlog requirments. I've emailed Gorry and Ana to see if we can get one defined in the doc so we can add it here,.

Pnce those minor things are done, the PR is likely good to go

PR #2126, opened 2025-08-14
Clean up streams in `h3::Connection` when done

maurer:
Currently, while `quiche::Connection` will clean up streams from its `StreamMap`, `quiche::h3::Connection` does not. As a result, long lived connections will slowly leak memory through the stream map in the connection.

This patch forwards the `collect()` event from the `quiche::Connection` to the `quiche::h3::Connection` through the `h3::Connection::poll()` call, allowing the `h3::Connection` to clean up its unused streams.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -2070,6 +2070,10 @@ impl Connection {
             return Ok((finished, Event::Finished));
         }
 
+        for stream_id in conn.streams.collected_streams() {
+            self.streams.remove(&stream_id);
+        }
+
         Err(Error::Done)
     }
 
@@ -3210,6 +3214,13 @@ impl Connection {
                 .peer_qpack_streams
                 .decoder_stream_bytes,
         }
+    }
+
+    #[cfg(test)]
+    /// Indicates how many streams are currently being tracked for use
+    /// in resource tests.
+    pub fn stream_count(&self) -> usize {
+        self.streams.len()
     }
 }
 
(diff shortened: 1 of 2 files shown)


ghedo:
Thnak you for the PR, and sorry for the delay. Looking at this my main worry was that an application not using the quiche h3 implementation (e.g. implementing a different application protocol on top of QUIC) would have to drain the collected streams queue itself or risk a memory leak of its own. That said, given that we keep a map of collected streams anyway this is probably not that big of an issue.

We do probably want to expose that API to applications directly, though an issue there would be both application and quiche::h3 draining the queue which could lead to leaking memory anyway. We'll need to think about how the API should work a bit more maybe.

I also spent some time looking at a different fix that wouldn't need to keep a queue of streams (https://github.com/cloudflare/quiche/compare/clean-streams-2?expand=1) though that doesn't quite handle some edge cases very well yet, so it wouldn't be a full replacement of this PR I think.

maurer:
Sorry, mixup with the git branch accidentally closed things.

I tried some things along the lines of what you were tracking with `local_finished` where I tried to avoid adding variable-sized data to the structure, but could never get it to actually stop the memory growth in practice. It sounds like you already understand what some of these cases are. If not, I could try to extend my test cases to produce them.

The trouble I ran into without actually having a list of pending streams was that it didn't seem like I had enough information at the `h3` module to figure out when a stream was actually done. I'd either fail to collect idle-but-completed streams (our DoH app would continue to grow), or I'd be too aggressive and remove things that your test suite forbids (this did work in practice for our app though).

If you get to the point with your alternate implementation that you'd like me to test it for leakage on our test app I'll give it a shot. If you're curious about the how we're using it, something very close to its current state is [here](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/DnsResolver/doh/). The leak case occurs when an extremely reliable network connection keeps making queries - every query ends up forcing `h3::Conn` to keep the stream for that query alive.

quiche/src/h3/mod.rs
```diff
@@ -7349,6 +7360,68 @@ mod tests {
...
     }
+
+    #[test]
+    fn h3_cleanup_streams() {
+        let mut buf = [0; 65535];
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        let init_streams_client = s.client.stream_count();
+        let init_streams_server = s.client.stream_count();
```
LPardue, line 7371: Think this should be server?

```suggestion
        let init_streams_server = s.server.stream_count();
```

quiche/src/stream/mod.rs
```diff
@@ -647,6 +655,12 @@ impl<F: BufFactory> StreamMap<F> {
                 self.local_max_streams_uni - self.peer_opened_streams_uni
     }
 
+    /// Return streams we have collected but not yet notified the transport
```
LPardue, line 658: I don't understand what "not yet notified the transport" means here? The quiche transport lib is the thing that would be calling `collect()` that would trigger adding to `pending_collected`.

quiche/src/h3/mod.rs
```diff
@@ -7349,6 +7360,68 @@ mod tests {
...
+        assert_eq!(s.recv_body_client(stream, &mut recv_buf), Ok(body.len()));
+
+        // The server stream should be gone now
+        assert_eq!(s.client.stream_count(), init_streams_client + 1);
+        assert_eq!(s.server.stream_count(), init_streams_server);
+
+        // Polling again should clean up the client
+        assert_eq!(s.poll_client(), Ok((stream, Event::Finished)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+        assert_eq!(s.client.stream_count(), init_streams_client);
```
LPardue, line 7423: I'd like to see tests that cover reset and stopped streams cases too

PR #2127, opened 2025-08-19
Add startup_full_bw_rounds to BbrParams internal override

antoniovicente:


quiche/src/recovery/gcongestion/mod.rs
```diff
@@ -151,6 +151,10 @@ pub struct BbrParams {
     /// Controls the BBR full bandwidth threshold.
     pub full_bw_threshold: Option<f32>,
 
+    /// Controls the number of rounds to stay in STARTUP before
+    /// exiting due to bandwidth plateau.
```
cfjubacf, line 155: nit: would it make sense to specify in the comment that the default is 3?

cfjubacf (approved):

quiche/src/recovery/gcongestion/mod.rs
antoniovicente: The default being a struct and a constant removed from this comment makes it very likely for comments and default values to drift.  Granted, the default here comes directly from the BBR RFC draft so there's not that much of a chance for drift.

It would be nice if we could set defaults here directly, but unfortunately serde doesn't support things like `#[serde(default_value="0.1")]`

https://github.com/serde-rs/serde/issues/368

PR #2128, opened 2025-08-19
BbrParams for BBR ProbeBw and ProbeRtt stages

antoniovicente:
* Split ProbeBW cwnd_gain parameter for ProbeBW_UP phase from param used for other ProbeBW phases.
* Add parameters for ProbeRTT pacing gain and cwnd gain.

Some of these params are inconsistent with the BBR RFC draft.  Having them as config is the first step towards getting the inconsistency fixed.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2/mode.rs
+++ b/quiche/src/recovery/gcongestion/bbr2/mode.rs
@@ -56,11 +56,18 @@ pub(super) enum CyclePhase {
 }
 
 impl CyclePhase {
-    pub(super) fn gain(&self, params: &Params) -> f32 {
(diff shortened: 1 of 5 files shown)


quiche/src/recovery/gcongestion/bbr2/mode.rs
```diff
@@ -56,13 +56,20 @@ pub(super) enum CyclePhase {
 }
 
 impl CyclePhase {
-    pub(super) fn gain(&self, params: &Params) -> f32 {
+    pub(super) fn pacing_gain(&self, params: &Params) -> f32 {
```
toidiu, line 59: yes ty!

toidiu (approved):

PR #2132, opened 2025-08-21
h3i: remove sync feature

ghedo:
When building with no default features (i.e. neither sync not async features are enabled) h3i fails to build. It seems like a strange footgun to have, and since `sync` doesn't really add any more dependencies than `async` there doesn't seem to be much point in being able to disable it anyway.


ghedo:
Merged via https://github.com/cloudflare/quiche/pull/2131 (not sure why GitHub didn't detect this...).

PR #2136, opened 2025-08-22
Implementation of the Ack Frequency draft for delayed ack

mactul:
Hello,

While creating flamegraphs of cURL in HTTP/3 with quiche, I noticed that in certain situations, processing acknowledgements took as long as processing the incoming stream. So I wanted to explore the changes that implementing the [draft for delayed ACKs](https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency-11) would bring.

I couldn't find anything in the quiche code to delay ACKs, except for commit b3e2bb4f7d2d8.

So I'm offering my contribution to fill this gap.

I made sure that as long as the extension is disabled (min_ack_delay parameter not specified), quiche's behaviour remains unchanged.

The extension works well, the tests I have done seem to show that the acks are well delayed and, above all, I observe a difference in the cURL flamegraph in which the impact of sendmsg is greatly reduced.

Unfortunately, the overall performance is not up to par.

The acks sometimes seem to be delayed too much, which triggers a flow control response. Ultimately, downloading 2GB of data sometimes takes 10% longer with the extension enabled. I assume it should be possible to fix my code to resolve this issue, which is why I am opening this PR as a draft.


frochet:
To add a bit more context, @mactul was a summer intern at UNamur, and we aimed at analyzing reasons for the low performance of Curl's usage of HTTP/3 and quiche displayed in @icing's results (https://github.com/icing/blog/blob/main/curl-h3-performance.md). We have seen several corners to cut. In quiche, support of delayed acknowledgement would help Curl's client reducing the cost of  sendmsg() sending acknowledgments.

The Ack frequency draft isn't trivial to implement, and many threshold values are left to the implementer to choose. These may need to be tweaked from further experiments that may involve real world setups. I wonder if you could help us here finishing this PR :-)

Thanks!

PR #2138, opened 2025-08-26
BbrParam to scale pacing rate when MSS changes

antoniovicente:
Pacing rate limits BBR3's ability to measure bandwidth increases during startup.  If we do not scale pacing up, we can't take advantage of bandwidth gains from a higher MTU until the end of the startup phase.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -186,6 +186,9 @@ struct Params {
     /// the initial pacing rate, which is calculated by dividing the
     /// initial cwnd by the first RTT estimate.
     initial_pacing_rate_bytes_per_second: Option<u64>,
+
+    /// If true, scale the pacing rate when updating mss when doing pmtud.
+    scale_pacing_rate_by_mss: bool,
 }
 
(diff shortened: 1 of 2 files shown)


quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -713,6 +719,10 @@ impl CongestionControl for BBRv2 {
             (self.cwnd as u64 * new_mss as u64 / self.mss as u64) as usize;
         self.initial_cwnd = (self.initial_cwnd as u64 * new_mss as u64 /
             self.mss as u64) as usize;
+        if self.params.scale_pacing_rate_by_mss {
+            self.pacing_rate =
```
cfjubacf, line 723: Is it worth it to check if `(new_mss as f64 / self.mss as f64)` is within a certain range before setting the `pacing_rate`?

antoniovicente: MSS can vary between 1200 and a bit less than 1500 bytes.  It is just the packet size discovered by path MTU discovery.

LPardue (approved):

PR #2080, opened 2025-06-13
remove vendored BoringSSL option

ghedo:
We should just use the one provided via boring/boring-sys crates. It should still be possible to build a different BoringSSL version through the facilities provided by those crates.

PR #2142, opened 2025-08-27
WIP: ci: raw cross-build

ghedo:
The stable version of the cargo-cross tool uses an Ubuntu version with an ancient LLVM version which is too old for cbindgen, and newer cargo-cross versions are broken in other ways. So switch to building the cross-compilation environment from scratch.

---

Note that this is a dependency for making cross-builds using the `boring` crate work (which is needed for https://github.com/cloudflare/quiche/pull/2080), just splitting the change into its own PR to make it easier to debug build failures and review.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -221,7 +221,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


ghedo:
Replaced by https://github.com/cloudflare/quiche/pull/2457.

PR #2144, opened 2025-08-27
Add max_bandwidth and startup exit bandwidth stats

antoniovicente:


quiche/src/recovery/congestion/recovery.rs
```diff
@@ -873,6 +873,10 @@ impl RecoveryOps for LegacyRecovery {
         self.congestion.delivery_rate()
     }
 
+    fn max_bandwidth(&self) -> Option<Bandwidth> {
+        None
+    }
```
toidiu: I assume we eventually want to come back and add logic here. Can you add a TODO comment saying that we only support gcongestion at the moment.

quiche/src/recovery/mod.rs
```diff
@@ -637,13 +640,23 @@ pub struct StartupExit {
     /// The congestion_window recorded at Startup exit.
     pub cwnd: usize,
 
+    /// The bandwidth estimate recorded at Startup exit.
+    pub bandwidth: Option<u64>,
+
```
toidiu, line 645: Any reason we cant keep the rich `Bandwidth` type. We should only need to convert it to u64 for logging and in the meantime other uses can benefit from the richer type.

quiche/src/path.rs
```diff
@@ -977,6 +982,11 @@ pub struct PathStats {
     /// [Pacing]: index.html#pacing
     pub delivery_rate: u64,
 
+    /// The maximum bandwidth estimate for the connection in bytes/s.
+    ///
+    /// Note: not all congestion control algorithms provide this metric.
+    pub max_bandwidth: Option<u64>,
```
toidiu: Similar comment as below: can we expose the rich `Bandwidth` type here instead?

> Note: not all congestion control algorithms provide this metric.

Can you specify that only `gcongestion` supports this value.

quiche/src/recovery/mod.rs
antoniovicente: The Bandwidth type is not exposed outside cloudflare/quiche.  For stuff exposed outside this crate we need the simpler type.

quiche/src/recovery/congestion/recovery.rs
antoniovicente: Possibly.  It is a big lift to provide this info for cubic since right now we do not make any attempts to compute bandwidth for cubic/reno.

Ideally we'ld shared the bandwidth computation across all congestion control algorithms when we invest in merging the contents of the congestion and gcongestion directories.

quiche/src/path.rs
antoniovicente: It would require exposing `Bandwidth` outside this crate.  We could consider that refactor in a followup PR.

Added bbr2_gcongestion to the comment.

PR #2145, opened 2025-08-28
fix: dynamically adjust time threshold on packet reordering

toidiu:
Dynamic time threshold logic from
https://github.com/cloudflare/quiche/pull/470 was accidentally removed.

The goal is to make time-based detection less sensitive when a spurious loss is detected.

## Reviewing
It would be best to review this change as separate commits:
- 1st commit adds the time_thresh change
- 2nd commit refactors an existing reordering tests to make it easier to read
- 3rd commit adds the new test to test the time_thresh change

## Testing
I also added the same test to the [previous version (0.21.0)](https://github.com/cloudflare/quiche/compare/akothari/test_time_thresh_0.21.0?expand=1) of code to confirm the behavior.

### network simulator testing
I did some simulation testing to measure the effectiveness of this fix for BBRv3 in the gconsgestion branch. Because reordering is hard to control for there is a lot of variability between runs. However, across ~5 runs each the branch with the fix seems more resilient to packet reordering.

**Setup**
[tc](https://man7.org/linux/man-pages/man8/tc.8.html) based simulator with
- fq used to enable packet pacing
- htb used for rate limiting (`htb quantum 1514 rate 20mbit` + queue `pfifo limit 10800`)
- netem used for delay + packet reordering (`netem limit 2000000 delay 52ms 2ms distribution pareto`). (To eliminate reordering also apply `rate 100gbit`)

**Before fix**
before_time_thresh_fix

**After fix**
after_time_thresh_fix

### DIFF
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -711,7 +711,11 @@ impl RecoveryOps for GRecovery {
         }
 
         if self.newly_acked.is_empty() {
-            return Ok(OnAckReceivedOutcome::default());
+            return Ok(OnAckReceivedOutcome {
+                acked_bytes,
+                spurious_losses,
+                ..Default::default()
+            });
         }
 
         self.bytes_in_flight.saturating_subtract(acked_bytes, now);
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -1,5 +1,7 @@
 use crate::packet;
 use crate::recovery::OnLossDetectionTimeoutOutcome;
+use crate::Error;
+use crate::Result;
 
 use std::collections::VecDeque;
 use std::time::Duration;
@@ -35,8 +37,7 @@ use crate::recovery::INITIAL_TIME_THRESHOLD;
 use crate::recovery::MAX_OUTSTANDING_NON_ACK_ELICITING;
 use crate::recovery::MAX_PACKET_THRESHOLD;
 use crate::recovery::MAX_PTO_PROBES_COUNT;
-use crate::Error;
-use crate::Result;
+use crate::recovery::PACKET_REORDER_TIME_THRESHOLD;
 
 use super::bbr2::BBRv2;
 use super::pacer::Pacer;
@@ -706,6 +707,7 @@ impl RecoveryOps for GRecovery {
         if let Some(thresh) = spurious_pkt_thresh {
             self.pkt_thresh =
                 self.pkt_thresh.max(thresh.min(MAX_PACKET_THRESHOLD));
+            self.time_thresh = PACKET_REORDER_TIME_THRESHOLD;
         }
 
         if self.newly_acked.is_empty() {
@@ -993,6 +995,11 @@ impl RecoveryOps for GRecovery {
     #[cfg(test)]
     fn pkt_thresh(&self) -> u64 {
         self.pkt_thresh
+    }
+
+    #[cfg(test)]
+    fn time_thresh(&self) -> f64 {
+        self.time_thresh
     }
 
     #[cfg(test)]
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -52,6 +52,19 @@ const MAX_PACKET_THRESHOLD: u64 = 20;
 
 const INITIAL_TIME_THRESHOLD: f64 = 9.0 / 8.0;
 
+// Reduce the sensitivity to packet reordering after the first reordering event.
+//
+// Packet reorder is not a real loss event so quickly reduce the sensitivity to
+// avoid penializing subsequent packet reordering.
+//
+// https://www.rfc-editor.org/rfc/rfc9002.html#section-6.1.2
+//
+// Implementations MAY experiment with absolute thresholds, thresholds from
+// previous connections, adaptive thresholds, or the including of RTT variation.
+// Smaller thresholds reduce reordering resilience and increase spurious
+// retransmissions, and larger thresholds increase loss detection delay.
+const PACKET_REORDER_TIME_THRESHOLD: f64 = 5.0 / 4.0;
+
 const GRANULARITY: Duration = Duration::from_millis(1);
 
 const MAX_PTO_PROBES_COUNT: usize = 2;
@@ -251,6 +264,9 @@ pub trait RecoveryOps {
 
     #[cfg(test)]
     fn pkt_thresh(&self) -> u64;
+
+    #[cfg(test)]
+    fn time_thresh(&self) -> f64;
 
     #[cfg(test)]
     fn lost_spurious_count(&self) -> usize;
@@ -1381,6 +1397,7 @@ mod tests {
         acked.insert(0..2);
 
         assert_eq!(r.pkt_thresh(), INITIAL_PACKET_THRESHOLD);
+        assert_eq!(r.time_thresh(), INITIAL_TIME_THRESHOLD);
 
         assert_eq!(
             r.on_ack_received(
@@ -1411,6 +1428,7 @@ mod tests {
 
         // Packet threshold was increased.
         assert_eq!(r.pkt_thresh(), 4);
+        assert_eq!(r.time_thresh(), PACKET_REORDER_TIME_THRESHOLD);
 
         // Wait 1 RTT.
         now += r.rtt();
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -1245,133 +1245,30 @@ mod tests {
         assert_eq!(r.sent_packets_len(packet::Epoch::Application), 0);
 
         // Start by sending a few packets.
-        let p = Sent {
-            pkt_num: 0,
-            frames: smallvec![],
-            time_sent: now,
-            time_acked: None,
-            time_lost: None,
-            size: 1000,
-            ack_eliciting: true,
-            in_flight: true,
-            delivered: 0,
-            delivered_time: now,
-            first_sent_time: now,
-            is_app_limited: false,
-            tx_in_flight: 0,
-            lost: 0,
-            has_data: false,
-            is_pmtud_probe: false,
-        };
+        //
+        // pkt number: [0, 1, 2, 3]
+        for i in 0..4 {
+            let p = test_utils::helper_packet_sent(i, now, 1000);
+            r.on_packet_sent(
+                p,
+                packet::Epoch::Application,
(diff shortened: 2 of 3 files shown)


quiche/src/recovery/mod.rs
```diff
@@ -1398,6 +1415,7 @@ mod tests {
 
         // Packet threshold was increased.
         assert_eq!(r.pkt_thresh(), 4);
+        assert_eq!(r.time_thresh(), PACKET_REORDER_TIME_THRESHOLD);
```
antoniovicente, line 1331: It would be good to have tests in this file that explore the differences in `detect_lost_packets` behavior when time_thresh is `INITIAL_TIME_THRESHOLD` vs `PACKET_REORDER_TIME_THRESHOLD`

I see that the gcongestion branch uses self.rtt_stats.loss_delay() to compute the threshold while the congestion branch rolls it's own loss_delay calculation.

toidiu: Added a test for the time-based loss logic.

quiche/src/recovery/mod.rs
```diff
@@ -1416,6 +1334,175 @@ mod tests {
         }
     }
 
+    // TODO: `congestion` and `gcongestion` behave differently. That might be ok
+    // given the different algorithms but it would be ideal to merge and share
+    // the logic.
```
toidiu: The logic is quite different and the code is many layers deep. We want to merge the two modules but lets do that separately.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -706,10 +707,15 @@ impl RecoveryOps for GRecovery {
...
+            self.time_thresh = PACKET_REORDER_TIME_THRESHOLD;
         }
 
         if self.newly_acked.is_empty() {
-            return Ok(OnAckReceivedOutcome::default());
+            return Ok(OnAckReceivedOutcome {
+                acked_bytes,
+                spurious_losses,
+                ..Default::default()
+            });
```
toidiu, line 718: Found a bug while testing. Without this we dont report the spurious count in the following case:

```
Ack packet: 1
[0, 1, 2, 3, 4, 5]
    ^

Ack packet: 0           // this is spurious loss, but since we didn't ack new packets we return 0 because of `default()`.
[0, 1, 2, 3, 4, 5]
 ^  x
```

antoniovicente: What is the meaning of the ascii art?

quiche/src/recovery/mod.rs
```diff
@@ -1232,133 +1232,30 @@ mod tests {
...
+        }
 
-        // Wait for 10ms.
+        // Wait for 10ms after sending.
         now += Duration::from_millis(10);
 
-        // ACKs are reordered.
+        // Recieve reordered ACKs, i.e. pkt_num [2, 3]
         let mut acked = RangeSet::default();
         acked.insert(2..4);
```
antoniovicente, line 1272: ` acked.insert(2..=3);` would be more clear.  Or  `acked.insert([2, 3]);`

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+        #[values("bbr2_gcongestion")] cc_algorithm_name: &str,
+    ) {
+        let mut cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        assert_eq!(cfg.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
+
+        let mut r = Recovery::new(&cfg);
+
+        assert_eq!(r.rtt(), Duration::from_millis(333));
+        let initial_thresh = 333 as f64 * INITIAL_TIME_THRESHOLD;
+        let spurious_loss_thresh = 333 as f64 * PACKET_REORDER_TIME_THRESHOLD;
```
antoniovicente: Consider using `const` for these two constants.  It's a bit odd to make asserts on them, but it does help with documentation.

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+    #[rstest]
+    fn time_thresholds_on_reordering(
+        #[values("bbr2_gcongestion")] cc_algorithm_name: &str,
+    ) {
+        let mut cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        assert_eq!(cfg.set_cc_algorithm_name(cc_algorithm_name), Ok(()));
+
+        let mut r = Recovery::new(&cfg);
+
+        assert_eq!(r.rtt(), Duration::from_millis(333));
```
antoniovicente: would be good to have a newline after this assert and remove the newline before this assert.

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+        assert_eq!(spurious_loss_thresh.round(), 416.0);
+
+        #[rustfmt::skip]
+        // Pick time between and above thresholds for testing threshold increase.
+        //
+        //```
+        // |--- INITIAL_PACKET_THRESHOLD --- PACKET_REORDER_TIME_THRESHOLD ------- |
+        //                                ^                                  ^
+        //                  (time_between_thresh)    (time_between_thresh + plus_overhead)
+        // ```
```
antoniovicente, line 1376: I don't understand this comment.

Also, too wide - doesn't fit in my review window.

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+        // Wait for `time_between_thresh` after sending to trigger loss based on
+        // loss threshold.
+        now += time_between_thresh;
+
+        // Ack packet: 1
+        //
+        // [0, 1, 2, 3, 4, 5]
+        //     ^
+        let mut acked = RangeSet::default();
+        acked.insert(1..2);
```
antoniovicente, line 1419: `acked.insert(1..=1);` would be more clear

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+        assert_eq!(spurious_loss_thresh.round(), 416.0);
+
+        // Wait for `time_between_thresh` after sending. However, since the
+        // threshold has increased, we do not expect loss.
+        now += time_between_thresh;
+
+        // Ack packet: 3
+        //
+        // [0, 1, 2, 3, 4, 5]
+        //  x  x     ^
```
antoniovicente: Use of 'A' to indicate acked instead of 'x' may be easier to follow.

Consider removing the acked 0 and 1.

`[2, 3, 4, 5]`

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
...
+        #[rustfmt::skip]
+        // Pick time between and above thresholds for testing threshold increase.
+        //
+        //```
+        // |--- INITIAL_PACKET_THRESHOLD --- PACKET_REORDER_TIME_THRESHOLD ------- |
+        //                                ^                                  ^
+        //                  (time_between_thresh)    (time_between_thresh + plus_overhead)
+        // ```
+        let time_between_thresh = Duration::from_millis(400);
+        let plus_overhead = Duration::from_millis(20);
```
antoniovicente: Consider computing `time_between_thresh` based on `initial_thresh` and making assertions on:
- `time_between_thresh > initial_thresh`
- `time_between_thresh < spurious_loss_thresh`
- `time_between_thresh + plus_overhead  > spurious_loss_thresh`

quiche/src/recovery/mod.rs
```diff
@@ -1334,6 +1334,175 @@ mod tests {
         }
     }
 
+    // TODO: `congestion` and `gcongestion` behave differently. That might be ok
+    // given the different algorithms but it would be ideal to merge and share
+    // the logic.
+    #[rstest]
+    fn time_thresholds_on_reordering(
+        #[values("bbr2_gcongestion")] cc_algorithm_name: &str,
```
antoniovicente, line 1357: Was the intent to run this test against multiple congestion algorithms and have alternate assertions for each algorithm?

toidiu: Yep, there is a TODO comment to add other algos. I spent quite some time trying to fix congestion but the logic is too different and would require a large change. I didn't want to block on that.

This will prob be part of the "lets merge the two modules" work.

quiche/src/recovery/mod.rs
toidiu: It wants a [`Range`](https://github.com/cloudflare/quiche/blob/master/quiche/src/ranges.rs#L114) rather than a `RangeInclusive`.

quiche/src/recovery/mod.rs
toidiu: It wants a [`Range`](https://github.com/cloudflare/quiche/blob/master/quiche/src/ranges.rs#L114) rather than a `RangeInclusive`.

quiche/src/recovery/mod.rs
toidiu: The `A` looks too much like the `^` so that wont work. I also tried adding `v` and `A` above the range but that also looks weird. I am inclined to leave it as it is.

+1 on removing `0, 1`

quiche/src/recovery/mod.rs
toidiu: Updated. `mul_f64` is a non-const fn so couldnt convert all into const.

quiche/src/recovery/gcongestion/recovery.rs
toidiu: - first ack packet 1
  - 0 gets marked as lost (time_thresh)
- receive ack for 0 (**this should be a spurious loss event but wasn't before**)

quiche/src/recovery/mod.rs
antoniovicente: The TODO should be more clear about the desire to run this test against other algorithms.

quiche/src/recovery/mod.rs
```diff
@@ -1416,6 +1335,180 @@ mod tests {
...
+        //```
+        //              between_thresh_ms
+        //                         |
+        //    initial_thresh_ms    |     spurious_thresh_ms
+        //      v                  v             v
+        // --------------------------------------------------
+        //      | ................ | ..................... |
+        //            THRESH_GAP         THRESH_GAP
+        // ```
+        // Initial RTT in millis.
```
antoniovicente: Comment seems out of place.  The next line is THRESH_GAP, not RTT related.

antoniovicente:
Looks good.

antoniovicente (approved):

PR #2147, opened 2025-08-30
h3: add support for streaming HEADERS sends

LPardue:
Previously, calls to send_request(), send_response(), send_response_with_priority(),
and send_additional_headers() could fail with a StreamBlocked
occurred, applications were expected to retry the operation when
the stream was later reported as writable. However, certain
conditions could mean that sufficient capacity was never made
available, effectively permenantly blocking header sends.

The root cause of this problem was the choice to enforce that
a HEADERS frame was always sent (buffered into a quiche stream)
in whole.

This change switches to a streaming send design. Methods that
send headers will produce a HEADERS frame that can be sent
in whole or in part, depending on available capacity. When
a frame is only partly sent, applications are notified and

prevented. HEADERS frames must be sent completly before

### DIFF
--- a/quiche/src/h3/ffi.rs
+++ b/quiche/src/h3/ffi.rs
@@ -227,7 +227,7 @@ pub extern "C" fn quiche_h3_send_request(
     let req_headers = headers_from_ptr(headers, headers_len);
 
     match conn.send_request(quic_conn, &req_headers, fin) {
-        Ok(v) => v as i64,
+        Ok((stream_id, _hdrs_fully_sent)) => stream_id as i64,
 
         Err(e) => e.to_c() as i64,
     }
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -113,7 +113,7 @@
 //!     quiche::h3::Header::new(b"user-agent", b"quiche"),
 //! ];
 //!
-//! let stream_id = h3_conn.send_request(&mut conn, &req, false)?;
+//! let (stream_id, _hdrs_completely_sent) = h3_conn.send_request(&mut conn, &req, false)?;
 //! h3_conn.send_body(&mut conn, stream_id, b"Hello World!", true)?;
 //! # Ok::<(), quiche::h3::Error>(())
 //! ```
@@ -424,6 +424,9 @@ pub enum Error {
     /// The requested operation cannot be served over HTTP/3. Peer should retry
     /// over HTTP/1.1.
     VersionFallback,
+
+    /// TODO
+    PartialHeader,
 }
 
 /// HTTP/3 error codes sent on the wire.
@@ -503,6 +506,7 @@ impl Error {
             Error::MessageError => WireErrorCode::MessageError as u64,
             Error::ConnectError => WireErrorCode::ConnectError as u64,
             Error::VersionFallback => WireErrorCode::VersionFallback as u64,
+            Error::PartialHeader => 0x1000,
         }
     }
 
@@ -529,6 +533,7 @@ impl Error {
             Error::MessageError => -18,
             Error::ConnectError => -19,
             Error::VersionFallback => -20,
+            Error::PartialHeader => -21,
 
             Error::TransportError(quic_error) => quic_error.to_c() - 1000,
         }
@@ -1096,12 +1101,18 @@ impl Connection {
 
     /// Sends an HTTP/3 request.
     ///
-    /// The request is encoded from the provided list of headers without a
-    /// body, and sent on a newly allocated stream. To include a body,
-    /// set `fin` as `false` and subsequently call [`send_body()`] with the
-    /// same `conn` and the `stream_id` returned from this method.
+    /// The request is encoded from the provided list of headers without a body,
+    /// and sent on a newly allocated stream. To include a body, set `fin` as
+    /// `false` and subsequently call [`send_body()`] with the same `conn` and
+    /// the `stream_id` returned from this method.
     ///
-    /// On success the newly allocated stream ID is returned.
+    /// On success the newly allocated stream ID is returned, together with a
+    /// bool that indicates if the headers were fully sent or not. If this value
+    /// is false, it indicates the underlying stream or connection does not have
+    /// enough capacity to send the entire remaining header bytes.
+    ///
+    /// The application must call [`continue_partial_headers()`] once the stream
+    /// is reported as writable again.
     ///
     /// The [`StreamBlocked`] error is returned when the underlying QUIC stream
     /// doesn't have enough capacity for the operation to complete. When this
@@ -1109,10 +1120,12 @@ impl Connection {
     /// reported as writable again.
     ///
     /// [`send_body()`]: struct.Connection.html#method.send_body
+    /// [`continue_partial_headers()`]:
+    ///     struct.Connection.html#method.continue_partial_headers
     /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
     pub fn send_request<T: NameValue, F: BufFactory>(
         &mut self, conn: &mut super::Connection<F>, headers: &[T], fin: bool,
-    ) -> Result<u64> {
+    ) -> Result<(u64, bool)> {
         // If we received a GOAWAY from the peer, MUST NOT initiate new
         // requests.
         if self.peer_goaway_id.is_some() {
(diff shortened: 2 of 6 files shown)


quiche/src/h3/ffi.rs
```diff
@@ -227,7 +227,7 @@ pub extern "C" fn quiche_h3_send_request(
     let req_headers = headers_from_ptr(headers, headers_len);
 
     match conn.send_request(quic_conn, &req_headers, fin) {
-        Ok(v) => v as i64,
```
LPardue, line 230: TODO: if th FFI swallows this, then apps can't use it. Consider a way to expose it to apps

tools/http3_test/src/lib.rs
```diff
@@ -467,7 +467,7 @@ impl Http3Test {
 
             info!("sending HTTP request {:?}", req.hdrs);
 
-            let s =
+            let (s, _hdrs_fully_sent) =
```
LPardue, line 470: We want to kill http3_test anyway, so I don't see much point perfecting this client

quiche/src/h3/stream.rs
```diff
@@ -165,11 +165,25 @@ pub struct Stream {
...
+    trailers_in_flight: bool,
+
     /// Whether a trailing HEADER field has been sent.
     trailers_sent: bool,
 
     /// Whether a trailing HEADER field has been received.
     trailers_received: bool,
+
+    /// Tracking for outgoing frame header.
+    pub outgoing_frame_header: Vec<u8>,
```
LPardue, line 178: all of these were made `pub` to make adding the feature easier, they can be put behind accessors once we're happy with general approach

quiche/src/h3/mod.rs
```diff
@@ -1096,58 +1101,65 @@ impl Connection {
...
+        let stream = stream::Stream::new(stream_id, true);
 
         // The underlying QUIC stream does not exist yet, so calls to e.g.
         // stream_capacity() will fail. By writing a 0-length buffer, we force
         // the creation of the QUIC stream state, without actually writing
-        // anything.
-        if let Err(e) = conn.stream_send(stream_id, b"", false) {
-            self.streams.remove(&stream_id);
+        // anything. Errors on this call don't make sense to bubble up.
+        conn.stream_send(stream_id, b"", false).ok();
```
gregor-cf, line 1142: I think you'll need to bubble up errors, since an error here would mean that the stream hasn't been been created in the quiche connection, so the stream_id is still unused.

quiche/src/h3/mod.rs
```diff
@@ -424,6 +424,9 @@ pub enum Error {
     /// The requested operation cannot be served over HTTP/3. Peer should retry
     /// over HTTP/1.1.
     VersionFallback,
+
+    /// TODO
+    PartialHeader,
```
gregor-cf, line 429: nit: since this is PR is sem-ver breaking already, might be good to add a `StreamLimit` error as well. Right now that's encoded as `TrasportError(StreamLimit)` it would be nice for clients to have it directly (similar to how we wrap StreamBlocked). Happy to make that change myself.

quiche/src/h3/mod.rs
```diff
@@ -1393,70 +1394,54 @@ impl Connection {
...
-        // Sending header block separately avoids unnecessary copy.
-        conn.stream_send(stream_id, &header_block, fin)?;
+        stream.outgoing_frame_header = frame_header;
+        stream.outgoing_frame_header_off = 0;
+
+        frame_payload.truncate(payload_len);
+
+        stream.outgoing_frame_payload = frame_payload;
+        stream.outgoing_frame_payload_off = 0;
+        stream.fin_after_outgoing_frame = fin;
```
gregor-cf, line 1438: you can put the frame headers and the frame payload in a single `Vec<u8>` (or actually `VecDeque<u8>`), that would then simplify your sending logic.
```
        const MAX_FRAME_HEADER_SIZE: usize = 16; // max varint length of two u64
        let mut frame_data = vec![0; headers_len + MAX_FRAME_HEADER_SIZE];
        // Keep enough space at the beginning of the buffer 
        let frame_payload = &mut frame_data[MAX_FRAME_HEADER_SIZE..];
        let payload_len = self
            .qpack_encoder
            .encode(headers, &mut frame_payload)
            .map_err(|_| Error::InternalError)?;

        let frame_header_size = octets::varint_len(frame::HEADERS_FRAME_TYPE_ID) +
            octets::varint_len(payload_len as u64);
            
         let frame_header_start = MAX_FRAME_HEADER_SIZE - frame_header_size;
         // write the frame_header to the `freame_header_size` bytes
         let frame_header =  &mut frame_data[frame_header_start..MAX_FRAME_HEADER_SIZE]

        let mut b = octets::OctetsMut::with_slice(&mut frame_header);
        b.put_varint(frame::HEADERS_FRAME_TYPE_ID)?;
        b.put_varint(payload_len as u64)?;
        debug_assert_eq!(b.off(), frame_header_size));

        stream.outgoing_frame_data = frame_data;
        stream.outgoing_frame_off = frame_header_start;
```

then you don't need to handle sending the frame header and frame payload separately below

quiche/src/h3/mod.rs
```diff
@@ -1474,19 +1459,168 @@ impl Connection {
...
+                    trace!(
+                        "only sent {} of {} frame header bytes",
+                        v,
+                        stream.outgoing_frame_header.len()
+                    );
+                    return Err(Error::PartialHeader);
+                },
+
+            Err(e) => {
+                // TODO: check its actually sensible to return all errors
```
gregor-cf, line 1486: `Done` and `StreamBlocked` should be turned into `Error::PartialHeader`

quiche/src/h3/stream.rs
```diff
@@ -165,11 +165,25 @@ pub struct Stream {
     /// Whether a DATA frame has been received.
     data_received: bool,
 
+    /// Whether a trailing HEADER field is in flight.
+    trailers_in_flight: bool,
```
gregor-cf, line 169: I think it makes sense to remove `trailers_if_flight`, `trailers_sent`, `local_initialize`, etc. and instead combine them into a `StreamSendState` enum. IMHO that will make checking if a particular call is allowed a lot easier. I think you might also want to update the checks in the various `send_foo()` functions, e.g., at least `send_response()` and `send_body()` would now allow sending even if partial headers are pending.

quiche/src/h3/mod.rs
```diff
@@ -1474,19 +1459,168 @@ impl Connection {
...
                 frame,
                 ..Default::default()
             });
 
             q.add_event_data_now(ev_data).ok();
         });
 
-        if let Some(s) = self.streams.get_mut(&stream_id) {
-            s.initialize_local();
+        match conn.stream_send(stream_id, &stream.outgoing_frame_header, false) {
```
gregor-cf, line 1470: I think you can just call `continue_partial_headers()` here, then you don't need to duplicate the send logic.

quiche/src/h3/mod.rs
```diff
@@ -1474,19 +1459,168 @@ impl Connection {
...
+
+        if !stream.outgoing_frame_header.is_empty() {
+            match conn.stream_send(
+                stream_id,
+                &stream.outgoing_frame_header[stream.outgoing_frame_header_off..],
+                false,
+            ) {
+                Ok(v) =>
+                    if v != stream.outgoing_frame_header.len() {
+                        stream.outgoing_frame_header_off += v;
```
gregor-cf, line 1569: FYI, this check is wrong. It should be: 
```
stream.outgoing_frame_header_off += v;
if (stream.outgoing_frame_header_off != stream.outgoing_frame_header.len())
```

quiche/src/h3/mod.rs
LPardue: Good idea

quiche/src/h3/stream.rs
LPardue: Tests indicate they are already prevented but I'm all for more readable and maintainable code, which bool flags aren't really.

quiche/src/h3/mod.rs
LPardue: Good catch!

quiche/src/h3/mod.rs
LPardue: I think we should consider that separately from the changes for streaming headers, I suspect it woulf have a few knock ons to the app ecosystem that I'd rather not have to deal with as part of this work.

quiche/src/h3/stream.rs
gregor-cf: JFYI, I think there are a couple of oder-of-calls that aren't covered by tests and would work but break the HTTP spec. I think the current test mostly relies on `local_initialized` so they only protect if the initial headers are pending, but not additional headers. 
E.g., 
* if `send_response()` returns HeaderPending, it can be called again without error
* if `send_additional_headers()` returns HeaderPending,  you can call `send_additional_headers()`, `send_response()`, and `send_body()` without error

hmmm, maybe add an additional check to `stream.outgoing_frame_header` and `outgoing_frame_payload` are empty in `send_headers()` and `send_body()`

LPardue: Yeah I intend to harden the code to ensure there are no invalid sequences before promoting the PR from draft. Thanks for the pointers

quiche/src/h3/mod.rs
LPardue: Actioned this over on https://github.com/cloudflare/quiche/pull/2153

quiche/src/h3/mod.rs
LPardue: Actioned this a different way over on https://github.com/cloudflare/quiche/pull/2153

quiche/src/h3/mod.rs
LPardue: actioned this over on https://github.com/cloudflare/quiche/pull/2153

quiche/src/h3/stream.rs
LPardue: This will be done as part of https://github.com/cloudflare/quiche/pull/2153

LPardue:
Closing this in favor of https://github.com/cloudflare/quiche/pull/2153

PR #2148, opened 2025-09-02
h3: add support for streaming HEADERS sends with new functions

LPardue:
NB: this is an alternative to https://github.com/cloudflare/quiche/pull/2147, it adds streaming to new functions in order to keep previous behaviour. New functions names can be bikeshedded.

Calls to send_request(), send_response(), send_response_with_priority(),
and send_additional_headers() can fail with a StreamBlocked
occurs, applications ar expected to retry the operation when
the stream is later reported as writable.

However, certain conditions could mean that sufficient capacity
might never be made available, effectively permenantly blocking
header sends. The root cause of this problem was the choice to enforce
that a HEADERS frame is always sent (buffered into a quiche stream)
in whole.

This change adds new variants of the header sending functions that
support a streaming send design. These will produce a HEADERS frame
that can be sent in whole or in part, depending on available capacity.
When a frame is only partly sent, applications are notified and

prevented. HEADERS frames must be sent completly before

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -1112,13 +1112,29 @@ typedef struct {
     bool incremental;
 } quiche_h3_priority;
 
+// send_request2 return values.
+typedef struct {
+    uint64_t stream_id;
+    bool headers_fully_sent;
+} send_request2_result;
+
 // Sends an HTTP/3 request.
 int64_t quiche_h3_send_request(quiche_h3_conn *conn, quiche_conn *quic_conn,
                                const quiche_h3_header *headers, size_t headers_len,
                                bool fin);
 
+// Sends an HTTP/3 request.
+int64_t quiche_h3_send_request2(quiche_h3_conn *conn, quiche_conn *quic_conn,
+                               const quiche_h3_header *headers, size_t headers_len,
+                               bool fin, send_request2_result* result);
+
 // Sends an HTTP/3 response on the specified stream with default priority.
 int quiche_h3_send_response(quiche_h3_conn *conn, quiche_conn *quic_conn,
+                            uint64_t stream_id, const quiche_h3_header *headers,
+                            size_t headers_len, bool fin);
+
+// Sends an HTTP/3 response on the specified stream with default priority.
+int quiche_h3_send_response2(quiche_h3_conn *conn, quiche_conn *quic_conn,
                             uint64_t stream_id, const quiche_h3_header *headers,
                             size_t headers_len, bool fin);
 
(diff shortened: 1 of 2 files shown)


quiche/src/h3/mod.rs
```diff
@@ -1393,100 +1644,252 @@ impl Connection {
...
+        let mut b = octets::OctetsMut::with_slice(&mut frame_header);
+        b.put_varint(frame::HEADERS_FRAME_TYPE_ID)?;
+        b.put_varint(payload_len as u64)?;
 
-        // Headers need to be sent atomically, so make sure the stream has
-        // enough capacity.
-        match conn.stream_writable(stream_id, overhead + header_block.len()) {
-            Ok(true) => (),
+        stream.outgoing_frame_header = frame_header;
+        stream.outgoing_frame_header_off = 0;
```
LPardue, line 1701: TODO: when streaming mode is false, don't set these

quiche/src/h3/mod.rs
```diff
@@ -1393,100 +1644,252 @@ impl Connection {
...
-            Ok(false) => return Err(Error::StreamBlocked),
+        frame_payload.truncate(payload_len);
 
-            Err(e) => {
-                if conn.stream_finished(stream_id) {
-                    self.streams.remove(&stream_id);
-                }
+        stream.outgoing_frame_payload = frame_payload;
+        stream.outgoing_frame_payload_off = 0;
+        stream.fin_after_outgoing_frame = fin;
```
LPardue, line 1707: TODO: when streaming mode is false, don't set these

quiche/src/h3/mod.rs
```diff
@@ -1393,100 +1644,252 @@ impl Connection {
...
+                frame,
+                ..Default::default()
+            });
+
+            q.add_event_data_now(ev_data).ok();
+        });
+
+        match conn.stream_send(stream_id, &stream.outgoing_frame_header, false) {
+            Ok(v) =>
+                if v != stream.outgoing_frame_header.len() {
```
LPardue, line 1741: TODO: only do this when streaming mode is true

quiche/src/h3/mod.rs
```diff
@@ -1393,100 +1644,252 @@ impl Connection {
...
+        stream.outgoing_frame_header_off = 0;
+
+        // Sending header block separately avoids unnecessary copy.
+        match conn.stream_send(
+            stream_id,
+            &stream.outgoing_frame_payload,
+            stream.fin_after_outgoing_frame,
+        ) {
+            Ok(v) =>
+                if v != stream.outgoing_frame_payload.len() {
```
LPardue, line 1770: TODO: only do this when streaming mode is true

gregor-cf:
It's probably a good idea to keep the old functions and add new ones like this PR.  With #2147 the functions' contract would change but not their signature, so code using them would still compile which could lead to unexpected errors and bugs down the line. 

I'd probably add a `#[deprecated(note="use the new foo() function")]` attribute to the old functions though. 

you can avoid the code duplication if you  add private implementations of your functions like:
```
pub fn send_response(....) {
     send_response_impl(false)
}
pub fn send_response2(...) {
     send_response_impl(true)
}
fn send_response_impl(..., streaming_mode: bool)
```

LPardue:
Closing this in favor of https://github.com/cloudflare/quiche/pull/2153

PR #2149, opened 2025-09-02
feat: expose configuration to enable relaxed loss threshold detection on spurious loss

toidiu:
- [x] Based on https://github.com/cloudflare/quiche/pull/2145

Packet re-ordering (spurious loss) is common to certain networks and can
cause severe degradation to a connection.  When enabled, the relaxed
mode will disable pkt_thresh loss detection on the first spurious loss
and then double time_thresh on subsequent spurious losses.

## Reviewing
Review commit by commit.
- The first commit is a refactor which abstracts out the loss threshold logic.
- The second introduces the new relaxed logic.

### DIFF
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -1,5 +1,7 @@
 use crate::packet;
 use crate::recovery::OnLossDetectionTimeoutOutcome;
+use crate::recovery::INITIAL_TIME_THRESHOLD_OVERHEAD;
+use crate::recovery::TIME_THRESHOLD_OVERHEAD_MULTIPLIER;
 use crate::Error;
 use crate::Result;
 
@@ -258,7 +260,7 @@ impl RecoveryEpoch {
     }
 
     fn detect_and_remove_lost_packets(
-        &mut self, loss_delay: Duration, pkt_thresh: u64, now: Instant,
+        &mut self, loss_delay: Duration, pkt_thresh: Option<u64>, now: Instant,
         newly_lost: &mut Vec<Lost>,
     ) -> LossDetectionResult {
         newly_lost.clear();
@@ -276,9 +278,13 @@ impl RecoveryEpoch {
             }
 
             if let SentStatus::Sent { time_sent, .. } = status {
-                if *time_sent <= lost_send_time ||
-                    largest_acked >= *pkt_num + pkt_thresh
-                {
+                let loss_by_time = *time_sent <= lost_send_time;
+                let loss_by_pkt = match pkt_thresh {
+                    Some(pkt_thresh) => largest_acked >= *pkt_num + pkt_thresh,
+                    None => false,
+                };
+
+                if loss_by_time || loss_by_pkt {
                     if let SentStatus::Sent {
                         in_flight,
                         sent_bytes,
@@ -351,6 +357,72 @@ impl RecoveryEpoch {
     }
 }
 
+struct LossThreshold {
+    pkt_thresh: Option<u64>,
+    time_thresh: f64,
+
+    // # Experiment: enable_relaxed_loss_threshold
+    //
+    // If `Some` this will disable pkt_thresh on the first loss and then double
+    // time_thresh on subsequent loss.
+    //
+    // The actual threshold is calcualted as `1.0 +
+    // INITIAL_TIME_THRESHOLD_OVERHEAD` and equivalent to the initial value
+    // of INITIAL_TIME_THRESHOLD.
+    time_thresh_overhead: Option<f64>,
+}
+
+impl LossThreshold {
+    fn new(recovery_config: &RecoveryConfig) -> Self {
+        let time_thresh_overhead =
+            if recovery_config.enable_relaxed_loss_threshold {
+                Some(INITIAL_TIME_THRESHOLD_OVERHEAD)
+            } else {
+                None
+            };
+        LossThreshold {
+            pkt_thresh: Some(INITIAL_PACKET_THRESHOLD),
+            time_thresh: INITIAL_TIME_THRESHOLD,
+            time_thresh_overhead,
+        }
+    }
+
+    fn pkt_thresh(&self) -> Option<u64> {
+        self.pkt_thresh
+    }
+
+    fn time_thresh(&self) -> f64 {
+        self.time_thresh
+    }
+
+    fn on_spurious_loss(&mut self, new_pkt_thresh: u64) {
+        match &mut self.time_thresh_overhead {
+            Some(time_thresh_overhead) => {
+                if self.pkt_thresh.is_some() {
+                    // Disable packet threshold on first spurious loss.
+                    self.pkt_thresh = None;
+                } else {
+                    // Double time threshold but cap it at `1.0`, which ends up
+                    // being 2x the RTT.
+                    *time_thresh_overhead *= TIME_THRESHOLD_OVERHEAD_MULTIPLIER;
+                    *time_thresh_overhead = time_thresh_overhead.min(1.0);
+
+                    self.time_thresh = 1.0 + *time_thresh_overhead;
+                }
+            },
+            None => {
+                let new_packet_threshold = self
+                    .pkt_thresh
+                    .expect("packet threshold should always be Some when `enable_relaxed_loss_threshold` is false")
+                    .max(new_pkt_thresh.min(MAX_PACKET_THRESHOLD));
+                self.pkt_thresh = Some(new_packet_threshold);
+
+                self.time_thresh = PACKET_REORDER_TIME_THRESHOLD;
+            },
+        }
+    }
+}
+
 pub struct GRecovery {
     epochs: [RecoveryEpoch; packet::Epoch::count()],
 
@@ -366,9 +438,7 @@ pub struct GRecovery {
 
     pub lost_spurious_count: usize,
 
-    pkt_thresh: u64,
-
-    time_thresh: f64,
+    loss_thresh: LossThreshold,
 
     bytes_in_flight: BytesInFlight,
 
@@ -423,9 +493,7 @@ impl GRecovery {
             lost_count: 0,
             lost_spurious_count: 0,
 
-            pkt_thresh: INITIAL_PACKET_THRESHOLD,
-            time_thresh: INITIAL_TIME_THRESHOLD,
-
+            loss_thresh: LossThreshold::new(recovery_config),
             bytes_in_flight: Default::default(),
             bytes_sent: 0,
             bytes_lost: 0,
@@ -456,17 +524,18 @@ impl GRecovery {
     fn detect_and_remove_lost_packets(
         &mut self, epoch: packet::Epoch, now: Instant,
     ) -> (usize, usize) {
+        let loss_delay =
+            self.rtt_stats.loss_delay(self.loss_thresh.time_thresh());
         let lost = &mut self.lost_reuse;
 
         let LossDetectionResult {
             lost_bytes,
             lost_packets,
-
             pmtud_lost_bytes,
             pmtud_lost_packets,
         } = self.epochs[epoch].detect_and_remove_lost_packets(
-            self.rtt_stats.loss_delay(self.time_thresh),
-            self.pkt_thresh,
+            loss_delay,
+            self.loss_thresh.pkt_thresh(),
             now,
             lost,
         );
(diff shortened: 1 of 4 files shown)


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -350,6 +354,67 @@ impl RecoveryEpoch {
     }
 }
 
+struct LossThreshold {
```
antoniovicente, line 360: Could you add some unit tests for this struct?

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -368,15 +385,33 @@ impl LossThreshold {
...
+        match &mut self.time_thresh_overhead {
+            Some(time_thresh_overhead) => {
+                if self.pkt_thresh != DISABLE_PKT_THRESHOLD {
+                    // Disable packet threshold on first spurious loss.
+                    self.pkt_thresh = DISABLE_PKT_THRESHOLD;
+                } else {
+                    // Double time threshold but cap it at `1.0`, which ends up
+                    // being 2x the RTT.
+                    *time_thresh_overhead *= 2.0;
+                    *time_thresh_overhead = time_thresh_overhead.min(1.0);
```
antoniovicente, line 408: Consider adding:
  `self.time_thresh = 1.0 + *time_thresh_overhead;`

And simplifying `fn time_thresh(&self) -> f64` to just return `self.time_thresh`

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -368,15 +385,33 @@ impl LossThreshold {
...
-            .pkt_thresh
-            .max(new_pkt_thresh.min(MAX_PACKET_THRESHOLD));
-
-        self.time_thresh = PACKET_REORDER_TIME_THRESHOLD;
+        match &mut self.time_thresh_overhead {
+            Some(time_thresh_overhead) => {
+                if self.pkt_thresh != DISABLE_PKT_THRESHOLD {
+                    // Disable packet threshold on first spurious loss.
+                    self.pkt_thresh = DISABLE_PKT_THRESHOLD;
+                } else {
```
antoniovicente, line 404: Should the code below always run?

Seems like the `self.time_thresh_overhead.is_none()` branch changes self.time_thresh on the first reorder, but this version waits until the second reorder.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -354,12 +357,26 @@ impl RecoveryEpoch {
 struct LossThreshold {
     pkt_thresh: u64,
```
antoniovicente: Should the type of pkt_thresh be changed to `Option<u64>` in order to avoid DISABLE_PKT_THRESHOLD?

quiche/src/recovery/mod.rs
```diff
@@ -1525,6 +1537,201 @@ mod tests {
         );
     }
 
+    // TODO: `congestion` and `gcongestion` behave differently. That might be ok
+    // given the different algorithms but it would be ideal to merge and share
+    // the logic.
```
antoniovicente: Please cleanup this TODO text.  It is hard to follow.

// TODO: Implement `enable_relaxed_loss_threshold` and enable this test for the `congestion` branch.

quiche/src/recovery/mod.rs
```diff
@@ -1525,6 +1537,201 @@ mod tests {
...
+    #[rstest]
+    fn relaxed_thresholds_on_reordering(
+        #[values("bbr2_gcongestion")] cc_algorithm_name: &str,
+    ) {
+        let mut cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        cfg.enable_relaxed_loss_threshold = true;
```
antoniovicente, line 1553: Worth considering parametrizing the existing test with `#[values(false, true)] enable_relaxed_loss_threshold` ?

It is currently hard to tell how this very long test differs in behavior for the two cases.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -351,6 +354,67 @@ impl RecoveryEpoch {
...
+    fn on_spurious_loss(&mut self, new_pkt_thresh: u64) {
+        match &mut self.time_thresh_overhead {
+            Some(time_thresh_overhead) => {
+                if self.pkt_thresh != DISABLE_PKT_THRESHOLD {
+                    // Disable packet threshold on first spurious loss.
+                    self.pkt_thresh = DISABLE_PKT_THRESHOLD;
+                } else {
+                    // Double time threshold but cap it at `1.0`, which ends up
+                    // being 2x the RTT.
+                    *time_thresh_overhead *= 2.0;
```
antoniovicente: There are no tests that cover the `time_thresh_overhead.min(1.0);` portion of this computation.

Please add an unit test for that.

quiche/src/recovery/gcongestion/recovery.rs
toidiu: Yea considered that to make it more obvious. That might be cleaner than the `DISABLE_PKT_THRESHOLD`. Let me explore what that looks like.

Updated in https://github.com/cloudflare/quiche/pull/2149/commits/0c1c935fd2321acd288aceaf41bc8bbbdf93db2d

quiche/src/recovery/gcongestion/recovery.rs
toidiu: addressed in https://github.com/cloudflare/quiche/pull/2149/commits/56ae6ff02ca7ba4c3b059d2dd5b330e5e77cda46

quiche/src/recovery/mod.rs
toidiu: addressed in https://github.com/cloudflare/quiche/pull/2149/commits/56ae6ff02ca7ba4c3b059d2dd5b330e5e77cda46

quiche/src/recovery/gcongestion/recovery.rs
toidiu: Good suggestion! Addressed in https://github.com/cloudflare/quiche/pull/2149/commits/56ae6ff02ca7ba4c3b059d2dd5b330e5e77cda46

quiche/src/recovery/gcongestion/recovery.rs
toidiu: addressed in https://github.com/cloudflare/quiche/pull/2149/commits/56ae6ff02ca7ba4c3b059d2dd5b330e5e77cda46

quiche/src/recovery/gcongestion/recovery.rs
toidiu: My goal was to implement logic similar to [h2o](https://github.com/h2o/quicly/pull/512#issuecomment-1092301408), which doesn't change time_thresh on the first loss.

Since we are disabling packet threshold, increasing time threshold could possibly be considered "too" relaxed.. dunno. I say we try and this and consider if we also want to change the time threshold if it shows benefits.

quiche/src/recovery/mod.rs
toidiu: Because we don't change time_thresh on the first loss here the two tests vary significantly. That plus sending/acking in tests is quite verbose, trying to capture both cases is going to make both tests unreadable IMO.

This test is important to test the experiment override is working + thresholds change in a relaxed way. The unit tests I added actually do a much better job of illustrating/testing the specific threshold behavior.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1172,3 +1176,91 @@ impl std::fmt::Debug for GRecovery {
...
+        assert_eq!(loss_thresh.time_thresh(), INITIAL_TIME_THRESHOLD);
+
+        // First spurious loss.
+        loss_thresh.on_spurious_loss(INITIAL_PACKET_THRESHOLD);
+        assert_eq!(loss_thresh.pkt_thresh(), INITIAL_PACKET_THRESHOLD);
+        assert_eq!(loss_thresh.time_thresh(), PACKET_REORDER_TIME_THRESHOLD);
+
+        // Subsequent spurious loss.
+        for packet_gap in 0..=MAX_PACKET_THRESHOLD {
+            loss_thresh.on_spurious_loss(INITIAL_PACKET_THRESHOLD + packet_gap);
```
antoniovicente: Is a packet gap of 0 valid?

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1172,3 +1176,91 @@ impl std::fmt::Debug for GRecovery {
...
+
+            // New threshold is the calcualted based on the max packet gap, but
+            // capped at MAX_PACKET_THRESHOLD.
+            let new_pkt_threshold =
+                (INITIAL_PACKET_THRESHOLD + packet_gap).min(MAX_PACKET_THRESHOLD);
+            assert_eq!(loss_thresh.pkt_thresh(), new_pkt_threshold);
```
antoniovicente: We are not properly asserting in cases where `INITIAL_PACKET_THRESHOLD + packet_gap` < `MAX_PACKET_THRESHOLD`

consider something like:
```
loss_thresh.on_spurious_loss(INITIAL_PACKET_THRESHOLD + 2);
assert_eq!(loss_thresh.pkt_thresh(), INITIAL_PACKET_THRESHOLD + 2);
assert!(loss_thresh.pkt_thresh() < MAX_PACKET_THRESHOLD);
```

or

```
assert!(MAX_PACKET_THRESHOLD - INITIAL_PACKET_THRESHOLD > 0);
if packet_gap < MAX_PACKET_THRESHOLD - INITIAL_PACKET_THRESHOLD {
  assert_eq!(loss_thresh.pkt_thresh(), INITIAL_PACKET_THRESHOLD + packet_gap);
} else {
  assert_eq!(loss_thresh.pkt_thresh(), MAX_PACKET_THRESHOLD);
}
```

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1172,3 +1176,91 @@ impl std::fmt::Debug for GRecovery {
...
+
+    #[test]
+    fn relaxed_loss_threshold() {
+        // The max time threshold when operating in relaxed loss mode.
+        const MAX_TIME_THRESHOLD: f64 = 2.0;
+
+        let mut config = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        config.set_enable_relaxed_loss_threshold(true);
+        let recovery_config = RecoveryConfig::from_config(&config);
+        assert_eq!(recovery_config.enable_relaxed_loss_threshold, true);
```
antoniovicente: Surprised that clippy allows use of assert_eq! on boolean types.

quiche/src/recovery/mod.rs
```diff
@@ -1663,7 +1661,7 @@ mod tests {
         //
         // Packet threshold should be disabled. Time threshold overhead should
         // stay the same.
-        assert_eq!(r.pkt_thresh(), DISABLE_PKT_THRESHOLD);
+        assert!(r.pkt_thresh().is_none());
```
antoniovicente: `assert_eq!(r.pkt_thresh(), None);` would provide better debug info.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1172,3 +1176,91 @@ impl std::fmt::Debug for GRecovery {
...
+            let new_time_threshold = {
+                // Double the overhead
+                let mut calculate_new_overhead =
+                    loss_thresh.time_thresh_overhead.unwrap() *
+                        TIME_THRESHOLD_OVERHEAD_MULTIPLIER;
+                // Cap at `1.0`.
+                calculate_new_overhead = calculate_new_overhead.min(1.0);
+                // Actual threshold is calculated as `1.0 + overhead`
+                1.0 + calculate_new_overhead
+            };
```
antoniovicente, line 1279: This test effectively copies the logic in the implementation.  If one is wrong, so is the other.

I recommend instead something like:
```
let new_time_threshold = if iter_count < 4 {
  1.0 + INITIAL_TIME_THRESHOLD_OVERHEAD * 2.pow(iter_count)
} else {
  2.0
};
```

toidiu: Updated in https://github.com/cloudflare/quiche/pull/2149/commits/76a8b5fb2c7296b2a2e6545f84480d0008abd22e

antoniovicente (approved):

PR #2150, opened 2025-09-02
tq: add monitoring of time taken to flush response headers

LPardue:
Sending headers in quiche means buffering from the application
layer into quiche. It is not always possible to do this due to
capacity at any moment in time.

Tokio-quiche currently sends response headers by processing the
writable frame queue (process_write_frame()) and acting on an
OutboundFrame::Headers. This supports informational headers,
final headers, and trailing headers. Each type might fail,
which results in leaving the OutboundFrame on the queue to be
retried later.

This change adds some time tracking to each stream. If the first
attempt to send headers fails, the time is marked. On completion,
a duration is calculated and the stream's aggregate flushing
time is incremented. This ensures retrying of any/all response
messages is accounted for.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -39,6 +39,7 @@ use std::error::Error;
 use std::fmt;
 use std::marker::PhantomData;
 use std::sync::Arc;
+use std::time::Instant;
 
 use datagram_socket::StreamClosureKind;
 use foundations::telemetry::log;
@@ -636,7 +637,7 @@ impl<H: DriverHooks> H3Driver<H> {
             OutboundFrame::Headers(headers, priority) => {
                 let prio = priority.as_ref().unwrap_or(&DEFAULT_PRIO);
 
-                if ctx.initial_headers_sent {
+                let res = if ctx.initial_headers_sent {
                     // Initial headers were already sent, send additional
                     // headers now.
                     conn.send_additional_headers_with_priority(
@@ -648,7 +649,24 @@ impl<H: DriverHooks> H3Driver<H> {
                         qconn, stream_id, headers, prio, false,
                     )
                     .inspect(|_| ctx.initial_headers_sent = true)
+                };
+
+                if let Err(h3::Error::StreamBlocked) = res {
+                    ctx.first_full_headers_flush_fail_time
+                        .get_or_insert(Instant::now());
                 }
+
+                if res.is_ok() {
+                    if let Some(first) =
+                        ctx.first_full_headers_flush_fail_time.take()
+                    {
+                        ctx.audit_stats.add_header_flush_duration(
+                            Instant::now().duration_since(first),
+                        );
+                    }
+                }
+
+                res
             },
 
             OutboundFrame::Body(body, fin) => {
--- a/tokio-quiche/src/http3/driver/streams.rs
+++ b/tokio-quiche/src/http3/driver/streams.rs
@@ -29,6 +29,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::Context;
 use std::task::Poll;
+use std::time::Instant;
 
 use tokio::sync::mpsc;
 use tokio_util::sync::PollSender;
@@ -52,6 +53,8 @@ pub(crate) struct StreamCtx {
     pub(crate) audit_stats: Arc<H3AuditStats>,
     /// Indicates the stream sent initial headers.
     pub(crate) initial_headers_sent: bool,
+    /// First time that a HEADERS frame was not fully flushed.
+    pub(crate) first_full_headers_flush_fail_time: Option<Instant>,
     /// Indicates the stream received fin. No more data will be received.
     pub(crate) fin_recv: bool,
     /// Indicates the stream sent fin. No more data will be sent.
@@ -77,6 +80,7 @@ impl StreamCtx {
             audit_stats: Arc::new(H3AuditStats::new(stream_id)),
 
             initial_headers_sent: false,
+            first_full_headers_flush_fail_time: None,
 
             fin_recv: false,
             fin_sent: false,
--- a/tokio-quiche/src/http3/stats.rs
+++ b/tokio-quiche/src/http3/stats.rs
@@ -27,6 +27,7 @@
 use std::sync::atomic::AtomicI64;
 use std::sync::atomic::AtomicU64;
 use std::sync::atomic::Ordering;
+use std::time::Duration;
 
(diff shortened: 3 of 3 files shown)


tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -274,7 +275,11 @@ impl H3Event {
 #[derive(Debug)]
 pub enum OutboundFrame {
     /// Response headers to be sent to the peer, with optional priority.
-    Headers(Vec<h3::Header>, Option<quiche::h3::Priority>),
+    Headers {
+        headers: Vec<h3::Header>,
+        priority: Option<quiche::h3::Priority>,
+        first_non_fully_flush_time: Option<Instant>,
```
evanrittenhouse: nit: maybe `first_partial_flush_time`?

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -648,7 +657,21 @@ impl<H: DriverHooks> H3Driver<H> {
                         qconn, stream_id, headers, prio, false,
                     )
                     .inspect(|_| ctx.initial_headers_sent = true)
+                };
+
+                if res.is_err() && first_non_fully_flush_time.is_none() {
```
evanrittenhouse: It may make the code's intention clearer to only set this value if the error is transient, like `StreamBlocked`. Currently for example, we'll set this on a `MessageError`, but we'd never actually expect to flush that frame since `process_writeable_stream` will just reset it.

LPardue: Good point, I was being a bit lazy

tokio-quiche/src/http3/driver/mod.rs
LPardue: With the quiche current API, there is no partial flush - its full or nothing

evanrittenhouse: Sorry yeah I assumed this would apply to streaming writes in the future. Maybe `first_flush_attempt_time`?

LPardue: I was aiming to only do work when its needed, so maybe first_full_flush_fail_time. It's all a bit of a mouthful though 😄

ghedo:
Why do we need to track the duration? I imagine we just need a counter for how many times send_response was blocked, which could just be a connection-level stat.

LPardue:
This mimics other kinds of observability we have in non-TQ uses. Furthermore, we have TQ use cases where we measure duration of serving the entire request. My thinking was that we could break that timing down into constituent parts further with this change.

When switching from atomic to streaming, its possible the counts will change in ways that wont necessarily tell us much information. However, it's a less disruptive change so happy to start that way and do this as a later followup if we think it's better.

ghedo:
If we do want to track the duration anyway I think we need a better approach here, because as it stands this will require every application to manually initialize the new field to `None` since `OutboundFrame` is a public API... it may be better to handle this at the `process_write_frame` caller site https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/http3/driver/mod.rs#L923-L924 and store the timestamp inside the stream context object instead.

LPardue:
Yeah that's fair, let me try the othervway

ghedo:
> its possible the counts will change in ways that wont necessarily tell us much information

Not sure I understand this though, could you elaborate? Like, I think the same applies to duration, because if a stream is initially blocked and then gets unblocked, the timing wouldn't really change between atomic and streaming approaches I think. The only difference would be for requests that can never be sent because the stream capacity never increases enough.

LPardue:
Possibly yeah, it was more a gut feeling than anything. 

The primary benefit is the higher level of granularity of response timing breakdown we could get

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -52,6 +53,8 @@ pub(crate) struct StreamCtx {
     pub(crate) audit_stats: Arc<H3AuditStats>,
     /// Indicates the stream sent initial headers.
     pub(crate) initial_headers_sent: bool,
+    /// First time that a HEADERS frame was not fully flushed.
+    pub(crate) first_full_flush_fail_time: Option<Instant>,
```
ghedo: Nit: Maybe `last_headers_flush_fail_time`? Since there could be more than 1 HEADERS frames sent "first" seems confusing here, and also to make it clearer that it's specifically for HEADERS.

tokio-quiche/src/http3/stats.rs
```diff
@@ -193,4 +209,10 @@ impl H3AuditStats {
     pub fn set_sent_stream_fin(&self, sent_stream_fin: StreamClosureKind) {
         self.sent_stream_fin.store(sent_stream_fin);
     }
+
+    #[inline]
+    pub fn add_header_flush_duration(&self, duration: Duration) {
+        let current = self.headers_flush_duration.load();
+        self.headers_flush_duration.store(current + duration);
```
ghedo, line 222: This looks a bit sketchy TBH since the load and store are not atomic, but I think it's actually fine? As in, we are only ever sending one HEADERS at a time for any given stream so this is not going to be called concurrently, but maybe worth adding a comment to avoid future confusion.

tokio-quiche/src/http3/stats.rs
```diff
@@ -140,6 +147,15 @@ impl H3AuditStats {
         self.sent_stream_fin.load()
     }
 
+    /// Cumulative time between HEADERS failed full flush and complete.
```
ghedo: Nit: Not entirely sure what "full" means here TBH, can maybe drop that.

ghedo (changes requested):
lgtm overall, just added a couple of nits.

tokio-quiche/src/http3/driver/streams.rs
LPardue: The intent is to capture the first moment in time that an instance of a HEADERS frame could not be sent fully. Once that instance is emitted, the duration of time is measured and aggregated in `audit_stats`. Once emitted, the Option is cleared and ready to measure any further blocks for other HEADERS on the stream. 

I'll add `headers` in the name but think `first` is still the most representative label for the logic.

tokio-quiche/src/http3/stats.rs
LPardue: Tweaked the one-liner and added more details in the description

tokio-quiche/src/http3/stats.rs
LPardue: added

ghedo (approved):

PR #2151, opened 2025-09-03
PMTUD Fixes

cf-rhett:
Two small fixes

1. No longer update max datagram size from PMTUD on every sent packet. This was unnecessary to begin with but not harmful in most cases, however, in some cases this sets the max datagram size to 1200 if there wasn't a successful probe which can interfere with the PMTUD process.

2. Remove a condition on when PMTUD probing should happen as pmtud.should_probe() handles this and the current condition blocks revalidation of PMTUD from occurring.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -5162,13 +5162,6 @@ impl<F: BufFactory> Connection<F> {
             self.ack_eliciting_sent = true;
         }
 
-        let active_path = self.paths.get_active_mut()?;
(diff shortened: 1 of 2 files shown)


ghedo (approved):

ghedo:
Merged, thanks @cf-rhett!

For the future, it'd be better to put the description of the change in the commit message too, and also split fixes into separate commits (though they were pretty small in this case, so not strictly necessary...).

PR #2153, opened 2025-09-09
h3: add support for streaming HEADERS with new sequence

LPardue:
Calls to send_request(), send_response(), send_response_with_priority(),
and send_additional_headers() can fail with a StreamBlocked
occurs, applications are expected to retry the operation when
the stream is later reported as writable.

However, certain conditions could mean that sufficient capacity
might never be made available, effectively permenantly blocking
header sends. The root cause of this problem was the choice to enforce
that a HEADERS frame is always sent (buffered into a quiche stream)
in whole.

This change adds new functions related to stream management and header
sending to support a streaming send design. These will produce a HEADERS
frame that can be sent in whole or in part, depending on available
capacity. When a frame is only partly sent, applications are notified and

The new functions unify how clients and servers send headers.
For a client, the expected sequence is now something like:

0. Decide to initate a request
1. reserve_request_stream() - reserves a stream if limits allow
2. stream_headers() - inititates streaming of a HEADERS frame
3. continue_partial_headers() - if stream_headers() returned Error::PartialHeader

For a server, the expected sequence is now something like:

0. Receive a request via the poll() function and decide to respond
1. stream_priority() - set the stream's sending priority per RFC 9218
3. stream_headers() - inititates streaming of a HEADERS frame
4. continue_partial_headers() - if stream_headers return Error::PartialHeader

prevented. HEADERS frames must be sent completely before

Co-authored-by: Gregor Maier <gregor@cloudflare.com>

### DIFF
--- a/apps/src/common.rs
+++ b/apps/src/common.rs
@@ -334,6 +334,8 @@ pub trait HttpConn {
         &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
     );
 
+    fn retry_partial_headers(&mut self, conn: &mut quiche::Connection);
+
     fn handle_responses(
         &mut self, conn: &mut quiche::Connection, buf: &mut [u8],
         req_start: &std::time::Instant,
@@ -369,12 +371,19 @@ pub struct Http09Request {
     response_writer: Option<std::io::BufWriter<std::fs::File>>,
 }
 
+enum HeadersSendStatus {
+    NotStarted,
+    InProgress,
+    Completed,
+}
+
 /// Represents an HTTP/3 formatted request.
 struct Http3Request {
     url: url::Url,
     cardinal: u64,
     stream_id: Option<u64>,
     hdrs: Vec<quiche::h3::Header>,
+    hdrs_send_status: HeadersSendStatus,
     priority: Option<Priority>,
     response_hdrs: Vec<quiche::h3::Header>,
     response_body: Vec<u8>,
@@ -476,6 +485,8 @@ impl HttpConn for Http09Conn {
 
         self.reqs_sent += reqs_done;
     }
+
+    fn retry_partial_headers(&mut self, _conn: &mut quiche::Connection) {}
 
     fn handle_responses(
         &mut self, conn: &mut quiche::Connection, buf: &mut [u8],
@@ -757,7 +768,6 @@ fn make_h3_config(
 
 pub struct Http3Conn {
     h3_conn: quiche::h3::Connection,
-    reqs_hdrs_sent: usize,
     reqs_complete: usize,
     largest_processed_request: u64,
     reqs: Vec<Http3Request>,
@@ -831,6 +841,7 @@ impl Http3Conn {
                     url: url.clone(),
                     cardinal: i,
                     hdrs,
+                    hdrs_send_status: HeadersSendStatus::NotStarted,
                     priority,
                     response_hdrs: Vec::new(),
                     response_body: Vec::new(),
@@ -850,7 +861,6 @@ impl Http3Conn {
                     qpack_blocked_streams,
                 ),
             ).expect("Unable to create HTTP/3 connection, check the server's uni stream limit and window size"),
-            reqs_hdrs_sent: 0,
             reqs_complete: 0,
             largest_processed_request: 0,
             reqs,
@@ -882,7 +892,6 @@ impl Http3Conn {
 
         let h_conn = Http3Conn {
             h3_conn,
-            reqs_hdrs_sent: 0,
             reqs_complete: 0,
             largest_processed_request: 0,
             reqs: Vec::new(),
@@ -1131,18 +1140,40 @@ impl Http3Conn {
 }
 
 impl HttpConn for Http3Conn {
+    fn retry_partial_headers(&mut self, conn: &mut quiche::Connection) {
+        for req in self.reqs.iter_mut().filter(|r| {
+            matches!(r.hdrs_send_status, HeadersSendStatus::InProgress)
+        }) {
+            let stream_id = req
+                .stream_id
+                .expect("partial header resend but no stream ID");
+
+            debug!("retrying stream id={}", stream_id);
+            match self.h3_conn.continue_partial_headers(conn, stream_id) {
+                Ok(_) => {
+                    req.hdrs_send_status = HeadersSendStatus::Completed;
+                    debug!("Completed HTTP request {:?}", &req.hdrs);
+                },
+
+                Err(e) => {
+                    error!("retry failed to send request {:?}", e);
+                    break;
+                },
+            }
+        }
+    }
+
     fn send_requests(
         &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
     ) {
-        let mut reqs_done = 0;
+        // First retry partial headers
+        self.retry_partial_headers(conn);
 
-        // First send headers.
-        for req in self.reqs.iter_mut().skip(self.reqs_hdrs_sent) {
-            let s = match self.h3_conn.send_request(
-                conn,
-                &req.hdrs,
-                self.body.is_none(),
-            ) {
+        // Then send new headers.
+        for req in self.reqs.iter_mut().filter(|r| {
+            matches!(r.hdrs_send_status, HeadersSendStatus::NotStarted)
+        }) {
+            let stream_id = match self.h3_conn.reserve_request_stream(conn) {
                 Ok(v) => v,
 
                 Err(quiche::h3::Error::TransportError(
@@ -1158,28 +1189,52 @@ impl HttpConn for Http3Conn {
                 },
 
                 Err(e) => {
+                    error!("failed to reserve stream {e:?}");
+                    break;
+                },
+            };
+
+            // Store the stream ID and set up response/body tracking
+            // immediately — the stream has been reserved and these must be
+            // initialized regardless of what stream_headers returns.
+            req.stream_id = Some(stream_id);
+            req.response_writer =
+                make_resource_writer(&req.url, target_path, req.cardinal);
+            self.sent_body_bytes.insert(stream_id, 0);
+
+            match self.h3_conn.stream_headers(
+                conn,
+                stream_id,
+                &req.hdrs,
+                false,
+                self.body.is_none(),
+            ) {
+                Ok(()) => {
+                    req.hdrs_send_status = HeadersSendStatus::Completed;
+                    debug!("Completed HTTP request {:?}", &req.hdrs);
+                },
+
+                Err(quiche::h3::Error::PartialHeader) => {
+                    debug!(
+                        "sent partial headers, marking for later continuation"
+                    );
+                    req.hdrs_send_status = HeadersSendStatus::InProgress;
+                    break;
+                },
+
+                Err(e) => {
                     error!("failed to send request {e:?}");
                     break;
                 },
             };
 
-            debug!("Sent HTTP request {:?}", &req.hdrs);
-
             if let Some(priority) = &req.priority {
                 // If sending the priority fails, don't try again.
                 self.h3_conn
-                    .send_priority_update_for_request(conn, s, priority)
+                    .send_priority_update_for_request(conn, stream_id, priority)
                     .ok();
             }
-
-            req.stream_id = Some(s);
-            req.response_writer =
-                make_resource_writer(&req.url, target_path, req.cardinal);
-            self.sent_body_bytes.insert(s, 0);
-
-            reqs_done += 1;
         }
-        self.reqs_hdrs_sent += reqs_done;
 
         // Then send any remaining body.
         if let Some(body) = &self.body {
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -1032,6 +1032,10 @@ enum quiche_h3_error {
     // over HTTP/1.1.
     QUICHE_H3_ERR_VERSION_FALLBACK = -20,
 
+    // An attempt to send a HEADERS frame did not fully complete. The local
+    // endpoint should try to continue sending the remainder at a future time.
+    QUICHE_H3_ERR_PARTIAL_HEADER = -21,
+
     // The following QUICHE_H3_TRANSPORT_ERR_* errors are propagated
     // from the QUIC transport layer.
 
@@ -1197,6 +1201,20 @@ typedef struct {
 int64_t quiche_h3_send_request(quiche_h3_conn *conn, quiche_conn *quic_conn,
                                const quiche_h3_header *headers, size_t headers_len,
                                bool fin);
+
+// Reserves an HTTP/3 request stream.
+int64_t quiche_h3_reserve_request_stream(quiche_h3_conn *conn,
+                            quiche_conn *quic_conn);
+
+// Initiates streaming of a new HTTP/3 HEADERS frame on the given stream.
+int quiche_h3_stream_headers(quiche_h3_conn *conn,
+                             quiche_conn *quic_conn, uint64_t stream_id,
+                             quiche_h3_header *headers, size_t headers_len,
+                             bool is_trailer_section, bool fin);
+
+// Continues sending a HTTP/3 HEADERS frame on the given stream.
+int quiche_h3_continue_partial_headers(quiche_h3_conn *conn,
+                            quiche_conn *quic_conn, uint64_t stream_id);
 
 // Sends an HTTP/3 response on the specified stream with default priority.
 int quiche_h3_send_response(quiche_h3_conn *conn, quiche_conn *quic_conn,
--- a/quiche/src/h3/ffi.rs
+++ b/quiche/src/h3/ffi.rs
@@ -238,6 +238,48 @@ pub extern "C" fn quiche_h3_send_request(
 }
 
 #[no_mangle]
+pub extern "C" fn quiche_h3_reserve_request_stream(
+    conn: &mut h3::Connection, quic_conn: &mut Connection,
+) -> i64 {
+    match conn.reserve_request_stream(quic_conn) {
+        Ok(v) => v as i64,
+
+        Err(e) => e.to_c() as i64,
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn quiche_h3_stream_headers(
+    conn: &mut h3::Connection, quic_conn: &mut Connection, stream_id: u64,
+    headers: *const Header, headers_len: size_t, is_trailer_section: bool,
+    fin: bool,
+) -> c_int {
+    let headers = headers_from_ptr(headers, headers_len);
+
+    match conn.stream_headers(
+        quic_conn,
+        stream_id,
+        &headers,
+        is_trailer_section,
+        fin,
+    ) {
+        Ok(_) => 0,
+
+        Err(e) => e.to_c() as c_int,
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn quiche_h3_continue_partial_headers(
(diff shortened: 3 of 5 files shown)


LPardue:
Still needs a bit more testing / hardening but is ready for initial review

quiche/src/h3/ffi.rs
```diff
@@ -288,6 +288,37 @@ pub extern "C" fn quiche_h3_send_additional_headers(
     }
 }
 
+#[no_mangle]
```
gregor-cf: I think you're missing the `reserve_request_stream` implementation.

tokio-quiche/src/http3/driver/client.rs
```diff
@@ -152,9 +152,12 @@ impl ClientHooks {
         let body_finished = request.body_writer.is_none();
 
         // TODO: retry the request if the error is not fatal
-        let stream_id = driver.conn_mut()?.send_request(
+        let stream_id = driver.conn_mut()?.reserve_request_stream(qconn)?;
+        driver.conn_mut()?.stream_headers(
```
gregor-cf: I would leave tokio-quiche as-is for now. It can continue to use the old functions until we update all of it to use the new ones.

quiche/src/h3/mod.rs
```diff
@@ -503,6 +506,7 @@ impl Error {
             Error::MessageError => WireErrorCode::MessageError as u64,
             Error::ConnectError => WireErrorCode::ConnectError as u64,
             Error::VersionFallback => WireErrorCode::VersionFallback as u64,
+            Error::PartialHeader => 0x1000,
```
gregor-cf: nit: do you want to use a larger number here to avoid possible future conflicts?

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+        &mut self, conn: &mut super::Connection,
+    ) -> Result<u64> {
+        if self.is_server {
+            return Err(Error::StreamCreationError);
+        }
+
+        // If we received a GOAWAY from the peer, MUST NOT initiate new
+        // requests.
+        if self.peer_goaway_id.is_some() {
+            return Err(Error::StreamCreationError);
```
gregor-cf, line 1230: Should we use a different error for this to distinguish it from the `self.is_server` check about? The "old" send_request used `FrameUnexpected`

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+    /// A stream's priority determines the order in which stream data is sent
+    /// on the wire (streams with lower priority are sent first).
+    ///
+    /// The target stream is created if it did not exist before calling this
+    /// method.
+    pub fn stream_priority<F: BufFactory>(
+        conn: &mut crate::Connection<F>, stream_id: u64, priority: &Priority,
+    ) -> Result<()> {
+        // Clamp and shift urgency into quiche-priority space
+        let urgency = priority
```
gregor-cf, line 1250: nit: should we check if the stream is in `self.streams`?

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+
+    /// Reserves an HTTP/3 request stream.
+    ///
+    /// This is only for use by client endpoints, who are required to reserve a
+    /// stream before sending headers via the [`stream_headers()`] function.
+    ///
+    /// The [`StreamLimit`] error is returned when the stream cannot be created
+    /// due to stream limits.
+    ///
+    /// [`StreamLimit`]: ../enum.Error.html#variant.StreamLimit
```
gregor-cf, line 1219: This variant doesn't exist yet. We should add it (currently if quiche returns `StreamLimit` it get translated into `TransportError(StreamLimit)`.

Can you add `StreamCreationError` to the docs too.

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+    /// as `false`.
+    ///
+    /// Additional headers can only be sent during certain phases of an HTTP/3
+    /// message exchange, see [Section 4.1 of RFC 9114]. The [`FrameUnexpected`]
+    /// error is returned if expectations are not met.
+    ///
+    /// On success the newly allocated stream ID is returned, together with a
+    /// bool that indicates if the headers were fully sent or not. If this value
+    /// is false, it indicates the underlying stream or connection does not have
+    /// enough capacity to send the entire remaining header bytes.
```
gregor-cf: that's no longer correct, can be removed.

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+                    return Err(Error::FrameUnexpected);
+                }
+
+                // Initial HEADERS must have been sent.
+                if is_trailer_section && !s.local_initialized() {
+                    return Err(Error::FrameUnexpected);
+                }
+
+                // Only one trailing HEADERS allowed.
+                if is_trailer_section && s.trailers_sent() {
```
gregor-cf: I think this should just be `if s.trailers_sent() {`, since after the trailers we can't send anyting else, right?

quiche/src/h3/mod.rs
```diff
@@ -1382,81 +1519,84 @@ impl Connection {
...
 
-        // Headers need to be sent atomically, so make sure the stream has
-        // enough capacity.
-        match conn.stream_writable(stream_id, overhead + header_block.len()) {
-            Ok(true) => (),
+        // max varint length of two u64
+        const MAX_FRAME_HEADER_SIZE: usize = 16;
+        let mut frame_data = vec![0; MAX_FRAME_HEADER_SIZE + headers_len];
+        // Keep enough space at the beginning of the buffer
+        let mut frame_payload = &mut frame_data[MAX_FRAME_HEADER_SIZE..];
```
gregor-cf:
```suggestion
        let frame_payload = &mut frame_data[MAX_FRAME_HEADER_SIZE..];
```

quiche/src/h3/mod.rs
```diff
@@ -1382,81 +1519,84 @@ impl Connection {
...
 
-            Ok(false) => return Err(Error::StreamBlocked),
+        let payload_len = self
+            .qpack_encoder
+            .encode(headers, &mut frame_payload)
```
gregor-cf: I think that and the above change will make the compiler happy :-)
```suggestion
            .encode(headers, frame_payload)
```

quiche/src/h3/mod.rs
```diff
@@ -1185,8 +1327,10 @@ impl Connection {
     ///     struct.Connection.html#method.send_additional_headers
     /// [`send_response_with_priority()`]:
     ///     struct.Connection.html#method.send_response_with_priority
+    /// [`continue_partial_headers()`]:
+    ///     struct.Connection.html#method.continue_partial_headers
     /// [`FrameUnexpected`]: enum.Error.html#variant.FrameUnexpected
-    /// [`StreamBlocked`]: enum.Error.html#variant.StreamBlocked
+    /// [`PartialHeader`]: enum.Error.html#variant.PartialHeader
```
gregor-cf: leave the `StreamBlocked` and remove the link `continue_partial_headers()`

quiche/src/h3/mod.rs
```diff
@@ -1474,25 +1614,106 @@ impl Connection {
...
 
             q.add_event_data_now(ev_data).ok();
         });
 
-        if let Some(s) = self.streams.get_mut(&stream_id) {
-            s.initialize_local();
+        self.send_headers_to_quiche(
+            conn,
+            stream_id,
+            is_trailer_section,
```
gregor-cf: I don't think you need to pass `is_trailer_section` to `send_headers_to_quiche()`. You can directly call `stream.mark_trailers_sent()` here (and you don't need `trailers_in_flight`). 

If I remember the H3 spec correctly, after this point the only the valid thing for the user todo on this stream is to call `continue_partial_headers()` so I don't think we need to distinguish `trailers_in_flight` from `trailers_sent`

quiche/src/h3/mod.rs
```diff
@@ -1474,25 +1614,106 @@ impl Connection {
...
 
+    /// Continues sending headers on the given stream.
+    ///
+    /// The [`PartialHeader`] error is returned when the underlying QUIC stream
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application should retry the operation once the stream is
+    /// reported as writable again.
+    ///
+    /// [`PartialHeader`]: enum.Error.html#variant.PartialHeaders
+    /// [`Done`]: enum.Error.html#variant.Done
```
gregor-cf: I don't think we ever return Done

quiche/src/h3/mod.rs
```diff
@@ -1138,18 +1143,155 @@ impl Connection {
...
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application must call [`continue_partial_headers()`] once
+    /// the stream is reported as writable again.
+    ///
+    /// [`continue_partial_headers()`]:
+    ///     struct.Connection.html#method.continue_partial_headers
+    /// [`PartialHeader`]: enum.Error.html#variant.PartialHeader
+    pub fn stream_headers<T: NameValue, F: BufFactory>(
+        &mut self, conn: &mut super::Connection<F>, stream_id: u64,
+        headers: &[T], is_trailer_section: bool, fin: bool,
```
gregor-cf, line 1282: is it possible / legal to call this with is_trailer_section == true and fin == false? Should we check for it?

gregor-cf (commented):
LGTM. Had a couple of smaller questions and comments but I'm really liking this new API.

apps/src/common.rs
```diff
@@ -1120,19 +1132,80 @@ impl Http3Conn {
 }
 
 impl HttpConn for Http3Conn {
+    fn retry_partial_headers(&mut self, conn: &mut quiche::Connection) {
+        for req in self.reqs.iter_mut().filter(|r| {
```
antoniovicente, line 1144: Possible efficiency concern: this filter seems to go through the full list of active streams even if none of them is in the HeadersSendStatus::InProgess state.

Is that a problem?

Related question: is Http3Conn expected to be used in high performance applications?

apps/src/common.rs
```diff
@@ -1120,19 +1132,80 @@ impl Http3Conn {
...
+            }
+        }
+    }
+
     fn send_requests(
         &mut self, conn: &mut quiche::Connection, target_path: &Option<String>,
     ) {
-        let mut reqs_done = 0;
+        // First retry partial headers
+        self.retry_partial_headers(conn);
```
antoniovicente, line 1170: Should this code proceed to the 2nd loop in cases where continue_partial_headers failed in a retryable way?

I would have expected that at most 1 stream be in the `HeadersSendStatus::InProgess` state, but it seems like this code allows for multiple streams to move to the `HeadersSendStatus::InProgess` state

quiche/include/quiche.h
```diff
@@ -1134,6 +1137,16 @@ int quiche_h3_send_additional_headers(quiche_h3_conn *conn,
...
 
+// Initiates streaming of a new HTTP/3 HEADERS frame.
+int quiche_h3_stream_headers(quiche_h3_conn *conn,
+                            quiche_conn *quic_conn, uint64_t stream_id,
+                            quiche_h3_header *headers, size_t headers_len,
+                            bool is_trailer_section, bool fin);
+
+// Continues sending headers on the given stream.
+int quiche_h3_continue_partial_headers(quiche_h3_conn *conn,
+                            quiche_conn *quic_conn, uint64_t stream_id);
```
antoniovicente: weird indentation

quiche/include/quiche.h
```diff
@@ -1134,6 +1137,16 @@ int quiche_h3_send_additional_headers(quiche_h3_conn *conn,
                             quiche_h3_header *headers, size_t headers_len,
                             bool is_trailer_section, bool fin);
 
+// Initiates streaming of a new HTTP/3 HEADERS frame.
+int quiche_h3_stream_headers(quiche_h3_conn *conn,
+                            quiche_conn *quic_conn, uint64_t stream_id,
```
antoniovicente: weird indentation; 1 space missing on this line and next 2.

quiche/src/h3/mod.rs
```diff
@@ -1474,25 +1614,106 @@ impl Connection {
...
+    ///
+    /// The [`PartialHeader`] error is returned when the underlying QUIC stream
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application should retry the operation once the stream is
+    /// reported as writable again.
+    ///
+    /// [`PartialHeader`]: enum.Error.html#variant.PartialHeaders
+    /// [`Done`]: enum.Error.html#variant.Done
+    pub fn continue_partial_headers(
+        &mut self, conn: &mut super::Connection, stream_id: u64,
```
antoniovicente, line 1756: Can we avoid this `stream_id` argument?  And limit the number of streams that can be in the partial header sent state to 1?

quiche/src/h3/mod.rs
```diff
@@ -1474,25 +1614,106 @@ impl Connection {
...
+    /// Continues sending headers on the given stream.
+    ///
+    /// The [`PartialHeader`] error is returned when the underlying QUIC stream
+    /// doesn't have enough capacity for the operation to complete. When this
+    /// happens the application should retry the operation once the stream is
+    /// reported as writable again.
+    ///
+    /// [`PartialHeader`]: enum.Error.html#variant.PartialHeaders
+    /// [`Done`]: enum.Error.html#variant.Done
+    pub fn continue_partial_headers(
```
antoniovicente, line 1755: Do we need a check to make sure that `send_body` is not called on a stream that is waiting for calls to `continue_partial_headers`?

quiche/src/h3/mod.rs
gregor-cf: I think we need to allow multiple streams to be in the partial header state. Otherwise `stream_headers()` would need to return an error if there's already a stream in a partial state and the application would need to deal with retries.

apps/src/common.rs
LPardue: The examples are not intended to be performant, and ideally provide a simple demonstration of the API.  While quiche-client can make multiple parallel requests its not really intended to be used as a load tester; I expect most uses of would be single digits and wouldn't find the looping to be a bottleneck.

A different approach could be to create multiple collections that track state independently but I suspect some methods will upset the borrow checker. I'm not sure the effort is really worth it.

apps/src/common.rs
LPardue: It seems fine to try it IMO. The reason a headers send might fail could be due to an individual stream's flow control, for example. 

A production application would probably want to avoid doing unnecessary work if the reason a send failed was due to no capacity driven by connection-level flow control. But I suspect that verges more into your recent territory :smile:  I.e. that seems like an opportunity we can pursue separately.

Finally, this sub-optimal naiive behavour might help catch cases where a QUIC connection is behaving unexpectedly, by logging when sending didn't work.

quiche/src/h3/mod.rs
LPardue: Yeah this definately needs to support multiple buffered streams, otherwise it introduces a form of HoL blocking. For a server, the amount of state would be bounded by the number of concurrent requests that was advertised to a peer (i.e. controlled locally). For a client, I'll double check that its also bounded by the concurrency limit. We might also consider if there should be some client-local config to set a value different from the server's advertised limit - although I think applications can implement that limit themselves easily enough.

tokio-quiche/src/http3/driver/client.rs
LPardue: Yeah, I'm not sure why I did this in the first place. Seems much better to land the change to quiche and tokio-quiche as separate PRs.

quiche/include/quiche.h
LPardue: This one is consistent with other functions that have long names and several params

quiche/src/h3/mod.rs
LPardue: I've tweaked the docs to make it clear this is a TransportError with value StreamLimit (there was also a prior case of this in unrelated code, so fixed that too). Added docs for StreamCreationError, which I think addresses https://github.com/cloudflare/quiche/pull/2153/files#r2337882084 too

quiche/src/h3/mod.rs
LPardue: I'm satisfied with the use of `StreamCreationError` here. It really is the inability to make new streams, rather than try to emit something on them (which is what send_request did combined). Errors often need context to understand - a client can never be a server, so it would only see this after receiving a goaway. I've now documented this, so it should be clearer.

quiche/src/h3/mod.rs
LPardue: I've removed this function from the PR, since it is not strictly required for anything to do with streaming headers.

LPardue: Sorry, I confused myself by removing the FFI part. I don't think this needs to be pub as part of this PR but it does need to do a stream check as you highlighted, so I removed the `pub` and added the check. Thx

quiche/src/h3/mod.rs
LPardue: good spot!

quiche/src/h3/mod.rs
LPardue: good spot~

quiche/src/h3/mod.rs
LPardue: This is a "private" error type that is never intended to be communicated to a peer. I've pushed up a tweak that makes this `unimplemented!` just in case someone wants to try this illegal operation. Not sure I like that approach but not sure what's the best way (while avoiding serious surgery on the error codes)

quiche/src/h3/mod.rs
LPardue: 1) I forgot a `not` (`!`) in the stream check, derp
2) When I corrected that, it still tripped up some tests. Investigating that, we have code that does some quite aggressive state tidyup in https://github.com/cloudflare/quiche/pull/2153/files#diff-851097865419440f06ef5f0c6fcbca011863b552fe556fdcded5c368fdeb9715R1700

TL;DR the current quiche implementation currently relies on setting the transport priority after data is buffered into quiche. This kinda makes sense because you want to control the frame and packet scheduler only after there's stuff to send. So I think I'll back the stream existing check out.

It adds an annoying wrinkle to proceedings, since its made me remember I need to do some work on controlling the priority of the streaming sends too. The simplest thing to do might be to keep the h3 `stream_priority` method private and add `with_priority` variants for the new streaming headers functions. This would open the door for clients to do some interesting things if they wanted (aka. a possible footgun) but means we can keep the model a bit more consistent. Will mull over this a bit.

quiche/src/h3/mod.rs
LPardue: I got claude to help me make the headers sending state machine clearer and cover of all edge cases, should be addressed now.

quiche/src/h3/mod.rs
LPardue: I got claude to help me make the headers sending state machine clearer and cover of all edge cases, should be addressed now.

quiche/src/h3/mod.rs
LPardue: I got claude to help me make the headers sending state machine clearer and cover of all edge cases. As part of this some of the internal APIs were tweaked and I think addresses the first point.

Regarding the second point, we do want to distinguish in_flight vs sent, because there may be a need to send extension HTTP/3 frames in some future point. Seems easiest to make the state machine clean now rather than leave a quasi state we might trip on later

PR #2155, opened 2025-09-10
Cap the value to ensure it fits into a varint as otherwise quiche wil…

normanmaurer:
…l crash when trying to encode it

Motivation:

We need to ensure we cap various values that are passed via u64 as we need to be able to encude these via varint encoding. If we fail to cap these we will crash quiche when using a too big value:

```
thread '<unnamed>' panicked at octets/src/lib.rs:691:9:
internal error: entered unreachable code
stack backtrace:
   0:        0x1420af66c - <unknown>
   1:        0x1420ee99c - <unknown>
   2:        0x1420a581c - <unknown>
   3:        0x1420af520 - <unknown>
   4:        0x1420b1810 - <unknown>
   5:        0x1420b164c - <unknown>
   6:        0x1420b23d8 - <unknown>
   7:        0x1420b1f7c - <unknown>
   8:        0x1420afb20 - <unknown>
   9:        0x1420b1c5c - <unknown>
  10:        0x14210c9e8 - <unknown>
  11:        0x14210ca58 - <unknown>
  12:        0x142027734 - <unknown>
  13:        0x141f84c88 - <unknown>
  14:        0x141f711cc - <unknown>
  15:        0x141fb755c - <unknown>
  16:        0x141e31904 - <unknown>
fatal runtime error: failed to initiate panic, error 5
```

Modifications:

Cap the values to a value that can be encoded via varint

Result:

No more crashes


quiche/src/lib.rs
```diff
@@ -512,6 +512,9 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
+// Maximum value that can be encoded via varint.
+const MAX_VAR_INT: u64 = 4_611_686_018_427_387_903;
```
LPardue: Worth making this public in case people want to use it themselves?

normanmaurer: no strong preference here :) Just let me know and I will change it

quiche/src/lib.rs
```diff
@@ -1137,14 +1140,15 @@ impl Config {
     ///
     /// The default value is infinite, that is, no timeout is used.
     pub fn set_max_idle_timeout(&mut self, v: u64) {
-        self.local_transport_params.max_idle_timeout = v;
+        self.local_transport_params.max_idle_timeout = cmp::min(v, MAX_VAR_INT);
```
antoniovicente: Worth also checking against this max at serialization time?

We'ld still debug assert or similar in cases where the value plumbed through is above the allowed ranges so we can detect bugs.  But at least in release modes we'ld avoid crashing.

quiche/src/lib.rs
```diff
@@ -512,6 +512,9 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
+// Maximum value that can be encoded via varint.
```
antoniovicente: Would be good to have some tests to cover these functional changes.

quiche/src/lib.rs
normanmaurer: sure I can add some asserts as well.

quiche/src/lib.rs
normanmaurer: I can add some tests that would have crashed before. Might need to do this tomorrow tho

antoniovicente: No rush.

Thanks for looking into all of this.

quiche/src/lib.rs
ghedo: Yeah, makes sense, though should be in octets crate https://github.com/cloudflare/quiche/blob/master/octets/src/lib.rs

quiche/src/cid.rs
```diff
@@ -282,14 +283,16 @@ impl ConnectionIdentifiers {
             },
         );
 
+        // Guard against overflow.
+        let value =
+            (destination_conn_id_limit as u64) * RETIRED_CONN_ID_LIMIT_MULTIPLIER;
+        let size = cmp::min(usize::MAX as u64, value) as usize;
```
normanmaurer, line 289: This change was needed to guard against overflow on i686:

```
---- tests::configuration_values_are_limited_to_max_varint stdout ----



thread 'tests::configuration_values_are_limited_to_max_varint' panicked at quiche/src/cid.rs:291:17:

attempt to multiply with overflow

failures:

    tests::configuration_values_are_limited_to_max_varint
```

normanmaurer: Should we do this in a separate PR or is it fine to include this here ?

normanmaurer:
Can someone rerun the stuck test... this is not related

normanmaurer:
@ghedo @antoniovicente @LPardue ping...

ghedo:
Merged, thanks @normanmaurer!

PR #2158, opened 2025-09-17
fix: connection flow control updates with  STOP_SENDING and RESET_STREAM

gregor-cf:
This fixes issues in quiche, where quiche does not properly update connection level flow control when sending STOP_SENDING or receiving RESET_STREAM frames.

On calling `shutdown_stream(id, Shutdown::Read)`, quiche drops the receive buffer (correct) but it should also mark these bytes as "consumed" to the connection-level flow-control. 

On receiving a `RESET_STREAM` frame, quiche drops the receive buffer but it should also mark these dropped bytes as "consumed" to the connection-level flow-control. 

When Quiche updates "consumed" in the two above cases, it should also check if needs to send a MAX_DATA frame after updating "consumed".

Note, the two added `quiche::h3` tests are not directly related but I started there because I first suspected streams not getting collected. I did add tests to quiche itself that test my changes though.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -4987,6 +4987,9 @@ mod tests {
 
         // No event generated at server
         assert_eq!(s.poll_server(), Err(Error::Done));
+
+        assert!(s.pipe.server.streams.is_collected(0));
+        assert!(s.pipe.client.streams.is_collected(0));
     }
 
     #[test]
@@ -7120,6 +7123,170 @@ mod tests {
         );
 
         assert_eq!(s.poll_server(), Err(Error::Done));
+    }
+
+    /// The client shuts down the stream's write direction, the server
+    /// shuts down its side with fin
+    #[test]
+    fn client_shutdown_write_server_fin() {
+        let mut buf = [0; 65535];
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        // Client sends request.
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        // Server sends response and closes stream.
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        let resp = s.send_response(stream, true).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: resp,
+            more_frames: false,
+        };
+
+        assert_eq!(s.poll_client(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_client(), Ok((stream, Event::Finished)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+
+        // Client shuts down stream ==> sends RESET_STREAM
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Write, 42),
+            Ok(())
+        );
+        assert_eq!(s.advance(), Ok(()));
+
+        // Server sees the Reset event for the stream.
+        assert_eq!(s.poll_server(), Ok((stream, Event::Reset(42))));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Streams have been collected by quiche
+        assert!(s.pipe.server.streams.is_collected(stream));
+        assert!(s.pipe.client.streams.is_collected(stream));
+
+        // Client sends another request, server sends response without fin
+        //
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        // Check that server has received the request.
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Server sends reponse without closing the stream.
+        let resp = s.send_response(stream, false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: resp,
+            more_frames: true,
+        };
+
+        assert_eq!(s.poll_client(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+
+        // Client shuts down stream ==> sends RESET_STREAM
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Write, 42),
+            Ok(())
+        );
+        assert_eq!(s.advance(), Ok(()));
+
+        // Server sees the Reset event for the stream.
+        assert_eq!(s.poll_server(), Ok((stream, Event::Reset(42))));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Server sends body and closes the stream.
+        s.send_body_server(stream, true).unwrap();
+
+        // Stream has been collected on server by quiche
+        assert!(s.pipe.server.streams.is_collected(stream));
+        // Client stream has not been collected, the client needs to
+        // read the fin from the stream first.
+        assert!(!s.pipe.client.streams.is_collected(stream));
+        assert_eq!(s.poll_client(), Ok((stream, Event::Data)));
+        s.recv_body_client(stream, &mut buf).unwrap();
+        assert_eq!(s.poll_client(), Ok((stream, Event::Finished)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+        assert!(s.pipe.client.streams.is_collected(stream));
+    }
+
+    #[test]
+    fn client_shutdown_read() {
+        let mut buf = [0; 65535];
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        // Client sends request and leaves stream open.
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        // Server sends response and leaves stream open.
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        let resp = s.send_response(stream, false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: resp,
+            more_frames: true,
+        };
+
+        assert_eq!(s.poll_client(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+        // Client shuts down read
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Read, 42),
+            Ok(())
+        );
+        assert_eq!(s.advance(), Ok(()));
+
+        // Stream is writable on server side, but returns StreamStopped
+        assert_eq!(s.poll_server(), Err(Error::Done));
+        let writables: Vec<u64> = s.pipe.server.writable().collect();
+        assert!(writables.contains(&stream));
+        assert_eq!(
+            s.send_body_server(stream, false),
+            Err(Error::TransportError(crate::Error::StreamStopped(42)))
+        );
+
+        // Client needs to finish its side by sending a fin
+        assert_eq!(
+            s.client.send_body(&mut s.pipe.client, stream, &[], true),
+            Ok(0)
+        );
+        assert_eq!(s.advance(), Ok(()));
+        // Note, we get an Event::Data for an empty buffer today. But it
+        // would also be fine to not get it.
+        assert_eq!(s.poll_server(), Ok((stream, Event::Data)));
+        assert_eq!(s.recv_body_server(stream, &mut buf), Err(Error::Done));
+        assert_eq!(s.poll_server(), Ok((stream, Event::Finished)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Since the client has already send a fin, the stream is collected
+        // on both client and server
+        assert!(s.pipe.client.streams.is_collected(stream));
+        assert!(s.pipe.server.streams.is_collected(stream));
     }
 
     #[test]
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -5728,7 +5728,11 @@ impl<F: BufFactory> Connection<F> {
 
         match direction {
             Shutdown::Read => {
-                stream.recv.shutdown()?;
+                let consumed = stream.recv.shutdown()?;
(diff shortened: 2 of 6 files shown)


LPardue:
I'm not sure this sullfy satisfies RFC 9000 requirements when using stop_sending because, IIUC, you're accounting for bytes in the connection-level flow control before receiving the final size

> An endpoint will know the final size for a stream when the receiving part of the stream enters the "Size Known" or "Reset Recvd" state ([Section 3](https://datatracker.ietf.org/doc/html/rfc9000#stream-states)). The receiver MUST use the final size of the stream to account for all bytes sent on the stream in its connection-level flow controller.

I might be wrong but worth double checking. If im right, then data in-flight that increased the final size before the stopper received it, could result in the endpoints getting out of sync with connection-level FC

gregor-cf:
> I'm not sure this sullfy satisfies RFC 9000 requirements when using stop_sending because, IIUC, you're accounting for bytes in the connection-level flow control before receiving the final size
> 

I think my logic is correct. When sending the STOP_SENDING frame, we drop all currently received bytes and I return FC credit based on the *highest offset received*. This is correct IMHO, it's like the application had read the full buffer (the difference is that we don't about gaps in the STOP_SENDING case). Since whatever final size we later receive MUST be larger or equal than this offest, otherwise we close due to protocol violation. We also mark the send-buffer as "draining" (that is existing code): any additional stream data we receive with a higher offset we record in the buffer's max_off and in the connection level rx_data counter. And we also immediately mark it as read to the connection level FC too (since we are not buffering it, like the app had read it). Eventually we will receive a FIN or a RESET_STREAM for this stream. At this time we know the final size and verify it's >= than the highest received offset. If the final size is larger than highest received, we update the conn's rx_data and FC accordingly. 
In other words: after sending STOP_SENDING, we essentially pretend that any newly arriving data is immediately read by the app (and we also pretend that we've received any gaps) and we still wait for a fin or RESET_FRAME to know the final size. 

One thing to note: the current code doesn't do a sanity check with the final size. It keeps a running count of received data and just verifies that the final size is consistent with what has been received previously.

LPardue:
Thanks for the details. Can we add a test case for the scenario where data arrives after one of the endpoints has stopped reading to ensure that the connection FC counters are tracked appropriately (or confirm that one of the existing tests does it) please?

gregor-cf:
I've updated the test `flow_control_drain()` (which wasn't really testing flow control) and added the test `flow_control_reset_stream()`

gregor-cf:
gentle ping @LPardue 
any more thoughts on this PR?

quiche/src/stream/recv_buf.rs
```diff
@@ -266,14 +272,18 @@ impl RecvBuf {
             return Err(Error::FinalSize);
         }
 
-        // Calculate how many bytes need to be removed from the connection flow
-        // control.
-        let max_data_delta = final_size - self.len;
-
         if self.error.is_some() {
-            return Ok(max_data_delta as usize);
```
LPardue, line 274: In the old code, the return value here is used while processing RESET_STEAM frames. Does this change imply the old code had a bug whereby you could send multiple RESET_STREAMS for the same stream and get it to inflate the amount of `rx_data` reported in the connection?

quiche/src/stream/recv_buf.rs
```diff
@@ -314,8 +324,10 @@ impl RecvBuf {
         self.flow_control.autotune_window(now, rtt);
     }
 
-    /// Shuts down receiving data.
-    pub fn shutdown(&mut self) -> Result<()> {
+    // Shuts down receiving data and returns the number of bytes
```
LPardue:
```suggestion
    /// Shuts down receiving data and returns the number of bytes
```

quiche/src/lib.rs
```diff
@@ -7757,18 +7761,28 @@ impl<F: BufFactory> Connection<F> {
...
-                self.rx_data += max_off_delta;
+                self.rx_data += max_data_delta;
+                // We dropped the recevice buffer, return connection level
+                // flow-control
+                self.flow_control.add_consumed(consumed_flowcontrol);
+
+                // ... and check if need to send an updated MAX_DATA frame
+                if self.should_update_max_data() {
+                    self.almost_full = true;
+                }
```
LPardue, line 7785: From a code reuse perspective, there's a very similar block in the Shutdown::Read handler - https://github.com/cloudflare/quiche/pull/2158/files#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R5731. Maybe it would make sense to stick that in a function?

In fact, all non-test uses self.should_update_max_data()  just result in setting self.almost_full = true, so it seems a bit weird it is how is. But I can live with not changing it as part of this PR.

quiche/src/lib.rs
```diff
@@ -7757,18 +7761,28 @@ impl<F: BufFactory> Connection<F> {
...
                     return Err(Error::FlowControl);
                 }
 
                 if !was_readable && stream.is_readable() {
                     self.streams.insert_readable(&priority_key);
                 }
 
-                self.rx_data += max_off_delta;
+                self.rx_data += max_data_delta;
+                // We dropped the recevice buffer, return connection level
```
```suggestion
                // We dropped the receive buffer, return connection level
```

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+        // Client shuts down stream ==> sends RESET_STREAM
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Write, 42),
+            Ok(())
+        );
+        assert_eq!(s.advance(), Ok(()));
+
+        // Server issues Reset event for the stream.
```
LPardue: This comment is confusing. The server receives the reset sent by the client.

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+
+        // Streams have been collected by quiche
+        assert!(s.pipe.server.streams.is_collected(stream));
+        assert!(s.pipe.client.streams.is_collected(stream));
+
+        // Client sends another request, server sends response without fin
+        //
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
```
LPardue, line 7180: nit: you could avoid redeclaring the event here if you assigned different names to the request or response event headers and then cloned them in the asserts.

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+            list: req,
+            more_frames: true,
+        };
+
+        // Server sends response and closes stream.
```
LPardue: I don't think this comment is accurate. The response is sent with `false` so the stream is not FIN'd and I don't see it being reset either.

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+        // Client shuts down stream ==> sends RESET_STREAM
+        assert_eq!(
+            s.pipe
+                .client
+                .stream_shutdown(stream, crate::Shutdown::Write, 42),
+            Ok(())
+        );
+        assert_eq!(s.advance(), Ok(()));
+
+        // Server issues Reset event for the stream.
```
LPardue: similar to above, this is the server receiving the reset

quiche/src/h3/mod.rs
LPardue: Oh sorry, I see it is FIN'd via send_body later on. Maybe would help to tweak the comment a little to avoid future me getting confused again :smile:

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+        assert!(s.pipe.client.streams.is_collected(stream));
+    }
+
+    #[test]
+    fn client_shutdown_read() {
+        let mut buf = [0; 65535];
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        // Client sends request with fin set.
```
LPardue: Similar to above, consider tweaking the comment to make it a little clearer the fin happens later on with the body, not with the headers.

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        // Server sends response and leaves stream open.
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        // assert_eq!(s.poll_server(), Ok((stream, Event::Finished)));
```
LPardue: nit: commented out code

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+            Err(Error::TransportError(crate::Error::StreamStopped(42)))
+        );
+
+        // Client needs to finish its side but sending a fin
+        assert_eq!(
+            s.client.send_body(&mut s.pipe.client, stream, &[], true),
+            Ok(0)
+        );
+        assert_eq!(s.advance(), Ok(()));
+        // TODO: why do we get an Event::Data event for an empty buffer?
```
LPardue: This is expected per the current design but a potential improvement (separate to this PR) could revist optimizing it out

quiche/src/h3/mod.rs
```diff
@@ -7122,6 +7125,168 @@ mod tests {
...
+        assert_eq!(s.poll_server(), Ok((stream, Event::Data)));
+        assert_eq!(s.recv_body_server(stream, &mut buf), Err(Error::Done));
+        assert_eq!(s.poll_server(), Ok((stream, Event::Finished)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // Since the client has already send a fin, the stream is collected
```
LPardue, line 7286: suggestion: you could also poll_client to confirm it received a Reset in response to its STOP_SENDING

quiche/src/stream/recv_buf.rs
gregor-cf: No. But it's not obvious. This return statement always returned 0. On the first RESET_STREAM, we'd enqueue a 0-length buf (line 286) with fin and the final offset, which causes self.len to be set to final size.

quiche/src/tests.rs
```diff
@@ -1444,34 +1444,249 @@ fn flow_control_update(
...
+    assert_eq!(r.next(), Some(0));
     assert_eq!(r.next(), None);
 
-    // Flush connection.
-    assert_eq!(pipe.advance(), Ok(()));
+    // check flow control accounting
+    assert_eq!(pipe.server.rx_data, 5);
+    assert_eq!(pipe.server.flow_control.consumed(), 0);
+
+    // Helper function that sends STREAM frames to the server on stream 4
```
LPardue: The comment is a lie! :cake: 

There is already a send_pkt_to_server helper, so I'm wondering if this really does anything different and we can cut down the code a bit. 

If it is different for some reason, we should at least cut out the duplicate helper in flow_control_drain via extraction/parameterization

quiche/src/tests.rs
```diff
@@ -2267,6 +2490,11 @@ fn reset_stream_flow_control(
         pipe.send_pkt_to_server(pkt_type, &frames, &mut buf),
         Err(Error::FlowControl),
     );
+
+    // stream 4 had 15 bytes, that should be consumed
+    assert_eq!(pipe.server.flow_control.consumed(), 15);
+    // a total of 30 bytes: 15 + 15 + 1 (streams 0, 4, and 8)
```
LPardue: doesn't that add up to 31?

quiche/src/tests.rs
```diff
@@ -2776,8 +3007,16 @@ fn stream_shutdown_read_after_fin(
     let mut r = pipe.server.readable();
     assert_eq!(r.next(), None);
 
-    // Server has nothing to send.
-    assert_eq!(pipe.server.send(&mut buf), Err(Error::Done));
+    // Server sends a flow control update, but it does NOT send
+    // STOP_SENDING
```
LPardue: Maybe worth clarifying that this is because the stream was FIN'd by the client

quiche/src/h3/mod.rs
gregor-cf: I actually have three different ev_headers in this test and that's the pattern used by all the other tests too :-)

quiche/src/h3/mod.rs
gregor-cf: good catch

quiche/src/lib.rs
gregor-cf: yeah. I might leave it as-is for this PR.

quiche/src/tests.rs
gregor-cf: > The comment is a lie! 🍰
:-). zero, four, what's the difference? 

`send_pkt_to_server` will do a `conn.recv()` followed by `conn.send()` to read any "response" packets into buf and I also wanted to remove the boiler-plate of creating the frames, etc. to keep the test more readable. 

Regarding the code duplication: In tests I tend to do more copy-paste if I think it makes it easier to understand the test. In this case, it makes it very clear what I'm sending exactly (just a single frame, not any ACKs, no bidirectional exchange) and it's just a couple of lines. I don't feel strongly about it though. If you prefer I'll factor it out.

quiche/src/tests.rs
gregor-cf: good catch. 
I'll actually remove these two asserts. This test actually tests that we get an `Error::FlowControl`, after which the connection should be closed so asserting on internal state of the connection probably doesn't really make sense.

quiche/src/h3/mod.rs
gregor-cf: The client doesn't actually recevice another event in `poll_client()`, but the 
```
assert!(s.pipe.client.streams.is_collected(stream));
```
makes sure that the client has received the RESET_STREAM.

LPardue: That seems a little unexpected but perhaps we swallow reset events if triggered by a local stop sending (i honestly can't remember). I agree the collected check is strong though. So we can probably just resolve this.

quiche/src/lib.rs
LPardue: Thats ok. Happy to close comment with no action.

quiche/src/stream/recv_buf.rs
LPardue: Ack!

quiche/src/tests.rs
LPardue: Ack

quiche/src/tests.rs
LPardue: Thanks for the explanation. I'm happy to leave it as is right now.

jrouviere (approved):

PR #2161, opened 2025-09-18
feat: expose option to disable probeDown early-exit behavior

toidiu:
BBRv3 probeDown currently exits based on time duration. This is Google experimental code which is not mentioned in the BBR RFC. This PR introduces an option to disable the experimental behavior.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
+++ b/quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
@@ -320,7 +320,10 @@ impl ProbeBW {
             return;
         }
 
-        if self.has_stayed_long_enough_in_probe_down(congestion_event) {
+        // This exit condition is experimental code from Google quiche which
+        // diverges from the RFC. Use `disable_probe_down_early_exit` to override
+        // the behavior.
+        if self.has_stayed_long_enough_in_probe_down(congestion_event, params) {
             self.enter_probe_cruise(congestion_event.event_time);
             return;
         }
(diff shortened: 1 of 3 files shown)


quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
```diff
@@ -560,11 +564,15 @@ impl ProbeBW {
...
     // long, as seen in some multi-sender simulator tests.
     fn has_stayed_long_enough_in_probe_down(
-        &self, congestion_event: &BBRv2CongestionEvent,
+        &self, congestion_event: &BBRv2CongestionEvent, params: &Params,
     ) -> bool {
-        // Stay in PROBE_DOWN for at most the time of a min rtt, as it is done in
-        // BBRv1.
-        self.has_phase_lasted(self.model.min_rtt(), congestion_event)
+        if params.disable_probe_down_early_exit {
+            false
```
antoniovicente: I would have considered checking the disable_probe_down_early_exit param at the one call site or done "return false;" on this line to avoid unnecessary and undesirable whitespace changes in the rest of this function.

Heavily indented code is harder to read and comprehend that code that is less indented.  Adding to the indentation level of the old version of the code adds unnecessary bulk to this code and the whitespace changes create unwanted churn on the code which affects git blame history.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
```diff
@@ -320,7 +320,11 @@ impl ProbeBW {
             return;
         }
 
-        if self.has_stayed_long_enough_in_probe_down(congestion_event) {
+        // This is experimental code from Google quiche and diverges from the RFC.
```
antoniovicente: It may make sense to move this comment inside the function as justification for the early return when disable_probe_down_early_exit is set.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
toidiu: Done. The comment still shows a diff but atleast the code is still there.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
toidiu: I also added a comment to the function but want to keep it here since this is the call-site in probeDown where it has effect and where a reader is likely to run into it.

antoniovicente (approved):

PR #2162, opened 2025-09-19
tokio-quiche: fix possible infinite loop in H3Driver

gregor-cf:
When the H3Driver sends a `fin` on a stream, it calls `close()` on associated `Receiver<OutboundFrame>`. But tokio has an inconsistency between `mpsc::Receiver::try_recv()` and `mpsc::Receiver::poll_recv()` when called on a closed Receiver (not closed Sender).
* try_recv() will always return TryRecvError::Empty
* poll_recv() will always return Poll::Ready(None) (indicating the channel is closed)

This leads to a infite loop. In `process_writeable_stream()` we use `try_recv()`, see empty and put the stream into `waiting_streams`. There we call `poll_recv()`, see that the channel is closed, and call `process_writable_stream()` again.

The fix is easy: drop the Receiver instead of calling close on it. (It's already in an Option)

see https://github.com/tokio-rs/tokio/issues/7631

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -659,10 +659,12 @@ impl<H: DriverHooks> H3Driver<H> {
                     return Ok(());
                 }
                 if *fin {
-                    // If this is the last body frame, close the receiver in the
-                    // stream map to signal that we shouldn't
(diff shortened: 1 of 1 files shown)


evanrittenhouse (approved):
Nice find!

PR #2165, opened 2025-09-22
tokio-quiche: better RESET_STREAM and STOP_SENDING handling

gregor-cf:
TL;DR:
* When receivign RESET_STREAM or STOP_SENDING, respond with RESET_STREAM and STOP_SENDING, respectively.
* When InboundFrame or OutboundFrame channels are closed without a fin, send STOP_SENDING or RESET_STREAM to close that half of the stream.
* Add a testing framework similar to quiche::test_utils::Pipe to allow unit testing of H3Driver.

This contains two changes to tokio-quiche's handling of "abnormal" stream termination via RESET_STREAM and STOP_SENDING. A stream is only fully completed if its final size in both directions is know and the final size is only communicated via a fin or RESET_STREAM frame. Note that quiche will automatically reply with a RESET_STREAM when it receives a STOP_SENDING frame.

That means that receiving a RESET_STREAM or STOP_SENDING only shuts down one direction of the stream, so tokio-quiche must also close the other direction. This PR does so by responding with RESET_STREAM or STOP_SENDING. Note, that I chose RequestCancelled as the error code in these messages, since I think that's a better fit than simply echoing back the error we receivd from the peer (since it's our choice to shut down the stream, we could have also continued normally until a fin).

When the H3Driver detects that either the InboundFrame or OutboundFrame channels have been closed without a fin, the driver shuts down that direction. This is necessary to allow quiche to release the streams resources (flow control credit). Since the channel is gone, it's otherwise impossible to finish the stream.


LPardue:
I'd need to analyse the code a bit more but a bit concerned this is a could be a broad stroke that isn't appropriate for all use cases. Resetting streams prevents delivery of stream application data and there might be cases where it is more appropriate to fin the stream than reset it (e.g. serving the remainder of a response).  I think you've mostly covered that off but I'd want to be sure. For example, if a client started to get the response to an upload that was like "4xx with problem details" and it responded with a client reset to avoid wasting bandwidth on any in-flight data, a server issuing a reset in reaction could interupt delivery of the problem body.

This is why quiche H3 non-TQ API surfaced the reset to applications, to let them decide precisely how to react to the read side getting reset.

gregor-cf:
> I'd need to analyse the code a bit more but a bit concerned this is a could be a broad stroke that isn't appropriate for all use cases. Resetting streams prevents delivery of stream application data and there might be cases where it is more appropriate to fin the stream than reset it (e.g. serving the remainder of a response). I think you've mostly covered that off but I'd want to be sure. For example, if a client started to get the response to an upload that was like "4xx with problem details" and it responded with a client reset to avoid wasting bandwidth on any in-flight data, a server issuing a reset in reaction could interupt delivery of the problem body.
> 
> This is why quiche H3 non-TQ API surfaced the reset to applications, to let them decide precisely how to react to the read side getting reset.

Good points. I actually wrote a version that didn't echo back the RESET or STOP_SENDING first but I shelved it.... I'll put it up as an alternative. That said, not echoing back RESET or STOP_SENDING would be a behavioral change for TQ (today, it simply drops its stream state and closes the InboundFrame/OutboundFrame to the application). So this PR mimics TQ's current behavior, expect that it ensure quiche can clean up its state.

LPardue:
Heh yeah, there's some legacy here and I've always had some issues with TQ's opinions in this area (before it was even called that). Thanks for trying to thread the needle here, there are definately improvements to be made.

gregor-cf:
#2166 is the version that doesn't blindly echo back RESET_STREAM or STOP_SENDING. 

Also, TQ doesn't have a good way to signal a STOP_SENDING back the application. Closing the InboudFrame channel is interpreted by the app as regular fin (and indeed there is a code path in TQ today where a regular fin is not reported to the application as InboundFrame::Body but only by closing the channel.  A RESET_STREAM is easier to disambiguate in the app)

gregor-cf:
Closing in favor of #2166

PR #2166, opened 2025-09-22
tokio-quiche: better RESET_STREAM and STOP_SENDING handling

gregor-cf:
Alternative version to #2165 (This PR does not echo back a received RESET_FRAME or STOP_SENDING and instead lets the application close the other stream direction on its own. )

TL;DR:
* When receiving RESET_STREAM or STOP_SENDING, we mark the appropriate
  direction of the stream as closed.
* When InboundFrame or OutboundFrame channels are closed without a fin,
  send STOP_SENDING or RESET_STREAM to close that half of the stream.
* Add a testing framework similar to quiche::test_utils::Pipe to allow
  unit testing of H3Driver.

This contains two changes to tokio-quiche's handling of "abnormal" stream
termination via RESET_STREAM and STOP_SENDING. A stream is only fully
completed if its final size in both directions is know and the final size
is only communicated via a fin or RESET_STREAM frame. Note that quiche
will automatically reply with a RESET_STREAM when it receives a STOP_SENDING
frame.

That means that receiving a RESET_STREAM or STOP_SENDING only shuts down one
direction of the stream, so tokio-quiche must also close the other direction.
This PR does so by marking the appropriate direction of the stream closed when
receiving these frames. The other direction must still be closed (fin or reset).

When the H3Driver detects that either the InboundFrame or OutboundFrame channels
have been closed without a fin, the driver shuts down that direction by sending
STOP_SENDING or RESET STREAM. This is
necessary to allow quiche to release the streams resources (flow control credit).
Since the channel is gone, it's otherwise impossible to finish the stream.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -33,6 +33,10 @@ mod datagram;
 mod hooks;
 mod server;
 mod streams;
+#[cfg(test)]
+pub mod test_utils;
+#[cfg(test)]
+mod tests;
 
 use std::collections::BTreeMap;
 use std::error::Error;
@@ -443,6 +447,7 @@ impl<H: DriverHooks> H3Driver<H> {
 
         enum StreamStatus {
             Done { close: bool },
+            Reset { wire_err_code: u64 },
             Blocked,
         }
 
@@ -453,27 +458,52 @@ impl<H: DriverHooks> H3Driver<H> {
                 break StreamStatus::Done { close: false };
             };
 
-            let permit = match sender.try_reserve() {
+            let try_reserve_result = sender.try_reserve();
+            let permit = match try_reserve_result {
                 Ok(permit) => permit,
                 Err(TrySendError::Closed(())) => {
+                    // The channel has closed before we delivered a fin or reset
+                    // to the application.
+                    if !ctx.fin_or_reset_recv &&
+                        ctx.associated_dgram_flow_id.is_none()
+                    // The channel might be closed if the stream was used to
+                    // initiate a datagram exchange.
+                    // TODO: ideally, the application would still shut down the
+                    // stream properly. Once applications code
+                    // is fixed, we can remove this check.
+                    {
+                        let err = h3::WireErrorCode::RequestCancelled as u64;
+                        let _ = qconn.stream_shutdown(
+                            stream_id,
+                            quiche::Shutdown::Read,
+                            err,
+                        );
+                        drop(try_reserve_result); // needed to drop the borrow on ctx.
+                        ctx.handle_sent_stop_sending(err);
+                        // TODO: should we send an H3Event event to
+                        // h3_event_sender? We can only get here if the app
+                        // actively closed or dropped
+                        // the channel so any event we send would be more for
+                        // logging or auditing
+                    }
                     break StreamStatus::Done {
-                        close: ctx.fin_sent && ctx.fin_recv,
+                        close: ctx.both_directions_done(),
                     };
                 },
                 Err(TrySendError::Full(())) => {
-                    if ctx.fin_recv || qconn.stream_readable(stream_id) {
+                    if ctx.fin_or_reset_recv || qconn.stream_readable(stream_id) {
                         break StreamStatus::Blocked;
                     }
                     break StreamStatus::Done { close: false };
                 },
             };
 
-            if ctx.fin_recv {
+            if ctx.fin_or_reset_recv {
                 // Signal end-of-body to upstream
                 permit
                     .send(InboundFrame::Body(BufFactory::get_empty_buf(), true));
                 break StreamStatus::Done {
-                    close: ctx.fin_sent,
+                    close: ctx.fin_or_reset_sent,
                 };
             }
 
@@ -497,6 +527,13 @@ impl<H: DriverHooks> H3Driver<H> {
                 },
                 Err(h3::Error::Done) =>
                     break StreamStatus::Done { close: false },
+                Err(h3::Error::TransportError(quiche::Error::StreamReset(
+                    code,
+                ))) => {
+                    break StreamStatus::Reset {
+                        wire_err_code: code,
+                    };
+                },
                 Err(_) => break StreamStatus::Done { close: true },
             }
         };
@@ -515,8 +552,18 @@ impl<H: DriverHooks> H3Driver<H> {
                 // blocked. qconn.stream_finished() will guarantee
                 // that we've fully parsed the body as it only returns true
                 // if we've seen a Fin for the read half of the stream.
-                if !ctx.fin_recv && qconn.stream_finished(stream_id) {
+                if !ctx.fin_or_reset_recv && qconn.stream_finished(stream_id) {
                     return self.process_h3_fin(qconn, stream_id);
+                }
+            },
+            StreamStatus::Reset { wire_err_code } => {
+                debug_assert!(ctx.send.is_some());
+                ctx.handle_recvd_reset(wire_err_code);
+                self.h3_event_sender
+                    .send(H3Event::ResetStream { stream_id }.into())
+                    .map_err(|_| H3ConnectionError::ControllerWentAway)?;
+                if ctx.both_directions_done() {
+                    return self.finish_stream(qconn, stream_id, None, None);
                 }
             },
             StreamStatus::Blocked => {
@@ -531,13 +578,16 @@ impl<H: DriverHooks> H3Driver<H> {
     fn process_h3_fin(
         &mut self, qconn: &mut QuicheConnection, stream_id: u64,
     ) -> H3ConnectionResult<()> {
-        let ctx = self.stream_map.get_mut(&stream_id).filter(|c| !c.fin_recv);
+        let ctx = self
+            .stream_map
+            .get_mut(&stream_id)
+            .filter(|c| !c.fin_or_reset_recv);
         let Some(ctx) = ctx else {
             // Stream is already finished, nothing to do
             return Ok(());
         };
 
-        ctx.fin_recv = true;
+        ctx.fin_or_reset_recv = true;
         ctx.audit_stats
             .set_recvd_stream_fin(StreamClosureKind::Explicit);
 
@@ -577,15 +627,36 @@ impl<H: DriverHooks> H3Driver<H> {
             h3::Event::Finished => self.process_h3_fin(qconn, stream_id),
 
             h3::Event::Reset(code) => {
-                if let Some(ctx) = self.stream_map.get(&stream_id) {
-                    ctx.audit_stats.set_recvd_reset_stream_error_code(code as _);
+                if let Some(ctx) = self.stream_map.get_mut(&stream_id) {
+                    ctx.handle_recvd_reset(code);
+                    // See if we are waiting on this stream and close the channel
+                    // if we are. If we are not waiting, `handle_recvd_reset()`
+                    // will have taken care of closing.
+                    for pending in self.waiting_streams.iter_mut() {
+                        match pending {
+                            WaitForStream::Upstream(
+                                WaitForUpstreamCapacity {
+                                    stream_id: id,
+                                    chan: Some(chan),
+                                },
+                            ) if stream_id == *id => {
+                                chan.close();
+                            },
+                            _ => {},
+                        }
+                    }
+
+                    self.h3_event_sender
+                        .send(H3Event::ResetStream { stream_id }.into())
+                        .map_err(|_| H3ConnectionError::ControllerWentAway)?;
+                    if ctx.both_directions_done() {
+                        return self.finish_stream(qconn, stream_id, None, None);
+                    }
                 }
 
-                self.h3_event_sender
-                    .send(H3Event::ResetStream { stream_id }.into())
-                    .map_err(|_| H3ConnectionError::ControllerWentAway)?;
-
-                self.finish_stream(qconn, stream_id, None, None)
+                // TODO: if we don't have the stream in our map: should we
+                // send the H3Event::ResetStream?
+                Ok(())
             },
 
             h3::Event::PriorityUpdate => Ok(()),
@@ -630,6 +701,11 @@ impl<H: DriverHooks> H3Driver<H> {
         let Some(frame) = &mut ctx.queued_frame else {
             return Ok(());
         };
+        debug_assert!(
+            ctx.recv.is_some(),
+            "If we have a queued frame in the context, we MUST NOT be waiting
+             on data from the channel"
+        );
 
         let audit_stats = &ctx.audit_stats;
         let stream_id = audit_stats.stream_id();
@@ -701,10 +777,10 @@ impl<H: DriverHooks> H3Driver<H> {
                     Err(h3::Error::StreamBlocked)
                 } else {
                     if *fin {
-                        ctx.fin_sent = true;
+                        ctx.fin_or_reset_sent = true;
                         audit_stats
                             .set_sent_stream_fin(StreamClosureKind::Explicit);
-                        if ctx.fin_recv {
+                        if ctx.fin_or_reset_recv {
                             // Return a TransportError to trigger stream cleanup
                             // instead of h3::Error::Done
                             return Err(h3::Error::TransportError(
@@ -954,8 +1030,12 @@ impl<H: DriverHooks> H3Driver<H> {
                 Err(h3::Error::TransportError(quiche::Error::StreamStopped(
                     e,
                 ))) => {
-                    ctx.audit_stats.set_recvd_stop_sending_error_code(e as i64);
-                    return self.finish_stream(qconn, stream_id, Some(e), None);
+                    ctx.handle_recvd_stop_sending(e);
+                    if ctx.both_directions_done() {
+                        return self.finish_stream(qconn, stream_id, None, None);
+                    } else {
+                        return Ok(());
+                    }
                 },
                 Err(h3::Error::TransportError(
                     quiche::Error::InvalidStreamState(stream),
@@ -970,6 +1050,11 @@ impl<H: DriverHooks> H3Driver<H> {
             let Some(recv) = ctx.recv.as_mut() else {
                 // This stream is already waiting for data or we wrote a fin and
                 // closed the channel.
+                debug_assert!(
+                    ctx.queued_frame.is_none(),
+                    "We MUST NOT have a queued frame if we are already waiting on 
+                    more data from the channel"
+                );
                 return Ok(());
             };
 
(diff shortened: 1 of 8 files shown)


tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,66 @@ impl StreamCtx {
...
+    pub(crate) fn handle_recvd_stop_sending(&mut self, wire_err_code: u64) {
+        debug_assert!(!self.fin_or_reset_sent);
+        // We received a STOP_SENDING frame. This indicates that the
+        // write direction has closed. We still need to continue
+        // reading from the stream until we receive a RESET_FRAME or
+        // `fin`.
+        self.audit_stats
+            .set_recvd_stop_sending_error_code(wire_err_code as i64);
+        self.audit_stats
+            .set_sent_stream_fin(StreamClosureKind::Implicit);
```
LPardue: Setting sent stream fin as StreamClosureKind::Implicit seems new. I'm not 100% what it is supposed to mean. The only other use of this kind is in a `drop()`.

Currently, when the quiche library receives a STOP_SENDING with an error code, it will implicitly reset the stream using the same error code.

From an audit stats perspective, it would be more technically correct to instead call `set_sent_reset_stream_error_code(wire_err_code)` here. However, the quiche behavior has been like that for a long time, so I'm not sure if TQ explicitly made the decision not to record such a reset in the audit stats, in order to not confuse it with explicit calls to do a reset. In an ideal world the reset type would articulate implicit or explicit but we have legacy use cases for the stats that wouldn't like that change.

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,66 @@ impl StreamCtx {
...
+    pub(crate) fn handle_recvd_reset(&mut self, wire_err_code: u64) {
+        debug_assert!(!self.fin_or_reset_recv);
+        // We received a RESET_STREAM frame, which closes the read direction
+        // but not the write direction. If the peer wants to shut down write,
+        // it must also send STOP_SENDING
+        self.audit_stats
+            .set_recvd_reset_stream_error_code(wire_err_code as i64);
+        // TODO: is StreamClosureKind::Implicit the right choice here?
+        self.audit_stats
+            .set_recvd_stream_fin(StreamClosureKind::Implicit);
```
LPardue: Related to my earlier comment, I'm not sure we have a good handle on what the implicit fins mean for consumers of our audit stats. My memory is that we were trying to help differentiate between "clean" finishes and stopped/reset and then added Implicit to cover the case where the process just tidied stuff away without any on-wire messages. However, my memory might be hazy or other events might have invalidated it.

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,66 @@ impl StreamCtx {
...
+        self.fin_or_reset_recv = true;
+        self.send = None;
+    }
+
+    pub(crate) fn hanle_sent_reset(&mut self, wire_err_code: u64) {
```
LPardue: nit: 
```suggestion
    pub(crate) fn handle_sent_reset(&mut self, wire_err_code: u64) {
```

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,66 @@ impl StreamCtx {
...
+        debug_assert!(!self.fin_or_reset_sent);
+        self.audit_stats
+            .set_sent_reset_stream_error_code(wire_err_code as i64);
+        self.audit_stats
+            .set_sent_stream_fin(StreamClosureKind::Implicit);
```
LPardue: Similar comment about `StreamClosureKind::Implicit` :smile:

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,66 @@ impl StreamCtx {
...
+
+    pub(crate) fn handle_sent_stop_sending(&mut self, wire_err_code: u64) {
+        debug_assert!(!self.fin_or_reset_recv);
+        self.audit_stats
+            .set_sent_stop_sending_error_code(wire_err_code as i64);
+        // It is ok for us to set the `fin_reset_recv` flag here.  While the peer
+        // must still send a fin or reset_stream with its final size, we don't
+        // need to read it from the stream. Quiche will take care of that.
+        self.audit_stats
+            .set_recvd_stream_fin(StreamClosureKind::Implicit);
```
LPardue: Similar comment about `StreamClosureKind::Implicit` :smile:

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -56,9 +57,9 @@ pub(crate) struct StreamCtx {
     /// First time that a HEADERS frame was not fully flushed.
     pub(crate) first_full_headers_flush_fail_time: Option<Instant>,
     /// Indicates the stream received fin. No more data will be received.
```
LPardue: nit: 
```suggestion
    /// Indicates the stream received fin or reset. No more data will be received.
```

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -56,9 +57,9 @@ pub(crate) struct StreamCtx {
...
-    pub(crate) fin_recv: bool,
+    pub(crate) fin_or_reset_recv: bool,
     /// Indicates the stream sent fin. No more data will be sent.
```
LPardue: nit: 
```suggestion
    /// Indicates the stream sent fin or reset. No more data will be sent.
```

tokio-quiche/src/http3/driver/test_utils.rs
```diff
@@ -0,0 +1,395 @@
+use std::time::Instant;
```
LPardue, line 1: Nice work on adding tests/test utils. I think @evanrittenhouse probably has some feedback to add here.

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
+use crate::http3::driver::client::ClientHooks;
+use crate::http3::driver::server::ServerHooks;
+use assert_matches::assert_matches;
+
+use super::test_utils::*;
+use super::*;
+
+#[test]
+fn simple_client_driver() {
```
LPardue: This is an "unconventional" test in HTTP terms. Neither endpoint seems to send a content-length and they both exchange headers before sending content. That sort of thing is legal and useful for a bidirectional stream model (e.g. CONNECT-based tunnels), but I think it might be a stretch to name it `simple_client_driver`. Happy to keep the test under a different name, since it exercises the machinery we care about. Maybe something like "client_fin_before_server_body`?

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+    // => {
+    //    assert_eq!(buf.into_inner().into_vec().len(), 0);
+    //    assert!(fin);
+    //});
+    assert_matches!(from_server.try_recv(), Err(TryRecvError::Disconnected));
+    assert_eq!(helper.driver.stream_map.len(), 0);
+}
+
+#[test]
+fn simple_server_driver() {
```
LPardue: similar comment about test name here.

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+        Err(h3::Error::Done)
+    );
+    assert_eq!(helper.peer_client_poll(), Ok((0, h3::Event::Finished)));
+
+    assert_eq!(helper.driver.stream_map.len(), 0);
+}
+
+// This test verifies https://github.com/cloudflare/quiche/pull/2162
+#[test]
+fn sending_fin() {
```
LPardue: nit: lets make the name a bit more specific since it seems focused on the client NOT sending FIN.  e.g. something like

```suggestion
fn server_fin_before_client_fin() {
```

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+    assert_eq!(body, &[1; 5]);
+    assert!(fin);
+
+    // Stream is done
+    assert_eq!(helper.driver.stream_map.len(), 0);
+}
+
+/// Test the case where the peer sends a STOP_SENDING quiche frame.
+#[test]
+fn peer_sends_stop_sending() {
```
LPardue: nit: `peer` is ambiguous since it requires knowledge of a vantage point, using the actual role means is easier to eyeball the test intent (useful in CI for example).

```suggestion
fn client_sends_stop_sending_before_server_fin() {
```

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+    assert_eq!(audit_stats.recvd_stream_fin(), StreamClosureKind::Explicit);
+    assert_eq!(audit_stats.sent_stream_fin(), StreamClosureKind::Implicit);
+    assert_eq!(audit_stats.downstream_bytes_recvd(), 3);
+    assert_eq!(audit_stats.downstream_bytes_sent(), 0);
+}
+
+/// Test the case where the peer sends a RESET_STREAM quiche frame.
+/// The peer sends its reset before we send a fin
+#[test]
+fn peer_sends_reset_stream_before_our_fin() {
```
LPardue: nit 

```suggestion
fn client_sends_reset_stream_before_server_fin() {
```

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+    assert_eq!(audit_stats.recvd_stream_fin(), StreamClosureKind::Implicit);
+    assert_eq!(audit_stats.sent_stream_fin(), StreamClosureKind::Explicit);
+    assert_eq!(audit_stats.downstream_bytes_recvd(), 0);
+    assert_eq!(audit_stats.downstream_bytes_sent(), 8);
+}
+
+/// Test the case where the peer sends a RESET_STREAM quiche frame.
+/// We send a fin before the peer sends reset
+#[test]
+fn peer_sends_reset_stream_after_our_fin() {
```
LPardue:
```suggestion
fn client_sends_reset_stream_after_server_fin() {
```

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+}
+
+/// Test the case where the peer sends a RESET_STREAM quiche frame while we're
+/// in the middle of reading data. We want to excercise the code-path where
+/// `upstream_ready` is called before `process_reads`. If `process_reads()` is
+/// called first, it will get the Reset event. If `upstream_ready()` is called
+/// first, it will attempt to read from the h3::Connection and will get a
+/// `TransportError(StreamReset(code))`
+#[test]
+fn peer_sends_reset_stream_while_reading_wait_for_data() {
```
LPardue:
```suggestion
fn client_sends_reset_stream_while_reading_wait_for_data() {
```

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+    assert_eq!(audit_stats.sent_stream_fin(), StreamClosureKind::Explicit);
+    assert_eq!(audit_stats.downstream_bytes_recvd(), 5);
+    assert_eq!(audit_stats.downstream_bytes_sent(), 8);
+}
+
+/// Test the case where the peer sends a RESET_STREAM quiche frame while we're
+/// in the middle of reading data. We want to excercise the code-path where
+/// where we call `process_reads` before `upstream_ready()`.
+#[test]
+fn peer_sends_reset_stream_while_reading_process_reads() {
```
LPardue:
```suggestion
fn client_sends_reset_stream_while_reading_process_reads() {
```

tokio-quiche/src/http3/driver/streams.rs
gregor-cf: Yeah. I wasn't sure what StreamClosureKind to pick. 
Not setting it here would probably most closely mimmic the current behavior, so I'll change it to that.

tokio-quiche/src/http3/driver/tests.rs
gregor-cf: Good point. My test naming is a bit inconsistent. I really want to express two different things: (a) whether the H3Driver is used on the client or server and (b) which side sends the STOP_SENDING. I think I might just use submodules.
```
mod server_driver {
    #[test]
    fn clients_sends_stop_sending_before_fin() { ...}
}
```

gregor-cf:
> Overall looks like on the right track. Most comments are minor or repetitive. I'd like some more of the team's opinion on the testing inside TQ - something that's been sorely lacking to do, so thanks for all those additions.

One question about the testing is, whether we should use a H3Driver for both the client and the server in the tests. I'm using an H3Driver for just one side and make direct use of quiche::h3::Connection for the other side. IMHO that makes writing the tests a bit less verbose but YMMV.

evanrittenhouse:
> One question about the testing is, whether we should use a H3Driver for both the client and the server in the tests. I'm using an H3Driver for just one side and make direct use of quiche::h3::Connection for the other side. IMHO that makes writing the tests a bit less verbose but YMMV.

The testing additions are great, thanks for adding them! I agree with the current approach as it also helps isolate the two cases.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -980,7 +1063,32 @@ impl<H: DriverHooks> H3Driver<H> {
...
+                        // The channel closed without having written a fin. Send a
+                        // RESET_STREAM to indicate we won't be writing anything
+                        // else
+                        let err = h3::WireErrorCode::RequestCancelled as u64;
+                        let _ = qconn.stream_shutdown(
+                            stream_id,
+                            quiche::Shutdown::Write,
+                            err,
+                        );
+                        ctx.hanle_sent_reset(err);
```
LPardue: nit 
```suggestion
                        ctx.handle_sent_reset(err);
```

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -453,27 +458,50 @@ impl<H: DriverHooks> H3Driver<H> {
...
                 Err(TrySendError::Closed(())) => {
+                    // The channel has closed before we delivered a fin or reset
+                    // to the application.
+                    if !ctx.fin_or_reset_recv &&
+                        ctx.associated_dgram_flow_id.is_none()
+                    // The channel might be closed if the stream was used to
+                    // initiate a datagram exchange.
+                    // TODO: ideally, the application would still shut down the
+                    // stream properly. Once applications code
+                    // is fixed, we can remove this check.
```
jrouviere, line 473: What's the plan for removing that check? Do we have a way to know when applications are fixed?

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -577,15 +625,36 @@ impl<H: DriverHooks> H3Driver<H> {
...
+                    // if we are. If we are not waiting, `handle_recvd_reset()`
+                    // will have taken care of closing.
+                    for pending in self.waiting_streams.iter_mut() {
+                        match pending {
+                            WaitForStream::Upstream(
+                                WaitForUpstreamCapacity {
+                                    stream_id: id,
+                                    chan: Some(chan),
+                                },
+                            ) if stream_id == *id => {
```
jrouviere, line 642: nit: maybe it's me, but I find this match branch very tricky to read. Might be worth splitting in two:

```suggestion
                            WaitForStream::Upstream(cap) => {
                                if stream_id == cap.id {
                                    cap.chan.close()
                                    ...
                                }
                             }
```

tokio-quiche/src/http3/driver/mod.rs
gregor-cf: I think we can fix the one we know of (httx) and the update the code here....

tokio-quiche/src/http3/driver/mod.rs
gregor-cf: `cap.chan` is an Option<>, so I'd also need to match on it to make sure it's some. I played around with a couple of different variants, but didn't feel like it helped readbility too much. (rustfmt is set to 82 chars width which doesn't help for readability in either case). So I'm incline to leave it as-is

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -453,27 +458,50 @@ impl<H: DriverHooks> H3Driver<H> {
...
+                    {
+                        let err = h3::WireErrorCode::RequestCancelled as u64;
+                        let _ = qconn.stream_shutdown(
+                            stream_id,
+                            quiche::Shutdown::Read,
+                            err,
+                        );
+                        drop(try_reserve_result); // needed to drop the borrow on ctx.
+                        ctx.handle_sent_stop_sending(err);
+                        // TODO: should we send an H3Event event to
```
evanrittenhouse, line 483: If we send a STOP_SENDING frame, we could theoretically still have data to write back to the peer. Removing the stream from the httx server map will kill the request handler function and lead to us sending a STOP_SENDING+RESET_STREAM in return, so we'd never actually get to finish sending that data.

That said, I'm having trouble thinking of a case where an H3Body receiver would close without us having seen a fin or the request handler ending, so maybe it's still fine to leave-is? FWIW `ResetStream` and `StreamClosed` are treated identically.

Similar to the below comment, there will probably be more iteration here

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -577,15 +625,36 @@ impl<H: DriverHooks> H3Driver<H> {
...
+                                },
+                            ) if stream_id == *id => {
+                                chan.close();
+                            },
+                            _ => {},
+                        }
+                    }
 
-                self.h3_event_sender
-                    .send(H3Event::ResetStream { stream_id }.into())
```
evanrittenhouse, line 585: I think this may actually be a small issue in the current implementation. If we're acting as a server and receive a RST_STREAM, that just means the client isn't going to write anymore. By sending the ResetStream event upstream we'll kill the request handler and respond with a STOP_SENDING+RESET_STREAM (I think the same issue would apply to the `StreamStatus::Reset` part above, after `recv_body`). This is similar to what Nginx does, but I don't think it's technically correct. I think, in the client-side case, it's fine - receiving a RST_STREAM from the server would mean that the server isn't going to send anything more, so killing the request future is okay. 

cc @LPardue not sure if we want to change this now or as follow-up (or at all). Personally, I think we're good for now, but we should file a followup to ensure this is tracked (we probably are going to want to change how servers handle `ResetStream`s)

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -106,6 +107,57 @@ impl StreamCtx {
             chan: self.recv.take(),
         })
     }
+
+    /// Handle the case when we received a STOP_SENDING frame. Note, that
```
evanrittenhouse, line 111: Can we leave a comment noting that Quiche will automatically respond to these with a RST_STREAM?

tokio-quiche/src/http3/driver/mod.rs
LPardue: Im happy to ship this as is. I suspect there'll be more iterations in this area in future anyway and I don't think blocking this PR any longer will change that

evanrittenhouse (approved):

tokio-quiche/src/http3/driver/mod.rs
evanrittenhouse: Cool, agreed. Approved

tokio-quiche/src/http3/driver/mod.rs
jrouviere: Ha fair enough, missed the extra `Some()`, thanks for taking a look, that sounds good.

tokio-quiche/src/http3/driver/mod.rs
gregor-cf: With the current logic in H3Body it doesn't matter. The only way for H3Body to "close" one of its channels w/o having seen / sent a  fin is if the H3Body is dropped. 
(In "datagram mode" H3Body could drop the channels independently but that's not relevant for streams obviously)
H3Body (and its users) also can't distinguish between seeing a FIN and a RESET on a stream today. Which is something we should probably tackle at some point.

gregor-cf: I think the only use-case of sending an event would be for logging / auditing, since we can only get here by an action initiated by the app (closing / dropping the channel). I'll tweak the TODO a bit.

PR #2167, opened 2025-09-22
tokio-quiche: ignore unrecognized cmsgs

evanrittenhouse:
Router currently processes a variety of cmsgs. It returns an EINVAL if it encounters a cmsg it doesn't care about. This means that we won't create an Incoming or send the packet to the corresponding IOW.

Rather than black-holing packets with cmsgs we don't care about, we should just no-op.

### DIFF
--- a/tokio-quiche/src/quic/router/mod.rs
+++ b/tokio-quiche/src/quic/router/mod.rs
@@ -527,10 +527,9 @@ where
                                     // messages because
                                     // we set IP_PKTINFO on the socket.
                                 },
-                                _ => {
-                                    return Poll::Ready(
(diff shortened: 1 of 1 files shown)


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -528,9 +528,8 @@ where
                                     // we set IP_PKTINFO on the socket.
                                 },
                                 _ => {
-                                    return Poll::Ready(
-                                        Err(Errno::EINVAL.into()),
-                                    );
+                                    // We received an unrecognized cmsg. Don't
+                                    // do anything.
```
rushilmehra: A debug log could be nice

PR #2168, opened 2025-09-23
qlog: update to latest drafts

LPardue:
This updates the qlog crate and all repo dependents (h3i, qlog-dancer, quiche) to use the schemas defined in the following docs

* https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-13
* https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-quic-events-12
* https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-h3-events-12


qlog/src/events/http3.rs
```diff
@@ -236,19 +244,35 @@ pub struct H3FrameCreated {
...
 
     pub raw: Option<RawInfo>,
 }
 
+#[serde_with::skip_serializing_none]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Default)]
+pub struct DatagramCreated {
+    pub quarter_stream_id: u64,
+    pub datagram: Option<Vec<String>>,
+    pub raw: Option<RawInfo>,
```
antoniovicente, line 265: This change to RawInfo instead of len seems to be part of version 13 of the draft, not version 12.

Could you include in the PR description links to the relevant qlog drafts?

quiche/src/h3/frame.rs
```diff
@@ -27,7 +27,7 @@
 use super::Result;
 
 #[cfg(feature = "qlog")]
-use qlog::events::h3::Http3Frame;
+use qlog::events::http3::Http3Frame;
```
antoniovicente, line 30: Having pure renames and file moves in separate commits could help with the review process.

This is a big refactor.  One of my biggest questions right now is the issue of backwards compatibility with older qlogs(or is the only way to render older logs is to use older versions of qlog-dancer) and any existing tools that exist to upgrade old qlogs to the new format so the updated tools can be used to render old logs.

LPardue:
The concerns on backwards compatibility are valid.

Unfortunately years out of date with the state of the current drafts (along with many people). This is BAD for the standards process. It risks qlog dying because of lack of uptake. Which is partly driven by lack of tools that support the new format.

I want to be bold here and break the deadlock. I want to move at speed and not have the overheads of backwards compatibility for maintenance.

Backwards compatibility actually works againt my goals, because it provides an incentive not to upgrade. Its like the damaging effects of Postels law.

All that said, breaking changes like this closer to the RFC publication date  or post RFC need to be taken a lot more seriously. But hopefully the changes etc will be far smaller, the qlog authors dont plan any more huge refactors

qlog/src/events/http3.rs
LPardue: Yeah I can add more descriptions but in this case version 12 is correct 

https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-h3-events-12#name-datagram_created

antoniovicente:
> The concerns on backwards compatibility are valid.
> 
> Unfortunately years out of date with the state of the current drafts (along with many people). This is BAD for the standards process. It risks qlog dying because of lack of uptake. Which is partly driven by lack of tools that support the new format.
> 
> I want to be bold here and break the deadlock. I want to move at speed and not have the overheads of backwards compatibility for maintenance.
> 
> Backwards compatibility actually works againt my goals, because it provides an incentive not to upgrade. Its like the damaging effects of Postels law.
> 
> All that said, breaking changes like this closer to the RFC publication date or post RFC need to be taken a lot more seriously. But hopefully the changes etc will be far smaller, the qlog authors dont plan any more huge refactors

Makes a lot of sense to me.  I'm all in favor of moving forward with this change - backwards compat is a trap that we shouldn't allow to slow us down.

I think the main request is to link to the draft from the PR description.  And update title to refer to version 13 instead of 12 - there seems to be multiple drafts involved with this change; versioning of the docs is not in sync.

qlog/src/events/http3.rs
antoniovicente: I think I was looking at a different document.  I was referring to the `raw` field replacing `len` from https://github.com/quicwg/qlog/pull/495

h3i/src/recordreplay/qlog.rs
```diff
@@ -387,7 +395,7 @@ impl From<&PacketSent> for H3Actions {
                         ..
                     } => actions.push(Action::ResetStream {
                         stream_id: *stream_id,
-                        error_code: *error_code,
+                        error_code: error_code.unwrap_or_default(),
```
antoniovicente, line 403: Consider being explicit about setting error code to NOERROR in this case.

h3i/src/recordreplay/qlog.rs
```diff
@@ -396,7 +404,7 @@ impl From<&PacketSent> for H3Actions {
                         ..
                     } => actions.push(Action::StopSending {
                         stream_id: *stream_id,
-                        error_code: *error_code,
+                        error_code: error_code.unwrap_or_default(),
```
antoniovicente, line 412: Consider being explicit about setting error code to NOERROR in this case.

h3i/src/recordreplay/qlog.rs
LPardue: so I don't think its safe to assume any known code if the log file omitted it. The default for integer is 0, which is not mapped to any HTTP/3 code, which is probably the best approximation of unknown code.

antoniovicente: I was confusing the transport error code NO_ERROR = 0x0 with the H3_NO_ERROR = 0x100

Seems fine to have use 0 as the sentinel when unknown / not set.

PR #2169, opened 2025-09-23
tokio-quiche: limit unknown-cmsg logging

evanrittenhouse:


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -71,6 +71,9 @@ use tokio::sync::mpsc;
 
 type ConnStream<Tx, M> = mpsc::Receiver<io::Result<InitialQuicConnection<Tx, M>>>;
 
+#[cfg(target_os = "linux")]
+const UNKNOWN_CMSG_LOG_LIMIT: u8 = 10;
```
rushilmehra, line 75: 10 seems a bit arbitrary, wdyt about 1 debug log per unknown CMSG observed? In production these logs will be off anyways since they're debug

rushilmehra: That may be overkill so no strong opinions really, but seems a bit more useful given you're more often than not going to have 10 packets setting the same unknown CMSG

evanrittenhouse: Yeah, definitely an arbitrary number. I'm moreso worried about traffic in test sites, which do have debug logging enabled, stressing the pipeline. We could track the distribution of unknown cmsgs but we'll only ever receive 5 of them at the moment, so this felt okay. It's not in production so I'm not overtly worried, just want some basic rate limiting

rushilmehra: Is 10 per connection still too much?

evanrittenhouse: It's at the router level, so you'd effectively log the first 10 unknown cmsgs per listening socket

PR #2171, opened 2025-09-27
tokio-quiche: allow users to read SO_MARK control messages

evanrittenhouse:


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -522,8 +535,10 @@ where
                                 },
                                 ControlMessageOwned::Ipv4PacketInfo(_) |
                                 ControlMessageOwned::Ipv6PacketInfo(_) => {
-                                    // We only want the destination address from
-                                    // IP_RECVORIGDSTADDR, but we'll get these
+                                    // We only want the destination address
+                                    // from
+                                    // IP_RECVORIGDSTADDR, but we'll get
+                                    // these
```
toidiu: fmt comment. Also above

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -413,8 +415,6 @@ where
                 return self.poll_recv_from(cx);
             };
 
-            self.reusable_cmsg_space.clear();
```
toidiu, line 416: What was it doing and why dont we need it now?

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -452,7 +452,18 @@ where
...
-                        for cmsg in r.cmsgs() {
+                        let Ok(cmsgs) = r.cmsgs() else {
+                            // Best-effort if we can't read cmsgs.
+                            return Poll::Ready(Ok(PollRecvData {
+                                bytes,
+                                src_addr: peer_addr,
+                                dst_addr_override,
+                                rx_time,
+                                gro,
+                            }));
```
toidiu, line 475: > Best-effort

As in we attempt to read the cms and if we can't then we just skip reading? Is this the correct.. as in should we fail here instead rather than silently passing?

Can you help me understand what functionality will be missing if we return here? We might want to keep the behavior but document it better.

tokio-quiche/src/quic/router/mod.rs
evanrittenhouse: The previous version of nix built the vector based on the capacity of the passed buffer, while the new version does it based on length. clear() sets the length to 0, so using that ends up throwing ENOBUFS when we call cmsgs(). That said I did have a fill(0) here which must've gotten removed in a rebase, I'll put that back 

E: actually buffer-pool may work well here too, though I'm not quite sure having _any_ sort of clear/fill matters - current `master` only clears the vector on the first `recvmsg` call, since it's outside the loop.

tokio-quiche/src/quic/router/mod.rs
evanrittenhouse: I was in two minds about this - cmsgs() only errors with ENOBUFS, which should mean the only failure mode is if reusable_cmsg_space is too short (and thus library misconfiguration). Part of me just wants to panic in that case, though I'm not sure if misconfiguration is the only trigger which is why I passed through

To answer your question though, if we return an error here we would skip processing the packet

datagram-socket/src/socket_stats.rs
```diff
@@ -242,6 +250,18 @@ impl QuicAuditStats {
     pub fn max_loss_pct(&self) -> u8 {
         self.max_loss_pct.load(Ordering::Acquire)
     }
+
+    #[inline]
+    #[cfg(target_os = "linux")]
+    pub fn set_initial_so_mark_data(&self, value: Option<Vec<u8>>) {
+        *self.initial_so_mark.write().unwrap() = value;
```
LPardue: Since this is supposed to be the initial, should this check that the value was not already set before allowing the write?

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -522,11 +547,25 @@ where
...
                                     // messages because
                                     // we set IP_PKTINFO on the socket.
                                 },
+                                ControlMessageOwned::Unknown(raw_cmsg) => {
+                                    let UnknownCmsg {
+                                        cmsg_header,
+                                        data_bytes,
+                                    } = raw_cmsg;
+
+                                    if cmsg_header.cmsg_level == SOL_SOCKET &&
```
LPardue, line 559: Don't you need some `#[cfg(target_os = "linux")]` protection here?

LPardue: Also, consider putting unknown cmsg handling in a function for code clarity, since I guess we might need more of these in future

datagram-socket/src/socket_stats.rs
evanrittenhouse: Good catch 😅  this should be a `OnceLock`

E: changed

tokio-quiche/src/quic/router/mod.rs
evanrittenhouse: > Don't you need some #[cfg(target_os = "linux")] protection here?

The whole block should already be Linux-protected (see L411)

> Also, consider putting unknown cmsg handling in a function for code clarity, since I guess we might need more of these in future

I think it could be a bit annoying for now because of the scoping issues - are we okay leaving as-is for now and then adding a helper function if/when another cmsg type is required?

tokio-quiche/src/socket/capabilities.rs
```diff
@@ -268,8 +295,15 @@ impl<'s> SocketCapabilitiesBuilder<'s> {
     /// will only check their status. **If neither of them is enabled, the
     /// `PKTINFO` sockopts will cause errors when sending packets.**
     pub fn allows_nonlocal_source(&self) -> io::Result<bool> {
-        Ok(getsockopt(self.socket.as_raw_fd(), IpFreebind)? ||
-            getsockopt(self.socket.as_raw_fd(), IpTransparent)?)
+        Ok(getsockopt(&self.socket.as_fd(), IpFreebind)? ||
+            getsockopt(&self.socket.as_fd(), IpTransparent)?)
+    }
+
+    pub fn mark(&mut self) -> io::Result<()> {
```
LPardue: nit: should this be called recv_mark instead of just mark?

LPardue (dismissed):
LGTM modulo naming nit

tokio-quiche/src/socket/capabilities.rs
evanrittenhouse: Changed

tokio-quiche/src/quic/connection/mod.rs
```diff
@@ -196,7 +196,16 @@ pub struct Incoming {
     pub buf: PooledBuf,
     /// If set, then `buf` is a GRO buffer containing multiple packets.
     /// Each individual packet has a size of `gso` (except for the last one).
-    pub gro: Option<u16>,
+    pub gro: Option<i32>,
+    /// [SO_MARK] control message value received from the socket.
+    ///
+    /// This will only be populated for the first packet on a given connection.
+    /// This will always be `None` after the connection has been spawned as
+    /// the message is `take()`d before spawning.
```
toidiu: > This will only be populated for the first packet on a given connection.

How important is it for only the first packet to be marked? What happens if we accidentally remove the `take`? Will things break or just keep working?

Do we need a test for this case. The current test checks that we have this stat on the first packet.

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -524,8 +547,20 @@ where
...
+                                        cmsg_header.cmsg_type == SO_MARK
+                                    {
+                                        let _ = mark_bytes.insert(data_bytes);
+                                    }
```
toidiu: Do we need to validate the size of `data_bytes`. Since this grows the `reusable_cmsg_space` on a socket, and since the socket is reused, this might be important.

If we decide to restrict size, does it make sense to make this a static slice?

tokio-quiche/src/socket/capabilities.rs
```diff
@@ -116,6 +117,32 @@ impl SetSockOpt for Ipv6MtuDiscoverProbe {
...
+            )
+        };
+
+        match ret {
+            0 => Ok(()),
+            _ => Err(Errno::last()),
+        }
+    }
+}
+
```
toidiu, line 148: nit:

- Can you add documentation to this and link the Linux man page (couldnt actually find the linux docs on this). e.g. why setting value to `1` is appropriate. 
- Can we pull out `1` as a named const

tokio-quiche/src/socket/capabilities.rs
```diff
@@ -353,6 +392,7 @@ impl SocketCapabilities {
         #[cfg(feature = "perf-quic-listener-metrics")]
         let _ = b.rxtime();
         let _ = b.gro();
+        let _ = b.rcvmark();
```
toidiu, line 398: do we need to handle the error if it doesn't get applied? Maybe metrics in the future?

evanrittenhouse: I don't think we do, none of the others are handled

tokio-quiche/src/quic/connection/mod.rs
evanrittenhouse: Removed that sentence, relic from a previous implementation

tokio-quiche/src/quic/router/mod.rs
toidiu: spoke offline. The SO_MARK cmsg will be marked by the kernel and is expected to be 32 bits.

evanrittenhouse: I can change it to a `[u8; 4]` for ergonomics in userspace, since we know that `SO_MARK` is a u32

toidiu (approved):

tokio-quiche/src/socket/capabilities.rs
evanrittenhouse: Ended up linking to the source code that sets this

toidiu (approved):

toidiu:
Looks great!

PR #2172, opened 2025-09-27
feat: disable dynamically changing the loss thresholds on spurious loss

toidiu:
By default quiche increases the packet and time loss thresholds on spurious loss. This change disables the dynamic behavior to allow for running experiments

### DIFF
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -370,6 +370,11 @@ struct LossThreshold {
     // INITIAL_TIME_THRESHOLD_OVERHEAD` and equivalent to the initial value
     // of INITIAL_TIME_THRESHOLD.
     time_thresh_overhead: Option<f64>,
+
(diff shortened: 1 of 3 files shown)


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -396,6 +403,10 @@ impl LossThreshold {
     }
 
     fn on_spurious_loss(&mut self, new_pkt_thresh: u64) {
+        if self.disable_dynamic_loss_threshold {
+            return;
+        }
+
```
antoniovicente, line 409: Would be good to have a second option to just disable this line:
`self.time_thresh = PACKET_REORDER_TIME_THRESHOLD;`

We'ld like to measure the benefits of this PR: https://github.com/cloudflare/quiche/pull/2145/files#

toidiu: Yea good point.

PR #2178, opened 2025-10-02
Defer log formatting until it's actually used

gregor-cf:
slog doesn't guard log statements by log-level. Instead it relies on arguments to its logging macros to defer work until needed.

### DIFF
--- a/tokio-quiche/src/quic/router/mod.rs
+++ b/tokio-quiche/src/quic/router/mod.rs
@@ -695,7 +695,7 @@ where
                     buf.truncate(bytes);
 
                     let send_from = if let Some(dst_addr) = dst_addr_override {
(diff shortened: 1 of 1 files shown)


LPardue (approved):

evanrittenhouse (dismissed):

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -694,8 +694,9 @@ where
                     );
                     buf.truncate(bytes);
 
+                    log::error!("FOO FOO "; "addr" => server_addr);
```
LPardue: FOO FOO 🤣

gregor-cf: and of course after I removed that leftover log, I pushed it to the wrong branch / PR. Probably the most revisions I've ever done on a oneline change ¯\_(ツ)_/¯

ghedo (approved):

PR #2181, opened 2025-10-06
Use longer packet number length encoding when necessary

### LINKED ISSUE #2180: Incorrect packet number encoding
Originally filed under https://github.com/netty/netty-incubator-codec-quic/issues/827

The encoding used by a packet number is sometimes too short to ensure correct decoding.

On a machine talking http3 to a Quiche implementation we observed long sequences of packets dropped because of tag mismatch. Example situation:

- the largest processed (and ACKed) PN observed on the client side is 0x11eaa
- the server sends a packet with 1-byte packet number 2e
- the packet number is decoded to 0x11e2e
- packet decoding fails, because the packet number was supposed to be 0x11f2e

The situation repeats for each packet number until the packet number reaches a8 (0x11ea8), after which the server starts sending 2-byte packet numbers and the transfer continues.

[RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-sample-packet-number-encodi) dictates that the sender should start using 2-byte packet number encoding as soon as the packet number is more than 127 numbers away from the last number acked by peer.

The problem is caused by an off-by-one error in pkt_num_len:
https://github.com/cloudflare/quiche/blob/ad489e3e173443b3623e4a23401a421341512892/quiche/src/packet.rs#L571-L572

Compare https://www.rfc-editor.org/rfc/rfc9000.html#name-sample-packet-number-encodi
> min_bits = log(num_unacked, 2) + 1

(the RFC could have been more clear about the rounding mode of log)

To confirm the problem, use the following piece of code:
```
		// roundtrip
		let base = 0xdeadbeef;
		for i in 1..255 {
			let pn = base + i;
			let num_len = pkt_num_len(pn, base);
			if num_len == 1 {
				let decoded = decode_pkt_num(base, pn & 0xff, num_len);
    			assert_eq!(decoded, pn);
			} else {
				assert_eq!(num_len, 2);
				let decoded = decode_pkt_num(base, pn & 0xffff, num_len);
    			assert_eq!(decoded, pn);
			}
		}
```

With the current implementation of pkt_num_len, the above test fails.

djelinski:
The decoding method described in RFC 9000 treats the encoded packet number as a signed integer. We need to encode the sign bit to ensure correct decoding when 128+ packets are lost.

This PR fixes the packet number length computation, and adds a test to verify that the returned packet length is sufficient to decode the next packet number assuming a loss of 1..254 packets.

Fixes #2180


ghedo:
As far as I understand, this will mean we never use 1-byte encoding? It's not necessarily a huge issue, but I also don't really think it's the correct fix...

djelinski:
Thanks @ghedo for the review. We still use 1-byte encoding for packet numbers between base+1 and base+126; the change makes the encoding one bit longer, not one byte longer.

quiche/src/packet.rs
```diff
@@ -568,8 +568,8 @@ impl std::fmt::Debug for Header<'_> {
 
 pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
     let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
-    // computes ceil of num_unacked.log2()
-    let min_bits = u64::BITS - num_unacked.leading_zeros();
+    // computes ceil of num_unacked.log2() + 1
+    let min_bits = u64::BITS - num_unacked.leading_zeros() + 1;
```
antoniovicente, line 572: Would be good to have a test that exercises the pkt_num_len computation for known cases.

Seems like 127 is the first index that requires a 2 byte representation?  This seemed unexpected to me.  I thought it was 128?

quiche/src/packet.rs
```diff
@@ -1380,6 +1380,20 @@ mod tests {
         let hdr_num = u64::from(b.get_u24().unwrap());
         let pn = decode_pkt_num(0xace9fa, hdr_num, num_len);
         assert_eq!(pn, 0xace9fe);
+        // roundtrip
+        let base = 0xdeadbeef;
```
antoniovicente, line 1384: Given the +1 in decode_pkt_num we should consider bases that end in 0xff in addition to other bases.

But I was not able to trigger errors when I changed the last byte to other values.  Need to look into this more closely.

quiche/src/packet.rs
```diff
@@ -568,8 +568,8 @@ impl std::fmt::Debug for Header<'_> {
 
 pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
     let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
```
antoniovicente, line 570: I think we're using 2 byte rep for 127 also because of the +1 in this line.

quiche/src/packet.rs
djelinski: Thanks @antoniovicente for the review.

> Would be good to have a test that exercises the pkt_num_len computation for known cases.

The cases from RFC 9000 section A.2 are already tested, see https://github.com/cloudflare/quiche/blob/104d3c9d06956841f89a8c0d28d0e9d2b401ee85/quiche/src/packet.rs#L1366 and https://github.com/cloudflare/quiche/blob/104d3c9d06956841f89a8c0d28d0e9d2b401ee85/quiche/src/packet.rs#L1375. The latter contains a copy/paste bug (`0xace9fe` should have been `0xace8fe`), but the test looks OK otherwise. What other cases would you like to test?

Note that there are cases where we might want to use longer encoding than necessary, for example when probing a path with a much longer RTT than the current one.

> Seems like 127 is the first index that requires a 2 byte representation? This seemed unexpected to me. I thought it was 128?

See the response to the other comment.

quiche/src/packet.rs
djelinski: That's correct. Should I fix this here?

quiche/src/packet.rs
djelinski: The exact bits of base don't matter much; the decoding algorithm only takes the distance between the base and the decoded PN into account.

quiche/src/packet.rs
antoniovicente: Let's try to get this fix in as is, and fix that in a followup.

Seems like you need a rebase; could you click on the "update branch" button?

quiche/src/packet.rs
antoniovicente: I assume that the value of the last byte would factor into the distance calculation, but I may be wrong.

quiche/src/packet.rs
djelinski: OK. Branch updated now.

quiche/src/packet.rs
djelinski: The last byte doesn't play any special role in the distance calculation.

The distance is the difference `pn - base`; based on that, the number of bits needed to encode the packet number is calculated, Based on the necessary number of bits, the number of bytes is selected, and then as many bytes of `pn` as needed are taken as the encoded packet number.

I'm using a nonzero base here because the decoding algorithm never produces a negative packet number, even if it's less distant than the following positive number. For example, `FF` decodes to `FF` in base `0`, even though `-1` is the less distant alternative. Zero base would not demonstrate the correctness of the fix.

PR #2184, opened 2025-10-06
docs(flowcontrol): correct window increase factor value

mxinden:
`WINDOW_INCREASE_FACTOR` is `2`, not `1.5`.


LPardue:
Just noticed too, thanks !

LPardue (approved):

PR #2185, opened 2025-10-08
Parse Version Information Transport Parameter

hawkinsw:
Does not (yet!) incorporate parsed Version Information Transport Parameter into version negotiotiation.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -8867,6 +8867,19 @@ impl<'a> Iterator for UnknownTransportParameterIterator<'a> {
     }
 }
 
+/// QUIC Version Information Transport Parameter
+#[derive(Clone, Debug, PartialEq)]
+pub struct VersionNegotiationParameter {
+    /// Version of the QUIC protocol chosen by the sender for this connection.
+    pub chosen_version: u32,
+    /// Semantics depends upon identity of sender (client or server)
+    ///
+    /// If the sender is the client, available_versions lists the QUIC versions
+    /// compatible with the ongoing first flight. If the sender is the server,
+    /// available_versions lists all its fully deployed versions.
+    pub available_versions: Option<Vec<u32>>,
+}
+
 /// QUIC Transport Parameters
 #[derive(Clone, Debug, PartialEq)]
 pub struct TransportParams {
@@ -8906,6 +8919,8 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
+    /// Information for version negotiation, if any.
+    pub version_information: Option<VersionNegotiationParameter>,
     /// Unknown peer transport parameters and values, if any.
     pub unknown_params: Option<UnknownTransportParameters>,
     // pub preferred_address: ...,
@@ -8931,6 +8946,7 @@ impl Default for TransportParams {
             initial_source_connection_id: None,
             retry_source_connection_id: None,
             max_datagram_frame_size: None,
+            version_information: None,
             unknown_params: Default::default(),
         }
     }
@@ -9089,6 +9105,29 @@ impl TransportParams {
 
                 0x0020 => {
                     tp.max_datagram_frame_size = Some(val.get_varint()?);
+                },
+
+                0x0011 => {
+                    let mut version_information = VersionNegotiationParameter {
+                        chosen_version: val.get_u32()?,
+                        available_versions: None,
+                    };
+
+                    // It is semantically invalid to have an empty list of
+                    // available versions when this transport
+                    // parameter is sent by a client (RFC9538, Sec. 3, para. 6),
+                    // but there is no such restriction when
+                    // it is sent by a server. Therefore, this code
+                    // allows for an empty list and places the error-checking onus
+                    // on users.
+                    while val.off() < val.len() {
+                        let version = val.get_u32()?;
+                        version_information
+                            .available_versions
+                            .get_or_insert_default()
+                            .push(version);
+                    }
+                    tp.version_information = Some(version_information);
                 },
 
                 // Track unknown transport parameters specially.
--- a/quiche/src/tests.rs
+++ b/quiche/src/tests.rs
@@ -51,6 +51,7 @@ fn transport_params() {
         initial_source_connection_id: Some(b"woot woot".to_vec().into()),
         retry_source_connection_id: Some(b"retry".to_vec().into()),
         max_datagram_frame_size: Some(32),
+        version_information: None,
         unknown_params: Default::default(),
     };
 
@@ -81,6 +82,7 @@ fn transport_params() {
         initial_source_connection_id: Some(b"woot woot".to_vec().into()),
         retry_source_connection_id: None,
         max_datagram_frame_size: Some(32),
+        version_information: None,
         unknown_params: Default::default(),
     };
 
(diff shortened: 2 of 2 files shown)


hawkinsw:
@LPardue and @ghedo -- as usual, thank you both for this amazing implementation! I have added support for encoding/parsing the Version Information Transport Parameter. I realize there are additional changes that could be incorporated into this PR (namely, using the Version Information in the process of version negotiation) and I would be _more_ than happy to talk that. But, I didn't want to get too far ahead in case this PR was unwelcome or you already had something planned. 

If you think that what is here so far is reasonable and would like me to press on, just give me the nod! I would love to help however I can!

quiche/src/lib.rs
```diff
@@ -9094,6 +9106,23 @@ impl TransportParams {
...
+
+                    let mut available_versions: Option<Vec<u32>> = None;
+                    while val.off() < val.len() {
+                        let version = val.get_u32()?;
+                        if let Some(available_versions) = &mut available_versions
+                        {
+                            available_versions.push(version);
+                        } else {
+                            available_versions = Some(vec![version]);
+                        }
```
LPardue: nit: more idiomatic way to do this is https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_default i.e.

```
available_verseions.get_or_insert_default().push(version);
```

LPardue:
Thanks for the PR Will. I think its probably ok to parse the received TP for debug/analysis reasons. However, adding full version negotiation support is a bit more work and something that has no been a high priority for us. 

It would probably be best to keep this PR focuses just on parsing received TPs and not provide an API to allow apps to set them (and hence no encoding). That would let us land the PR soonish. Then follow up work could consider a more-complete version negotiation approach, which will need lots of testing etc

hawkinsw:
> Thanks for the PR Will. I think its probably ok to parse the received TP for debug/analysis reasons. However, adding full version negotiation support is a bit more work and something that has no been a high priority for us.
> 
> It would probably be best to keep this PR focuses just on parsing received TPs and not provide an API to allow apps to set them (and hence no encoding). That would let us land the PR soonish. Then follow up work could consider a more-complete version negotiation approach, which will need lots of testing etc

Funny that is your suggestion, because that is my preferred plan, too!! As you know, my use case is _parsing_ TPs, so this will do _exactly_ what I want. I will modify the PR (according to this plan and the helpful suggestion above) and resubmit!

Thanks, as usual!

quiche/src/lib.rs
hawkinsw: Thank you for noting that! I agree. I saw this method for the first time on Friday -- not sure why I have never needed/seen it before!

hawkinsw:
Let me know if you think that this version is better!

hawkinsw:
I don't want to overstep, but if you would like me to add to the PR an update to the minimum supported Rust version (to solve this failure: [45b6494902ce0dc40fe8543f6d42f36385cd0e1e](https://github.com/cloudflare/quiche/actions/runs/18479878515/job/52652312302?pr=2185#step:10:417)), I would be more than happy to do that!

LPardue:
> I don't want to overstep, but if you would like me to add to the PR an update to the minimum supported Rust version (to solve this failure: [45b6494902ce0dc40fe8543f6d42f36385cd0e1e](https://github.com/cloudflare/quiche/actions/runs/18479878515/job/52652312302?pr=2185#step:10:417)), I would be more than happy to do that!

Yeah a separate PR to bump th MSRV to 1.83.0 would be welcomed!

hawkinsw:
> > I don't want to overstep, but if you would like me to add to the PR an update to the minimum supported Rust version (to solve this failure: [45b6494902ce0dc40fe8543f6d42f36385cd0e1e](https://github.com/cloudflare/quiche/actions/runs/18479878515/job/52652312302?pr=2185#step:10:417)), I would be more than happy to do that!
> 
> Yeah a separate PR to bump th MSRV to 1.83.0 would be welcomed!

Let me know if https://github.com/cloudflare/quiche/pull/2197 helps!

quiche/src/lib.rs
```diff
@@ -9279,6 +9294,20 @@ impl TransportParams {
             b.put_varint(max_datagram_frame_size)?;
         }
 
+        if let Some((left, right)) = &tp.version_information {
```
LPardue: I think we can drop this too now that the PR is focused on parsing only?

hawkinsw: Sorry I completely missed this when converting the PR to handle parsing only. In the most recent version of the PR I removed the encoding functionality and added a test. Thank you for the feedback!

quiche/src/lib.rs
```diff
@@ -8909,6 +8909,8 @@ pub struct TransportParams {
     pub retry_source_connection_id: Option<ConnectionId<'static>>,
     /// DATAGRAM frame extension parameter, if any.
     pub max_datagram_frame_size: Option<u64>,
+    /// Information for version negotiation, if any.
+    pub version_information: Option<(u32, Option<Vec<u32>>)>,
```
LPardue: nit: the structure is a bit complex. I'd suggest defining a new struct like

```
struct VersionNegotiationParameter {
  chosen_version: u32,
  available_versions: Option<Vec<u32>>,
}
```

quiche/src/lib.rs
```diff
@@ -9094,6 +9097,18 @@ impl TransportParams {
                     tp.max_datagram_frame_size = Some(val.get_varint()?);
                 },
 
+                0x0011 => {
+                    let chosen_version = val.get_u32()?;
```
LPardue: If we use the structure suggested above, you can then do something like

```
VersionNegotiationParameter {
  chosen_version: val.get_u32()?,
  available_versions: None,
}
```

and just populate available_versions below directly on the struct

hawkinsw: Great idea(s)! Will implement ASAP! I am _very_ sorry for the delay!

hawkinsw: I think that I got it done!

LPardue (approved):

quiche/src/tests.rs
```diff
@@ -94,6 +96,38 @@ fn transport_params() {
     assert_eq!(new_tp, tp);
 }
 
+#[test]
+fn transport_params_decode_version_information() {
+    // Given an encoded version_information transport parameter, ...
+    let raw_version_information_params = [
+        0x11, 0x10, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0xfa, 0x5a,
+        0x0a, 0x7a, 0xff, 0x00, 0x00, 0x1d,
+    ];
```
antoniovicente, line 105: Would be good to have some tests for the error cases where this input is truncated in the middle of the chosen version or one of the other versions.

antoniovicente (approved):
Thanks for the change!

quiche/src/tests.rs
hawkinsw: I agree and will do!!

hawkinsw: Sorry it took me so long! Let me know if that additional tests (and added comment in the decoding implementation) are acceptable!

hawkinsw:
Hope that the additional tests are helpful!

hawkinsw:
Just thought I would pop my head up here again and see whether there is anything else that we need to take care of on this PR! I hope everyone is having a nice holiday!!

hawkinsw:
> Just thought I would pop my head up here again and see whether there is anything else that we need to take care of on this PR! I hope everyone is having a nice holiday!!

Hope everyone is having a good beginning to 2026!

PR #2187, opened 2025-10-10
h3: move Huffman encoding / decoding to octets crate

ghedo:
Huffman encoding is used by HPACK as well, so it makes sense to share the implementation rather than duplicate all this code.

This also somewhat simplifies the decoding code which was split across multiple functions and structures for not much benefit (if anything it made the code harder to follow).


octets/src/lib.rs
```diff
@@ -248,6 +251,49 @@ impl<'a> Octets<'a> {
         self.get_bytes(len as usize)
     }
 
+    /// Decodes a Huffman-encoded value from the current offset.
+    pub fn get_huffman_bytes(&mut self) -> Result<Vec<u8>> {
```
antoniovicente: Would it be possible to include "decode" in the function name to indicate that this is decoding?

Suggestion: `pub fn get_huffman_decoded(&mut self) -> Result<Vec<u8>> {`

quiche/src/h3/qpack/encoder.rs
```diff
@@ -152,22 +152,20 @@ pub fn encode_str<const LOWER_CASE: bool>(
 ) -> Result<()> {
     // Huffman-encoding generally saves space but in some cases it doesn't, for
     // those just encode the literal string.
-    match super::huffman::encode_output_length::<LOWER_CASE>(v) {
+    match octets::huffman_encoding_len::<LOWER_CASE>(v) {
         Ok(len) => {
             encode_int(len as u64, first | (1 << prefix), prefix, b)?;
-            super::huffman::encode::<LOWER_CASE>(v, b)?;
+            b.put_huffman_bytes::<LOWER_CASE>(v)?;
```
antoniovicente: Possible followup: It seems to me that computing the len is as expensive as doing the encoding.  Should we consider just doing the encoded version and rewind and add the unencoded version if it generated something longer than the original?

octets/src/lib.rs
```diff
@@ -248,6 +251,49 @@ impl<'a> Octets<'a> {
...
+        const FLAG_ERR: u8 = 4;
+
+        // Max compression ratio is >= 0.5.
+        let mut out = Vec::with_capacity(self.cap() << 1);
+
+        let mut state = 0;
+        let mut eos = false;
+
+        while let Ok(byte) = self.get_u8() {
+            let (next, sym, flags) = DECODE_TABLE[state][(byte >> 4) as usize];
```
antoniovicente: Would this be simpler and less redundant?

```
for data in [byte >> 4, byte & 0xf] {
  let (next, sym, flags) = DECODE_TABLE[state][(data) as usize];
  if flags & FLAG_ERR == FLAG_ERR {
    // Data followed the "end" marker.
    return Err(BufferTooShortError);
  } else if flags & FLAG_SYM == FLAG_SYM {
    out.push(sym);
  }

  state = next;
  // eos only correct when handling the byte & 0xf case; ignored and overwritten in the byte >> 4 case.
  eos = flags & FLAG_END == FLAG_END;
}
```

quiche/src/h3/qpack/encoder.rs
LPardue: The current design was done as a quick fix for a real bug where the output buffer was too small for the huffman encoded version. I don't know if it would be more performant to do it as suggested but I'd probably want to see some benchmarks. Part of the issue is that you need to write the length first into the output buffer. An improvement might be to support more of a "write the length after doing the data buffering" approach.

LPardue:
Seems rather strange to put it in octets. We have various users of that crate that will never care about huffman encoding. Consider instead making this a new crate.

antoniovicente:
> Seems rather strange to put it in octets. We have various users of that crate that will never care about huffman encoding. Consider instead making this a new crate.

Being able to interact with huffman encoded bytes by simply calling member functions of `Octets` seems quite nice.

LPardue:
Needs to be an optional feature at least. There are many protocols that use QUIC's varint encoding now but lots of them don't need access to the static huffman stuff

octets/src/lib.rs
ghedo: Done. Also rename `put_huffman_bytes()` -> `put_huffman_encoded()`.

quiche/src/h3/qpack/encoder.rs
ghedo: Yeah. Also, I think the main issue is not so much CPU, but memory allocation since the encoding code allocates memory as it goes (once it exceed the pre-allocated size), so I'd be worried about potentially allocating too much memory which would need additional tests.

Not against looking into it as a follow-up, but not something that is high priority.

octets/src/lib.rs
ghedo: Yeah, good idea. Done.

ghedo:
I gated this behind a feature now as well.

octets/Cargo.toml
```diff
@@ -9,6 +9,10 @@ license = { workspace = true }
 keywords = { workspace = true }
 categories = { workspace = true }
 
+[features]
+# Build the Huffman encoding support as defined by HPACK (RFC7541).
+huffman = []
```
LPardue: nit: maybe call it `hpack_huffman` in case we ever wanted to add some sliding huffman or something

ghedo: Done.

quiche/src/h3/qpack/encoder.rs
```diff
@@ -152,22 +152,20 @@ pub fn encode_str<const LOWER_CASE: bool>(
...
     // those just encode the literal string.
-    match super::huffman::encode_output_length::<LOWER_CASE>(v) {
+    match octets::huffman_encoding_len::<LOWER_CASE>(v) {
         Ok(len) => {
             encode_int(len as u64, first | (1 << prefix), prefix, b)?;
-            super::huffman::encode::<LOWER_CASE>(v, b)?;
+            b.put_huffman_encoded::<LOWER_CASE>(v)?;
         },
 
-        Err(super::Error::InflatedHuffmanEncoding) => {
```
LPardue, line 161: nit: you can probably remove `InflatedHuffmanEncoding` from the qpack module too

ghedo: Ah yeah, good point. Done.

PR #2188, opened 2025-10-14
task-killswitch: Fix ordering edge case and improve worker throughput

TheJokr:
### task-killswitch: Fix remove-before-add edge case

The `RemoveOnDrop` guard can execute before we send the `Add` operation
on the killswitch channel. This causes the JoinHandle from the `Add`
operation to be kept around forever, leaking memory.

To fix this, we introduce a new `TaskEntry` enum with a tombstone
variant. If we receive a `Remove` operation for an ID before the
corresponding `Add` operation, we instead insert the tombstone into our
task map. Once we receive the `Add` operation, we check for the
tombstone's presence and discard the handle if it exists.

### task-killswitch: Use AbortHandle to reduce drop cost

JoinHandle is expensive to drop as it allows awaiting a task and
accessing its output. By keeping it around in the task map, we
centralize this drop cost in a single thread with cold memory.

An AbortHandle instead only keeps around a cheap(er) reference to the
task. The JoinHandle is dropped almost immediately after the task is
spawned, while it is still fresh in memory/caches. Tokio can take
advantage of the dropped JoinHandle by avoiding unncessary work.

(Alternatively, this would also allow us to return the JoinHandle to the
caller.)

### DIFF
--- a/task-killswitch/src/lib.rs
+++ b/task-killswitch/src/lib.rs
@@ -26,7 +26,7 @@
 
 use tokio::sync::mpsc;
 use tokio::sync::watch;
-use tokio::task::JoinHandle;
+use tokio::task::AbortHandle;
 
 use std::collections::HashMap;
 use std::future::Future;
@@ -35,7 +35,7 @@ use std::sync::atomic::Ordering;
 use std::sync::LazyLock;
 
 enum ActiveTaskOp {
-    Add { id: u64, handle: JoinHandle<()> },
+    Add { id: u64, handle: AbortHandle },
     Remove { id: u64 },
 }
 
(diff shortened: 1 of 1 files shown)


task-killswitch/src/lib.rs
```diff
@@ -138,19 +147,48 @@ impl ActiveTasks {
...
+                // We assert against duplicate IDs in `spawn_task`. Panicing here
+                // wouldn't do anything besides stopping the killswitch loop, so
+                // just overwrite the handle.
+                e.insert_entry(TaskEntry::Handle(handle));
+            },
+        }
+    }
+
+    fn remove_task(&mut self, id: u64) {
+        use std::collections::hash_map::Entry::*;
```
ghedo, line 183: Might as well just import this at module level? So we don't need to import twice.

TheJokr: These are enum variants. I generally dislike importing those at file scope since their use is very localized to the `match` statements. At file scope I'd import `Entry`, but that name is so generic that it also gets confusing.

PR #2189, opened 2025-10-14
tokio-quiche: store latest PRIORITY_UPDATE value on incoming request

LPardue:
Priority signals take the form of HTTP headers and/or frames. Some clients only send frames, while some clients can both. At the point in time a server reads request HEADERS, a client may sent a PRIORITY_UPDATE frame that should take precedence.

With this change, we read take the last PRIORITY_UPDATE processed by quiche, and attach it to the IncomingH3Headers so that Tokio-Quiche-based apps can use it if they wish.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -46,6 +46,7 @@ use foundations::telemetry::log;
 use futures::FutureExt;
 use futures_util::stream::FuturesUnordered;
 use quiche::h3;
+use quiche::h3::Priority;
 use tokio::select;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::error::TryRecvError;
@@ -190,6 +191,9 @@ pub struct IncomingH3Headers {
     pub read_fin: bool,
     /// Handle to the [`H3AuditStats`] for the message's stream.
     pub h3_audit_stats: Arc<H3AuditStats>,
+    /// The latest PRIORITY_UPDATE frame value, if any. Always `None` for
+    /// clients.
+    pub latest_priority_update: Option<Priority>,
 }
 
 impl fmt::Debug for IncomingH3Headers {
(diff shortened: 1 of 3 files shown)


tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -190,6 +191,8 @@ pub struct IncomingH3Headers {
     pub read_fin: bool,
     /// Handle to the [`H3AuditStats`] for the message's stream.
     pub h3_audit_stats: Arc<H3AuditStats>,
+    /// The latest PRIORITY_UPDATE frame value, if any.
```
ghedo: Maybe worth mentioning this is a server-only field.

LPardue: Good point, done.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -190,6 +191,9 @@ pub struct IncomingH3Headers {
     pub read_fin: bool,
     /// Handle to the [`H3AuditStats`] for the message's stream.
     pub h3_audit_stats: Arc<H3AuditStats>,
+    /// The latest PRIORITY_UPDATE frame value, if any. Always `None` for
```
evanrittenhouse, line 194: One thing we could do here is extract `ServerH3Event::Core(H3Event::IncomingHeaders { .. })` into `ServerH3Event::Headers { incoming_headers, priority }`. We'd still grab the priority off of the driver, but that would prevent us from leaking this priority into the client-side API.

Alternatively we could consider adding a new `ServerH3Event` variant called `Priority`, then send that to the listening `Server` just before sending the `Headers`. Doing so would ensure that the listening server application doesn't have any race conditions while processing priorities vs. headers, but that seems more complex

LPardue: Thanks for the input. I like the idea of not leaking things to a client that are never set. Given the benefits of that API, it sounds like we should create a PR that does that just for comparison before merging this.

Ideally we could surface all PRIORITY_UPDATE events at any point in a request lifecycle but like you say it adds a lot of complexity and is not quite required right now. Perhaps we can add another event in future, since the concept layering kind of allows that

LPardue:
Closing this in favor of https://github.com/cloudflare/quiche/pull/2195

PR #2190, opened 2025-10-14
task-killswitch: operation queue metrics

TheJokr:
### task-killswitch: Add optional metrics to monitor operation queue

The `metrics` feature enables metrics collection via foundations.

### DIFF
--- a/task-killswitch/Cargo.toml
+++ b/task-killswitch/Cargo.toml
@@ -8,7 +8,11 @@ keywords = { workspace = true }
 categories = { workspace = true }
 description = "Abort all tokio tasks at once"
 
+[features]
+metrics = ["dep:foundations"]
+
 [dependencies]
+foundations = { workspace = true, optional = true, features = ["metrics"] }
 parking_lot = { workspace = true }
 tokio = { workspace = true, features = ["rt", "sync"] }
 
--- a/task-killswitch/src/lib.rs
+++ b/task-killswitch/src/lib.rs
@@ -34,6 +34,9 @@ use std::sync::atomic::AtomicU64;
 use std::sync::atomic::Ordering;
 use std::sync::LazyLock;
 
+#[cfg(feature = "metrics")]
+mod metrics;
+
 enum ActiveTaskOp {
     Add { id: u64, handle: AbortHandle },
     Remove { id: u64 },
@@ -49,6 +52,8 @@ impl Drop for RemoveOnDrop {
     fn drop(&mut self) {
         if let Some(tx) = self.task_tx_weak.upgrade() {
             let _ = tx.send(ActiveTaskOp::Remove { id: self.id });
+            #[cfg(feature = "metrics")]
+            metrics::killswitch_queue::ops_sent(metrics::OP_KIND_REMOVE).inc();
         }
     }
 }
@@ -106,6 +111,8 @@ impl TaskKillswitch {
         .abort_handle();
 
         let _ = task_tx.send(ActiveTaskOp::Add { id, handle });
+        #[cfg(feature = "metrics")]
+        metrics::killswitch_queue::ops_sent(metrics::OP_KIND_ADD).inc();
     }
 
     fn activate(&self) {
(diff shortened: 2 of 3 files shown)


task-killswitch/src/lib.rs
```diff
@@ -143,8 +143,11 @@ struct ActiveTasks {
 
 impl ActiveTasks {
     async fn collect(mut self) {
-        while let Some(op) = self.task_rx.recv().await {
-            self.handle_task_op(op);
+        let mut ops = Vec::with_capacity(16);
```
gregor-cf: instead of running in a tokio task, would it make sense to spawn a dedicated thread that works the queue? While processing many ops at once helps with the coop budget, each poll() is also taking longer. 
(assuming a task can be aborted via the handle from outside the runtime).

task-killswitch/src/lib.rs
```diff
@@ -49,6 +52,8 @@ impl Drop for RemoveOnDrop {
     fn drop(&mut self) {
         if let Some(tx) = self.task_tx_weak.upgrade() {
             let _ = tx.send(ActiveTaskOp::Remove { id: self.id });
+            #[cfg(feature = "metrics")]
+            crate::metrics::killswitch_queue::ops_sent("remove").inc();
```
bobrik: Make `add` and `remove` constants?

task-killswitch/src/lib.rs
bobrik: I too was thinking about having a separate thread for consumption. The only async op is `recv()`, so we'll have to switch to a sync channel.

It might be a reasonable idea, but I would separate this in another PR.

task-killswitch/src/lib.rs
bobrik: Also: does it need `crate::`? You already have `mod metrics`.

TheJokr: not necessary, removed it to make it less verbose

task-killswitch/src/lib.rs
TheJokr: makes sense, i don't see any reason why a thread + std::sync::mpsc would be worse. i'll make another PR tomorrow.

LloydW93: I'd prefer to, at least as a configurable, be able to disable coop mode for the task, through https://docs.rs/tokio/latest/tokio/task/coop/fn.unconstrained.html or similar.

With a multi-consumer channel, the task then being smart enough to realise it is getting a lot of work and spawning more of itself, would be great, but unconstrained helps a lot.

bobrik: I did suggest unconstrained, but I'm also interested getting some metrics first so that we could compare before and after. We're doing `recv_many()` here, so maybe the ship has sailed.

gregor-cf: We should not use unconstrained(). Then we are turning this into a purely CPU bound tasks and we shouldn't run those on the same workers as IO tasks. 

> The only async op is recv(), so we'll have to switch to a sync channel.

no need. The tokio receiver has `blocking_recv()` and `blocking_recv_many()`.

TheJokr: > I'd prefer to be able to disable coop mode for the task

I agree with Gregor here. Without knowing whether the task is CPU-bound or just isn't getting polled enough, slapping `unconstrained` on it has the potential to starve IO tasks. With a separate thread, we can `nice()` it to behave.

> We're doing recv_many() here, so maybe the ship has sailed.

I'll drop that commit in favor of the separate thread approach (in a separate PR). This will allow you to get metrics with the old `recv()` code if you want, or switch to threads directly.

E: Opened #2194

task-killswitch/Cargo.toml
```diff
@@ -8,9 +8,13 @@ keywords = { workspace = true }
 categories = { workspace = true }
 description = "Abort all tokio tasks at once"
 
+[features]
+metrics = ["dep:foundations"]
+
 [dependencies]
 parking_lot = { workspace = true }
 tokio = { workspace = true, features = ["rt", "sync"] }
+foundations = { workspace = true, optional = true, features = ["metrics"] }
```
ghedo: Nit: please keep alphabetical ordering.

TheJokr: fixed, though the consensus seems to be to move forward with #2196, where these metrics don't make sense anymore.

ghedo: Oh, even better. I guess we can just close this and https://github.com/cloudflare/quiche/pull/2194 then?

PR #2191, opened 2025-10-14
h3: clear streams when send finishes before recv

ghedo:
h3 streams are meant to be dopped from the internal stream map when both the send and recv side of the stream are finished (this means both sending and received a `fin` for example).

However currently this only works if the recv side is finished first, and the clean-up happens on the next `send_body()` or `send_response()` call. When the opposite happens (e.g. `fin` is sent before receiving it) the the stream leaks.

This change adds logic to properly check for a stream being completed in that case as well, which involves tracking a "local finished" state for each stream.

When returning a `Finished` event, the corresponding stream is checked again and if its send-side is also finished then the stream's state is cleared.

Note that this change doesn't cover the case of stream being reset by the local application itself, as the h3 layer doesn't get any signal about that happening right now, so it will need follow-up changes to be fixed.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -282,6 +282,7 @@
 //! [`send_response()`]: struct.Connection.html#method.send_response
 //! [`send_body()`]: struct.Connection.html#method.send_body
 
+use std::collections::hash_map;
 use std::collections::HashSet;
 use std::collections::VecDeque;
 
@@ -977,6 +978,10 @@ pub struct Connection {
 
     max_push_id: u64,
 
+    // Streams whose peer send side has finished and still need a Finished
+    // event. If the local send side is already done, poll() removes the H3
+    // stream state when returning the event. Otherwise the send path removes it
+    // when the local side finishes later.
     finished_streams: VecDeque<u64>,
 
     frames_greased: bool,
@@ -1496,12 +1501,10 @@ impl Connection {
             q.add_event_data_now(ev_data).ok();
         });
 
-        if let Some(s) = self.streams.get_mut(&stream_id) {
+        if fin {
+            self.finish_local_stream(conn, stream_id, true);
+        } else if let Some(s) = self.streams.get_mut(&stream_id) {
             s.initialize_local();
-        }
-
-        if fin && conn.stream_finished(stream_id) {
-            self.streams.remove(&stream_id);
         }
 
         Ok(())
@@ -1743,8 +1746,8 @@ impl Connection {
             let _ = conn.stream_writable(stream_id, overhead + 1);
         }
 
-        if fin && written == len && conn.stream_finished(stream_id) {
-            self.streams.remove(&stream_id);
+        if fin && written == len {
+            self.finish_local_stream(conn, stream_id, false);
         }
 
         Ok(ret)
@@ -2097,8 +2100,8 @@ impl Connection {
         }
 
         // Process finished streams list.
-        if let Some(finished) = self.finished_streams.pop_front() {
-            return Ok((finished, Event::Finished));
+        if let Some(ev) = self.pop_finished_stream(conn) {
+            return Ok(ev);
         }
 
         // Process HTTP/3 data from readable streams.
@@ -2112,8 +2115,11 @@ impl Connection {
 
                 // Return early if the stream was reset, to avoid returning
                 // a Finished event later as well.
-                Err(Error::TransportError(crate::Error::StreamReset(e))) =>
-                    return Ok((s, Event::Reset(e))),
+                Err(Error::TransportError(crate::Error::StreamReset(e))) => {
+                    self.remove_local_finished_stream(s);
+
+                    return Ok((s, Event::Reset(e)));
+                },
 
                 Err(e) => return Err(e),
             };
@@ -2131,17 +2137,8 @@ impl Connection {
         // Process finished streams list once again, to make sure `Finished`
         // events are returned when receiving empty stream frames with the fin
         // flag set.
-        if let Some(finished) = self.finished_streams.pop_front() {
-            if conn.stream_readable(finished) {
-                // The stream is finished, but is still readable, it may
-                // indicate that there is a pending error, such as reset.
-                if let Err(crate::Error::StreamReset(e)) =
-                    conn.stream_recv(finished, &mut [])
-                {
-                    return Ok((finished, Event::Reset(e)));
-                }
-            }
-            return Ok((finished, Event::Finished));
+        if let Some(ev) = self.pop_finished_stream(conn) {
+            return Ok(ev);
         }
 
         Err(Error::Done)
@@ -2890,6 +2887,58 @@ impl Connection {
 
             _ => (),
         };
+    }
+
+    fn finish_local_stream<F: BufFactory>(
+        &mut self, conn: &super::Connection<F>, stream_id: u64,
+        initialize_local: bool,
+    ) {
+        let hash_map::Entry::Occupied(mut stream) = self.streams.entry(stream_id)
+        else {
+            return;
+        };
+
+        {
+            let stream = stream.get_mut();
+
+            if initialize_local {
+                stream.initialize_local();
+            }
+
+            stream.finish_local();
+        }
+
+        if conn.stream_finished(stream_id) {
+            stream.remove();
+        }
+    }
+
+    fn remove_local_finished_stream(&mut self, stream_id: u64) {
+        if let hash_map::Entry::Occupied(stream) = self.streams.entry(stream_id) {
+            if stream.get().local_finished() {
+                stream.remove();
+            }
+        }
+    }
+
+    fn pop_finished_stream<F: BufFactory>(
+        &mut self, conn: &mut super::Connection<F>,
+    ) -> Option<(u64, Event)> {
+        let finished = self.finished_streams.pop_front()?;
+
+        self.remove_local_finished_stream(finished);
+
+        if conn.stream_readable(finished) {
+            // The stream is finished, but is still readable, it may indicate
+            // that there is a pending error, such as reset.
+            if let Err(crate::Error::StreamReset(e)) =
+                conn.stream_recv(finished, &mut [])
+            {
+                return Some((finished, Event::Reset(e)));
+            }
+        }
+
+        Some((finished, Event::Finished))
     }
 
     fn process_frame<F: BufFactory>(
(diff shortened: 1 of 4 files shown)


quiche/src/h3/mod.rs
```diff
@@ -7349,6 +7384,124 @@ mod tests {
...
+        s.send_response(stream, false).unwrap();
+        s.pipe
+            .server
+            .stream_shutdown(stream, crate::Shutdown::Write, 0)
+            .unwrap();
+
+        s.advance().ok();
+
+        // TODO: need to notify resets better.
+        let _ = s.send_body_server(stream, true);
```
LPardue, line 8079: I don't follow how this triggers a reset in this test

LPardue: ohh the comment at https://github.com/cloudflare/quiche/pull/2191/files#diff-851097865419440f06ef5f0c6fcbca011863b552fe556fdcded5c368fdeb9715R7484 is misleading, consiser changing it

ghedo: Yeah good point. I added more details to the TODO comment and fixed the one above as well.

ghedo: But it's basically the problem I mention in the PR description about stream resets.

LPardue (approved):

quiche/src/lib.rs
```diff
@@ -6538,6 +6538,23 @@ impl<F: BufFactory> Connection<F> {
...
+            return true;
+        }
+
+        let stream = match self.streams.get(stream_id) {
+            Some(v) => v,
+
+            None => return false,
+        };
+
+        stream.recv.is_fin() && stream.send.is_fin()
```
LPardue: I think this is  a buggy version of what you want to accomplish. It doesn't seem to work for unidirectinal streams for instance; we already have stream.is_complete() that can deal with stream types. MY friend clause suggests send.is_complete() is preferred to is_fin() too, since the semantics are different

LPardue (changes requested):
think there's a bug in the second commit

quiche/src/lib.rs
LPardue: Something like this is probably better

```
fn stream_closed(&self, stream_id: u64) -> bool {
    if self.streams.is_collected(stream_id) {
        return true;
    }
    self.streams.get(stream_id).map_or(false, |s| s.is_complete())
}
```

ghedo: It's a bit trickier than that because `is_complete()` checks on the send-side that all data has also been acked, rather than just fin being seen. Though I can update this menthod to make it work for uni streams too.

LPardue: I figured it was actually more correct to wait for th fin ack before collecting 🤔

ghedo: Made it support uni streams, but kept `send.is_fin()`.

ghedo: > I figured it was actually more correct to wait for th fin ack before collecting 🤔

Yes, for collecting the stream, but this is only used to avoid re-creating the state for the same stream by accident when receiving a priority update after the stream is closed. It's related to the collection issue, but not quite the same (here we just prevent the state from being re-created to begin with, so that it's not leaked).

LPardue: aha, ok.

LPardue (approved):
LGTM now

quiche/src/h3/mod.rs
```diff
@@ -2132,6 +2160,12 @@ impl Connection {
         // events are returned when receiving empty stream frames with the fin
         // flag set.
         if let Some(finished) = self.finished_streams.pop_front() {
+            if let Some(stream) = self.streams.get_mut(&finished) {
+                if stream.local_finished() {
+                    self.streams.remove(&finished);
+                }
+            }
+
             if conn.stream_readable(finished) {
```
antoniovicente: Pre-existing issue:

This readable check is not included in the first time that `self.finished_streams.pop_front()` appears in this poll.  Why the inconsistency between the first and second call?

quiche/src/h3/stream.rs
```diff
@@ -204,8 +207,11 @@ impl Stream {
             frame_type: None,
 
             is_local,
+
             remote_initialized: false,
+
             local_initialized: false,
+            local_finished: false,
```
antoniovicente, line 214: nit: the addition of newlines between some of the existing fields, but not before this newly added field seems arbitrary.

quiche/src/lib.rs
```diff
@@ -6542,6 +6542,39 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// This also returns true if the stream has already been collected.
+    #[inline]
+    pub fn stream_closed(&self, stream_id: u64) -> bool {
+        if self.streams.is_collected(stream_id) {
+            return true;
+        }
+
+        let Some(stream) = self.streams.get(stream_id) else {
+            return false;
```
antoniovicente: It is surprising that it returns false for unknown streams.

quiche/src/lib.rs
```diff
@@ -6542,6 +6542,39 @@ impl<F: BufFactory> Connection<F> {
...
+    /// Returns true if the specified stream is closed.
+    ///
+    /// For bidirectional streams this happens when both the receive and send
+    /// sides have signaled `fin`. For unidirectional streams only the
+    /// relevant direction is checked, depending on whether the stream was
+    /// created locally or not.
+    ///
+    /// This also returns true if the stream has already been collected.
+    #[inline]
+    pub fn stream_closed(&self, stream_id: u64) -> bool {
```
antoniovicente, line 6555: Unit tests?

quiche/src/h3/mod.rs
```diff
@@ -5221,6 +5219,55 @@ mod tests {
...
+            Header::new(b":status", b"200"),
+            Header::new(b"server", b"quiche-test"),
+        ];
+
+        s.server
+            .send_response(&mut s.pipe.server, stream, &resp, true)
+            .unwrap();
+
+        assert_eq!(s.server.streams.len(), init_streams_server);
+        assert!(!s.pipe.server.streams.is_collected(stream));
```
antoniovicente, line 5274: Why is the stream not collected at this point?

Consider adding assertions on `conn.stream_closed` and `is_fin`, and comments explaining why the stream is still around.

quiche/src/h3/mod.rs
```diff
@@ -1496,11 +1496,17 @@ impl Connection {
             q.add_event_data_now(ev_data).ok();
         });
 
+        let recv_finished = conn.stream_finished(stream_id);
+
         if let Some(s) = self.streams.get_mut(&stream_id) {
             s.initialize_local();
+
+            if fin && !recv_finished {
```
antoniovicente: why `!recv_finished`?

This seems like a confusing optimization: we avoid setting local_finished on the stream if we're about to remove it from the map shortly in a few lines.

quiche/src/h3/mod.rs
```diff
@@ -1743,7 +1749,16 @@ impl Connection {
             let _ = conn.stream_writable(stream_id, overhead + 1);
         }
 
-        if fin && written == len && conn.stream_finished(stream_id) {
+        let send_finished = fin && written == len;
+        let recv_finished = conn.stream_finished(stream_id);
+
+        if let Some(s) = self.streams.get_mut(&stream_id) {
+            if send_finished && !recv_finished {
```
antoniovicente: See https://github.com/cloudflare/quiche/pull/2191/changes/02cdc449bd61bad048e104b190ba45119b2a6754#r3270711710

quiche/src/h3/mod.rs
```diff
@@ -2098,6 +2113,12 @@ impl Connection {
 
         // Process finished streams list.
         if let Some(finished) = self.finished_streams.pop_front() {
+            if let Some(stream) = self.streams.get_mut(&finished) {
+                if stream.local_finished() {
+                    self.streams.remove(&finished);
+                }
+            }
```
antoniovicente: The previous 5 lines appear multiple times in this method.  Is it time to factor it out to a helper function?

quiche/src/h3/mod.rs
```diff
@@ -2098,6 +2113,12 @@ impl Connection {
 
         // Process finished streams list.
         if let Some(finished) = self.finished_streams.pop_front() {
```
antoniovicente: Do we need a comment on `self.finished_streams` above describing what it means to be in this Vec?

I find it confusing that not all finished_streams have `stream.local_finished()` set.  In the cases where it is not set, who is in charge of removing the stream from the map?

quiche/src/h3/stream.rs
ghedo: The blank lines are intentional here to group the local/remote initialization flags separately from the event state.

quiche/src/h3/mod.rs
ghedo: Factored both finished-stream pops through pop_finished_stream(), so they now both apply the same readable/reset check before returning Finished.

quiche/src/lib.rs
ghedo: Kept never-opened streams distinct from collected streams so PRIORITY_UPDATE can still arrive before the request stream exists. I updated the docs to spell out that behavior and changed the lookup order so active streams avoid the collected-set lookup.

quiche/src/lib.rs
ghedo: Added stream_closed_bidi and stream_closed_uni tests covering never-opened streams, partially closed streams, and closed streams for both directions.

quiche/src/h3/mod.rs
ghedo: Added assertions for stream_finished(), stream_closed(), the recv/send FIN state, and the not-yet-collected transport stream. The comment now explains that H3 has dropped its stream state, while QUIC keeps the stream until the response FIN is acknowledged.

quiche/src/h3/mod.rs
ghedo: Removed the !recv_finished special case. finish_local_stream() now marks the local side finished and removes the H3 stream immediately when the transport receive side is already finished.

quiche/src/h3/mod.rs
ghedo: Same fix as the HEADERS path: the send-body path now calls finish_local_stream(), which marks the local side finished and removes the H3 stream if the receive side is already finished.

quiche/src/h3/mod.rs
ghedo: Factored the repeated cleanup into pop_finished_stream() and remove_local_finished_stream(), and used the same helper for both finished-stream pops and the reset path.

quiche/src/h3/mod.rs
ghedo: Added a comment on finished_streams explaining that it tracks streams waiting for a Finished event. If the local send side is already done, poll() removes the H3 stream when returning the event. Otherwise the send path removes it when the local side finishes later.

antoniovicente (approved):

PR #2193, opened 2025-10-14
h3: remove duplicated stream type validation

ghedo:
The `Stream::set_frame_type()` already checks whether a frame is allowed on a specific type of stream, so we don't have to check it again while processing the frame.

### DIFF
--- a/quiche/src/h3/stream.rs
+++ b/quiche/src/h3/stream.rs
@@ -349,6 +349,12 @@ impl Stream {
                         (frame::MAX_PUSH_FRAME_TYPE_ID, _) =>
                             return Err(Error::FrameUnexpected),
 
+                        (frame::PRIORITY_UPDATE_FRAME_REQUEST_TYPE_ID, _) =>
+                            return Err(Error::FrameUnexpected),
+
+                        (frame::PRIORITY_UPDATE_FRAME_PUSH_TYPE_ID, _) =>
+                            return Err(Error::FrameUnexpected),
+
                         // All other frames can be ignored regardless of stream
                         // state.
                         _ => (),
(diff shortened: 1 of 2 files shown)


quiche/src/h3/stream.rs
```diff
@@ -349,6 +349,12 @@ impl Stream {
                         (frame::MAX_PUSH_FRAME_TYPE_ID, _) =>
                             return Err(Error::FrameUnexpected),
 
+                        (frame::PRIORITY_UPDATE_FRAME_REQUEST_TYPE_ID, _) =>
+                            return Err(Error::FrameUnexpected),
+
+                        (frame::PRIORITY_UPDATE_FRAME_PUSH_TYPE_ID, _) =>
+                            return Err(Error::FrameUnexpected),
```
jrouviere, line 356: Feels like that might be easy to forget a type of frame here. Would that be simpler to invert the logic and only allow the types we expect?

LPardue: Its a good question. Unlike QUIC, HTTP/3 allows senders to send unknown frame types without negotiation. So there are effectively 3 types of check happening here:

1. Is the frame known and allowed on the request stream?
2. Is the frame known and not allowed on the request stream?
3. Is the frame unknown and hence we have to permit it on this stream and skip over it to read the next frame.

A catch-all pattern matching risks conflating (2) and (3).

LPardue: Also, new frames don't occur that often, and I'd expect a PR to add one to quiche to also add a unit test to validate the stream expectations in the spec

jrouviere (approved):

quiche/src/h3/stream.rs
jrouviere: Got it, makes sense.

PR #2195, opened 2025-10-15
tokio-quiche: add ServerH3Event::Headers

LPardue:
Priority signals take the form of HTTP headers and/or frames. Some clients only send frames, while some clients can both. At the point in time a server reads request HEADERS, a client may sent a PRIORITY_UPDATE frame that should take precedence.

With this change, we read take the last PRIORITY_UPDATE processed by quiche, and pass it alongside IncomingH3Headers in a new ServerH3Event:Headers event o that Tokio-Quiche-based server can use it if they wish.

This is the alternative design proposed by @evanrittenhouse on https://github.com/cloudflare/quiche/pull/2189/files#r2430013438

### DIFF
--- a/tokio-quiche/examples/async_http3_server/server.rs
+++ b/tokio-quiche/examples/async_http3_server/server.rs
@@ -44,6 +44,7 @@ use tokio_quiche::http3::driver::H3Event;
 use tokio_quiche::http3::driver::IncomingH3Headers;
 use tokio_quiche::http3::driver::OutboundFrame;
 use tokio_quiche::http3::driver::OutboundFrameSender;
+use tokio_quiche::http3::driver::RawPriorityValue;
 use tokio_quiche::http3::driver::ServerEventStream;
 use tokio_quiche::http3::driver::ServerH3Event;
 use tokio_quiche::BoxError;
@@ -102,20 +103,17 @@ where
     ) -> QuicResult<()> {
         loop {
             match h3_event_receiver.recv().await {
-                Some(event) => self.handle_h3_event(event).await?,
+                Some(event) => self.handle_server_h3_event(event).await?,
                 None => return Ok(()), /* The sender was dropped, implying
                                         * connection was terminated */
             }
         }
     }
 
-    /// Handle a [`ServerH3Event`].
+    /// Handle a [`H3Event`].
     ///
     /// For simplicity's sake, we only handle a couple of events here.
-    // TODO(evanrittenhouse): support POST requests
-    async fn handle_h3_event(&mut self, event: ServerH3Event) -> QuicResult<()> {
-        let ServerH3Event::Core(event) = event;
-
+    fn handle_h3_event(event: H3Event) -> QuicResult<()> {
         match event {
             // Received an explicit connection level error. Not much to do here.
             H3Event::ConnectionError(err) => QuicResult::Err(Box::new(err)),
@@ -130,15 +128,29 @@ where
                 QuicResult::Err(err)
             },
 
-            // Received headers for a new stream from the H3Driver.
-            H3Event::IncomingHeaders(headers) => {
-                self.handle_incoming_headers(headers).await;
-
-                Ok(())
-            },
-
             _ => {
                 log::info!("received unhandled event: {event:?}");
+                Ok(())
+            },
+        }
+    }
+
+    /// Handle a [`ServerH3Event`].
+    ///
+    /// TODO(evanrittenhouse): support POST requests
+    async fn handle_server_h3_event(
+        &mut self, event: ServerH3Event,
+    ) -> QuicResult<()> {
+        match event {
+            ServerH3Event::Core(event) => Self::handle_h3_event(event),
+
+            ServerH3Event::Headers {
+                incoming_headers,
+                priority,
+            } => {
+                // Received headers for a new stream from the H3Driver.
+                self.handle_incoming_headers(incoming_headers, priority)
+                    .await;
                 Ok(())
             },
         }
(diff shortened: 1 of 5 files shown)


tokio-quiche/src/http3/driver/server.rs
```diff
@@ -64,11 +64,23 @@ pub type ServerEventStream = mpsc::UnboundedReceiver<ServerH3Event>;
 #[derive(Debug)]
 pub enum ServerH3Event {
     Core(H3Event),
+
+    Headers {
+        incoming_headers: IncomingH3Headers,
+        /// The latest PRIORITY_UPDATE frame value, if any.
+        priority: Option<Vec<u8>>,
```
LPardue: I need some input on the type we use. The PRIORITY_UPDATE frame jsut carries bytes. We could instead parse them directly into a `Priority` object using try_from defined here https://github.com/cloudflare/quiche/blob/master/quiche/src/h3/mod.rs#L842, however that is guarded behind the sfv feature (which pulls in the sfv crate). Do we want to make TQ dependent on this feature?

tokio-quiche/tests/fixtures/mod.rs
```diff
@@ -133,21 +133,28 @@ pub async fn serve_connection_details(
...
-            Some(frame) = event_rx.recv() => {
-                let ServerH3Event::Core(frame) = frame;
-                match frame {
-                    H3Event::IncomingSettings {..} | H3Event::BodyBytesReceived { .. } | H3Event::StreamClosed { .. } => {},
-                    H3Event::IncomingHeaders(headers) => {
+            Some(event) = event_rx.recv() => {
+                match event {
+                    ServerH3Event::Core(event) => {
+                            match event {
+                            H3Event::IncomingSettings {..} | H3Event::BodyBytesReceived { .. } | H3Event::StreamClosed { .. } => {},
```
evanrittenhouse: nit: formatting - clippy sometimes has issues in macros

LPardue: Oops, I meant to clean up this area anyhow

tokio-quiche/examples/async_http3_server/server.rs
```diff
@@ -114,31 +114,36 @@ where
...
+                        log::info!("received unhandled event: {event:?}");
+                        Ok(())
+                    },
+                }
             },
 
-            _ => {
-                log::info!("received unhandled event: {event:?}");
+            ServerH3Event::Headers {
+                incoming_headers, ..
```
antoniovicente: So we're ignoring `priority` for now?

Worth a TODO?

tokio-quiche/tests/fixtures/mod.rs
```diff
@@ -133,21 +133,25 @@ pub async fn serve_connection_details(
...
         select! {
-            Some(frame) = event_rx.recv() => {
-                let ServerH3Event::Core(frame) = frame;
-                match frame {
-                    H3Event::IncomingSettings {..} | H3Event::BodyBytesReceived { .. } | H3Event::StreamClosed { .. } => {},
-                    H3Event::IncomingHeaders(headers) => {
+            Some(event) = event_rx.recv() => {
+                match event {
+                    ServerH3Event::Core(event) => {
+                        match event {
```
antoniovicente, line 139: Holy nested `match`es batman!

Could something be done to make this test code more readable?

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -130,6 +160,12 @@ impl ServerHooks {
             stream_ctx.associated_dgram_flow_id = Some(flow_id);
         }
 
+        let latest_priority_update: Option<RawPriorityValue> = driver
+            .conn_mut()?
+            .take_last_priority_update(stream_id)
+            .ok()
+            .map(|v| v.into());
```
antoniovicente, line 167: I think this is the main functional change in this PR.  But I think there are no assertions that would detect if you were to change the right hand side of this expression to `None`

I assume there are planned followup changes to use this priority information.

tokio-quiche/examples/async_http3_server/server.rs
LPardue: Sure can do

tokio-quiche/src/http3/driver/server.rs
LPardue: It's up to apps to use it if they want. I guess the todo comment above will be an indicator that you can do something in that event.

tokio-quiche/tests/fixtures/mod.rs
LPardue: Since its just test code I didn't want to spend to much time refining it but I'm open to suggestions. The obvious tweak would be to add functions for each ServerH3Event but its only marginal

tokio-quiche/src/http3/driver/server.rs
antoniovicente: Having some tests that verify that this plumbing works as advertised would be nice.

tokio-quiche/examples/async_http3_server/server.rs
antoniovicente: I missed this is example code.  At least we should consider including the priority field and consider ignoring it more explicitly.

tokio-quiche/src/http3/driver/server.rs
LPardue: Yeah, the difficulty is in needing a client that could send the PRIORITY_UPDATE frame. That's not something we expose in the TQ client already and I something that is tangential to current needs. It is possible in h3i but adding that as a test capability seems a lot of work for the benefit t might provide. Not sure if @evanrittenhouse has considered h3i here before.

antoniovicente (approved):

PR #2196, opened 2025-10-15
task-killswitch: Switch to dashmap instead of worker task

TheJokr:
As an alternative to #2194, it is worth considering a different design for task-killswitch. Using a concurrent hashmap like `dashmap`, each `spawn_with_killswitch()` call pays the cost of adding a task itself. Task cleanup meanwhile becomes the responsibility of the task itself, when it returns its result.

dashmap internally is set up as a set of `4 * nproc` RwLock-ed regular hashmaps. Keys are sharded into maps by their hash. This means collisions between different tasks being inserted or removed concurrently are quite unlikely, since they use different keys (provided the hash function is good.)

To get an idea of potential performance, I ran the benchmarks linked from dashmap's repo locally. With a 50/50 insert/remove workload using random u64 keys, dashmap had an average latency per operation between 100-200 ns and a throughput of millions of ops/s per thread. I think this is more than fine for the task-killswitch use case.

### DIFF
--- a/task-killswitch/src/lib.rs
+++ b/task-killswitch/src/lib.rs
@@ -24,99 +24,114 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use tokio::sync::mpsc;
+use dashmap::DashMap;
+use parking_lot::Mutex;
 use tokio::sync::watch;
+use tokio::task;
 use tokio::task::AbortHandle;
 
-use std::collections::HashMap;
 use std::future::Future;
-use std::sync::atomic::AtomicU64;
+use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::LazyLock;
-
-enum ActiveTaskOp {
-    Add { id: u64, handle: AbortHandle },
-    Remove { id: u64 },
-}
 
 /// Drop guard for task removal. If a task panics, this makes sure
 /// it is removed from [`ActiveTasks`] properly.
 struct RemoveOnDrop {
-    id: u64,
-    task_tx_weak: mpsc::WeakUnboundedSender<ActiveTaskOp>,
+    id: task::Id,
+    storage: &'static ActiveTasks,
 }
 impl Drop for RemoveOnDrop {
     fn drop(&mut self) {
-        if let Some(tx) = self.task_tx_weak.upgrade() {
-            let _ = tx.send(ActiveTaskOp::Remove { id: self.id });
-        }
+        self.storage.remove_task(self.id);
     }
 }
 
 /// A task killswitch that allows aborting all the tasks spawned with it at
-/// once. The implementation strives to not introduce any in-band locking, so
-/// spawning the future doesn't require acquiring a global lock, keeping the
-/// regular pace of operation.
+/// once. The implementation strives to minimize in-band locking. Spawning a
+/// future requires a single sharded lock from an internal [`DashMap`].
+/// Conflicts are expected to be very rare (dashmap defaults to `4 * nproc`
+/// shards, while each thread can only spawn one task at a time.)
 struct TaskKillswitch {
-    // NOTE: use a lock without poisoning here to not panic all the threads if
-    // one of the worker threads panic.
-    task_tx: parking_lot::RwLock<Option<mpsc::UnboundedSender<ActiveTaskOp>>>,
-    task_counter: AtomicU64,
+    // Invariant: If `activated` is true, we don't add new tasks anymore.
+    activated: AtomicBool,
+    storage: &'static ActiveTasks,
+
+    /// Watcher that is triggered after all kill signals have been sent (by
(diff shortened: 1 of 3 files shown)


task-killswitch/src/lib.rs
```diff
@@ -24,99 +24,107 @@
...
+/// Conflicts are expected to be very rare (dashmap defaults to `4 * nproc`
+/// shards, while each thread can only spawn one task at a time.)
 struct TaskKillswitch {
-    // NOTE: use a lock without poisoning here to not panic all the threads if
-    // one of the worker threads panic.
-    task_tx: parking_lot::RwLock<Option<mpsc::UnboundedSender<ActiveTaskOp>>>,
-    task_counter: AtomicU64,
+    // Invariant: If `activated` is true, we don't add new tasks anymore.
+    activated: AtomicBool,
+    id_counter: AtomicU64,
```
gregor-cf: you can avoid this atomic and the associated atomic increment by using `AbortHandle::id()` (and `tokio::task::id()` inside the tasks) as the key for the map. (see https://docs.rs/tokio/latest/tokio/task/struct.Id.html for unique-ness)

task-killswitch/src/lib.rs
```diff
@@ -24,99 +24,107 @@
...
+        if self.was_activated() {
             return;
-        };
+        }
 
-        let id = self.task_counter.fetch_add(1, Ordering::SeqCst);
-        assert!(id < u64::MAX, "task-killswitch ID counter wrapped around!");
-        let task_tx_weak = task_tx.downgrade();
+        let id = self.id_counter.fetch_add(1, Ordering::Relaxed);
+        let storage = Arc::clone(&self.storage);
```
gregor-cf: There is a very small chance that the atomic ref count for this Arc could be a point of contention. (unlikely because the existing code already performs atomic increments and it does so with SeqCst), but something to potentially look for in profiles/flamegraphs. 

The other option is to use `&'static` for the `storage` (or `self`). Since we already use a global task-killswitch that would be easy to do and it avoids any potential contention on the atomic ref count here.

task-killswitch/src/lib.rs
```diff
@@ -24,99 +24,107 @@
...
-    task_counter: AtomicU64,
+    // Invariant: If `activated` is true, we don't add new tasks anymore.
+    activated: AtomicBool,
+    id_counter: AtomicU64,
+    storage: Arc<ActiveTasks>,
+
+    // NOTE: We only need something like a "TakeOnce" for the sender here, but
+    // adding another dependency doesn't seem worth it. The lock will be unlocked
+    // anyway.
+    signal_killed: Mutex<Option<watch::Sender<()>>>,
```
gregor-cf: nit: you don't need this mutex. The atomic compare-and-swap on `activated` ensures it can only be taken once.

task-killswitch/src/lib.rs
```diff
@@ -24,99 +24,107 @@
...
+        let signal_killed = Mutex::new(Some(signal_killed));
 
         Self {
-            task_tx: parking_lot::RwLock::new(Some(task_tx)),
-            task_counter: Default::default(),
+            activated: AtomicBool::new(false),
+            id_counter: AtomicU64::default(),
+            storage: Arc::new(ActiveTasks::default()),
+            signal_killed,
             all_killed,
```
gregor-cf, line 80: nit: it's existing code, but it might be good to add a comment to `all_killed` or `fn killed()` to clarify that it will signal as soon as the abort signal has been *sent*, not when the tasks have actually *finished*

task-killswitch/src/lib.rs
bobrik: `arc-swap` is less contentious than `Arc` if you want to avoid globals.

task-killswitch/src/lib.rs
```diff
@@ -135,52 +143,45 @@ enum TaskEntry {
     Tombstone,
```
bobrik, line 150: Do we still need these? There aren't any `Add` messages anymore.

task-killswitch/src/lib.rs
TheJokr: I know, but i suppose the comment doesnt quite make that clear. I need interior mutability for `Option::take`, which I can do either via a raw `UnsafeCell` or a mutex. I chose the mutex because 1) as you said, it will be unlocked anyway, and 2) introducing `unsafe` into the codebase increases the barrier to entry for no relevant gain.

task-killswitch/src/lib.rs
TheJokr: Good point! There are comments on the public functions, but i'll add some here as well.

task-killswitch/src/lib.rs
TheJokr: We do, but only in pathological cases. The kernel scheduler could suspend the thread calling `spawn_task()` after `tokio::spawn()` and before `self.storage.add_task_if()`. I'll fix the comment.

task-killswitch/src/lib.rs
TheJokr: The struct is only used in the crate's own `static TASK_KILLSWITCH` and in tests, so switching to `&'static ActiveTasks` is trivial. I've done this now. If we ever need a TaskKillswitch with non-`'static` storage, we can still make `storage` generic.

PR #2217, opened 2025-10-24
BBR3: Only check for bandwidth growth in cases where last_packet_send_state is valid

antoniovicente:


quiche/src/recovery/mod.rs
```diff
@@ -2410,6 +2410,152 @@ mod tests {
...
+                120000
+            } else {
+                0
+            },
+        );
+
+        // Send some no "in_flight" packets
+        for iter in 3..1000 {
+            let mut p = test_utils::helper_packet_sent(next_packet, now, 1200);
+            p.in_flight = false;
```
antoniovicente, line 2505: I think `in_flight=false` means that the packet only contains ACKs.

https://datatracker.ietf.org/doc/html/rfc9002#name-congestion-control

quiche/src/recovery/gcongestion/bbr2/network_model.rs
```diff
@@ -618,6 +618,10 @@ impl BBRv2NetworkModel {
             return true;
         }
 
+        if !congestion_event.last_packet_send_state.is_valid {
```
antoniovicente, line 621: is_app_limited is incorrect if !is_valid

Logic below this line can trigger early startup exit.

evanrittenhouse (approved):

quiche/src/recovery/mod.rs
antoniovicente: Setting in_flight to false will send us down this branch and not add the packet number to connection_state_map

https://github.com/cloudflare/quiche/blob/3950b9fea821b68142453155df461d6bb0efa9b6/quiche/src/recovery/gcongestion/bbr/bandwidth_sampler.rs#L516

PR #2218, opened 2025-10-24
Simplify error handling by removing SendTimeState is_valid

antoniovicente:
Old code followed common C++ code patterns instead of Rust code patterns.

Not having the is_valid data member makes the code less error prone.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -406,13 +406,21 @@ pub(crate) struct BBRv2 {
     params: Params,
 }
 
-struct BBRv2CongestionEvent {
+struct BBRv2CongestionEventMeta {
     event_time: Instant,
 
     /// The congestion window prior to the processing of the ack/loss events.
     prior_cwnd: usize,
+
     /// Total bytes inflight before the processing of the ack/loss events.
     prior_bytes_in_flight: usize,
+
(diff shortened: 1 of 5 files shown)


quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -406,14 +406,22 @@ pub(crate) struct BBRv2 {
     params: Params,
 }
 
-struct BBRv2CongestionEvent {
+struct BBRv2CongestionEventMeta {
```
antoniovicente, line 409: Not too happy with this new struct.  Would consider just passing in the 4 members as arguments to `on_congestion_event_start`

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -636,20 +621,22 @@ impl CongestionControl for BBRv2 {
...
-        network_model.on_congestion_event_start(
+        let Some(mut congestion_event) = network_model.on_congestion_event_start(
+            meta,
             acked_packets,
             lost_packets,
-            &mut congestion_event,
             &self.params,
-        );
+        ) else {
+            return;
```
antoniovicente, line 638: This refactor is mostly removing is_valid in favor of Option.  Some behavior changes may be hiding in the details, including this early return.

The original version could be changed to return here early if `!is_valid`.

PR #2219, opened 2025-10-24
test: expose functionality to pass/retrieve in session info

toidiu:


antoniovicente (approved):

tokio-quiche/src/quic/mod.rs
```diff
@@ -220,6 +220,12 @@ where
 
     log::info!("created unestablished quiche::Connection"; "scid" => ?scid);
 
+    if let Some(session) = &params.session {
+        quiche_conn
+            .set_session(session)
+            .map_err(|_err| quiche::Error::CryptoFail)?;
```
evanrittenhouse: Since this is a user-provided parameter, it may be helpful to surface the actual error via `log::error`

h3i/src/client/connection_summary.rs
```diff
@@ -392,14 +392,18 @@ pub struct ConnectionCloseDetails {
...
 }
 
 impl ConnectionCloseDetails {
     pub fn new(qconn: &Connection) -> Self {
+        let session = qconn.session().map(|s| s.to_vec());
         Self {
             peer_error: qconn.peer_error().cloned(),
             local_error: qconn.local_error().cloned(),
             timed_out: qconn.is_timed_out(),
+            session,
```
evanrittenhouse, line 417: Can't these sessions be fairly large? This struct gets printed as part of the `ConnectionSummary`, so it may drown out the rest of the output. We may want to have a custom `Debug` implementation which ignores it - it'd still be accessible since it's a `pub` field.

antoniovicente: Printing the contents of the session struct to logs seems very bad from a security perspective also.  This struct should not implement debug.

tokio-quiche/src/quic/mod.rs
```diff
@@ -220,6 +220,13 @@ where
 
     log::info!("created unestablished quiche::Connection"; "scid" => ?scid);
 
+    if let Some(session) = &params.session {
+        quiche_conn.set_session(session).map_err(|_error| {
+            log::error!("application provided an invalid session");
```
evanrittenhouse:
```suggestion
            log::error!("application provided an invalid session"; "error"=>?error);
```

Issue #2200, opened 2025-10-16
--dump-responses argument


YoussefBaccouch:
I am conducting my bachelor thesis on quiche and I wanted to understand what can it be the cause that if i do not include the --dump-responses when I am running the quiche client binary the experiment of transferring a 100MB testfile from the server to the client on a 1 Gbit/s link takes around 54 seconds while when I include it the experient takes around 0.8 seconds as it should be.

Geun-Oh:
Hello.

I think it's because of `--dump-responses` uses `BufWriter.write_all()` without copying, and non `--dump-responses` does not.

So it may solved by removeing copy overhead (+string type conversions).

Would you let me fix this?

YoussefBaccouch:
Thanks for your help @Geun-Oh 👍.

PR #2221, opened 2025-10-25
Fix: Write stdout directly to avoid time comsumption without --dump-response option

### LINKED ISSUE #2220: Fix time consuming error without --dump-response argument
HI Cloudflare Team.

## Related 

This issue related with #2200 

## Description
When we don't use `--dump-response` overall time delays because of copying & converting types into String.

So I this issue fixes error by writing stdout even if we do not use `--dump-response`.

## Error Reproduce

Make dump file

```sh
dd if=/dev/zero of=/tmp/1GB.bin bs=1M count=1024
```

Open the server

```sh
cargo run --bin quiche-server -- \
--listen 127.0.0.1:4433 \
--cert apps/src/bin/cert.crt \
--key apps/src/bin/cert.key \
--root /tmp
```

Request with --dump-response

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify \
--dump-responses /tmp
```

Request without --dump-response

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify
```

===

We can check it takes more time when not using `--dump-response`

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify
   Compiling quiche_apps v0.1.0 (/Users/ohyeong-geun/rust/quiche/apps)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `target/debug/quiche-client 'https://127.0.0.1:4433/1GB.bin' --no-verify`
cargo run --bin quiche-client -- https://127.0.0.1:4433/1GB.bin --no-verify  5.43s user 8.30s system 81% cpu 16.792 total
```

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify \
--dump-responses /tmp
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/quiche-client 'https://127.0.0.1:4433/1GB.bin' --no-verify --dump-responses /tmp`
cargo run --bin quiche-client -- https://127.0.0.1:4433/1GB.bin --no-verify    0.15s user 0.08s system 31% cpu 0.729 total
```

Geun-Oh:
Hello team.

This PR Closes https://github.com/cloudflare/quiche/issues/2220

I've changed response writing logic without dump option to write down stdout directly without copying and converting types.

Now there are no dramatic timp delay between the two operations.

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify
   Compiling quiche_apps v0.1.0 (/Users/ohyeong-geun/rust/quiche/apps)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.57s
     Running `target/debug/quiche-client 'https://127.0.0.1:4433/1GB.bin' --no-verify`
cargo run --bin quiche-client -- https://127.0.0.1:4433/1GB.bin --no-verify  0.15s user 0.07s system 21% cpu 0.993 total
```

```sh
time cargo run --bin quiche-client -- \
https://127.0.0.1:4433/1GB.bin --no-verify \
--dump-responses /tmp
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/quiche-client 'https://127.0.0.1:4433/1GB.bin' --no-verify --dump-responses /tmp`
cargo run --bin quiche-client -- https://127.0.0.1:4433/1GB.bin --no-verify    0.15s user 0.06s system 41% cpu 0.489 total
```

Additionally, I've used BufWriter to minimize system calls (If we use ``std::io::stdout().write_all()`, there might be a x8 syscalls).

Please check and leave any comments.

Thanks.

### DIFF
--- a/apps/src/common.rs
+++ b/apps/src/common.rs
@@ -764,6 +764,7 @@ pub struct Http3Conn {
     dump_json: bool,
     dgram_sender: Option<Http3DgramSender>,
     output_sink: Rc<RefCell<dyn FnMut(String)>>,
+    stdout_writer: std::io::BufWriter<std::io::Stdout>,
 }
 
 impl Http3Conn {
@@ -857,6 +858,7 @@ impl Http3Conn {
             dump_json: dump_json.is_some(),
             dgram_sender,
             output_sink,
+            stdout_writer: std::io::BufWriter::new(std::io::stdout()),
         };
 
         Box::new(h_conn)
@@ -889,6 +891,7 @@ impl Http3Conn {
             dump_json: false,
             dgram_sender,
             output_sink,
+            stdout_writer: std::io::BufWriter::new(std::io::stdout()),
         };
 
         Ok(Box::new(h_conn))
@@ -1270,11 +1273,7 @@ impl HttpConn for Http3Conn {
 
                             None =>
                                 if !self.dump_json {
-                                    self.output_sink.borrow_mut()(unsafe {
-                                        String::from_utf8_unchecked(
-                                            buf[..read].to_vec(),
-                                        )
-                                    });
+                                    self.stdout_writer.write_all(&buf[..read]).ok();
                                 },
                         }
                     }


apps/src/common.rs
```diff
@@ -1270,11 +1273,7 @@ impl HttpConn for Http3Conn {
 
                             None =>
                                 if !self.dump_json {
-                                    self.output_sink.borrow_mut()(unsafe {
-                                        String::from_utf8_unchecked(
-                                            buf[..read].to_vec(),
-                                        )
-                                    });
+                                    self.stdout_writer.write_all(&buf[..read]).ok();
```
LPardue, line 1276: Without the from_utf8_unchecked() then we risk binary being printed to stdout, which can cause terminal corruption

Geun-Oh:
Hello @YoussefBaccouch .

I've been working with this issue (the one that you asked in https://github.com/cloudflare/quiche/issues/2200) and debugging all the internal logics.
But I discovered that when I call massive file from server, it takes pretty long time and there are no differences between `--dump-responses` argument usage.

I think it seems normal that it takes time to call massive files.
So can you just give me a reproducible example that I can debug real problem?

Thanks :)

apps/src/common.rs
Geun-Oh: Your right.
And I found that it seems normal to take time when calling large file. I need to check real issue situation and debug it again.

I've asked the one who issued this to provide reproducible example.

PR #2223, opened 2025-10-29
Add TimeHistogram for time spent in the sendmsg loop

antoniovicente:
We would like to understand how long connections can get stuck retrying the sendmsg.

### DIFF
--- a/tokio-quiche/src/quic/io/gso.rs
+++ b/tokio-quiche/src/quic/io/gso.rs
@@ -125,34 +125,34 @@ pub async fn send_to(
     // of timespec this allows to extract raw values from an [`Instant`]
     const INSTANT_ZERO: Instant = unsafe { std::mem::transmute(0u128) };
 
-    let iov = [std::io::IoSlice::new(send_buf)];
-    let segment_size_u16 = segment_size as u16;
-
-    let raw_time = tx_time
-        .map(|t| t.duration_since(INSTANT_ZERO).as_nanos() as u64)
-        .unwrap_or(0);
-
-    let pkt_info = from.map(PktInfo::from_socket_addr);
-
-    let mut cmsgs: SmallVec<[ControlMessage; 3]> = SmallVec::new();
-
-    // Create cmsg for UDP_SEGMENT.
-    cmsgs.push(ControlMessage::UdpGsoSegments(&segment_size_u16));
-
-    if tx_time.is_some() {
-        // Create cmsg for TXTIME.
-        cmsgs.push(ControlMessage::TxTime(&raw_time));
-    }
-
-    if let Some(pkt) = pkt_info.as_ref() {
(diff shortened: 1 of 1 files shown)


tokio-quiche/src/metrics/mod.rs
```diff
@@ -264,6 +272,11 @@ pub(crate) mod quic {
     /// Number of error and partial writes while sending QUIC packets
     pub fn write_errors(reason: labels::QuicWriteError) -> Counter;
 
+    /// Record timing information from sendmsg calls that return
+    /// WouldBlock and are retried in a loop.
+    #[ctor = HistogramBuilder { buckets: &[1E-6, 1E-5, 1E-4, 1E-3, 5E-3, 1E-2, 2E-2, 4E-2, 8E-2, 16E-2, 1.0], }]
+    pub fn send_to_wouldblock_duration() -> TimeHistogram;
```
evanrittenhouse: Can we do `send_to_wouldblock_duration_s` to signify that the units are in seconds?

tokio-quiche/src/quic/io/gso.rs
```diff
@@ -114,42 +115,44 @@ impl PktInfo {
...
+    if let Some(pkt) = pkt_info.as_ref() {
+        // Create cmsg for IP(V6)_PKTINFO.
+        cmsgs.push(pkt.make_cmsg());
+    }
 
-        let addr = SockaddrStorage::from(to);
+    let addr = SockaddrStorage::from(to);
 
+    let mut sendmsg_retry_timer = None;
+    loop {
```
evanrittenhouse: The IoSlice has an internal cursor which, at least in Linux, looks to be adjusted on each call to [`advance()`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/io/io_slice/iovec.rs#L29-L38), and we no longer recreate it on each iteration. 

I'm not actually sure how `IoSlice` interacts with `sendmsg` here. If the syscall returns EAGAIN after the cursor has somehow advanced, I think we'd lose bytes each time the socket buffer is full. I'm not actually sure that happens, but maybe it's worth pulling the `IoSlice` creation into the loop as well? 

Since everything else is dependent on the arguments to this function, I think we're fine with those being outside of the loop

antoniovicente: sendmsg arguments are immutable, so I think this is fine.

But I can also easily restore the old loop structure.  Done.

tokio-quiche/src/metrics/mod.rs
antoniovicente: done.

tokio-quiche/src/quic/io/gso.rs
evanrittenhouse: Very excited to see the results

jrouviere (approved):

PR #2224, opened 2025-10-31
fix: zero-initialize msghdr first for musl lib build compatibility 

### LINKED ISSUE #2222: Alpine musl libc build fails with `cannot construct `msghdr` with struct literal syntax due to private fields`
I'm trying to package quiche 0.24.5 for Alpine Linux , building with Rust 1.90 and latest libc crate 0.2.177. This errors with:
```
   Compiling serde_path_to_error v0.1.20
   Compiling datagram-socket v0.3.0 (/home/builder/package/testing/quiche/src/quiche-0.24.5/datagram-socket)
error: cannot construct `msghdr` with struct literal syntax due to private fields
  --> datagram-socket/src/mmsg.rs:59:26
   |
59 |                 msg_hdr: libc::msghdr {
   |                          ^^^^^^^^^^^^
   |
   = note: ...and other private fields `__pad1` and `__pad2` that were not provided

error: cannot construct `msghdr` with struct literal syntax due to private fields
   --> datagram-socket/src/mmsg.rs:119:26
    |
119 |                 msg_hdr: libc::msghdr {
    |                          ^^^^^^^^^^^^
    |
    = note: ...and other private fields `__pad1` and `__pad2` that were not provided

   Compiling slog-stdlog v4.1.1
error: could not compile `datagram-socket` (lib) due to 2 previous errors
```
msghdr cannot be directly instantiated on musl due to private padding fields, the issue https://github.com/rust-lang/libc/issues/2344 is marked "wontfix". Other projects have solved the issues using `mem::zeroed`, see https://github.com/anza-xyz/agave/pull/3836 and https://github.com/anza-xyz/agave/pull/4760

tedscode:
Hey there!

fixes #2222 
Like @strophy pointed out, there's an issue where **msghdr** can't be created directly on **musl** systems. The problem is that the musl libc has a couple extra private padding fields in the struct that aren't accessible.

#### Fix
This PR uses [std::mem::zeroed()](https://doc.rust-lang.org/std/mem/fn.zeroed.html) to initialize `msghdr` by just setting everything to `0`. This way quiche builds fine on musl-based distros.
If we take a look at the [msghdr](https://docs.rs/libc/latest/libc/struct.msghdr.html) struct, it's basically just **raw pointers** and **numbers**. By zeroing everything out (including those hidden padding fields), we are able to satisfy the kernel.

The original code **manually** set many fields to null pointers and `0`. With `zeroed()` doing this automatically, those manual assignments were commented out. (kept for reference)

#### Safety
Now, `std::mem::zeroed()` can be dangerous if used on the wrong types - like anything that can't be null or zero because that creates [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#invalid-values). 

But as mentioned before it's fine here since `msghdr` is just raw pointers and numeric types, which can all safely be `0`.

#### Performance
Ran some benchmarks with **criterion.rs** to compare how fast safe vs unsafe initialization of  msghdr struct are.
Optimizations stayed enabled since the goal was to test under realistic conditions.

The safe constructor (which doesn't work with musl) was indeed slower, but the difference wasn't as big as mentioned [here](https://github.com/anza-xyz/agave/pull/4760#issue-2828227375).
On average the unsafe approach is about **1.5-2%** faster than the version without `zeroed()`.
Between having redundant assignments vs just using `zeroed()`, the performance is virtually **identical**.

### DIFF
--- a/datagram-socket/src/mmsg.rs
+++ b/datagram-socket/src/mmsg.rs
@@ -25,9 +25,7 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use std::io::IoSlice;
-use std::io::{
-    self,
-};
+use std::io::{self};
 use std::os::fd::AsRawFd;
 use std::os::fd::BorrowedFd;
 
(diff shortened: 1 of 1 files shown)


ghedo:
Thanks for the contribution. #2576 addresses the same musl build issue through a shared initialization helper covering both the receive and send paths. Closing this PR in favor of #2576.

PR #2231, opened 2025-11-03
metrics: only log recurring WouldBlock errors.

toidiu:


tokio-quiche/src/quic/io/gso.rs
```diff
@@ -167,7 +169,20 @@ pub async fn send_to(
...
+                // loop.
+                //
+                // A WouldBlock error suggests that the library should try again
+                // later since it is expected that a socket will block
+                // occasionally.
+                //
+                // We would like to measure WouldBlock errors that result because
+                // of the hot `loop`ing calls to `sendmsg`.
+                if recurring_would_block {
+                    would_block_metric.inc();
```
antoniovicente, line 182: I think that send_to_wouldblock_duration_s_count gives you this...  I guess only once in cases where the same connection hits this connection multiple times in a loop.  Do we care about the difference?

antoniovicente (approved):

tokio-quiche/src/quic/io/gso.rs
toidiu: Given that we cant really control when a metal is overloaded and is returning WouldBlock, the best we can do is not stress the system and actually wait before the next sendmsg call. So I was interested in removing the noise of the first occurrence to see the true impact.

However, applying this change will hide the 1st WouldBlock occurrence, which is also quite insightful. so lets punt on this PR for now and evaluate after we see data from the histogram.

PR #2234, opened 2025-11-06
Allow to specify dcid when creating a client side connection

normanmaurer:
Motivation:

Sometimes the user want to have control over generating the dcid when creating a new client side connection. This was not possible before.

Modification:

- Only use dcid for local transport params when the connection is for server side
- Use provided odcid as dcid on the client side when given and if not just randomly generate one

Result:

More flexible usage possible

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1889,6 +1889,11 @@ impl<F: BufFactory> Connection<F> {
         client_dcid: Option<&ConnectionId>, local: SocketAddr, peer: SocketAddr,
         config: &Config, tls: tls::Handshake, is_server: bool,
     ) -> Result<Connection<F>> {
+        if retry_cids.is_some() && client_dcid.is_some() {
+            // These are exclusive, the caller should only specify one or the
+            // other.
+            return Err(Error::InvalidDcidInitialization);
+        }
         #[cfg(feature = "custom-client-dcid")]
         if let Some(client_dcid) = client_dcid {
             // The Minimum length is 8.


normanmaurer:
I really only care about be able to do this using the c headers but never less thought it might also be useful to be able to do this via the rust api so I added two new functions there as well.

LPardue:
The RFC places requirements for unpredictability and length on the client DCID field; see https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3

I'm concerned this change can make QUIC connections vulnerable, in ways described in the RFC security considerations, or in ways that are yet to be determined.

That said, there could be use cases where a deterministic CID (yet unpredictable to some actors) is useful.

Providing safety rails for the majority of default users is important. Perhaps this could be placed behind a very explicit opt-in feature flag with some clear warning are references to the relevant RFC text

normanmaurer:
> The RFC places requirements for unpredictability and length on the client DCID field; see https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
> 
> I'm concerned this change can make QUIC connections vulnerable, in ways described in the RFC security considerations, or in ways that are yet to be determined.

Fair enough...

> 
> That said, there could be use cases where a deterministic CID (yet unpredictable to some actors) is useful.


Yeah that's why I need it :)

> 
> Providing safety rails for the majority of default users is important. Perhaps this could be placed behind a very explicit opt-in feature flag with some clear warning are references to the relevant RFC text

That would be fine with me ... That said I think we could also add some validations for the length at least (must be at least 8 bytes).  Let me do this in any case.

normanmaurer:
@LPardue added the length check and also added some more docs. That said I would also be ok to just remove the two new rust functions completely as long as I can do it via FFI / c :)

quiche/src/lib.rs
```diff
@@ -1974,17 +2018,26 @@ impl Default for QlogInfo {
...
-        scid: &ConnectionId, odcid: Option<&ConnectionId>, local: SocketAddr,
+        scid: &ConnectionId, dcid: Option<&ConnectionId>, local: SocketAddr,
         peer: SocketAddr, config: &Config, tls: tls::Handshake, is_server: bool,
     ) -> Result<Connection<F>> {
+        if !is_server {
+            if let Some(dcid) = dcid {
+                // The Minimum length is 8.
+                // See https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
+                if dcid.to_vec().len() < 8 {
+                    return Err(Error::InvalidTransportParam);
```
normanmaurer: Not sure this is the best Error value to use...

normanmaurer:
Just to make it clear.... I basically (as a library user) would like to be able to control how this random value is generated for the DCID.

normanmaurer:
@LPardue @ghedo would be possible to add ?

normanmaurer:
ping again...

LPardue:
As noted in https://github.com/cloudflare/quiche/pull/2234#issuecomment-3495528844, I think this need to be put behind a feature flag. I'm not willing to make this the default behaviour and risk clients accidentally using the API incorrectly

normanmaurer:
@LPardue that's fair enough let me make the change later today

normanmaurer:
@LPardue I added the feature flag as requested... PTAL

quiche/src/lib.rs
```diff
@@ -1974,17 +2020,33 @@ impl Default for QlogInfo {
...
+                }
+            }
+            #[cfg(not(feature = "connect-with-dcid"))]
+            // Just error out in case we did not compile with feature enabled to
+            // guard against miss-use in the future.
+            if dcid.is_some() {
+                return Err(Error::InvalidTransportParam);
```
normanmaurer: we could also panic in this case if this would be better ?

quiche/src/lib.rs
toidiu: I think it would be better to put this validation in the new `connect_with_dcid*` functions since that is where the new logic is being added. It will also make this common function less busy.

quiche/src/lib.rs
toidiu: If we get here then its a library error so maybe we can just do a `debug_assert`. panic is not a great experience for users.

quiche/src/lib.rs
```diff
@@ -1974,17 +2020,33 @@ impl Default for QlogInfo {
...
         let tls = config.tls_ctx.new_handshake()?;
-        Connection::with_tls(scid, odcid, local, peer, config, tls, is_server)
+        Connection::with_tls(scid, dcid, local, peer, config, tls, is_server)
     }
 
     fn with_tls(
-        scid: &ConnectionId, odcid: Option<&ConnectionId>, local: SocketAddr,
+        scid: &ConnectionId, dcid: Option<&ConnectionId>, local: SocketAddr,
         peer: SocketAddr, config: &Config, tls: tls::Handshake, is_server: bool,
     ) -> Result<Connection<F>> {
```
toidiu, line 1891: The Option `odcid` is currently being used to detect if the connection is a Server or Client. i.e. if odcid is Some then implement some server specific logic. 

I dont love that we are using Option for this and now that the client can also provide a dcid, we now need a bunch of `if server` logic, which makes things brittle and harder to maintain.

Given that this is an internal function, I think to make this less error prone we should introduce a new parameter, `client_dcid`, here instead of reusing `odcid`. 

```
  fn with_tls(
        scid: &ConnectionId,
        odcid: Option<&ConnectionId>,
        client_dcid: Option<&ConnectionId>,
        local: SocketAddr,
        peer: SocketAddr,
        config: &Config,
        tls: tls::Handshake,
        is_server: bool,
    ) -> Result<Connection<F>>
```

quiche/Cargo.toml
```diff
@@ -30,6 +30,9 @@ rust-version = "1.83"
 [features]
 default = ["boringssl-vendored"]
 
+# Allow to provide generated dcid and connect with it.
+connect-with-dcid = []
```
toidiu: nit: what do you think about

```suggestion
# Allow client connections to provide a custom dcid.
custom-client-dcid = []
```

quiche/src/lib.rs
```diff
@@ -1771,6 +1771,29 @@ pub fn connect(
...
+/// Be aware the RFC places requirements for unpredictability and length
+/// on the client DCID field.
+/// [`RFC9000`]:  https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
+///
+/// The `scid` parameter is used as the connection's source connection ID,
+/// while the optional `server_name` parameter is used to verify the peer's
+/// certificate.
+#[cfg(feature = "connect-with-dcid")]
+#[cfg_attr(docsrs, doc(cfg(feature = "connect-with-dcid")))]
+pub fn connect_with_dcid(
```
toidiu, line 1628: It would be nice to also include a test for this. Although not sure if we currently have a great way of enable features in tests.

quiche/src/lib.rs
normanmaurer: @toidiu I did put it here to have the same safe-guards when using quiche via the C api (I use quiche via the C API) as this function is directly used from there. Otherwise I would have had to duplicate this which I think is more error-prone. WDYT ?

quiche/src/lib.rs
normanmaurer: sure that is fine with me as well... let me do this.

quiche/src/lib.rs
normanmaurer: @toidiu I am fine with this but then I will also need to add another FFI / c api to use it. I did aim to keep the changes minimal and that's why I did it this way. 

Basically I only found out that I need changes when I tried to pass in the dcid via the C api and it just didn't work. So I added the changes here to make it work without any changes in the FFI / c layer. 

Just let me know what you guys think should be done and I am happy to do so

quiche/src/lib.rs
normanmaurer: I didn't see how :( I did validate it via our integration in netty tho.

quiche/src/lib.rs
toidiu: Ahh that makes sense. Guessing that would be this [function](https://github.com/cloudflare/quiche/blob/master/quiche/src/ffi.rs#L608). I am less familiar with where ffi is used and if a breaking change is acceptable so I'll let @LPardue speak more on that.

From the Rust and lib.rs logic maintenance perspective, I would love to not mix `dcid` and `odcid`.

Another option could be to add a new function to the ffi layer.. something like `quiche_conn_new_with_client_dcid`.

quiche/src/lib.rs
LPardue: I think this is possible. There's a few examples of tests only being run behind certain features in quiche/src/tests.rs, e.g., https://github.com/cloudflare/quiche/blob/master/quiche/src/tests.rs#L7545

Then, we can add another CI test step. For example something like

add a new env in https://github.com/cloudflare/quiche/blob/master/.github/workflows/stable.yml#L9
```
env:
  DEFAULT_OPTIONS: "--features=async,ffi,qlog --workspace"
  NO_BORING_OPTIONS: "--features=ffi,qlog --workspace --exclude h3i --exclude tokio-quiche"
  CONNECT_WITH_DCID_OPTIONS: "--features=ffi,connect-with-dcid --workspace --exclude h3i --exclude tokio-quiche"
```

then add a new step around https://github.com/cloudflare/quiche/blob/master/.github/workflows/stable.yml#L85

```
- name: Run cargo test for connect-with-dcd
        if: ${{ matrix.tls-feature == 'boringssl-boring-crate' }}
        run: cargo test --verbose --all-targets --features=${{ matrix.tls-feature }} ${{ env.CONNECT_WITH_DCID_OPTIONS }}
```

quiche/src/lib.rs
LPardue: I'm a bit confused, are you not using [quiche_connect()](https://github.com/cloudflare/quiche/blob/master/quiche/src/ffi.rs#L533)?

normanmaurer: @LPardue no I don't do... I use `quiche_conn_new_with_tls` as I need to use my own existing `SSL` pointer: 
https://github.com/cloudflare/quiche/blob/master/quiche/src/ffi.rs#L608C19-L608C43

LPardue: But yeah I think I agree with @toidu that we can probably tweak the internal `with_tls` and then wrap it by add something in the FFI layer like below where returning null means something failed

```
pub extern "C" fn quiche_conn_new_with_tls_and_client_dcid(
    scid: *const u8, scid_len: size_t, dcid: *const u8, dcid_len: size_t,
    local: &sockaddr, local_len: socklen_t, peer: &sockaddr, peer_len: socklen_t,
    config: &Config, ssl: *mut c_void,
) -> *mut Connection {
  #[cfg(not(feature = "connect-with-dcid"))]
  {
    return ptr::null_mut();
  }
  
  let scid = unsafe { slice::from_raw_parts(scid, scid_len) };
  let scid = ConnectionId::from_ref(scid);

  let dcid = if !dcid.is_null() && dcid_len >= 8 {
        Some(ConnectionId::from_ref(unsafe {
            slice::from_raw_parts(dcid, dcid_len)
        }))
    } else {
        return ptr::null_mut();
    };

    let local = std_addr_from_c(local, local_len);
    let peer = std_addr_from_c(peer, peer_len);

    let tls = unsafe { tls::Handshake::from_ptr(ssl) };

    match Connection::with_tls(
        &scid,
       None, // odcid
        dcid.as_ref(),
        local,
        peer,
        config,
        tls,
        false, // only ever used for client connections
    ) {
        Ok(c) => Box::into_raw(Box::new(c)),

        Err(_) => ptr::null_mut(),
    }
}
```

quiche/src/lib.rs
LPardue: regarding error value, I would define a new one like `InvalidDcidInitialization`. This would be especially useful for implementing the suggestion to add a new `client_dcid: Option<&ConnectionId>,` param to the function. We can check that only one of `odcid` or `client_dcd` are Some and return that error if some expectation failed.

I'd prefer that level of robustness vs. just relying on debug_assert

quiche/Cargo.toml
LPardue: I kinda prefer this name

normanmaurer: @LPardue you prefer the suggested one correct ? I also like it more... Just wanted to check before starting to write the code

LPardue: Yes. lets go with this suggested one. Also while I'm here nitpicking, any comment that refers to DCID should be capitalized (there's a few of those through the PR. The feautre/functions themself obviously needs to be lowercase

quiche/src/lib.rs
normanmaurer: done

quiche/src/tests.rs
```diff
@@ -10393,3 +10394,71 @@ fn configuration_values_are_limited_to_max_varint() {
     // do not panic because of too large values that we try to encode via varint.
     assert_eq!(pipe.handshake(), Err(Error::InvalidTransportParam));
 }
+#[cfg(feature = "custom-client-dcid")]
+#[rstest]
+fn connect_custom_client_dcid(
+    #[values(true, false)] too_short: bool,
```
normanmaurer: @LPardue added the test here

quiche/src/lib.rs
normanmaurer: I didn't the following in the ffi layer as it produced warnings around unused params :/. Instead I just let it call `with_tls` which will fail. Let me know what you think @LPardue 

```
 #[cfg(not(feature = "connect-with-dcid"))]
  {
    return ptr::null_mut();
  }
```

normanmaurer:
@LPardue @toidiu this is now ready to review again.. sorry it took me longer then I expected due other priorities  :/

quiche/src/lib.rs
```diff
@@ -1989,17 +2026,42 @@ impl Default for QlogInfo {
...
+        config: &Config, tls: tls::Handshake, is_server: bool,
     ) -> Result<Connection<F>> {
+        #[cfg(feature = "custom-client-dcid")]
+        if let Some(client_dcid) = client_dcid {
+            // The Minimum length is 8.
+            // See https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
+            if client_dcid.to_vec().len() < 8 {
+                return Err(Error::InvalidDcidInitialization);
+            }
+        }
```
toidiu, line 1904: Thx for linking the RFC +1

quiche/src/lib.rs
```diff
@@ -1989,17 +2026,42 @@ impl Default for QlogInfo {
...
+        #[cfg(not(feature = "custom-client-dcid"))]
+        if client_dcid.is_some() {
+            return Err(Error::InvalidDcidInitialization);
+        }
```
toidiu, line 1908: The feature gate is useful for gating public access. This gating seems redundant since you are already gating `connect_with_dcid`. Having gating here just makes testing harder imo. Lmk if there is a good usecase for this check that I am missing.

toidiu:
> @LPardue @toidiu this is now ready to review again.. sorry it took me longer then I expected due other priorities :/

Looks good to me @normanmaurer. Nice job and ty!

Had a minor comment https://github.com/cloudflare/quiche/pull/2234#discussion_r2663118672 which I can see going either way. Going to stamp but also get a review from Lucas.

toidiu (approved):

quiche/src/lib.rs
normanmaurer: The gating still makes sense as this method is directly called from FFI / c:

https://github.com/cloudflare/quiche/pull/2234/changes/BASE..8c079dd48b408c0d0c9165f48a03d4c458c4c287#diff-0d38e6104a2bf7f84352c702a59efeb4eeab78f6114188cfafa82cf70cdc001aR631

normanmaurer:
@LPardue can I have another review please ?

normanmaurer:
@LPardue ping...

quiche/Cargo.toml
```diff
@@ -25,11 +25,14 @@ include = [
   "/quiche.svg",
   "/src",
 ]
-rust-version = "1.85"
+rust-version = "1.83"
 
 [features]
 default = ["boringssl-vendored"]
 
+# Allow client connections to provide a custom DCID.
```
LPardue: I'd like to put more warning signs here.

```suggestion
# Allow client connections to provide a custom DCID when initiating a
# connection. Be aware that RFC 9000 places requirements for unpredictability and
# length on the client DCID field. Enabling this feature can be dangerous if these
# requirements are not satisfied.
```

quiche/src/tests.rs
```diff
@@ -10410,3 +10393,69 @@ fn configuration_values_are_limited_to_max_varint() {
     // do not panic because of too large values that we try to encode via varint.
     assert_eq!(pipe.handshake(), Err(Error::InvalidTransportParam));
 }
+#[cfg(feature = "custom-client-dcid")]
+#[rstest]
+fn connect_custom_client_dcid(#[values(true, false)] too_short: bool) {
```
LPardue: Thanks for adding the paramaterization but I think it would make the test code clearer to split these tests out manually e.g. connect_custom_client_dcid_too_short and connect_custom_client_dcid. The too_short case doesn't need any server settings so will be a very succinct test.

quiche/src/tests.rs
```diff
@@ -10410,3 +10393,69 @@ fn configuration_values_are_limited_to_max_varint() {
...
+    let server_addr = "127.0.0.1:4321".parse().unwrap();
+
+    // 8 is the minimum required.
+    let mut client_dcid = [0; 8];
+    rand::rand_bytes(&mut client_dcid[..]);
+    let client_dcid = if too_short {
+        // Just use something which is smaller than 8 (which is the minimum
+        ConnectionId::from_ref(&client_dcid[0..6])
+    } else {
+        ConnectionId::from_ref(&client_dcid)
```
LPardue: Assuming we split the tests up as  suggested in the previous comment, I would suggest we use different values for DCID and SCID and an example of "good practice", rather than use the same value for both.

quiche/src/tests.rs
```diff
@@ -10410,3 +10393,69 @@ fn configuration_values_are_limited_to_max_varint() {
...
+            server: accept(
+                &server_scid,
+                None,
+                server_addr,
+                client_addr,
+                &mut server_config,
+            )
+            .unwrap(),
+        };
+        assert_eq!(pipe.handshake(), Ok(()));
```
LPardue: Rather than _just_ confirm the handshake succeeded, I think we need to explicitly check the client Initial contains the desired DCID. This will help avoid any future surprises if there is some kind of regression. You can look at the handshake_confirmation() test for an example of doing the handshake more step-by-step. In the client's first flight, you can check the packet header via `quiche::Header::from_slice` . We should still test that the handshake completes, as in handshake_confirmation.

quiche/src/lib.rs
```diff
@@ -1777,7 +1766,31 @@ pub fn connect(
...
+}
+
+/// Creates a new client-side connection using the given DCID initially.
+/// Be aware the RFC places requirements for unpredictability and length
+/// on the client DCID field.
+/// [`RFC9000`]:  https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
+///
+/// The `scid` parameter is used as the connection's source connection ID,
+/// while the optional `server_name` parameter is used to verify the peer's
+/// certificate.
```
LPardue, line 1623:
```suggestion
/// Creates a new client-side connection using the given DCID initially.
///
/// Be aware that [RFC 9000] places requirements for unpredictability and length
/// on the client DCID field. This function is dangerous if these  requirements
/// are not satisfied.
///
/// The `scid` parameter is used as the connection's source connection ID, while
/// the optional `server_name` parameter is used to verify the peer's
/// certificate.
///
/// [RFC 9000]: <https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3>
```

quiche/src/ffi.rs
```diff
@@ -604,6 +604,48 @@ pub extern "C" fn quiche_retry(
...
+            None
+        };
+
+        let local = std_addr_from_c(local, local_len);
+        let peer = std_addr_from_c(peer, peer_len);
+
+        let tls = unsafe { tls::Handshake::from_ptr(ssl) };
+
+        // This will return an error if we did not compile with
+        // --custom-client-dcid.
```
LPardue: nit: rather than rely on with_tls to do the work here, I'd rather have two versions of quiche_conn_new_with_tls_and_client_dcid based on the feature flag. E.g.

add 
```
#[cfg(not(feature = "custom-client-dcid"))]
pub extern "C" fn quiche_conn_new_with_tls_and_client_dcid(
    scid: *const u8, scid_len: size_t, dcid: *const u8, dcid_len: size_t,
    local: &sockaddr, local_len: socklen_t, peer: &sockaddr, peer_len: socklen_t,
    config: &Config, ssl: *mut c_void,
) -> *mut Connection {
  // It's always an errror to call this function without the custom-client-dcid feature enabled.
  ptr::null_mut()

```

and change this function to require #[cfg(feature = "custom-client-dcid")]

quiche/include/quiche.h
```diff
@@ -323,6 +326,12 @@ quiche_conn *quiche_conn_new_with_tls(const uint8_t *scid, size_t scid_len,
                                       const quiche_config *config, void *ssl,
                                       bool is_server);
 
+quiche_conn *quiche_conn_new_with_tls_and_client_dcid(const uint8_t *scid, size_t scid_len,
```
LPardue, line 329: We should add a short comment here that it requires the "custom-client-dcid" to be enabled, otherwise returns NULL

LPardue (commented):
Mostly minor ergonomics feedback

quiche/src/tests.rs
normanmaurer: we already use different values... Or I am missing something ?

normanmaurer:
@LPardue I think I addressed everything... PTAL

LPardue:
Thanks Norman. Unfortunately, it seems like some other changes to this area are now conflicting. If it's not a trivial conflict to resolve, I can find some time to analyse and propose a path forward

normanmaurer:
@LPardue let me try to fix the conflict... Is it ok to force-push as this might be easier ?

LPardue:
Yep, I'll review the whole thing top to bottom once it's in a good state so dont worry about commit lineage

normanmaurer:
@LPardue ready for review...

quiche/src/lib.rs
LPardue: yeah that's fine

quiche/src/lib.rs
```diff
@@ -1817,20 +1867,41 @@ impl Default for QlogInfo {
...
+            is_server,
         )
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn with_tls(
         scid: &ConnectionId, retry_cids: Option<RetryConnectionIds>,
-        local: SocketAddr, peer: SocketAddr, config: &Config,
-        tls: tls::Handshake, is_server: bool,
+        client_dcid: Option<&ConnectionId>, local: SocketAddr, peer: SocketAddr,
```
LPardue, line 1889: Now that we have the `retry_cids` parameter, should it be mutually exclusive with `client_dcid`?  In other words, it is not valid for both to be `Some` and the function could error with `InvalidDcidInitialization` if that happened.

 I know that this is currently enforced through the layers but adding the check in here might help avoid some footgun in future?

normanmaurer: done...

PR #2237, opened 2025-11-07
tokio-quiche: Add OutboundFrame::Trailers

jrouviere:
This allows sending HTTP trailers to the peer. The trailers is an additional headers with the trailers and fin flags set.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -779,16 +779,7 @@ impl<H: DriverHooks> H3Driver<H> {
                     Err(h3::Error::StreamBlocked)
                 } else {
                     if *fin {
-                        ctx.fin_or_reset_sent = true;
-                        audit_stats
-                            .set_sent_stream_fin(StreamClosureKind::Explicit);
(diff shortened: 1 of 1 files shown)


tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -792,6 +794,29 @@ impl<H: DriverHooks> H3Driver<H> {
...
 
+            OutboundFrame::Trailers(headers, priority) => {
+                let prio = priority.as_ref().unwrap_or(&DEFAULT_PRIO);
+
+                let res = conn.send_additional_headers_with_priority(
+                    qconn, stream_id, headers, prio, true, true,
+                );
+
+                // trailers always set fin=true
+                // here we run the same logic as OutboundFrame::Body with fin=true
```
LPardue: I think it would be nice to put some of this in a function called in both places. We then don't run the risk of the behaviour diverging

jrouviere: Yes, good idea, I've created a `on_fin_sent` helper and updated the PR.

evanrittenhouse (approved):

PR #2238, opened 2025-11-07
feat: add connection flow control and stream to stats

gregor-cf:
These stats would be really helpful for debugging issues we're seeing in prod.

### DIFF
--- a/datagram-socket/src/socket_stats.rs
+++ b/datagram-socket/src/socket_stats.rs
@@ -64,9 +64,32 @@ pub struct SocketStats {
     pub bytes_retrans: u64,
     pub bytes_unsent: u64,
     pub delivery_rate: u64,
+    pub pacing_rate_bytes_per_sec: u64,
     pub max_bandwidth: Option<u64>,
     pub startup_exit: Option<StartupExit>,
     pub bytes_in_flight_duration_us: u64,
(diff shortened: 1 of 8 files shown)


LPardue:
This seems like it is conflating two different use cases, one for general monitoring/observability of however a connection was used, another for helping applications make runtime choices.  

A good rule of thumb here might be that any number that only increases goes in stats, everything else can be a direct getter

gregor-cf:
> This seems like it is conflating two different use cases, one for general monitoring/observability of however a connection was used, another for helping applications make runtime choices.
> 
> A good rule of thumb here might be that any number that only increases goes in stats, everything else can be a direct getter

Is your concern with just the flow control windows in quiche::Stats? Instead of putting the window in the stats, I could also use `tx_data, max_tx_data, rx_data, max_rx_data` separately, then we'd have monotonically increasing values again. Alternatively, I could also remove the flow control windows from `quiche::Stats` but leave them in tokio_quiche's `QuicheConnectionStats / AsSocketStats`. 

(FWIW, we already have the congestion window in `quiche::PathStats` (and also in `AsSocketStats`). 

I need these stats in `AsSocketStats` (in some form), since we have plumbing in oxy and friends to query these stats, but don't have any plumbing to directly query the quiche connection itself.

PR #2246, opened 2025-11-11
tokio-quiche: release 0.11.1

jrouviere:


jrouviere:
Actually I think a patch release won't do since we are adding a new enum value. I'll recreate.

PR #2248, opened 2025-11-11
Return task ID from spawn_with_killswitch

scotchmist:
As part of the change to run Tokio task hooks in Oxy and FL2, we also need the task id within Oxy to map between tasks and the associated timing data. Therefore we need to have spawn_with_killswitch return the task_id from the spawned task.

### DIFF
--- a/task-killswitch/src/lib.rs
+++ b/task-killswitch/src/lib.rs
@@ -29,6 +29,7 @@ use parking_lot::Mutex;
 use tokio::sync::watch;
 use tokio::task;
 use tokio::task::AbortHandle;
+use tokio::task::Id;
 
 use std::future::Future;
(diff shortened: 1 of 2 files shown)


task-killswitch/src/lib.rs
```diff
@@ -28,7 +28,7 @@ use dashmap::DashMap;
 use parking_lot::Mutex;
 use tokio::sync::watch;
 use tokio::task;
-use tokio::task::AbortHandle;
+use tokio::task::{AbortHandle, Id};
```
LPardue: The CI formatter is complaining here

```suggestion
use tokio::task::AbortHandle;
use tokio::task::Id;
```

LPardue (approved):

jrouviere (approved):

PR #2249, opened 2025-11-11
Fix app_limited computation for the gcongestion BBR implementation

antoniovicente:
BBR is app-limited when there are not enough bytes in the tx buffer to use up the current cwnd. Simply paying attention to cwnd_available is not enough due to pacing.

### DIFF
--- a/quiche/src/tests.rs
+++ b/quiche/src/tests.rs
@@ -5015,14 +5015,15 @@ fn app_limited_false(
 
     // Server sends stream data bigger than cwnd.
     let send_buf1 = [0; 20000];
-    assert_eq!(
-        pipe.server.stream_send(0, &send_buf1, false),
-        if cc_algorithm_name == "cubic" {
-            Ok(12000)
-        } else {
-            Ok(13879)
-        }
-    );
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -4969,9 +4969,14 @@ impl<F: BufFactory> Connection<F> {
...
+        let sent_pkt_has_data = if path.recovery.gcongestion_enabled() {
+            has_data || dgram_emitted
+        } else {
+            has_data
+        };
+
+        if sent_pkt_has_data &&
+            cwnd_available > self.tx_buffered + frame::MAX_STREAM_OVERHEAD
         {
             path.recovery.on_app_limited();
```
antoniovicente, line 4981: Note, `on_app_limited` is only implemented by the gcongestion branch so functional changes in this PR are exclusive to gcongestion

https://github.com/cloudflare/quiche/blob/948f7fc/quiche/src/recovery/congestion/recovery.rs#L973

antoniovicente:
I think there are some errors in my logic still, sending back to draft.

antoniovicente:
Abandoning this version.  Should have a better fix in the next week.

PR #2251, opened 2025-11-13
fix: send MAX_STREAMS when available <= 50% of initial for long lived connections

ehaydenr:
The current logic sends MAX_STREAMS too aggressively. As the next MAX_STREAMS value increases, the condition for whether or not to send MAX_STREAMS becomes unconditionally true. This change fixes that so that MAX_STREAMS is only sent when 50% or more streams have completed.


quiche/src/stream/mod.rs
```diff
@@ -654,18 +662,28 @@ impl<F: BufFactory> StreamMap<F> {
...
+    ///
+    /// This only sends MAX_STREAMS when available capacity is at or below 50%
+    /// of the initial maximum streams target.
     pub fn should_update_max_streams_bidi(&self) -> bool {
+        let available = self
+            .local_max_streams_bidi
+            .saturating_sub(self.peer_opened_streams_bidi);
         self.local_max_streams_bidi_next != self.local_max_streams_bidi &&
-            self.local_max_streams_bidi_next / 2 >
-                self.local_max_streams_bidi - self.peer_opened_streams_bidi
```
antoniovicente, line 668: So the logic before was to send an increase following each stream completion?

Consider an application that requires 9 active requests at all times and the stream limit is 10.  If 2 of the streams close and the client attempts to re-create them, it will only be able to create 1 of the two and remain in this state until 3 other streams close and trigger a max stream update.

So the application would need 2 QUIC connections in order to work reliably in this case even though the number of streams it wants to keep open is less than the per connection stream limit.

ehaydenr: In that case, I would expect "available" to be 1 which is less than 10/2, so we should be sending MAX_STREAMS in that case. I added another test case to try and capture this scenario.

quiche/src/stream/mod.rs
```diff
@@ -135,10 +135,16 @@ pub struct StreamMap<F: BufFactory = DefaultBufFactory> {
     local_max_streams_bidi: u64,
     local_max_streams_bidi_next: u64,
 
+    /// Initial maximum bidirectional stream count
```
ghedo:
```suggestion
    /// Initial maximum bidirectional stream count.
```

quiche/src/stream/mod.rs
```diff
@@ -135,10 +135,16 @@ pub struct StreamMap<F: BufFactory = DefaultBufFactory> {
...
+    initial_max_streams_bidi: u64,
+
     /// Local maximum unidirectional stream count limit.
     local_max_streams_uni: u64,
     local_max_streams_uni_next: u64,
 
+    /// Initial maximum unidirectional stream count
```
ghedo:
```suggestion
    /// Initial maximum unidirectional stream count.
```

ghedo (approved):

quiche/src/stream/mod.rs
antoniovicente: Apologies.  I misread your patch and hadn't had a time to come back to this PR.

Thanks so much for this much needed fix.

PR #2256, opened 2025-11-21
tokio-quiche: Add SCID to retry token HMAC

TheJokr:
Due to tokio-quiche's stateless address validation via RETRY packets, we rely on the client to reflect our chosen SCID back to us. We should make sure that we actually generated that SCID by binding the retry token to the SCID.

I didn't want to duplicate the SCID inside the token, so I refactored the code a bit to only include data that is not available from elsewhere. This means we also don't need to include the client IP address in it anymore.

### DIFF
--- a/tokio-quiche/src/quic/addr_validation_token.rs
+++ b/tokio-quiche/src/quic/addr_validation_token.rs
@@ -25,10 +25,7 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use quiche::ConnectionId;
-use std::io::Write;
-use std::io::{
-    self,
-};
+use std::io;
 use std::net::IpAddr;
 use std::net::SocketAddr;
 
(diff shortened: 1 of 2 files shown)


tokio-quiche/src/quic/router/acceptor.rs
```diff
@@ -236,31 +236,31 @@ where
...
-            if token.is_empty() {
-                return self.stateless_retry(incoming, hdr);
-            }
+                if token.is_empty() {
+                    return self.stateless_retry(incoming, hdr);
+                }
+
+                let scid = hdr.dcid;
+                let original_dcid = self
+                .token_manager
```
TheJokr, line 252: nightly `rustfmt` insists on formatting it this way :/

LPardue:
I think this would cause some issues for the potential future use of NEW_TOKEN frame, so needs some consideration. Possibly just a TODO comment so we dont forget but needs a little bit of thought

https://datatracker.ietf.org/doc/html/rfc9000#section-19.7

TheJokr:
There are some alternatives to fix the issue, this one is just the first one I thought of without re-reading RFC 9000 in detail.

We could just generate a new SCID when receiving a valid retry token. My understanding is that the RFC [explicitly allows this](https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-7). However, quiche doesn't expose an interface to specify [the initial SCID and the retry SCID separately](https://github.com/cloudflare/quiche/blob/aecfd90abc30e0eb29b5028272f1cc1de11dfeb3/quiche/src/lib.rs#L2192-L2199), so I'd have to add that. How do you feel about something like `quiche::accept_retry(scid, rdcid, odcid, local, peer, config)`?

Alternatively, we could also validate the CID using `ConnectionIdGenerator::verify_connection_id`. This isn't as robust as the  retry token (32 byte HMAC vs <=20 bytes shared with other info). It's also limited by whatever format `ConnectionIdGenerator` implements, but may provide sufficient guarantees.

What do you think?

tokio-quiche/src/quic/addr_validation_token.rs
```diff
@@ -54,140 +51,135 @@ impl Default for AddrValidationTokenManager {
...
+        token_buf.extend_from_slice(ip_bytes);
+        token_buf.extend_from_slice(scid);
 
-        token.set_position(0);
-        token.write_all(tag.as_ref()).unwrap();
+        let tag = boring::hash::hmac_sha256(&self.sign_key, &token_buf).unwrap();
+        debug_assert_eq!(tag.len(), HMAC_TAG_LEN);
 
-        token.into_inner()
+        // Drop the non-payload parts of the HMAC and reuse the storage for the
```
ghedo, line 77: If we are dropping the non-CID parts of the buffer anyway I think we can just use the `Hasher::update()` API instead to avoid having to allocate the whole `token_buf` and then truncating it https://docs.rs/boring/latest/boring/hash/struct.Hasher.html#method.update

TheJokr: That's what I wanted to do originally, but boring's Hasher doesn't do HMACs natively. The rust crate only exposes the one-shot API.

TheJokr:
Closing in favor of #2325

PR #2257, opened 2025-11-22
Add headers pending flush to audit stats

antoniovicente:


LPardue:
Looks broadly good. It got me wondering if, while we're touching this area, maybe we should do a little more refactor to ensure the time metric is reported on stream teardown (whatever the cause)? i.e. on destruction if the time is Some e.g.

```
                    if let Some(first) =
                        ctx.first_full_headers_flush_fail_time.take()
                    {
                        ctx.audit_stats.add_header_flush_duration(
                            Instant::now().duration_since(first),
                        );
                    }
```

If so, we'd want to take avoid putting `self.headers_pending_flush.fetch_sub(1, Ordering::SeqCst);` in `add_header_flush_duration()`

tokio-quiche/src/http3/stats.rs
```diff
@@ -69,6 +69,11 @@ pub struct H3AuditStats {
     /// Measured across all HEADERS frames sent on the stream. A value of 0
     /// indicates there was no failed flushing.
     headers_flush_duration: AtomicCell<Duration>,
+    /// Number of stream headers entered the pending flush state but haven't
+    /// exited.
+    ///
+    /// This count includes streams that were reset before headers were flushed.
+    headers_pending_flush: AtomicU64,
```
LPardue: I think this can only ever be a maximum of `1`? While blocked on HEADERS, no other frame will be attempted. If the frame succeeds, the count is decremented, before attempting any subsequent HEADERS. Maybe easier to convert it to a bool and let apps do counting across all streams based on that.

antoniovicente:
> Looks broadly good. It got me wondering if, while we're touching this area, maybe we should do a little more refactor to ensure the time metric is reported on stream teardown (whatever the cause)? i.e. on destruction if the time is Some e.g.
> 
> ```
>                     if let Some(first) =
>                         ctx.first_full_headers_flush_fail_time.take()
>                     {
>                         ctx.audit_stats.add_header_flush_duration(
>                             Instant::now().duration_since(first),
>                         );
>                     }
> ```
> 
> If so, we'd want to take avoid putting `self.headers_pending_flush.fetch_sub(1, Ordering::SeqCst);` in `add_header_flush_duration()`

Reporting a duration on other teardown reasons seems fine to me, but I think we'ld like to also indicate if the flush succeeded or the stream was cancelled.  We do not want to mix together the success and failure cases.

Ultimately we want to detect cases where the connection has data to send but cannot make progress.  The reasons why we may not be able to make progress include flow control, congestion control or other reasons.

LPardue:
Right. The duration is cumulative across all HEADERS, while being actively blocked can happen only on an individual HEADERS. For example, 103 early hints, 200 OK, trailers. The audit stats already capture if the stream was stopped/reset/fin, apps can use that information in combination with your new metric introduced in the PR to figure out the terminal status.

antoniovicente:
I think there's a difference between the case that the headers were sent after some delay and the case headers were never sent after a delay.  Even if the client action is to reset the stream before FIN is sent.

LPardue:
I agree, that's why I think it would be better to switch the metric proposed here to a bool that indicates if HEADERS were blocked, or not, at the time the stream is terminated.

tokio-quiche/src/http3/stats.rs
antoniovicente: Changed to bool.

antoniovicente:
> Looks broadly good. It got me wondering if, while we're touching this area, maybe we should do a little more refactor to ensure the time metric is reported on stream teardown (whatever the cause)? i.e. on destruction if the time is Some e.g.
> 
> ```
>                     if let Some(first) =
>                         ctx.first_full_headers_flush_fail_time.take()
>                     {
>                         ctx.audit_stats.add_header_flush_duration(
>                             Instant::now().duration_since(first),
>                         );
>                     }
> ```
> 
> If so, we'd want to take avoid putting `self.headers_pending_flush.fetch_sub(1, Ordering::SeqCst);` in `add_header_flush_duration()`

Attempted to record blocked time in cases where the stream terminates.  I'll look into some more testing on Monday.

LPardue (approved):

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1258,10 +1258,13 @@ impl<H: DriverHooks> ApplicationOverQuic for H3Driver<H> {
...
 impl<H: DriverHooks> Drop for H3Driver<H> {
     fn drop(&mut self) {
-        for stream in self.stream_map.values() {
+        for stream in self.stream_map.values_mut() {
             stream
                 .audit_stats
                 .set_recvd_stream_fin(StreamClosureKind::Implicit);
+
+            // Update stats if there were pending header sends on this stream.
+            stream.full_headers_flush_failed();
```
antoniovicente: Inspired by https://github.com/cloudflare/quiche/pull/2259 https://github.com/cloudflare/quiche/pull/2259/commits/68fc918274c0433bc737c4ca13db535e2bc292f1

antoniovicente: This change is not correct, moving to draft.

antoniovicente: Fixed.

tokio-quiche/src/http3/stats.rs
```diff
@@ -221,4 +230,27 @@ impl H3AuditStats {
...
+        AtomicCell::<T>::is_lock_free()
+    }
+
+    #[test]
+    fn atomic_cells_are_lock_free() {
+        let stats = H3AuditStats::new(0x07ac0ca707ac0ca7);
+        assert!(is_lock_free(&stats.recvd_stream_fin));
+        assert!(is_lock_free(&stats.sent_stream_fin));
+        // BUG!  AtomicCell uses global lock.
+        assert!(!is_lock_free(&stats.headers_flush_duration));
```
toidiu, line 254: Seems bad? Can you create a ticket to track this and any info if you know what the fix is.

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -92,6 +92,37 @@ impl StreamCtx {
...
+    /// Signal that the headers flush was aborted due to stream being reset.
+    pub(crate) fn full_headers_flush_aborted(&mut self) {
+        self.maybe_update_header_flush_duration(false);
+    }
+
+    fn maybe_update_header_flush_duration(&mut self, flush_successful: bool) {
+        // Clear headers_pending_flush only if header were successfully flushed.
+        if flush_successful {
+            self.audit_stats.set_headers_pending_flush(false);
+        }
```
toidiu: can we move this up to `full_headers_flush_success` since that the only time this logic is invoked

tokio-quiche/src/http3/driver/streams.rs
```diff
@@ -92,6 +92,37 @@ impl StreamCtx {
...
+
+        if let Some(first) = self.first_full_headers_flush_fail_time.take() {
+            self.audit_stats
+                .add_header_flush_duration(Instant::now().duration_since(first));
+        }
```
toidiu, line 123: Why do we update flush duration on abort? Do we also perform the flush action when the connection is aborted?

tokio-quiche/tests/integration_tests/headers.rs
```diff
@@ -109,3 +113,70 @@ async fn test_additional_headers() {
...
+        .await
+        .expect("request failed");
+
+    let mut headers = summary.stream_map.headers_on_stream(0).into_iter();
+    assert!(headers.next().is_none());
+
+    let audit_stats = audit_stats.read().unwrap().clone().unwrap();
+    assert!(audit_stats.headers_pending_flush());
+    // Verify that headers_flush_duration is updated on connection drop.
+    assert!(audit_stats.headers_flush_duration() > Duration::ZERO);
```
toidiu, line 244: If I understand correctly this is coming from `full_headers_flush_aborted`? Is it also possible to test `full_headers_flush_success`?

antoniovicente: Testing this was hard enough and don't have the time to look into h3i changes required to test that case too.  I can file a ticket to followup, but I can't add those other tests in this PR due to time constraints.

tokio-quiche/src/http3/driver/streams.rs
antoniovicente: Because we'ld like to record that the slow operation happened.  Before this PR if we attempted to send a header but never did even after minutes or hours, and then reset/close the connection the event would never get recorded.

We could consider recording success and failures separately.

tokio-quiche/src/http3/driver/streams.rs
antoniovicente: I really rather not since it splits the logic in the helper.

tokio-quiche/src/http3/stats.rs
antoniovicente: The fix is to replace it with a dedicated RWLock.

A deeper fix would be to use a type that doesn't compile if the atomic operation can't be lock free.

tokio-quiche/src/http3/driver/streams.rs
toidiu: Got it. Lets keep the same metric for now

tokio-quiche/src/http3/driver/streams.rs
toidiu: But this line only ever called from the success case and never from the abort case so its just confusing.

toidiu (approved):

tokio-quiche/src/http3/driver/streams.rs
LPardue: Apps can determine the stream status (finished, reset) and combine this counter with that information to handle the metrics they might want. Makes it easier than trying to accomodate every eventuality in this layer.

antoniovicente: I think we only track today time that the connection spent waiting and if the most recent wait resulted in a successful flush.

It is not possible to distinguish between the case of a single header that blocks until the end of the connection, and the case of several header send attempts that eventually succeeded followed by an attempt that blocked until the end of the connection.

LPardue: Its a per-stream context, each one is track individually.

antoniovicente: oops, you're right.

tokio-quiche/src/http3/driver/streams.rs
antoniovicente: done

tokio-quiche/tests/integration_tests/headers.rs
```diff
@@ -109,3 +113,70 @@ async fn test_additional_headers() {
...
+    let summary = h3i_fixtures::request(&url, 1)
+        .await
+        .expect("request failed");
+
+    let mut headers = summary.stream_map.headers_on_stream(0).into_iter();
+    assert!(headers.next().is_none());
+
+    let audit_stats = audit_stats.read().unwrap().clone().unwrap();
+    assert!(audit_stats.headers_pending_flush());
+    // Verify that headers_flush_duration is updated on connection drop.
```
antoniovicente, line 243: The assertion in the next line is flaky according to CI; failed in mac and window builds.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1266,10 +1258,13 @@ impl<H: DriverHooks> ApplicationOverQuic for H3Driver<H> {
...
 impl<H: DriverHooks> Drop for H3Driver<H> {
     fn drop(&mut self) {
-        for stream in self.stream_map.values() {
+        for stream in self.stream_map.values_mut() {
             stream
                 .audit_stats
                 .set_recvd_stream_fin(StreamClosureKind::Implicit);
+
+            // Update stats if there were pending header sends on this stream.
+            stream.full_headers_flush_aborted();
```
antoniovicente, line 1267: The fix for flaky test could be to move the contents of this function to `IoWorker::close_connection` and wait until `metrics.connections_in_memory` is 0 before making assertions in the test.

tokio-quiche/tests/integration_tests/headers.rs
antoniovicente: Adding a sleep doesn't seem to mitigate the flakiness. I need to debug further.

PR #2264, opened 2025-12-04
qlog-dancer: Connection overview chart improvements

LPardue:
The connection overview attempts to give a high-level view of events related
to connection data, stream data, congestion control, and RTT.

Previously, the top panel combined all connection and stream events, making
it difficult to assess when send and write operations are large and/or
asymmetric.

This change splits out thecombined  plot into separate send and receive
plots. In addition, the send plot gets two new series: cumulative buffer
write, and cumulative buffer drop.


antoniovicente (approved):
Thanks for the visualization improvements.

PR #2267, opened 2025-12-04
feat[tokio-quiche]: implement server-side support for 0-RTT

toidiu:
This PR adds server support for 0-RTT. This is done by transitioning the Handshake to Application when we have 0-RTT keys. The PR also adds an extension header IsInEarlyData to forward the early data request status so application can detect when a request was received with 0-RTT keys (and take appropriate security precautions).

### DIFF
--- a/h3i/src/client/sync_client.rs
+++ b/h3i/src/client/sync_client.rs
@@ -105,6 +105,10 @@ fn create_config(args: &Config, should_log_keys: bool) -> quiche::Config {
     config.set_max_stream_window(args.max_stream_window);
     config.grease(false);
 
+    if args.enable_early_data {
+        config.enable_early_data();
+    }
+
     if args.enable_dgram {
         config.enable_dgram(
             true,
@@ -131,6 +135,16 @@ fn create_config(args: &Config, should_log_keys: bool) -> quiche::Config {
 /// Returns a [ConnectionSummary] on success, [ClientError] on failure.
 pub fn connect(
     args: Config, actions: Vec<Action>,
+    close_trigger_frames: Option<CloseTriggerFrames>,
+) -> std::result::Result<ConnectionSummary, ClientError> {
+    connect_with_early_data(args, None, actions, close_trigger_frames)
+}
+
+/// Connect to a server and execute provided early_action and actions.
+///
+/// See `connect` for additional documentation.
+pub fn connect_with_early_data(
+    args: Config, early_actions: Option<Vec<Action>>, actions: Vec<Action>,
     close_trigger_frames: Option<CloseTriggerFrames>,
 ) -> std::result::Result<ConnectionSummary, ClientError> {
     let mut buf = [0; 65535];
@@ -178,10 +192,15 @@ pub fn connect(
         return Err(ClientError::Other("invalid socket".to_string()));
     };
 
-    // Create a QUIC connection and initiate handshake.
+    // Create a new client-side QUIC connection.
     let mut conn =
         quiche::connect(connect_url, &scid, local_addr, peer_addr, &mut config)
             .map_err(|e| ClientError::Other(e.to_string()))?;
+
+    if let Some(session) = &args.session {
+        conn.set_session(session)
+            .map_err(|error| ClientError::Other(error.to_string()))?;
+    }
 
     if let Some(keylog) = &mut keylog {
         if let Ok(keylog) = keylog.try_clone() {
@@ -195,7 +214,29 @@ pub fn connect(
 
     let mut app_proto_selected = false;
 
+    // Send ClientHello and initiate the handshake.
     let (write, send_info) = conn.send(&mut out).expect("initial send failed");
+
+    let mut client = SyncClient::new(close_trigger_frames);
+    // Send early data if connection is_in_early_data (resumption with 0-RTT was
+    // successful) and if we have early_actions.
+    if conn.is_in_early_data() {
+        if let Some(early_actions) = early_actions {
+            let mut early_action_iter = early_actions.iter();
+            let mut wait_duration = None;
+            let mut wait_instant = None;
+            let mut waiting_for = WaitingFor::default();
+
+            check_duration_and_do_actions(
+                &mut wait_duration,
+                &mut wait_instant,
+                &mut early_action_iter,
+                &mut conn,
+                &mut waiting_for,
+                client.stream_parsers_mut(),
+            );
+        }
+    }
 
     while let Err(e) = socket.send_to(&out[..write], send_info.to) {
         if e.kind() == std::io::ErrorKind::WouldBlock {
@@ -216,7 +257,6 @@ pub fn connect(
     let mut wait_duration = None;
     let mut wait_instant = None;
 
-    let mut client = SyncClient::new(close_trigger_frames);
     let mut waiting_for = WaitingFor::default();
 
     loop {
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -7072,6 +7072,17 @@ impl<F: BufFactory> Connection<F> {
         self.handshake.is_in_early_data()
     }
 
+    /// Returns the early data reason for the connection.
+    ///
+    /// This status can be useful for logging and debugging. See [BoringSSL]
+    /// documentation for a definition of the reasons.
+    ///
+    /// [BoringSSL]: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#ssl_early_data_reason_t
+    #[inline]
+    pub fn early_data_reason(&self) -> u32 {
+        self.handshake.early_data_reason()
+    }
+
     /// Returns whether there is stream or DATAGRAM data available to read.
     #[inline]
     pub fn is_readable(&self) -> bool {
--- a/quiche/src/tls/boringssl.rs
+++ b/quiche/src/tls/boringssl.rs
@@ -274,6 +274,12 @@ impl Handshake {
     pub fn is_in_early_data(&self) -> bool {
         unsafe { SSL_in_early_data(self.as_ptr()) == 1 }
     }
+
+    pub fn early_data_reason(&self) -> u32 {
+        let reuse_reason_status =
+            unsafe { SSL_get_early_data_reason(self.as_ptr()) };
+        reuse_reason_status.0
+    }
 }
 
 pub(super) fn get_session_bytes(session: *mut SSL_SESSION) -> Result<Vec<u8>> {
@@ -293,6 +299,10 @@ pub(super) fn get_session_bytes(session: *mut SSL_SESSION) -> Result<Vec<u8>> {
 }
 pub(super) const TLS_ERROR: c_int = 3;
 
+#[allow(non_camel_case_types)]
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub struct ssl_early_data_reason_t(pub ::std::os::raw::c_uint);
 extern "C" {
     // SSL_METHOD specific for boringssl.
     pub(super) fn SSL_CTX_set_tlsext_ticket_keys(
@@ -340,6 +350,8 @@ extern "C" {
     fn SSL_reset_early_data_reject(ssl: *mut SSL);
 
     fn SSL_in_early_data(ssl: *const SSL) -> c_int;
+
+    fn SSL_get_early_data_reason(ssl: *const SSL) -> ssl_early_data_reason_t;
 
     fn SSL_SESSION_to_bytes(
         session: *const SSL_SESSION, out: *mut *mut u8, out_len: *mut usize,
--- a/tokio-quiche/src/http3/driver/server.rs
+++ b/tokio-quiche/src/http3/driver/server.rs
@@ -78,6 +78,24 @@ impl Deref for RawPriorityValue {
     }
 }
 
+/// The request was received during early data (0-RTT).
+#[derive(Clone, Debug)]
+pub struct IsInEarlyData(bool);
+
+impl IsInEarlyData {
+    fn new(is_in_early_data: bool) -> Self {
+        IsInEarlyData(is_in_early_data)
+    }
+}
+
+impl Deref for IsInEarlyData {
+    type Target = bool;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 /// Events produced by [ServerH3Driver].
 #[derive(Debug)]
 pub enum ServerH3Event {
@@ -87,15 +105,24 @@ pub enum ServerH3Event {
         incoming_headers: IncomingH3Headers,
         /// The latest PRIORITY_UPDATE frame value, if any.
         priority: Option<RawPriorityValue>,
+        is_in_early_data: IsInEarlyData,
     },
 }
 
 impl From<H3Event> for ServerH3Event {
     fn from(ev: H3Event) -> Self {
         match ev {
-            H3Event::IncomingHeaders(incoming_headers) => Self::Headers {
-                incoming_headers,
-                priority: None,
+            H3Event::IncomingHeaders(incoming_headers) => {
+                // Server `incoming_headers` are exclusively created in
+                // `ServerHooks::handle_request`, which correctly serializes the
+                // RawPriorityValue and IsInEarlyData values.
+                //
+                // See `H3Driver::process_read_event` for implementation details.
+                Self::Headers {
+                    incoming_headers,
+                    priority: None,
+                    is_in_early_data: IsInEarlyData::new(false),
+                }
             },
             _ => Self::Core(ev),
         }
@@ -205,6 +232,7 @@ impl ServerHooks {
             .send(ServerH3Event::Headers {
                 incoming_headers: headers,
                 priority: latest_priority_update,
+                is_in_early_data: IsInEarlyData::new(qconn.is_in_early_data()),
             })
             .map_err(|_| H3ConnectionError::ControllerWentAway)?;
         driver.hooks.requests += 1;
--- a/tokio-quiche/src/quic/io/connection_stage.rs
+++ b/tokio-quiche/src/quic/io/connection_stage.rs
@@ -123,7 +123,9 @@ impl ConnectionStage for Handshake {
         &mut self, qconn: &mut QuicheConnection,
         _ctx: &mut ConnectionStageContext<A>,
     ) -> ControlFlow<QuicResult<()>> {
-        if qconn.is_established() {
+        // Transition to RunningApplication if we have 1-RTT keys (handshake is
+        // complete) or if we have 0-RTT keys (in early data).
+        if qconn.is_established() || qconn.is_in_early_data() {
             ControlFlow::Break(Ok(()))
         } else {
             ControlFlow::Continue(())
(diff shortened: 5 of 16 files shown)


tokio-quiche/src/quic/io/connection_stage.rs
```diff
@@ -122,7 +122,7 @@ impl ConnectionStage for Handshake {
         &mut self, qconn: &mut QuicheConnection,
         _ctx: &mut ConnectionStageContext<A>,
     ) -> ControlFlow<QuicResult<()>> {
-        if qconn.is_established() {
+        if qconn.is_established() || qconn.is_in_early_data() {
```
toidiu, line 128: This is what helps us transition to Application space so we can then handle the early_data request.

h3i/src/client/sync_client.rs
```diff
@@ -210,6 +241,11 @@ pub fn connect(
         return Err(ClientError::Other(format!("send() failed: {e:?}")));
     }
 
+    println!(
```
antoniovicente: stray println! from debugging.

h3i/src/client/sync_client.rs
```diff
@@ -528,6 +564,8 @@ where
 
     // Send actions
     for action in iter {
+        println!("------- sync_client: handle_actions {:?}", action);
```
antoniovicente: stray println!

quiche/src/tls/boringssl.rs
```diff
@@ -274,6 +274,13 @@ impl Handshake {
     pub fn is_in_early_data(&self) -> bool {
         unsafe { SSL_in_early_data(self.as_ptr()) == 1 }
     }
+
+    pub fn early_reason(&self) -> u32 {
+        // unsafe { SSL_in_early_data(self.as_ptr()) == 1 }
```
antoniovicente: commented out code.

quiche/src/lib.rs
```diff
@@ -7129,6 +7129,60 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// // Value 11 is reserved. (It has historically |ssl_early_data_token_binding|.)
+    /// // The client and server ticket age were too far apart.
+    /// ssl_early_data_ticket_age_skew = 12,
+    ///
+    /// // QUIC parameters differ between this connection and the original.
+    /// ssl_early_data_quic_parameter_mismatch = 13,
+    ///
+    /// // The application settings did not match the session.
+    /// ssl_early_data_alps_mismatch = 14,
```
antoniovicente: Would it make sense to create a rust enum to wrap the early data reason and provide more user friendly print semantics?

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -96,6 +115,7 @@ impl From<H3Event> for ServerH3Event {
             H3Event::IncomingHeaders(incoming_headers) => Self::Headers {
                 incoming_headers,
                 priority: None,
+                is_in_early_data: IsInEarlyData::new(false),
```
antoniovicente: Not clear to me why is_in_early_data must be false here.  Can you add a comment explaining why?

tokio-quiche/src/quic/io/connection_stage.rs
antoniovicente: Thinking about your comment about requests after early data having trouble.  Is there something that prevents the handshake from continuing and completing after taking a break to handle early data?

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -824,6 +824,12 @@ where
     pub(crate) async fn run<A: ApplicationOverQuic>(
         mut self, mut qconn: QuicheConnection, mut ctx: ConnectionStageContext<A>,
     ) -> Closing<Tx, M, A> {
+        println!(
```
antoniovicente: stray println!

tokio-quiche/src/quic/io/connection_stage.rs
antoniovicente: Worth adding some logging on handshake completion?

tokio-quiche/src/http3/driver/server.rs
toidiu: Good question and I still need to double check this actually. Will keep this comment open.

It was a placeholder since the test didnt invoke this code path and there wasnt an obvious way to get the early-data status (unlike `handle_request` below).

tokio-quiche/src/quic/io/connection_stage.rs
toidiu: > Thinking about your comment about requests after early data having trouble. Is there something that prevents the handshake from continuing and completing after taking a break to handle early data?

Thats exactly what I need to debug next. My mental model is that it should invoke the work_loop and continue so need to double check that and compare against the no early_data variation.

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -158,8 +195,10 @@ impl ServerHooks {
         // the first one is an actual request. For now ignore any additional
         // HEADERS (e.g. "trailers").
         if driver.stream_map.contains_key(&stream_id) {
+            println!("ABORT -------------- 41 handle_request: receiving multiple headers on stream: {:?}", stream_id);
             return Ok(());
         }
```
toidiu, line 189: Narrowed it down to here. The early data ("early.test.com") is sent on stream: 0 so I need to send the app_request ("test.com") on a different stream.

However, changing the stream id for app_request causes failure ([comment](https://github.com/cloudflare/quiche/pull/2267/files#r2594034178)). Need to investigate why.. possibly need to open the stream first?

tokio-quiche/tests/integration_tests/zero_rtt.rs
```diff
@@ -0,0 +1,252 @@
...
+        ];
+
+        let mut h3i_config = setup_h3i(&url, "test.com");
+        // Provide session to the client to enable resumption.
+        h3i_config.session = Some(nst_data);
+        h3i_config.enable_early_data = true;
+        let summary = helper_connect_with_early_data_and_assert_status_code(
+            h3i_config,
+            Some(early_frame_actions),
+            frame_actions(0),
```
toidiu: This cant be on stream_id 0 since we already sent the early_data on stream 0.

antoniovicente: Just thinking out loud about ways to make this more debugable.

Did the early data have FIN?  Were your second set of frame actions flagged as protocol errors?

Wondering if qlog or wireshark could have told you more.  Or trace logging.

toidiu: https://github.com/cloudflare/quiche/pull/2267#discussion_r2594032669 We silently drop multiple headers here which made it difficult to track down. But also the h3i and tokio-quiche codebase is newish to me and the use of channels makes it hard to track side effects.

Wireshark no. qlog maybe, but prob not because the silent dropping was subtle and I am not use to reading qlog events. trace logging is essentially what I ended up doing with the print statements.

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -87,15 +105,21 @@ pub enum ServerH3Event {
...
-                incoming_headers,
-                priority: None,
+            H3Event::IncomingHeaders(incoming_headers) => {
+                // TODO: parse incoming_headers for "is_in_early_data"
+                // incoming_headers.headers.unwrap_or(&false);
+                Self::Headers {
+                    incoming_headers,
+                    priority: None,
+                    is_in_early_data: IsInEarlyData::new(false),
+                }
```
toidiu, line 125: @LPardue can you help me reason about why this is correct. I see that you defaulted to None for priority.

The issue being that we need a handle to quiche::Connection (which is not available in the `From` interface) to get priority and early_data status. I am wondering if we need a different trait to capture that a q_conn is required.

The conversion is used a few times in https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/http3/driver/mod.rs

LPardue: We never expect to really use this conversion, the only place we generate ServerH3Event::Headers is directly in the `handle_request` hook

tokio-quiche/tests/fixtures/h3i_fixtures.rs
```diff
@@ -60,11 +60,12 @@ pub fn h3i_config(url: &str) -> h3i::config::Config {
         .unwrap()
 }
 
-pub fn default_headers() -> Vec<Header> {
+pub fn default_headers(host: Option<&str>) -> Vec<Header> {
```
LPardue: The default should be sane, setting `None` is not. Rather than changing this, better to make a different method like `default_headers_with_authority` or something

h3i/src/client/sync_client.rs
```diff
@@ -130,7 +134,7 @@ fn create_config(args: &Config, should_log_keys: bool) -> quiche::Config {
 ///
 /// Returns a [ConnectionSummary] on success, [ClientError] on failure.
 pub fn connect(
-    args: Config, actions: Vec<Action>,
+    args: Config, early_actions: Option<Vec<Action>>, actions: Vec<Action>,
```
LPardue, line 147: Early data is going to be the exception. It would seem to me better to have a separate `connect_with_early_actions` or something rather than require changing calls to `connect` everywhere.

tokio-quiche/src/http3/driver/server.rs
toidiu: Left a comment explaining this to avoid stumbling on this in the future. However, it would be nice to assert this at the type level, which I think would require introducing different types for internal and external types.

tokio-quiche/src/http3/driver/server.rs
toidiu: See below: https://github.com/cloudflare/quiche/pull/2267#discussion_r2599916894

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -87,15 +105,24 @@ pub enum ServerH3Event {
...
 
 impl From<H3Event> for ServerH3Event {
     fn from(ev: H3Event) -> Self {
         match ev {
-            H3Event::IncomingHeaders(incoming_headers) => Self::Headers {
-                incoming_headers,
-                priority: None,
+            H3Event::IncomingHeaders(incoming_headers) => {
+                // Requests are exclusively handled by `Self::headers_received`,
+                // which correctly serializers the RawPriorityValue and
```
jannes: typo -> `serializes`

tokio-quiche/tests/integration_tests/zero_rtt.rs
```diff
@@ -0,0 +1,234 @@
...
+
+#[tokio::test]
+async fn handle_0_rtt_request() {
+    let context = Arc::new(Mutex::new(TestContext::default()));
+    let early_stream_id = 0;
+    let stream_id = 4;
+
+    let context_clone = context.clone();
+    let mut quic_settings = QuicSettings::default();
+    quic_settings.enable_early_data = true;
```
jannes, line 69: nit: could be
```
let quic_settings = QuicSettings {
  enable_early_data: true,
  ..Default::default()
};
```

toidiu: Since QuicSettings is marked #[non_exhaustive] the compiler complains when you try to instantiate the struct directly. Which btw is super cool!

`error[E0639]: cannot create non-exhaustive struct using struct expression`

jannes: oh ok. pattern matching is great in Rust indeed :)

tokio-quiche/src/settings/quic.rs
```diff
@@ -59,6 +59,12 @@ pub struct QuicSettings {
     #[serde(default = "QuicSettings::default_dgram_max_queue_len")]
     pub dgram_send_max_queue_len: usize,
 
+    /// Configures whether to enable early data (0-RTT) support.
```
evanrittenhouse:
```suggestion
    /// Configures whether to enable early data (0-RTT) support. Currently only supported for servers.
```

tokio-quiche/src/quic/io/connection_stage.rs
evanrittenhouse: The primary concern I see here is that `work_loop`'s wait phase will not call `wait_for_quiche`, which wakes the IOW when a TLS callback progresses, in the `RunningApplication` phase. It depends solely on the AOQ's `wait_for_data` implementation.`

I'm not super familiar with early data, is it possible that the 0-RTT request/response finishes before the PK callback? If so, we may delay the handshake until the next received packet/Quiche timer fires, rather than when the callback actually completes. From [Boring docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Early-data) it sounds like there's no guarantee about that: 

> As a server, if early data is accepted, [SSL_do_handshake](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_do_handshake) will complete as soon as the ClientHello is processed and server flight sent. [SSL_write](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_write) may be used to send half-RTT data. [SSL_read](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_read) will consume early data and transition to 1-RTT data as appropriate. Prior to the transition, [SSL_in_init](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_in_init) will report the handshake is still in progress. Callers may use it or [SSL_in_early_data](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_in_early_data) to defer or reject requests as needed.

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -78,6 +78,24 @@ impl Deref for RawPriorityValue {
     }
 }
 
+/// Custom header info to track early data requests.
```
evanrittenhouse:
```suggestion
/// [Extensions]-compatible boolean wrapper to track early data requests.
///
/// [Extensions]: https://docs.rs/http/1.3.1/http/request/struct.Request.html#method.extensions
```

tokio-quiche/src/quic/io/connection_stage.rs
toidiu: If I understand correctly: `wait_for_quiche` is used to progress the TLS Handshake callbacks. In `RunningApplication` phase, we do not call the `wait_for_quiche`. The concern is that we might stall the Handshake we transition to RunningApplication in early data (since technically the Handshake is not complete).

We currently use quiche for two callbacks private_key_callback (not called when we resume) and async_select_cert_callback (called when before processing ClientHello). So for our usecase this the Handshake should not stall.

One way to make this clean would be to and another phase Handshake -> ZeroRtt -> RunningApplication. But that doesn't seem necessary at the moment. I'll leave a comment documenting this behavior and something we might need to visit if we decide to use more callbacks.

toidiu:
@evanrittenhouse 👏  helped confirm that we want to record the handshake duration as soon as we resume the early connection.

---

@LPardue @ghedo 

@evanrittenhouse  and I discovered that we set the handshake duration right after resumption with 0-rtt. Wondering if this is ok or if we want to record the duration when the full handshake finishes.

https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/quic/io/worker.rs#L787-L788

tokio-quiche/src/http3/driver/server.rs
toidiu: I don't think we have a concept of header Extensions in quiche and even headers is too specific.

Changed to something more general:
`/// The request was received during early data (0-RTT).`

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -87,15 +105,24 @@ pub enum ServerH3Event {
...
 }
 
 impl From<H3Event> for ServerH3Event {
     fn from(ev: H3Event) -> Self {
         match ev {
-            H3Event::IncomingHeaders(incoming_headers) => Self::Headers {
-                incoming_headers,
-                priority: None,
+            H3Event::IncomingHeaders(incoming_headers) => {
+                // Requests are exclusively handled by `Self::headers_received`,
```
evanrittenhouse:
```suggestion
                // Server `incoming_headers` are exclusively created in `ServerHooks::handle_request`,
```

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -635,8 +635,18 @@ where
...
     ) -> QuicResult<()> {
         if quic_application.should_act() {
+            // Poll the application to make progress.
+            //
+            // Once the connection has been established (i.e. the handshake is
+            // complete), we only poll the application.
+            //
+            // The exception is 0-RTT in TLS 1.3, where the full handshake is
+            // still in progress but we have 0-RTT keys to process early data.
+            // This means TLS callbacks migth only be polled on the next timeout
```
evanrittenhouse: This shouldn't affect `select_cert`/`private_key` callbacks, right? Would be good to note that down since this is purely a concern for future implementation


```suggestion
            // This means TLS callbacks might only be polled on the next timeout
```

tokio-quiche/src/settings/quic.rs
```diff
@@ -59,6 +59,13 @@ pub struct QuicSettings {
     #[serde(default = "QuicSettings::default_dgram_max_queue_len")]
     pub dgram_send_max_queue_len: usize,
 
+    /// Configures whether to enable early data (0-RTT) support. Currently only
+    /// supported for servers.
+    ///
+    /// Defaults to `false`.
+    #[serde(default = "QuicSettings::default_enable_early_data")]
```
evanrittenhouse: Booleans default to false already, so not sure we need this

tokio-quiche/tests/integration_tests/zero_rtt.rs
```diff
@@ -0,0 +1,234 @@
...
+        let _summary = helper_connect_with_early_data(
+            h3i_config,
+            Some(early_frame_actions),
+            helper_frame_actions(stream_id),
+        )
+        .await;
+
+        {
+            let context = context.lock().unwrap();
+            assert_eq!(context.hosts_seen.len(), 2);
```
evanrittenhouse, line 125: Since these happen deterministically (`early.test.com` should always occur before `test.com`, since the former is an early request), can we just do:

```
assert_eq!(context.hosts_seen, vec!["early.test.com", "test.com"])
```

tokio-quiche/tests/fixtures/h3i_fixtures.rs
evanrittenhouse: Should this now shim to `default_headers_with_authority("test.com")`?

tokio-quiche/src/quic/io/worker.rs
toidiu: There are many callbacks so calling those out seems odd. I check that priv_key callback was not called. We do call the async_select_cert callback and this is actually where we enable 0-rtt on the connection so that does need to be called.

A cleaner abstraction in the future would be to add a separate EarlyData connection stage and handle progressing that explicitly rather than doing it via RunningApplication.

ghedo:
Looks like there is a conflicts with one of the other changes now :crying_cat_face:

Issue #2277, opened 2025-12-09
FINISHED vs RESET


icing:
In `curl` we have test cases against an `nghttpx` server that resets stream responses after the first 16k of DATA. This works reliably with curl using `ngtcp2`. Using `quiche` as the backend, we sometimes see `FINISHED` events instead of a `RESET` and assume wrongly that the transfer was complete.

Is there anything we might have missed in our quiche usage here? Or is this a bug in quiche's RESET handling? Thanks for the help.

Tested with quiche 0.24.4

LPardue:
Finished should fire if you've read all the data sent on the stream up until the Fin. 

Do you have a pcap or qlog of the case where you see a Finished when you'd expect a reset?

icing:
[6a5e1f05d4b1e353b26411b4f87cae6c38ab740d.sqlog.txt](https://github.com/user-attachments/files/24052685/6a5e1f05d4b1e353b26411b4f87cae6c38ab740d.sqlog.txt)

Attached a qlog where stream 72 gives `FINISHED` event and the qlog lists:

```
[{"frame_type":"reset_stream","stream_id":72,"error_code":258,"final_size":13992},{"frame_type":"padding","payload_length":1399}]
```

LPardue:
Hmm, that does seem weird. Do you have any instructions for running this test setup that would allow us to try and repro?

icing:
Well, you'd need the whole curl test environment for that. A httpd, an nghttpx with h3, pytest, etc.

If you want I can point you to the details, but it might take a bit.

LPardue:
I don't have many spare cycles to get anything complicated running, which is going to make investigating this hard. Maybe you can provide a curl trace log too just so I can see what the sequence of events is?

LPardue:
I _think_ I might have emulated the sequence in https://github.com/cloudflare/quiche/pull/2278 but it would be great to see if curl client sees the same order of events.

icing:
Confirming the fix makes the issue no longer reproduce on my system. Thanks.

vszakats:
Nice!

normanmaurer:
@LPardue should this one be closed as it seems like it was fixed

normanmaurer:
Actually scratch that... the PR was not merged.

PR #2278, opened 2025-12-09
Mid-DATA stream reset

LPardue:
This attempts to use a unit test to emulate the behaviour described in https://github.com/cloudflare/quiche/issues/2277, where the peer clams to send a DATA frame of length N but sends fewer bytes than that and resets the stream. This seems to trigger an erroneous Event::Finished event that confuses applications.

Ideally, we fix the underlying logic and thus will need to change the test to match.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -7168,6 +7168,74 @@ mod tests {
         assert_eq!(s.poll_server(), Err(Error::Done));
     }
 
+    #[test]
+    fn reset_mid_data_frame() {
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        // Client sends request.
+        let (stream, req) = s.send_request(false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        // Server sends response
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        let resp = s.send_response(stream, false).unwrap();
+
+        let ev_headers = Event::Headers {
+            list: resp,
+            more_frames: true,
+        };
+
+        assert_eq!(s.poll_client(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_client(), Err(Error::Done));
+
+        // We want to indicate a larger DATA frame than we will send, so
+        // sidestep quiche's h3 API and write H3 frame bytes directly to QUIC
+        // stream. Imagine a response body of 200 bytes, that would be sent
+        // across two DATA frames of 100 bytes, but quiche's stream write is
+        // limited to only emitting 95 bytes of the DATA frame before the stream
+        // is reset.
+        let mut d = [42; 10];
+        let mut b = octets::OctetsMut::with_slice(&mut d);
+        b.put_varint(frame::DATA_FRAME_TYPE_ID).unwrap();
+        b.put_varint(100).unwrap();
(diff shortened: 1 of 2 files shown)


LPardue:
@icing can you try this branch out to see if it addresses your failing curl tests?

icing:
Thanks @LPardue , with this patch I can no longer reproduce the bug.

I'll disable the test in curl until we see a quiche version higher than 0.24.4.

quiche/src/h3/mod.rs
```diff
@@ -2744,6 +2744,20 @@ impl Connection {
...
                     }
 
+                    // Do a dummy read to check if the stream has been reset
+                    // while waiting to receive more data from a declared DATA
+                    // frame length. Doing this prevents an Event::Data frame
+                    // being triggered, which then might be followed by an
+                    // immediate Event::Finished if an application has already
+                    // read exactly the stream final size.
+                    if let Err(crate::Error::StreamReset(e)) =
+                        conn.stream_recv(stream_id, &mut [])
```
antoniovicente, line 2754: Are there other ways to query if the stream was reset?

I think that this conn.stream_recv will trigger a call to self.streams.remove_readable and collect the stream if it has completed.

Does something elsewhere in the code depend on the stream being listed there in order to properly handle the stream reset?

LPardue: This is similar to how we detect if the stream was reset [here](https://github.com/cloudflare/quiche/blob/86a79a7a049701baf608d6e88a917fa96fe804b6/quiche/src/h3/mod.rs#L2065)                  

There's no other way to check that without, I beleive, without changing the quiche library to track and expose the state, which probably raises evem more questions.

I think it's important for the quiche state to get cleaned up, to prevent duplicate or erroneous events being raised (this issue is one case where that's happened but there's been others in the past, leading to the already convoluted code)

antoniovicente: This may be fine, although I'm curious what application hooks are or should be invoked when a stream is reset or finishes gracefully.  Won't block the PR on this.

antoniovicente (approved):

quiche/src/h3/mod.rs
LPardue: Well, that's up to the app to deal with when consuming the result of poll() 😄 

Some scenarios:

* Client resets a POST early on its own according.
* Client resets GET because server rejected it with STOP_SENDING
* Server resets a response because the origin abruptly terminated the response for whatever reason (the scenario identified in Stefan's issue)

It's important to distinguish between reset or finished especially with respect to if there is a content-length declared or not. We delegate a lot of the semantic responsibility to apps. In turn we should be expected to pass on the truth of what's happened "on the wire".

quiche/src/h3/mod.rs
```diff
@@ -7168,6 +7182,140 @@ mod tests {
...
+        assert_eq!(s.poll_client(), Err(Error::Done));
+
+        let mut read_buf = [0; 65535];
+        let read = s
+            .client
+            .recv_body(&mut s.pipe.client, 0, &mut read_buf)
+            .unwrap();
+        assert_eq!(read, 95);
+
+        let written = s.pipe.server.stream_send(0, &[], true).unwrap();
```
ghedo, line 7308: I think this needs to be treated as an error? https://datatracker.ietf.org/doc/html/rfc9114#section-7.1-6

Somewhat orthogonal to this PR though...

vszakats:
Is there a planned timeframe for finalizing and merging this fix?

(It's a blocker to upgrade quiche in the curl CI.)

PR #2282, opened 2025-12-11
Clippy Fixes

PSUdaemon:
Clippy fixes for Rust 1.92

https://rust-lang.github.io/rust-clippy/stable/index.html#/non_canonical_partial_ord_impl

### DIFF
--- a/quiche/src/stream/mod.rs
+++ b/quiche/src/stream/mod.rs
@@ -816,44 +816,41 @@ impl PartialEq for StreamPriorityKey {
 impl Eq for StreamPriorityKey {}
 
 impl PartialOrd for StreamPriorityKey {
-    // Priority ordering is complex, disable Clippy warning.
-    #[allow(clippy::non_canonical_partial_ord_impl)]
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
-        // Ignore priority if ID matches.
-        if self.id == other.id {
-            return Some(cmp::Ordering::Equal);
-        }
-
-        // First, order by urgency...
-        if self.urgency != other.urgency {
-            return self.urgency.partial_cmp(&other.urgency);
-        }
-
-        // ...when the urgency is the same, and both are not incremental, order
-        // by stream ID...
-        if !self.incremental && !other.incremental {
-            return self.id.partial_cmp(&other.id);
-        }
-
-        // ...non-incremental takes priority over incremental...
-        if self.incremental && !other.incremental {
-            return Some(cmp::Ordering::Greater);
-        }
-        if !self.incremental && other.incremental {
-            return Some(cmp::Ordering::Less);
-        }
-
-        // ...finally, when both are incremental, `other` takes precedence (so
-        // `self` is always sorted after other same-urgency incremental
-        // entries).
-        Some(cmp::Ordering::Greater)
+        Some(self.cmp(other))
     }
 }
 
 impl Ord for StreamPriorityKey {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
-        // `partial_cmp()` never returns `None`, so this should be safe.
-        self.partial_cmp(other).unwrap()
+        // Ignore priority if ID matches.
+        if self.id == other.id {
+            return cmp::Ordering::Equal;
+        }
+
+        // First, order by urgency...
+        if self.urgency != other.urgency {
+            return self.urgency.cmp(&other.urgency);
+        }
+
+        // ...when the urgency is the same, and both are not incremental, order
+        // by stream ID...
+        if !self.incremental && !other.incremental {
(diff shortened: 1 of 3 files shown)


tokio-quiche/src/quic/io/utilization_estimator.rs
```diff
@@ -147,9 +147,8 @@ impl PartialEq for Estimate {
 impl Eq for Estimate {}
 
 impl PartialOrd for Estimate {
-    #[allow(clippy::non_canonical_partial_ord_impl)]
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        self.bandwidth.partial_cmp(&other.bandwidth)
+        Some(self.cmp(other))
     }
 }
```
PSUdaemon, line 153: This is the `Ord` implementation:
```
impl Ord for Estimate {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        self.bandwidth.cmp(&other.bandwidth)
    }
}
```

quiche/src/stream/mod.rs
```diff
@@ -816,44 +816,41 @@ impl PartialEq for StreamPriorityKey {
 impl Eq for StreamPriorityKey {}
 
 impl PartialOrd for StreamPriorityKey {
-    // Priority ordering is complex, disable Clippy warning.
-    #[allow(clippy::non_canonical_partial_ord_impl)]
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
```
LPardue, line 820: Touching this always makes me a little nervous but we have test coverage and that seems to have passed

tokio-quiche/src/quic/io/utilization_estimator.rs
LPardue: @antoniovicente please check

LPardue (approved):

quiche/src/stream/mod.rs
```diff
@@ -816,44 +816,41 @@ impl PartialEq for StreamPriorityKey {
...
+    }
+}
+
+impl Ord for StreamPriorityKey {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
         // Ignore priority if ID matches.
         if self.id == other.id {
-            return Some(cmp::Ordering::Equal);
+            return cmp::Ordering::Equal;
```
antoniovicente, line 828: I'm worried about correctness due to this line.

Ignoring the other fields if the ids are equal seems very very bad as it doesn't allow for correct sorting for a collection of elements when there are 2 with same id but differences in other fields.

quiche/src/stream/mod.rs
```diff
@@ -816,44 +816,41 @@ impl PartialEq for StreamPriorityKey {
...
         // entries).
-        Some(cmp::Ordering::Greater)
-    }
-}
-
-impl Ord for StreamPriorityKey {
-    fn cmp(&self, other: &Self) -> cmp::Ordering {
-        // `partial_cmp()` never returns `None`, so this should be safe.
-        self.partial_cmp(other).unwrap()
+        cmp::Ordering::Greater
```
antoniovicente, line 853: I don't know what the requirements for cmp are and what are the consequences of having elements that return different answers if you swap the argument order.

I think this is a case where the id could be compared.

In the partial order case, this case may want to return None.

tokio-quiche/src/quic/io/utilization_estimator.rs
antoniovicente: Seems correct.  loss is ignored consistently for eq and cmp.

quiche/src/stream/mod.rs
antoniovicente: It seems like the before and after do the same, even if the semantics of the comparison are surprising.

quiche/src/stream/mod.rs
LPardue: Its kinda implicit in the overall design but where StreamPriorityKey would be used, there cannot be two objects with the same stream ID (If there were, we'd have bigger issues :joy: )

quiche/src/stream/mod.rs
LPardue: That would break the prioritization algorithm which needs these types of streams to be round-robined. This is a bit of a trick, newly added incremental streams of the same urgency to "go to the back", so by removing then adding a stream when it's been visited (e.g. to send data) we can ensure the next visit hits the next stream of this type.

antoniovicente: This version of the comparison function does not allow binary search to work correctly.  Also relevant to sorted data structures like trees.

I don't know enough about how we use these comparison and how we avoid using this type on a data structure that requires certain properties from the comparison function.  This may be touching on the differences between cmp and partial_cmp.

PR #2287, opened 2025-12-15
Fix StreamPriorityKey::cmp implementation

PSUdaemon:
Fix `StreamPriorityKey::cmp` implementation so it doesn't violate Rust's `Ord`
trait contract while keeping the intended round-robin behavior.

This is based on the discussion that was had in #2282. @antoniovicente had some
concerns about the implementation with respect to the `Ord` trait contract.

This is my attempt at fixing that and maintaining the round-robin requirement.

- Ordering comes from a derived `Ord` on a new position type, so the contract
  holds without a hand-written comparison.
- The sequence field is per queue, so reading from a stream doesn't reorder what
  we send on it.
- `cycle_flushable()` became `cycle_priority()`. It takes the queue that was
  serviced, and runs every time we service a stream rather than only when the
  stream stays queued.
- `stream_readable_next()` and `stream_writable_next()` take a stream off the
  queue as they hand it out. The old reinsert put it back by accident, and only
  for incremental streams, so membership follows eligibility now instead.
- Moved `stream_priority()` into `StreamMap::set_priority()`, so the priority
  tests don't need their own copy of it.

The h3 change is a test, see the comment on the thread.

The branch is rebased onto master and squashed, so the existing review threads
point at lines that no longer exist. Happy to re-open any of them.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -4134,6 +4134,47 @@ mod tests {
         assert_eq!(s.poll_client(), Ok((stream, Event::Finished)));
     }
 
+    /// A server polling concurrent requests sees the headers of every request
+    /// before the body of any of them, so it can route or reject on headers
+    /// alone instead of being made to consume one request's body first.
+    ///
+    /// The transport provides this: `stream_recv` cycles an incremental stream
+    /// to the back of its priority group each time it is serviced, and `poll()`
+    /// parses frames through `stream_recv`. HTTP/3 does not arrange the
+    /// interleaving itself.
+    ///
+    /// Asserted here rather than left to the transport tests because it is an
+    /// HTTP/3 property an application depends on, and because losing it is not
+    /// obvious from either layer alone: without the cycling this order becomes
+    /// per-stream FIFO, headers and body together, one request at a time.
+    #[test]
+    fn poll_yields_headers_of_all_requests_before_any_body() {
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        let mut streams = Vec::new();
+
+        for _ in 0..3 {
+            let (stream, _) = s.send_request(false).unwrap();
+            s.send_body_client(stream, true).unwrap();
+            streams.push(stream);
+        }
+
+        let mut headers = Vec::new();
+
+        for _ in 0..streams.len() {
+            match s.poll_server() {
+                Ok((stream, Event::Headers { .. })) => headers.push(stream),
+
+                other => panic!(
+                    "expected Headers for every request before any body, got {other:?}"
+                ),
+            }
+        }
+
+        assert_eq!(headers, streams);
+    }
+
     #[test]
     /// Send a request with multiple DATA frames, get a response with one DATA
     /// frame.
--- a/quiche/src/stream/mod.rs
+++ b/quiche/src/stream/mod.rs
@@ -26,6 +26,8 @@
 
 use std::cmp;
 
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
 use std::collections::hash_map;
@@ -110,6 +112,14 @@ type BuildStreamIdHasher = std::hash::BuildHasherDefault<StreamIdHasher>;
 
 pub type StreamIdHashMap<V> = HashMap<u64, V, BuildStreamIdHasher>;
 pub type StreamIdHashSet = HashSet<u64, BuildStreamIdHasher>;
+
+/// One of the priority queues a stream can be scheduled in.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum PriorityQueue {
+    Readable,
+    Writable,
+    Flushable,
+}
 
 /// Keeps track of QUIC streams and enforces stream limits.
 #[derive(Default)]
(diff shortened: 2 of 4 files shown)


quiche/src/h3/mod.rs
```diff
@@ -3825,33 +3825,30 @@ mod tests {
             more_frames: true,
         };
         assert_eq!(ev, ev_headers);
+        assert_eq!(s.poll_server(), Ok((0, Event::Data)));
```
LPardue: Needing to change this test seems a bit fishy to me. 

While RFC 9218 primarily targets prioritization of serving responses, we apply stream scheduling to all QUIC streams. The default priority for a new stream is, in quiche's model, for urgency=127 and incremental=true (see https://github.com/cloudflare/quiche/blob/86a79a7a049701baf608d6e88a917fa96fe804b6/quiche/src/stream/mod.rs#L798). That should trigger round-robin sending of all request data.

In the previous test, we saw the server reading events in the round-robin order of stream sends. Although that's only partly true because the recv_body() function attempts to greedily read all available data frames independent of the peer's sender scheduling. 

The changes to this test seem to imply that sending has turned into FIFO, where HEADERS and DATA are sent together (or, the peer is somehow receiving them together). It's not abundantly clear to me what's going on. I think it would pay to get some more understanding why this test triggered some change while others didn't

LPardue:
Thanks Phil. I appreciate the work to try and make this area more rusty. I'd like to dig in more to one of the tests as I commented above. But I think this is probably a path we can follow through with.

quiche/src/h3/mod.rs
PSUdaemon: The old "interleaved" test results were an artifact of undefined behavior from the broken `Ord` implementation. With the fix, the h3 test now shows correct receive side sequential (priority) ordering per-stream.

To demonstrate that round-robin actually works, I added a new end to end test (`incremental_stream_round_robin`) that tracks per-packet send order. It shows the expected interleaving: `[4, 8, 12, 4, 8, 12, 4, 8, 12]`.

LPardue: Interesting, thanks for the explanation. This sounds a bit like a Bob Ross happy accident. It _might_ be an undesirable change since it can be in the server's interest to "skim" across headers that is is receiving in order to do some useful work.

It might be that we need to make the choice of that behaviour more application configurable. That would potentially involves separate read and write priorities (which I appreciate is beyond the scope of a simple fix up).

A stop gap might be, on the reader side, to cycle the stream after generating an Event::Headers?

PSUdaemon: > A stop gap might be, on the reader side, to cycle the stream after generating an Event::Headers?

Yeah, that's an interesting idea. Are you looking to have that added in this PR, or as a followup one?

LPardue: I think for parity with the current behaviour, we'd need that in this PR. Would you mind trying it and see if it achieves the old ordering?

PSUdaemon: This was simpler than I expected.

quiche/src/stream/mod.rs
```diff
@@ -426,6 +456,90 @@ impl<F: BufFactory> StreamMap<F> {
...
+        self.flushable.insert(new_priority_key);
+
+        true
+    }
+
+    /// Cycles a readable stream to the back of the queue.
+    ///
+    /// This is used for round-robin scheduling: after processing headers from
+    /// a stream, it should be moved to the back of its priority group to give
+    /// other streams a chance.
```
LPardue: nit: the QUIC layer doesn't know anything about headers. Maybe we can say something like

```suggestion
    /// This is used for round-robin scheduling: after processing some data from
    /// a stream, it can be moved to the back of its priority group to give
    /// other streams a chance of being read.
```

quiche/src/tests.rs
```diff
@@ -7804,9 +7804,10 @@ fn last_tx_data_larger_than_tx_data(
     pipe.server.on_timeout();
 
     // Server sends PTO probe (not limited to cwnd),
-    // to update last_tx_data.
+    // to update last_tx_data. The exact size depends on stream scheduling
+    // order, which may vary; we just need some data to be sent.
     let (len, _) = pipe.server.send(&mut buf).unwrap();
-    assert_eq!(len, 1200);
+    assert!(len > 0);
```
LPardue: This also seems suspect - the round robin should be determininstic no?

PSUdaemon: It is, but I didn't think the specific value here was important and it could just cause churn in the future if other changes instigated a change here. I will change this to the new value.

PSUdaemon:
Happy new year! 🎆

Checking back in to see if there are any more concerns for this PR?

quiche/src/tests.rs
```diff
@@ -8251,7 +8251,7 @@ fn last_tx_data_larger_than_tx_data(
     // Server sends PTO probe (not limited to cwnd),
     // to update last_tx_data.
     let (len, _) = pipe.server.send(&mut buf).unwrap();
-    assert_eq!(len, 1200);
+    assert_eq!(len, 848);
```
LPardue: How come this decreased?

PSUdaemon: I believe the stream ordering changed. If you look at lines 8241-8242 above, you can see 1200 bytes vs 800 bytes in two different streams. It's 800 bytes + 48 bytes of overhead. Why not 1248 then? The buffer on line 8240 is only 1200 bytes.

If you would like I can change this test more to prove that out.

PSUdaemon: I looked into this in detail, and it's a little more complicated (dealing with PTO and retransmission), but switching the order that we buffer the streams shows that it's stream 8 being selected. Specifically this is because we had `emit_flight` sending on the other stream (4) which increases the sequence number (sends it to the back) and thus a different stream (8) is selected.

quiche/src/h3/mod.rs
PSUdaemon: The headers first behavior you asked for is now baked into the design, so I removed the explicit code in the h3 module. I replaced it with a test to make sure we don't accidentally lose it again.

PR #2289, opened 2025-12-17
tokio-quiche: Enable active connection migration for clients of tq's QUIC server

TheJokr:
This PR consists of 4 separate commits. The first two are refactors to simplify bits of the internal and public API of tokio-quiche. On the public side, I removed the `socket_cookie` parameter completely and made the CID generator a `QuicListener` property. This means our users are expected to provide a generator for each socket now, with TryFrom defaulting to `SimpleConnectionIdGenerator`. Users that relied on `socket_cookie` have more flexibility now, as they can embed the cookie (or any other data they may require) directly in the CID generator instance. _This is a breaking change._

The third commit changes the IoWorker loop to top up quiche's supply of SCIDs in every iteration, as well as removing retired SCIDs from the `ConnectionMap`. SCIDs are generated by the generator from the `QuicListener` (which was the reason for refactoring that API in the first place.) Each active connection receives its own Arc to the shared generator. This unblocks active connection migration for clients, as we already support passive migration where only the IP/port changes.

Finally, the fourth commit adds a test to verify that active migration works correctly. Since active and passive migration scenarios are extremely similar, and both require low-level quiche access, I decided to add the active-migration-specific bits to the existing passive migration test case rather than copy-pasting the entire thing.


TheJokr:
CI failures are due to missing apt packages, ~~I'm pretty sure it's a GitHub issue. Will re-trigger CI tomorrow.~~

Took some time to investigate, it's because we don't call `apt-get update`: #2292

tokio-quiche/tests/integration_tests/migration.rs
```diff
@@ -80,15 +84,15 @@ async fn test_passive_migration() {
     client_config.set_initial_max_stream_data_uni(1500);
     client_config.set_initial_max_streams_bidi(10);
     client_config.set_initial_max_streams_uni(3);
+    client_config.set_active_connection_id_limit(2);
+    // We don't need to allow the server to initiate connection migration
```
LPardue: nit: QUICv1 does not permit servers to initiate migration any way, so perhaps a comment like below is more accurate

```
    // Server's can't initiate migration but disable it explicitly anyway.
```

TheJokr: Replaced the comment and rebased the PR

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -605,9 +611,11 @@ where
     fn handle_conn_map_commands(&mut self) {
         while let Ok(req) = self.conn_map_cmd_rx.try_recv() {
             match req {
+                ConnectionMapCommand::MapCid {
+                    existing_cid,
+                    new_cid,
+                } => self.conns.map_cid(&existing_cid, &new_cid),
                 ConnectionMapCommand::UnmapCid(cid) => self.conns.unmap_cid(&cid),
-                ConnectionMapCommand::RemoveScid(scid) =>
-                    self.conns.remove(&scid),
```
ghedo, line 608: Maybe I'm missing something, but aren't we just leaking memory here as we are not removing connections from the map anymore?

TheJokr: RemoveScid did 2 things: remove the passed SCID and, using the integer ID stored for that SCID, remove the entry for the integer ID. The integer ID map doesn't exist anymore, so RemoveSCID is the same as UnmapCID.

ghedo (approved):

tokio-quiche/tests/integration_tests/migration.rs
```diff
@@ -80,15 +84,15 @@ async fn test_passive_migration() {
...
     client_config.verify_peer(false);
 
-    let mut client_scid = [0; quiche::MAX_CONN_ID_LEN];
-    boring::rand::rand_bytes(&mut client_scid[..]).unwrap();
-    let client_scid = quiche::ConnectionId::from_ref(&client_scid);
-    let client_addr = "127.0.0.1:12345".parse().unwrap();
+    let client_scid = SimpleConnectionIdGenerator.new_connection_id();
 
-    let socket = tokio::net::UdpSocket::bind(client_addr).await.unwrap();
+    let socket = tokio::net::UdpSocket::bind("127.0.0.1:0").await.unwrap();
```
LPardue: I'm not sure why you're delegating the port number selection. In the past, this sort of thing leads to flakes

TheJokr: IME hardcoded port numbers lead to flaky tests. If I left this with a hardcoded port, then either the active or the passive test would fail when run in parallel. (cargo runs tests in parallel by default.)

LPardue: That type of flake also makes sense. How about we use separate static ports for active vs passive migration then? I.e something like

```
let (initial_client_addr, migrated_client_addr) = if active { ...} else {... };

PR #2291, opened 2025-12-18
octets: fix slice_last() with non-zero offset

TerryTaoYY:
## Summary

Fix `Octets::slice_last()` and `OctetsMut::slice_last()` returning an incorrect tail slice when the internal offset has been advanced.

These helpers previously used `cap()` (remaining bytes after `off`) to compute the start index, but then applied that index to the full underlying buffer (`self.buf[...]`). When `off > 0`, this could include already-consumed bytes and return the wrong region.

## Reproduction

```rust
let mut b = Octets::with_slice(b"helloworld");
b.get_bytes(5).unwrap(); // advance offset past "hello"
assert_eq!(b.slice_last(4).unwrap(), b"orld");
```

Before this change, slice_last(4) could return the wrong region (e.g. starting at index 1), because cap() was treated as an absolute index into the full buffer.

## Fix

Compute `end` as `self.buf.len()` and slice `self.buf[end - len..end]` (and the mutable equivalent). The existing `len <= self.cap()` guard ensures `end - len >= off`, so the returned slice stays within the remaining window.

## Tests

Add regression cases that advance the offset via `get_bytes()` and then assert `slice_last()` returns the correct tail bytes for both `Octets` and `OctetsMut`.

Tested with:
- \ cargo test -p octets

### DIFF
--- a/octets/src/lib.rs
+++ b/octets/src/lib.rs
@@ -332,8 +332,8 @@ impl<'a> Octets<'a> {
             return Err(BufferTooShortError);
         }
 
-        let cap = self.cap();
-        Ok(&self.buf[cap - len..])
+        let end = self.buf.len();
+        Ok(&self.buf[end - len..end])
     }
 
     /// Advances the buffer's offset.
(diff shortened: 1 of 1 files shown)


ghedo (approved):

LPardue (approved):

ghedo:
Merged, thanks @TerryTaoYY!

PR #2294, opened 2025-12-18
refactor: simplify packet sending loop using core::slice::chunks

birneee:


apps/src/sendto.rs
```diff
@@ -109,22 +109,14 @@ pub fn send_to(
...
-    let mut off = 0;
-    let mut left = buf.len();
     let mut written = 0;
-
-    while left > 0 {
-        let pkt_len = cmp::min(left, segment_size);
-
-        match socket.send_to(&buf[off..off + pkt_len], send_info.to) {
+    for chunk in buf.chunks(segment_size) {
+        match socket.send_to(&chunk, send_info.to) {
```
antoniovicente: Nice simplification.  Please see CI failures due to compile warnings:

```
error: unused import: `std::cmp`
  --> apps/src/sendto.rs:27:5
   |
27 | use std::cmp;
```

apps/src/sendto.rs
```diff
@@ -109,22 +109,14 @@ pub fn send_to(
...
 
-    let mut off = 0;
-    let mut left = buf.len();
     let mut written = 0;
-
-    while left > 0 {
-        let pkt_len = cmp::min(left, segment_size);
-
-        match socket.send_to(&buf[off..off + pkt_len], send_info.to) {
+    for chunk in buf.chunks(segment_size) {
```
LPardue, line 122: According to rustdocs, Segment_size 0 will panic. We should probably have an earlier check for that. The lack of check in the old code would be a problem too but would manifest differently

jannes (approved):

PR #2297, opened 2025-12-23
Delete old BBR and pacer implementations in the recovery/congestion directory.

antoniovicente:
There are big known problems in these implementations.

We recommend use of:
* the cubic implementation, without pacing
* the bbr2 implementation from the gcongestion directory, with pacing.  Note that pacing only works correctly on Linux and if TxTime is available (qdisc fq kernel module enabled).

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -381,6 +381,7 @@ impl FromStr for CongestionControlAlgorithm {
         match name {
             "reno" => Ok(CongestionControlAlgorithm::Reno),
             "cubic" => Ok(CongestionControlAlgorithm::CUBIC),
+            "bbr" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
             "bbr2" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
             "bbr2_gcongestion" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
             _ => Err(crate::Error::CongestionControl),
@@ -727,11 +728,13 @@ mod tests {
         assert_eq!(algo, CongestionControlAlgorithm::CUBIC);
         assert!(!recovery_for_alg(algo).gcongestion_enabled());
 
+        let algo = CongestionControlAlgorithm::from_str("bbr").unwrap();
+        assert_eq!(algo, CongestionControlAlgorithm::Bbr2Gcongestion);
+        assert!(recovery_for_alg(algo).gcongestion_enabled());
+
         let algo = CongestionControlAlgorithm::from_str("bbr2").unwrap();
-        {
-            assert_eq!(algo, CongestionControlAlgorithm::Bbr2Gcongestion);
-            assert!(recovery_for_alg(algo).gcongestion_enabled());
-        }
+        assert_eq!(algo, CongestionControlAlgorithm::Bbr2Gcongestion);
+        assert!(recovery_for_alg(algo).gcongestion_enabled());
 
         let algo =
             CongestionControlAlgorithm::from_str("bbr2_gcongestion").unwrap();


reneleonhardt:
>     * the bbr2 implementation from the gcongestion directory, with pacing.

But this PR deletes brr and brr2? What about switching to brr3?

Other users report much better experiences with brr than cubic, what are those big known problems?

antoniovicente:
> > ```
> > * the bbr2 implementation from the gcongestion directory, with pacing.
> > ```
> 
> But this PR deletes brr and brr2? What about switching to brr3?
> 
> Other users report much better experiences with brr than cubic, what are those big known problems?

The main motivation behind this PR is to delete the pacer in the congestion directory.  Since BBR implementations in that directory have a hard dependency on it, I think it is time for them to also be deleted.

All of our investments are on making the bbr2 implementation in the gcongestion directory work well even in cases where Linux TxTime is not available.  I have several planned changes in the near future to fix some recently found bugs and make this implementation closer to the BBR RFC draft.

The config option to select "bbr2" still works.  I can also keep the "bbr" option, but point it to the bbr2 implementation in the gcongestion directory.

quiche/src/recovery/congestion/mod.rs
```diff
@@ -223,7 +191,7 @@ impl Congestion {
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn on_packet_sent(
         &mut self, bytes_in_flight: usize, sent_bytes: usize, now: Instant,
-        pkt: &mut Sent, rtt_stats: &RttStats, bytes_lost: u64, in_flight: bool,
+        pkt: &mut Sent, _rtt_stats: &RttStats, bytes_lost: u64, in_flight: bool,
```
LPardue: nit: it seems we can remove `rtt_stats`, that might also bring the number below clippy's threshold so we could drop the exception

antoniovicente: I think it's still used in quiche/src/recovery/gcongestion/mod.rs

quiche/src/recovery/congestion/mod.rs
```diff
@@ -348,11 +284,9 @@ impl From<CongestionControlAlgorithm> for &'static CongestionControlOps {
         match algo {
             CongestionControlAlgorithm::Reno => &reno::RENO,
             CongestionControlAlgorithm::CUBIC => &cubic::CUBIC,
-            CongestionControlAlgorithm::BBR => &bbr::BBR,
-            CongestionControlAlgorithm::BBR2 => &bbr2::BBR2,
             CongestionControlAlgorithm::Bbr2Gcongestion => {
                 debug_panic!("legacy implementation, not gcongestion");
-                &bbr2::BBR2
+                &cubic::CUBIC
```
LPardue: I'm not really sure how to rationalize what is intended to be the failure case here.  Walking through it, the only usage I can see is inside recovery/congestion/mod.rs::from_config() via

```
cc_ops:  recovery_config.cc_algorithm.into(),
```

Is the idea that this module is "legacy" and can't be configured with the gcongestion code? I wonder if it might make more sense to fail closed as `unreachable!` than pick a default. This seems to sort of be what GRecovery::new() is doing anyway by returning a `None` for any option that is not CongestionControlAlgorithm::Bbr2Gcongestion

quiche/src/recovery/congestion/mod.rs
antoniovicente: Actually, good catch.  Pushed a commit with the removals.

quiche/src/recovery/congestion/mod.rs
antoniovicente: Changed to unreachable.

quiche/src/recovery/congestion/mod.rs
```diff
@@ -348,12 +283,7 @@ impl From<CongestionControlAlgorithm> for &'static CongestionControlOps {
         match algo {
             CongestionControlAlgorithm::Reno => &reno::RENO,
             CongestionControlAlgorithm::CUBIC => &cubic::CUBIC,
-            CongestionControlAlgorithm::BBR => &bbr::BBR,
-            CongestionControlAlgorithm::BBR2 => &bbr2::BBR2,
-            CongestionControlAlgorithm::Bbr2Gcongestion => {
-                debug_panic!("legacy implementation, not gcongestion");
-                &bbr2::BBR2
-            },
+            CongestionControlAlgorithm::Bbr2Gcongestion => unreachable!(),
```
LPardue, line 290: nit: a comment about why this is unreachable might make sense, or some text in the unreachable call. But this is really a nit and I can live without either.

quiche/src/recovery/mod.rs
```diff
@@ -1861,7 +1842,7 @@ mod tests {
         assert_eq!(r.bytes_in_flight(), 6000);
         assert_eq!(r.bytes_in_flight_duration(), Duration::from_millis(50));
 
-        if cc_algorithm_name != "bbr2_gcongestion" {
+        if cc_algorithm_name == "cubic" || cc_algorithm_name == "reno" {
             // Pacing is not done during initial phase of connection.
```
LPardue: nit: pacing is never done, so you could update this comment or leave it as it because the statement is kind of accurate

quiche/src/recovery/mod.rs
```diff
@@ -1956,27 +1926,13 @@ mod tests {
...
-                let bw =
-                    r.cwnd() as f64 / Duration::from_millis(50).as_secs_f64();
-                (bw * PACING_MULTIPLIER) as u64
-            },
+            _ => 0,
         };
         assert_eq!(r.pacing_rate(), pacing_rate);
 
-        let scale_factor = if cc_algorithm_name == "bbr2_gcongestion" {
+        let scale_factor = if cc_algorithm_name == "bbr2_gcongestion" ||
```
LPardue, line 1936: nit: I don't know if it would help to have a helper function like "is_bbr_algo()" but figured I'd ask. Purely a nit.

quiche/src/tests.rs
```diff
@@ -9449,9 +9449,9 @@ fn resilience_against_migration_attack(
     let send1_bytes = pipe.server.stream_send(1, &buf, true).unwrap();
     assert_eq!(send1_bytes, match cc_algorithm_name {
         #[cfg(feature = "openssl")]
-        "bbr2" => 14041,
+        "bbr2" => 13966,
```
LPardue, line 9454: I don't really understand why the values here have changed.

quiche/src/tests.rs
```diff
@@ -6656,7 +6656,7 @@ fn dgram_send_app_limited(
             .expect("no active")
             .recovery
             .app_limited(),
-        cc_algorithm_name != "bbr2_gcongestion"
+        cc_algorithm_name == "cubic" || cc_algorithm_name == "reno"
```
LPardue: nit: maybe clearer to move this out to a variable like

```
let should_be_app_limited = match cc_algorithm_name {
  bbr | bbr2_gcongestion => false,
  cubic | reno => true
}

```

quiche/src/tests.rs
antoniovicente: The old BBR2 implementation that is being replaced by bbr2_gcongestion operated slightly differently.

quiche/src/recovery/mod.rs
antoniovicente: Could help as a test tool.  More cleanups are coming.

quiche/src/tests.rs
antoniovicente: Done.

quiche/src/recovery/mod.rs

quiche/src/recovery/congestion/mod.rs
antoniovicente: Done.

quiche/src/recovery/mod.rs
```diff
@@ -385,10 +381,6 @@ impl FromStr for CongestionControlAlgorithm {
         match name {
             "reno" => Ok(CongestionControlAlgorithm::Reno),
             "cubic" => Ok(CongestionControlAlgorithm::CUBIC),
-            "bbr" => Ok(CongestionControlAlgorithm::BBR),
-            #[cfg(not(feature = "gcongestion"))]
-            "bbr2" => Ok(CongestionControlAlgorithm::BBR2),
-            #[cfg(feature = "gcongestion")]
             "bbr2" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
             "bbr2_gcongestion" => Ok(CongestionControlAlgorithm::Bbr2Gcongestion),
```
toidiu, line 386: I would love to end up where we only offer `bbr`. What about keeping that and aliasing it to `Bbr2Gcongestion`.

antoniovicente: Added the alias.

PR #2300, opened 2026-01-07
Add the intended send_time arg to send_on_path to fix packet send times when pacing to fix RTT calculation when using BBR

antoniovicente:
The packet send time is computed deep in on_packet_sent, and this time differs from the one passed to the GSO burst which is based on the first packet in the train.  When ReleaseDecision.can_burst() returns true, the difference between the real send time and the one stored in the recovery module can grow arbitrarily large resulting in very low RTT estimates which BBR holds onto as minRTT from there on.  Incorrect RTT will result in the congestion window being set to artificially small values because of incorrect BDP calculations since BDP = minRTT * BW_estimate

### DIFF
--- a/apps/src/client.rs
+++ b/apps/src/client.rs
@@ -33,8 +33,6 @@ use std::rc::Rc;
 
 use std::cell::RefCell;
 
-use std::time::Instant;
-
 use ring::rand::*;
 
 const MAX_DATAGRAM_SIZE: usize = 1350;
@@ -499,13 +497,10 @@ pub fn connect(
 
             for peer_addr in conn.paths_iter(local_addr) {
                 loop {
-                    let now = Instant::now();
                     let (write, send_info) = match conn.send_on_path(
                         &mut out,
                         Some(local_addr),
                         Some(peer_addr),
-                        &quiche::TimeSent::new(&None, now),
-                        now,
                     ) {
                         Ok(v) => v,
 
--- a/h3i/src/client/sync_client.rs
+++ b/h3i/src/client/sync_client.rs
@@ -384,13 +384,10 @@ pub fn connect(
 
             for peer_addr in conn.paths_iter(local_addr) {
                 loop {
-                    let now = Instant::now();
                     let (write, send_info) = match conn.send_on_path(
                         &mut out,
                         Some(local_addr),
                         Some(peer_addr),
-                        &quiche::TimeSent::new(&None, now),
-                        now,
                     ) {
                         Ok(v) => v,
 
--- a/quiche/src/ffi.rs
+++ b/quiche/src/ffi.rs
@@ -828,8 +828,7 @@ pub extern "C" fn quiche_conn_send_on_path(
     let to = optional_std_addr_from_c(to, to_len);
     let out = unsafe { slice::from_raw_parts_mut(out, out_len) };
 
-    let now = Instant::now();
-    match conn.send_on_path(out, from, to, &TimeSent::new(&None, now), now) {
+    match conn.send_on_path(out, from, to) {
         Ok((v, info)) => {
             out_info.from_len = std_addr_to_c(&info.from, &mut out_info.from);
             out_info.to_len = std_addr_to_c(&info.to, &mut out_info.to);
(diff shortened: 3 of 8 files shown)


LPardue:
I'm struggling to follow this a bit. IIUC there's new params for send_on_path but not documentation on how to populate them. The only place in the PR that seems to populate the new time_sent param with a non-None value is the TQ worker. Reading between the lines, it optionally pulls a value out of quiche and then reuses that for all send_on_path calls. I guess this stickiness of the value is what the fix is all about? 

It seems unfortunate that we have push the responsibility for this sort of thing up to library users, especially if it forces everyone to have to add something like `&quiche::TimeSent::new(&None, now), now` for no added benefit and a potential footgun if they try to set something and do it incorrectly.

While it would be nice to keep these details all internal, I appreciate that might be difficult. I don't really fully understand the complexity of `get_next_release_time()`. Maybe some kind of alternative could be to have an internal notion of batch release time that only changes when a batch is completed, then send_on_path can pull that directly? 

Alternatively, if jumping out through the app is the only way, we might consider a new send_on_path_magic_bikeshed function, that takes time_sent. Then send_on_path can just call that with `&quiche::TimeSent::new(&None, now), now`, meaning all legacy apps work untouched. That would let apps explicitly move to this new function and at that time do the work to figure out how to use it correctly.

antoniovicente:
> I'm struggling to follow this a bit. IIUC there's new params for send_on_path but not documentation on how to populate them. The only place in the PR that seems to populate the new time_sent param with a non-None value is the TQ worker. Reading between the lines, it optionally pulls a value out of quiche and then reuses that for all send_on_path calls. I guess this stickiness of the value is what the fix is all about?
> 
> It seems unfortunate that we have push the responsibility for this sort of thing up to library users, especially if it forces everyone to have to add something like `&quiche::TimeSent::new(&None, now), now` for no added benefit and a potential footgun if they try to set something and do it incorrectly.
> 
> While it would be nice to keep these details all internal, I appreciate that might be difficult. I don't really fully understand the complexity of `get_next_release_time()`. Maybe some kind of alternative could be to have an internal notion of batch release time that only changes when a batch is completed, then send_on_path can pull that directly?
> 
> Alternatively, if jumping out through the app is the only way, we might consider a new send_on_path_magic_bikeshed function, that takes time_sent. Then send_on_path can just call that with `&quiche::TimeSent::new(&None, now), now`, meaning all legacy apps work untouched. That would let apps explicitly move to this new function and at that time do the work to figure out how to use it correctly.

I think we may want to consider moving `gather_data_from_quiche_conn` from tokio-quiche to quiche.  The problem I'm struggling with here is that the API from quiche is packet based, but packet timing info needs to be computed based on bursts of packets that are sent together via GSO.  I think that's the way to keep the `send_on_path` API as it is today.

antoniovicente:
> > I'm struggling to follow this a bit. IIUC there's new params for send_on_path but not documentation on how to populate them. The only place in the PR that seems to populate the new time_sent param with a non-None value is the TQ worker. Reading between the lines, it optionally pulls a value out of quiche and then reuses that for all send_on_path calls. I guess this stickiness of the value is what the fix is all about?
> > It seems unfortunate that we have push the responsibility for this sort of thing up to library users, especially if it forces everyone to have to add something like `&quiche::TimeSent::new(&None, now), now` for no added benefit and a potential footgun if they try to set something and do it incorrectly.
> > While it would be nice to keep these details all internal, I appreciate that might be difficult. I don't really fully understand the complexity of `get_next_release_time()`. Maybe some kind of alternative could be to have an internal notion of batch release time that only changes when a batch is completed, then send_on_path can pull that directly?
> > Alternatively, if jumping out through the app is the only way, we might consider a new send_on_path_magic_bikeshed function, that takes time_sent. Then send_on_path can just call that with `&quiche::TimeSent::new(&None, now), now`, meaning all legacy apps work untouched. That would let apps explicitly move to this new function and at that time do the work to figure out how to use it correctly.
> 
> I think we may want to consider moving `gather_data_from_quiche_conn` from tokio-quiche to quiche. The problem I'm struggling with here is that the API from quiche is packet based, but packet timing info needs to be computed based on bursts of packets that are sent together via GSO. I think that's the way to keep the `send_on_path` API as it is today.

I did a first attempt to move GSO packet creation to quiche from tokio-quiche.  Need to go back and do a few more cleanups.

antoniovicente:
I'm going to try a more minimal change, moving back to draft.

PR #2302, opened 2026-01-08
Update tx_buffered when bytes are acked or lost on closed streams.

antoniovicente:
tx_buffered tracks the number of bytes that the connection has in output buffers or the network.  Stream bytes on closed streams that are acked or declared lost must be removed from tx_buffered in order to avoid this count going out of sync.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -505,229 +505,6 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
-/// A specialized [`Result`] type for quiche operations.
-///
-/// This type is used throughout quiche's public API for any operation that
-/// can produce an error.
-///
-/// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
-pub type Result<T> = std::result::Result<T, Error>;
-
-/// A QUIC error.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum Error {
-    /// There is no more work to do.
-    Done,
-
-    /// The provided buffer is too short.
-    BufferTooShort,
-
-    /// The provided packet cannot be parsed because its version is unknown.
-    UnknownVersion,
-
-    /// The provided packet cannot be parsed because it contains an invalid
-    /// frame.
-    InvalidFrame,
-
-    /// The provided packet cannot be parsed.
-    InvalidPacket,
-
-    /// The operation cannot be completed because the connection is in an
-    /// invalid state.
-    InvalidState,
-
-    /// The operation cannot be completed because the stream is in an
-    /// invalid state.
-    ///
-    /// The stream ID is provided as associated data.
-    InvalidStreamState(u64),
-
-    /// The peer's transport params cannot be parsed.
-    InvalidTransportParam,
-
-    /// A cryptographic operation failed.
-    CryptoFail,
-
-    /// The TLS handshake failed.
-    TlsFail,
-
-    /// The peer violated the local flow control limits.
-    FlowControl,
-
-    /// The peer violated the local stream limits.
-    StreamLimit,
-
-    /// The specified stream was stopped by the peer.
-    ///
-    /// The error code sent as part of the `STOP_SENDING` frame is provided as
-    /// associated data.
-    StreamStopped(u64),
-
-    /// The specified stream was reset by the peer.
-    ///
-    /// The error code sent as part of the `RESET_STREAM` frame is provided as
-    /// associated data.
-    StreamReset(u64),
-
-    /// The received data exceeds the stream's final size.
-    FinalSize,
-
-    /// Error in congestion control.
-    CongestionControl,
-
-    /// Too many identifiers were provided.
-    IdLimit,
-
-    /// Not enough available identifiers.
-    OutOfIdentifiers,
-
-    /// Error in key update.
-    KeyUpdate,
-
-    /// The peer sent more data in CRYPTO frames than we can buffer.
-    CryptoBufferExceeded,
-
-    /// The peer sent an ACK frame with an invalid range.
-    InvalidAckRange,
-
-    /// The peer send an ACK frame for a skipped packet used for Optimistic ACK
-    /// mitigation.
-    OptimisticAckDetected,
-}
-
-/// QUIC error codes sent on the wire.
-///
-/// As defined in [RFC9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-error-codes).
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum WireErrorCode {
-    /// An endpoint uses this with CONNECTION_CLOSE to signal that the
-    /// connection is being closed abruptly in the absence of any error.
-    NoError              = 0x0,
-    /// The endpoint encountered an internal error and cannot continue with the
-    /// connection.
-    InternalError        = 0x1,
-    /// The server refused to accept a new connection.
-    ConnectionRefused    = 0x2,
-    /// An endpoint received more data than it permitted in its advertised data
-    /// limits; see Section 4.
-    FlowControlError     = 0x3,
-    /// An endpoint received a frame for a stream identifier that exceeded its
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -7862,6 +7918,26 @@ impl<F: BufFactory> Connection<F> {
                     self.tx_buffered =
                         self.tx_buffered.saturating_sub(unsent as usize);
 
+                    // These drops in qlog are a bit weird, but the only way to
+                    // ensure that all bytes that are moved from App to Transport
+                    // in stream_do_send are eventually moved from Transport to
+                    // Dropped.  Ideally we would add a Transport to Network
+                    // transition also as a way to indicate when bytes were
+                    // transmitted vs dropped without ever being sent.
```
LPardue, line 7701: Yeah we could add a to-network transition but 1) that adds even more log lines 2) STREAM frames already capture similar information and qlog tools can work with that to give us insighs we need.

I can live with things as they are for now.

quiche/src/lib.rs
```diff
@@ -8336,6 +8412,18 @@ impl<F: BufFactory> Connection<F> {
             cwin_available > 0
     }
 
+    fn check_tx_buffered_invariant(&mut self) {
```
LPardue, line 8190: one thought: do datagrams need any consideration in this calcuation? I think no, due to their different nature but just wanted to ask in case we were making an oversight.

quiche/src/lib.rs
```diff
@@ -8802,6 +8890,12 @@ pub struct Stats {
     /// Total duration during which this side of the connection was
     /// actively sending bytes or waiting for those bytes to be acked.
     pub bytes_in_flight_duration: Duration,
+
+    /// True if the connection detected inconsistencies in tx_buffered
+    /// which tracks the size of the connection's send buffer.
+    /// Inaccuracies in tx_buffered could lead to connection stalls or
+    /// excess buffering.
```
LPardue: nit: nearly all of these stats related to QUIC protocol concepts, while tx_buffered is an internal quiche detail. We could consider an alternative framing like

```
    /// Represents the status of quiche's connection send buffer tracking.
    /// Returns true  if there are consistency issues, which could lead to
    /// connection stalls  or excess buffering.
```

When presented like this, makes me wonder if we should expose the information as an enum such as 

```
pub enum TxBufferTrackingState {
  /// The send buffer is in a good state
  Ok,
  /// The send buffer is in an inconsistent state, which could lead to
  /// connection stalls  or excess buffering.
  Inconsistent,
}
```

Then we could simplify the description of the field on the stats object

quiche/src/lib.rs
```diff
@@ -8802,6 +8890,12 @@ pub struct Stats {
...
+    pub inconsistent_tx_buffered: bool,
```
LPardue: We have not done a great job of keeping the C Stats stuct in sync with the Rust one. Curoius if you wanted to add this to C or not.

quiche/src/lib.rs
antoniovicente: I don't know if datagrams increment tx_buffered.  They could increase it.  I don't think there's code to decrease on datagram ack or loss.  I know datagrams are not retransmitable.

quiche/src/lib.rs
antoniovicente: I'm happy with the current state.  We could consider options to enable more pedantic qlog but the current version is great.

quiche/src/lib.rs
antoniovicente: I see the structs in `ffi.rs`

No objections to trying to get the two sync'ed up.

quiche/src/lib.rs

quiche/src/lib.rs
antoniovicente: Possible to look into this in a followup?

quiche/src/lib.rs

quiche/src/lib.rs
LPardue: sure!

quiche/src/lib.rs
```diff
@@ -1449,6 +1449,17 @@ impl Config {
...
 }
 
+/// Tracks the health of the tx_buffered value.
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub enum TxBufferTrackingState {
+    /// The send buffer is in a good state
+    #[default]
+    Ok,
+    /// The send buffer is in an inconsistent state, which could lead to
+    /// connection stalls  or excess buffering.
```
LPardue: nit:

```suggestion
    /// connection stalls or excess buffering.
```

quiche/src/ffi.rs
```diff
@@ -1256,17 +1256,27 @@ pub struct Stats {
...
     stopped_stream_count_local: u64,
     reset_stream_count_remote: u64,
     stopped_stream_count_remote: u64,
+    data_blocked_sent_count: u64,
+    stream_data_blocked_sent_count: u64,
+    data_blocked_recv_count: u64,
+    stream_data_blocked_recv_count: u64,
+    path_challenge_rx_count: u64,
+    bytes_in_flight_duration_msec: u64,
+    tx_buffered_inconsistent: bool,
```
LPardue, line 1279: you'll need to update the C def of `quiche_stats` in include/quiche.h too

antoniovicente: yikes...

antoniovicente: Done.

quiche/src/lib.rs
```diff
@@ -8336,6 +8422,18 @@ impl<F: BufFactory> Connection<F> {
...
+    fn check_tx_buffered_invariant(&mut self) {
+        if self.tx_buffered > 0 &&
+            !self.streams.has_flushable() &&
+            !self
+                .paths
+                .iter()
+                .any(|(_, p)| p.recovery.bytes_in_flight() > 0)
+        {
+            self.tx_buffered_state = TxBufferTrackingState::Inconsistent;
+        }
```
toidiu, line 8207: Can you add a comment explaining what this is checking.

quiche/src/lib.rs
```diff
@@ -1449,6 +1449,17 @@ impl Config {
...
 
+/// Tracks the health of the tx_buffered value.
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub enum TxBufferTrackingState {
+    /// The send buffer is in a good state
+    #[default]
+    Ok,
+    /// The send buffer is in an inconsistent state, which could lead to
+    /// connection stalls or excess buffering.
+    Inconsistent,
```
toidiu, line 1225: Can you expand on why the state might be inconsistent. You have some info in your pr description.

e.g. This can happen if we have bytes in the tx buffer but the stream is closed.

quiche/src/lib.rs
```diff
@@ -4107,6 +4150,7 @@ impl<F: BufFactory> Connection<F> {
                 }
             }
         }
+        self.check_tx_buffered_invariant();
```
toidiu, line 3918: Does it make sense to check invariance each time we interact with `self.tx_buffer`?

You could create a new type for tx_buffer and do automatically.

```
struct TxBuffer {
  buffered: usize
}

impl TxBuffer {
  fn add() {
     // logic
     // check invariance
  }

  fn sub() {
     // logic
     // check invariance
  }
}
```

LPardue (approved):

quiche/src/lib.rs
antoniovicente: Checking invariant requires external state for the connection.

The invariant checks that I added run after handling ACKs and detecting lost packets; they are not directly tied to add/sub.

quiche/src/lib.rs
antoniovicente: This PR fixes the known bugs so stream close will no longer cause a discrepancy in tx_buffered

quiche/src/lib.rs
antoniovicente: Done.

quiche/src/lib.rs
toidiu: I think it might be worth checking on each operation so you dont have to manually do it in specific places. But we can revisit that in the future.

PR #2303, opened 2026-01-08
Add prometheus metrics to buffer-pool

sharrington22:
Add a few metrics to the buffer-pool crate, track total number of objects/bytes allocated or in a pool.

### DIFF
--- a/buffer-pool/src/buffer.rs
+++ b/buffer-pool/src/buffer.rs
@@ -27,6 +27,7 @@
 use std::ops::Deref;
 use std::ops::DerefMut;
 
+use crate::buffer_pool::consume_buffer_total_bytes;
 use crate::Reuse;
 
 /// A convinience wrapper around Vec that allows to "consume" data from the
@@ -58,8 +59,10 @@ impl DerefMut for ConsumeBuffer {
 
 impl Reuse for ConsumeBuffer {
     fn reuse(&mut self, val: usize) -> bool {
+        let old_capacity = self.inner.capacity();
         self.inner.clear();
         self.inner.shrink_to(val);
+        self.update_metrics_after_resize(old_capacity);
         self.head = 0;
         self.inner.capacity() > 0
     }
@@ -69,13 +72,25 @@ impl Reuse for ConsumeBuffer {
     }
 }
 
+impl Drop for ConsumeBuffer {
+    fn drop(&mut self) {
+        consume_buffer_total_bytes().dec_by(self.inner.capacity() as u64);
+    }
+}
+
 impl ConsumeBuffer {
     pub fn from_vec(inner: Vec<u8>) -> Self {
+        consume_buffer_total_bytes().inc_by(inner.capacity() as u64);
         ConsumeBuffer { inner, head: 0 }
     }
 
-    pub fn into_vec(self) -> Vec<u8> {
-        let mut inner = self.inner;
+    pub fn into_vec(mut self) -> Vec<u8> {
+        // As ConsumeBuffer has a Drop impl we have use `take` to get at the inner
+        // vec out instead of being able to move it out directly. This also means
+        // we need to update the metrics by hand as the buffer is no longer owned
+        // by us.
+        let mut inner = std::mem::take(&mut self.inner);
+        consume_buffer_total_bytes().dec_by(inner.capacity() as u64);
         inner.drain(0..self.head);
         inner
     }
@@ -86,7 +101,9 @@ impl ConsumeBuffer {
     }
 
     pub fn expand(&mut self, count: usize) {
+        let old_capacity = self.inner.capacity();
         self.inner.reserve_exact(count);
+        self.update_metrics_after_resize(old_capacity);
         // SAFETY: u8 is always initialized and we reserved the capacity.
         unsafe { self.inner.set_len(count) };
     }
@@ -105,10 +122,67 @@ impl ConsumeBuffer {
 
         true
     }
+
+    fn update_metrics_after_resize(&mut self, old_capacity: usize) {
+        let new_capacity = self.inner.capacity();
+        if new_capacity < old_capacity {
+            consume_buffer_total_bytes()
+                .dec_by((old_capacity - new_capacity) as u64);
+        } else if new_capacity > old_capacity {
+            consume_buffer_total_bytes()
+                .inc_by((new_capacity - old_capacity) as u64);
+        }
+    }
 }
 
 impl<'a> Extend<&'a u8> for ConsumeBuffer {
     fn extend<T: IntoIterator<Item = &'a u8>>(&mut self, iter: T) {
-        self.inner.extend(iter)
+        let old_capacity = self.inner.capacity();
+        self.inner.extend(iter);
+        self.update_metrics_after_resize(old_capacity);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_metrics() {
+        {
+            let mut buf = ConsumeBuffer::default();
+            assert_eq!(buf.inner.capacity(), 0);
+            assert_eq!(consume_buffer_total_bytes().get(), 0);
+
+            buf.extend(&[0, 1, 2, 3, 4]);
+            assert_eq!(
+                consume_buffer_total_bytes().get(),
+                buf.inner.capacity() as u64
+            );
+        }
+        // drop buf
+        assert_eq!(consume_buffer_total_bytes().get(), 0);
+
+        let mut buf_a = ConsumeBuffer::from_vec(vec![0, 1, 2, 3, 4]);
+        let buf_b = ConsumeBuffer::from_vec(vec![5, 6, 7]);
+        assert_eq!(
+            consume_buffer_total_bytes().get(),
+            (buf_a.inner.capacity() + buf_b.inner.capacity()) as u64
+        );
+
+        buf_a.expand(100000);
+        assert_eq!(
+            consume_buffer_total_bytes().get(),
+            (buf_a.inner.capacity() + buf_b.inner.capacity()) as u64
+        );
+
+        buf_b.into_vec();
+        assert_eq!(
+            consume_buffer_total_bytes().get(),
+            buf_a.inner.capacity() as u64
+        );
+
+        drop(buf_a);
+        assert_eq!(consume_buffer_total_bytes().get(), 0);
     }
 }
--- a/buffer-pool/src/lib.rs
+++ b/buffer-pool/src/lib.rs
@@ -49,6 +49,11 @@ pub mod buffer_pool {
     pub fn pool_idle_bytes(name: &'static str) -> Gauge;
     /// Number of objects currently active and in-use.
     pub fn pool_active_count(name: &'static str) -> Gauge;
+    /// Total number of bytes allocated across all `ConsumeBuffer` objects.
+    ///
+    /// We're not able to track this with better granularity because
+    /// the ConsumeBuffers may be resized, and they don't know their pools.
+    pub fn consume_buffer_total_bytes() -> Gauge;
 }
 
 /// A sharded pool of elements.


buffer-pool/src/lib.rs
```diff
@@ -35,9 +35,26 @@ use std::sync::atomic::Ordering;
...
+    /// Total number of objects available in the pool.
+    pub fn pool_objects(name: &'static str) -> Gauge;
+    /// Total number of bytes used by available objects.
+    pub fn pool_memory(name: &'static str) -> Gauge;
+    /// Total number of live objects, ie objects allocated out of pools.
+    pub fn live_objects(name: &'static str) -> Gauge;
+    /// Total number of bytes allocated across all `ConsumeBuffer` objects.
+    /// We're not able to track this with better granularity because
+    /// the ConsumeBuffers may be resized, and they don't know their pools.
+    pub fn total_consume_buffer_memory() -> Gauge;
```
toidiu: nit:
- Space after the first line. The first line is used as a summary in docs. [eg](https://docs.rs/quiche/latest/quiche/struct.Config.html#method.load_cert_chain_from_pem_file)

```suggestion
    /// Total number of bytes allocated across all `ConsumeBuffer` objects.
    ///
    /// We're not able to track this with better granularity because
    /// the ConsumeBuffers may be resized, and they don't know their pools.
    pub fn total_consume_buffer_memory() -> Gauge;
```

tokio-quiche/src/buf_factory.rs
```diff
@@ -55,16 +55,20 @@ const MAX_POOL_BUF_SIZE: usize = 64 * 1024;
 
 type BufPool = Pool<POOL_SHARDS, ConsumeBuffer>;
 
-static TINY_POOL: BufPool = BufPool::new(TINY_BUF_SIZE, TINY_BUF_SIZE);
-static SMALL_POOL: BufPool = BufPool::new(SMALL_BUF_SIZE, SMALL_BUF_SIZE);
-static MEDIUM_POOL: BufPool = BufPool::new(MEDIUM_BUF_SIZE, MEDIUM_BUF_SIZE);
+static TINY_POOL: BufPool =
```
toidiu, line 58: nit: I think this crate is actually more general than QUIC so maybe we remove the `quic_` prefix?

buffer-pool/src/lib.rs
```diff
@@ -35,9 +35,26 @@ use std::sync::atomic::Ordering;
...
 use crossbeam::queue::SegQueue;
 
+use foundations::telemetry::metrics::metrics;
+use foundations::telemetry::metrics::Gauge;
+
 pub use crate::buffer::*;
 pub use crate::raw_pool_buf_io::*;
 
+#[metrics]
+pub mod buffer_pool {
```
toidiu: do we know what the metric name will be? Will it be `buffer_pool_metric` or `buffer_pool_buffer_pool_metric`?

https://docs.rs/foundations/latest/foundations/telemetry/metrics/attr.metrics.html

There is also `#[metrics(unprefixed)]`

buffer-pool/src/lib.rs
```diff
@@ -91,11 +116,16 @@ impl<T: Default + Reuse> Drop for Pooled<T> {
...
+            name,
         } = self.pool;
+        // The memory associated with this object is no longer live.
+        buffer_pool::live_objects(name).dec();
         if self.inner.reuse(*trim) {
             if elem_cnt.fetch_add(1, Ordering::Acquire) < *max {
                 // If returning the element to the queue would not exceed max
                 // number of elements, return it
+                buffer_pool::pool_objects(name).inc();
+                buffer_pool::pool_memory(name).inc_by(self.inner.size() as u64);
```
toidiu: Do we also need to do accounting below since we do `elem_cnt.fetch_sub(1, Ordering::Release);`

buffer-pool/src/lib.rs
```diff
@@ -35,9 +35,26 @@ use std::sync::atomic::Ordering;
...
 pub use crate::raw_pool_buf_io::*;
 
+#[metrics]
+pub mod buffer_pool {
+    /// Total number of objects available in the pool.
+    pub fn pool_objects(name: &'static str) -> Gauge;
+    /// Total number of bytes used by available objects.
+    pub fn pool_memory(name: &'static str) -> Gauge;
+    /// Total number of live objects, ie objects allocated out of pools.
+    pub fn live_objects(name: &'static str) -> Gauge;
```
toidiu: What is "live" exactly?
I am missing something with "allocated out of pools" means.

buffer-pool/src/buffer.rs
```diff
@@ -58,20 +59,34 @@ impl DerefMut for ConsumeBuffer {
...
     pub fn from_vec(inner: Vec<u8>) -> Self {
+        total_consume_buffer_memory().inc_by(inner.capacity() as u64);
         ConsumeBuffer { inner, head: 0 }
     }
 
-    pub fn into_vec(self) -> Vec<u8> {
-        let mut inner = self.inner;
+    pub fn into_vec(mut self) -> Vec<u8> {
+        let mut inner = std::mem::take(&mut self.inner);
+        total_consume_buffer_memory().dec_by(inner.capacity() as u64);
```
toidiu: why this refactor to `mem::take`?

buffer-pool/src/buffer.rs
```diff
@@ -101,10 +118,66 @@ impl ConsumeBuffer {
...
+        assert_eq!(
+            total_consume_buffer_memory().get(),
+            (buf_a.inner.capacity() + buf_b.inner.capacity()) as u64
+        );
+
+        buf_a.expand(100000);
+        assert_eq!(
+            total_consume_buffer_memory().get(),
+            (buf_a.inner.capacity() + buf_b.inner.capacity()) as u64
+        );
```
toidiu, line 177: Any way to assert that mem grew by `100000`

buffer-pool/src/buffer.rs
```diff
@@ -101,10 +118,66 @@ impl ConsumeBuffer {
...
+        assert_eq!(total_consume_buffer_memory().get(), 0);
+
+        buf.extend(&[0, 1, 2, 3, 4]);
+        assert_eq!(
+            total_consume_buffer_memory().get(),
+            buf.inner.capacity() as u64
+        );
+
+        drop(buf);
+        assert_eq!(total_consume_buffer_memory().get(), 0);
```
toidiu: Instead of calling `drop` directly it might be more realistic to mimic drop like it will happen in real code. This also helps ensure you don't use buf after drop and IMO make reading test a bit easier.

```
{
    let mut buf = ConsumeBuffer::default();
    ...

    // auto drop `buf` at the end of this scope
}
assert_eq!(total_consume_buffer_memory().get(), 0);
```

buffer-pool/src/buffer.rs
sharrington22: I've added a drop impl to `ConsumeBuffer`, which stops into_vec from being able to move the inner vec out directly. The whole object needs to continue to exist so drop can run, so `take` is used instead, swapping the inner vec out for a default vec

buffer-pool/src/lib.rs
sharrington22: This is definitely not as clear as it could be, will think how I can rephrase this better! Object lifecycles:
1. Not yet allocated/Doesn't exist
2. In use
3. Returned to one of the pools for reallocation (can become 2 again)
4. Freed back to the system
This is tracking objects in category 2

buffer-pool/src/buffer.rs
sharrington22: Unfortunately not in any consistent way, it depends on how Vec & the underlying memory allocator allocates and extends the buffer. It's incredibly unlikely, but the original allocation might be large enough to satisfy the expansion. But it doesn't really matter, we're only testing that the metrics stay consistent with the buffer capacities.

buffer-pool/src/lib.rs
sharrington22: Nope, the accounting was done above - the `buffer_pool::live_objects(name).dec();`. These lines are only when we return an object to the buffer

buffer-pool/src/lib.rs
sharrington22: I think you're right, it'll end up as `buffer_pool_buffer_pool_metric`. I've added the `unprefixed` flag.

buffer-pool/src/lib.rs
sharrington22: I've renamed the metrics to hopefully make them more clear now

buffer-pool/Cargo.toml
```diff
@@ -10,3 +10,6 @@ description = "Pooled in-memory buffers"
 
 [dependencies]
 crossbeam = { workspace = true, features = ["alloc"] }
+foundations = { workspace = true, default-features = false, features = [
+  "server-client-common-default",
```
ghedo: We only really need `metrics` here I think.

buffer-pool/src/buffer.rs
```diff
@@ -58,20 +59,34 @@ impl DerefMut for ConsumeBuffer {
...
     fn reuse(&mut self, val: usize) -> bool {
+        let old_capacity = self.inner.capacity();
         self.inner.clear();
         self.inner.shrink_to(val);
+        self.update_metrics_after_resize(old_capacity);
         self.head = 0;
         self.inner.capacity() > 0
     }
+
+    fn size(&self) -> usize {
```
antoniovicente: Should this be capacity instead of size?

buffer-pool/src/lib.rs
```diff
@@ -229,17 +274,20 @@ mod tests {
...
             4,
+            POOL_NAME,
         )));
 
         let bufs = (0..SHARDS * 4).map(|_| pool.get()).collect::<Vec<_>>();
 
         for shard in pool.queues.iter() {
             assert_eq!(shard.elem_cnt.load(Ordering::Relaxed), 0);
         }
+        assert_eq!(buffer_pool::pool_idle_count(POOL_NAME).get(), 0);
```
antoniovicente, line 288: Would be good to have some assertions for `pool_active_count` and `pool_idle_bytes`

buffer-pool/src/lib.rs
```diff
@@ -35,9 +35,27 @@ use std::sync::atomic::Ordering;
...
 use crossbeam::queue::SegQueue;
 
+use foundations::telemetry::metrics::metrics;
+use foundations::telemetry::metrics::Gauge;
+
 pub use crate::buffer::*;
 pub use crate::raw_pool_buf_io::*;
 
+#[metrics(unprefixed)]
+pub mod buffer_pool {
```
toidiu, line 45: Would love it if we could this into its own file `metrics.rs` so there is less clutter here.

buffer-pool/src/lib.rs
```diff
@@ -171,7 +203,8 @@ impl<const S: usize, T: Default + Reuse> Pool<S, T> {
     pub fn from_owned(&'static self, inner: T) -> Pooled<T> {
         let shard = self.next_shard.fetch_add(1, Ordering::Relaxed) % S;
         let shard = &self.queues[shard];
-        Pooled { inner, pool: shard }
+        buffer_pool::pool_active_count(shard.name).inc();
+        Pooled::new(inner, shard)
```
toidiu: If I am understanding it correctly, this might be incrementing twice.? In this fn and also in Pooled::new(). Do we need a test for this?

Same above in `get_empty`

buffer-pool/src/buffer.rs
toidiu: This took me a while to understand so might benefit from a comment.

There is a slight risk if we ever add more logic to the Drop impl.

buffer-pool/src/lib.rs
```diff
@@ -35,9 +35,27 @@ use std::sync::atomic::Ordering;
...
+    pub fn pool_idle_count(name: &'static str) -> Gauge;
+    /// Memory footprint of objects currently in the pool.
+    pub fn pool_idle_bytes(name: &'static str) -> Gauge;
+    /// Number of objects currently active and in-use.
+    pub fn pool_active_count(name: &'static str) -> Gauge;
+    /// Total number of bytes allocated across all `ConsumeBuffer` objects.
+    ///
+    /// We're not able to track this with better granularity because
+    /// the ConsumeBuffers may be resized, and they don't know their pools.
+    pub fn consume_buffer_total_bytes() -> Gauge;
```
toidiu, line 56: Naming:

"Idle" seems odd to me in the context of a memory pool. How about "used" and "available"?

```suggestion
    /// Number of objects available for reuse in the pool.
    pub fn pool_available_obj_count(name: &'static str) -> Gauge;
    /// Total number of bytes available for reuse in the pool.
    pub fn pool_available_bytes(name: &'static str) -> Gauge;
    /// Number of objects currently active and in-use.
    pub fn pool_used_obj_count(name: &'static str) -> Gauge;
    /// Total number of bytes allocated across all `ConsumeBuffer` objects.
    ///
    /// We're not able to track this with better granularity because
    /// the ConsumeBuffers may be resized, and they don't know their pools.
    pub fn consume_buffer_used_bytes() -> Gauge;
    
```

buffer-pool/src/buffer.rs
sharrington22: Yep, fixed

buffer-pool/src/lib.rs
sharrington22: Done

buffer-pool/src/lib.rs
sharrington22: Yes, good catch! I've fixed and added tests

buffer-pool/src/lib.rs
sharrington22: A separate file feels like quite a lot for 14 lines of code

buffer-pool/src/lib.rs
sharrington22: idle/active are fairly common terms in this context

toidiu:
Looks good! Thanks for getting it across the line.

PR #2304, opened 2026-01-08
tokio-quiche: attach additional context when propagating errors.

toidiu:


tokio-quiche/src/result.rs
```diff
@@ -50,14 +51,68 @@ pub trait QuicResultExt<T, E> {
...
+                error_code: err.error_code,
+            });
+
+            CloseError {
+                work_loop_err: err.into(),
+                handshake_complete: qconn.is_established(),
+                local_err,
+                peer_err,
+            }
+            .into()
```
toidiu: **Problem:** I am wondering how useful this will be in practice. Many of the quiche and h3 Errors codes are rewritten to the WireErrorCode.

- h3 seems to preserve 1-1 [translation](https://github.com/cloudflare/quiche/blob/5852965d6f7c02e1c6e5dd1fd7bfd6678c038c99/quiche/src/h3/mod.rs#L483).
- quiche on the other hand maps a lot of the errors to [ProtocolViolation](https://github.com/cloudflare/quiche/blob/master/quiche/src/error.rs#L182). This is done to make it harder for attackers to inspect the internal protocol state.

**Potential solution:**
We could store the original error code and use that for logging here. However, it might be worth deploying this first and seeing if this provides sufficient info.

tokio-quiche/src/result.rs
```diff
@@ -50,14 +51,80 @@ pub trait QuicResultExt<T, E> {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+impl std::error::Error for CloseError {}
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub struct ConnectionError {
```
evanrittenhouse: Can we use `quiche::ConnectionError`?

toidiu: https://github.com/cloudflare/quiche/blob/master/quiche/src/error.rs#L258

I choose not to because that has a `reason: Vec<u8>` which could be useful but also would require allocation via clone. Maybe we do want it and its ok to pay the cost since the connection is closing?

toidiu: Auditing our usage looks like we generate small enough strings and also there is a max len ([RFC](https://www.rfc-editor.org/rfc/rfc9000.html#section-19.19-6.6.1 )) on the frame we receive so seems ok to clone and include it.

tokio-quiche/src/result.rs
```diff
@@ -50,14 +52,67 @@ pub trait QuicResultExt<T, E> {
...
+        })
+    }
 }
+
+#[derive(Debug)]
+#[allow(
+    dead_code,
+    reason = "Only used for logging so fields are not directly accessed."
+)]
+pub struct CloseError {
```
antoniovicente: Would it be possible to add a test and/or example to show the value of this change?  It's hard to understand the value of this PR without examples.

toidiu: At the moment the only place we use the error is for logging:
- logging: https://github.com/cloudflare/quiche/blob/bda63e05c472dce0e00cd31fb9667f702c10634b/tokio-quiche/src/http3/driver/mod.rs#L1224

We might add them to metrics but want to see what info is useful.

antoniovicente: Got some example output?

toidiu: Was able to add a test to demonstrate extracting Close error.

tokio-quiche/tests/integration_tests/connection_close.rs
```diff
@@ -168,3 +179,86 @@ async fn test_no_connection_close_frame_on_idle_timeout() -> QuicResult<()> {
...
+        let audit_stats = Arc::clone(h3_over_quic.audit_log_stats());
+        *audit_logs_clone.write().unwrap() = Some(audit_stats);
+        let _ = tokio::spawn(async move {
+            handle_connection(h3_over_quic).await;
+        })
+        .await;
+    });
+
+    let actions = vec![
+        Action::SendDatagram { payload: vec![] },
```
LPardue, line 234: The error below seems to be for an invalid DATA frame, so it seems like we don't need this datagram action.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -338,12 +339,17 @@ where
...
-            );
+            let outcome = self
+                .write_packet_to_buffer(
+                    qconn,
+                    send_buf,
+                    &mut send_info,
+                    segment_size,
+                )
+                .map_err(|err| {
+                    let msg = "Error writing packet. Opt: `AOQ::write_packet_to_buffer`";
```
LPardue, line 350: what does `Opt` mean?

toidiu: operation. i'll change the format so its more obvious.

tokio-quiche/tests/integration_tests/connection_close.rs
toidiu: The error seems to trigger only when I first send a Datagram frame. I think this works since I can send an arbitrary payload in the datagram frame and therefore mess with the server state.

toidiu:
replaced with https://github.com/cloudflare/quiche/pull/2307

PR #2306, opened 2026-01-12
fix race condition that caused the StreamStopped error to be lost

ghedo:
Currently if a STOP_SENDING frame is received, a RESET_STREAM frame is sent in response and the corresponding stream is marked as writable so that the STOP_SENDING error can be forwarded to the application.

However, the stream's state is collected once the RESET_STREAM frame is acked, and if the application hasn't yet received the corresponding error, it will never get it again.

Instead of collecting the stream immediately on receiving the RESET_STREAM ack, wait until the `StreamStopped` error is returned to the application (i.e. on calling `stream_send()`, `stream_capacity()` or `stream_writable()`).

The downside is that because of the need to modify the connection state to collect a stream, `stream_capacity()` now needs a mutable reference to the connection object, which is a breaking change.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3343,10 +3343,35 @@ impl<F: BufFactory> Connection<F> {
 
                         stream.send.ack_and_drop(offset, length);
 
+                        let priority_key = Arc::clone(&stream.priority_key);
+
                         // Only collect the stream if it is complete and not
-                        // readable. If it is readable, it will get collected when
-                        // stream_recv() is used.
-                        if stream.is_complete() && !stream.is_readable() {
+                        // readable or writable.
+                        //
+                        // If it is readable, it will get collected when
+                        // stream_recv() is next used.
+                        //
+                        // If it is writable, it might mean that the stream
+                        // has been stopped by the peer (i.e. a STOP_SENDING
+                        // frame is received), in which case before collecting
+                        // the stream we will need to propagate the
+                        // `StreamStopped` error to the application. It will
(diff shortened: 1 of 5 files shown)


quiche/src/lib.rs
```diff
@@ -3344,8 +3344,8 @@ impl<F: BufFactory> Connection<F> {
                         stream.send.ack_and_drop(offset, length);
 
                         // Only collect the stream if it is complete and not
-                        // readable. If it is readable, it will get collected when
-                        // stream_recv() is used.
+                        // readable. If it is readable, it will get collected
+                        // when stream_recv() is used.
                         if stream.is_complete() && !stream.is_readable() {
                             let local = stream.local;
                             self.streams.collect(stream_id, local);
```
antoniovicente, line 3376: You're still triggering collect() here when STOP_SENDING is received.

The app code on the server side is never woken up after the client sends a STOP_SENDING.

antoniovicente: `if stream.is_complete() && !stream.is_readable() && !is_writable {` seems to help in my test.

ghedo: Not sure I follow, this is part of the `frame::Frame::StreamHeader` branch, not STOP_SENDING...

ghedo: Oh, do you mean if both STREAM and STOP_SENDING are acked together?

ghedo: This should be fixed now. I also tightened the `is_writable` condition a bit to ensure that its only hit if the stream has actually been stopped.

PR #2307, opened 2026-01-12
tokio-quiche: attach additional context when logging errors.

toidiu:


tokio-quiche/tests/integration_tests/connection_close.rs
```diff
@@ -168,3 +178,76 @@ async fn test_no_connection_close_frame_on_idle_timeout() -> QuicResult<()> {
...
+    // Close error is also exposed from the Audit Log.
+    let audit_log = audit_log.read().unwrap();
+    let connection_close_reason =
+        audit_log.as_ref().unwrap().connection_close_reason();
+
+    let tq_error = (*connection_close_reason).as_ref().unwrap();
+    assert_eq!(
+        tq_error.to_string(),
+        "app_err=H3ConnectionError: TransportError(BufferTooShort) while processing data via AOQ::process_reads"
+    );
```
toidiu: CI fails here.. but it passes locally. So there is some non-determinism here. Also not sure if we need this test since its kinda complicated and only tests that we now add some error context.

toidiu: Removed the flaky test in https://github.com/cloudflare/quiche/pull/2307/commits/51d9b04086309b3609246933e9ae4d514029f1c6

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -338,12 +339,18 @@ where
...
+            let outcome = self
+                .write_packet_to_buffer(
+                    qconn,
+                    send_buf,
+                    &mut send_info,
+                    segment_size,
+                )
+                .map_err(|err| {
+                    to_box_error(
+                        format!("app_err={} while writing packets with AOQ::write_packet_to_buffer", err),
```
evanrittenhouse:
```suggestion
                        format!("app_err={} while writing packets with IOW::write_packet_to_buffer", err),
```

ghedo (approved):

PR #2308, opened 2026-01-13
Fix probe_up_acked adjustment in probe_inflight_high_upward to match the BBR RFC draft

antoniovicente:
probe_up_acked was expected to be in [0, probe_up_bytes) but could end up larger than probe_up_bytes in cases where delta > 1.

The effect of this bug is that inflight_hi would be adjusted upwards too fast in cases where delta > 1.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
+++ b/quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
@@ -613,10 +613,13 @@ impl ProbeBW {
             self.cycle.probe_up_acked += congestion_event.bytes_acked;
         }
 
-        if let Some(probe_up_bytes) = self.cycle.probe_up_bytes.as_mut() {
-            if self.cycle.probe_up_acked >= *probe_up_bytes {
-                let delta = self.cycle.probe_up_acked / *probe_up_bytes;
-                self.cycle.probe_up_acked -= delta * *probe_up_bytes;
+        if let Some(probe_up_bytes) = self.cycle.probe_up_bytes {
+            if self.cycle.probe_up_acked >= probe_up_bytes {
+                let delta = self.cycle.probe_up_acked / probe_up_bytes;
+                // probe_up_acked is set to the remainder mod probe_up_bytes;
+                // probe_up_acked >= delta * probe_up_bytes so this
+                // can't underflow.
+                self.cycle.probe_up_acked -= delta * probe_up_bytes;
                 let new_inflight_hi =
                     self.model.inflight_hi() + delta * DEFAULT_MSS;
                 if new_inflight_hi > self.model.inflight_hi() {


quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
```diff
@@ -616,7 +616,7 @@ impl ProbeBW {
         if let Some(probe_up_bytes) = self.cycle.probe_up_bytes.as_mut() {
             if self.cycle.probe_up_acked >= *probe_up_bytes {
                 let delta = self.cycle.probe_up_acked / *probe_up_bytes;
-                self.cycle.probe_up_acked -= *probe_up_bytes;
+                self.cycle.probe_up_acked -= delta * *probe_up_bytes;
```
antoniovicente: Code from the draft: https://github.com/ietf-wg-ccwg/draft-ietf-ccwg-bbr/blob/c4093de/draft-ietf-ccwg-bbr.md?plain=1#L2224

ghedo: Also at https://github.com/google/quiche/blob/main/quiche/quic/core/congestion_control/bbr2_probe_bw.cc#L390 FWIW.

LPardue: given the introduction of a muliplication factor, is there any risk this could underflow?

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
```diff
@@ -616,7 +616,7 @@ impl ProbeBW {
         if let Some(probe_up_bytes) = self.cycle.probe_up_bytes.as_mut() {
```
LPardue: While we're touch code, it doesn't seem we need mutability here
```suggestion
        if let Some(probe_up_bytes) = self.cycle.probe_up_bytes.as_ref() {
```

ghedo: Yeah, was thinking the same. It probably doesn't need to be a reference either since the value can just be copied (so wouldn't need derefence with `*` anymore), but not 100% sure.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
LPardue: I guess no since delta is always a sub-proportion of probe_up_acked. Please confirm my maths after a long day of work.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
LPardue: avoiding the deref would avoid me thinging the maths was `**` on first inspection :joy:

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
antoniovicente: added a comment.

quiche/src/recovery/gcongestion/bbr2/probe_bw.rs
antoniovicente: Good call.

PR #2309, opened 2026-01-14
Temporarily pin boring-sys to 4.19

LPardue:


Cargo.toml
```diff
@@ -78,3 +78,6 @@ debug = true
 
 [profile.release]
 debug = true
+
+[patch.crates-io]
+boring-sys = { version = "4.19.0" }
```
antoniovicente: Could use a comment explaining why.

Cargo.toml
```diff
@@ -78,3 +78,8 @@ debug = true
 
 [profile.release]
 debug = true
+
+# temporary workaround for quiche_windows (i686-pc-windows-gnu) failures
+# observed in boring-sys 4.20
+[patch.crates-io]
+boring-sys = { git = "https://github.com/cloudflare/boring", tag = "v4.19.0" }
```
ghedo: Why not actually pin the dependency? i.e. add `boring-sys = "=4.19.0"` to `workspace.dependencies`.

ghedo: Probably need to pin `boring` too.

ghedo: Oh, lol, I think you need to add `boring-sys` to `quiche/Cargo.toml` `dependencies too, otherwise nothing is using it :facepalm:

LPardue: Hmm, new failure

```
Analyzing dependencies of crates in this directory...
cargo-machete found the following unused dependencies in this directory:
quiche -- ./quiche/Cargo.toml:
	boring-sys
Done!
```

LPardue: Worked around it by constraining the dep to the windows target. It seems janky but works. This whole this is just a temporary fix.

PR #2311, opened 2026-01-16
Fix ABI tests on MinGW32

kornelski:


LPardue:
You might need to include a revert of the pin here? https://github.com/cloudflare/quiche/commit/c6a0d4c9cbfb8d710d4db0363b86cfb04b4a77e2

LPardue:
Sorry, I meant a literal rever, since the addition of the boring-sys lines to Cargo.toml was required to force the pin to work. So can you remove those, especially the one in quiche/Cargo.toml

PR #2312, opened 2026-01-16
Use now instead of get_next_release_time to compute time_sent for BBR packets

antoniovicente:
The packet send time is computed deep in on_packet_sent, and this time differs from the one passed to the GSO burst which is based on the first packet in the train. When ReleaseDecision.can_burst() returns true, the difference between the real send time and the one stored in the recovery module can grow arbitrarily large resulting in very low RTT estimates which BBR holds onto as minRTT from there on. Incorrect RTT will result in the congestion window being set to artificially small values because of incorrect BDP calculations since BDP = minRTT * BW_estimate.

Setting time_sent to now provides a more stable reference time that guarantees that the computed RTT is never under estimated, which is critical for BBR to operate correctly.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -169,6 +169,13 @@ jobs:
     strategy:
       matrix:
         target: ["x86_64-pc-windows-msvc", "i686-pc-windows-msvc", "x86_64-pc-windows-gnu", "i686-pc-windows-gnu"]
+        include:
+          - target: "i686-pc-windows-gnu"
+            cflags: "-mlong-double-64" # bindgen detects that Rust doesn't support 80-bit long double
+    env:
+      CFLAGS: ${{ matrix.cflags }}
+      CXXFLAGS: ${{ matrix.cflags }}
+      BINDGEN_EXTRA_CLANG_ARGS: ${{ matrix.cflags }}
     # Only run on "pull_request" event for external PRs. This is to avoid
     # duplicate builds for PRs created from internal branches.
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
@@ -201,13 +208,6 @@ jobs:
         uses: crazy-max/ghaction-chocolatey@v3
         with:
           args: install nasm
-
-      # https://github.com/rust-lang/cmake-rs/pull/259 breaks handling of long Windows paths
-      # https://github.com/cloudflare/boring/issues/414
-      # https://github.com/0x676e67/boring/commit/efacedb5bf409d0d773e8ce4f5080cb4b4c10f54
-      - name: Pin some dependencies, temporary patch for cmake breakage
-        run: |
-          cargo update --package cmake --precise 0.1.54
 
       - name: Run cargo build
         if: endsWith(matrix.target, '-gnu')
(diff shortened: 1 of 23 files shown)


quiche/src/recovery/congestion/recovery.rs
```diff
@@ -1027,7 +1027,7 @@ impl RecoveryOps for LegacyRecovery {
     fn get_next_release_time(&self) -> ReleaseDecision {
         ReleaseDecision {
             time: ReleaseTime::Immediate,
-            allow_burst: false,
+            allow_burst: true,
```
jrouviere: I notice you have a separate commit for this, is that related to the rest of the PR? If not, perhaps we should keep it separate so we can identify any regression quicker.

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -492,6 +507,10 @@ impl BBRv2 {
         }
     }
 
+    pub fn use_next_release_time_as_time_sent(&self) -> bool {
+        !self.params.set_time_sent_to_now
+    }
```
jrouviere: I find it a bit confusing that in some part of the code we use "next_release_time_as_time_sent" which is the opposite of "set_time_sent_to_now", unless I misunderstood the intent.

Could we use the same option name everywhere?

quiche/src/recovery/congestion/recovery.rs
antoniovicente: I'll come back to this at a future time when I try to cleanup the workloop implementation.  This mostly doesn't matter right now except if someone where to use CUBIC with the gcongestion rust feature.

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: Switched to set_time_sent_to_now everywhere.

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -330,6 +342,8 @@ const DEFAULT_PARAMS: Params = Params {
     scale_pacing_rate_by_mss: false,
 
     disable_probe_down_early_exit: false,
+
+    set_time_sent_to_now: true,
```
evanrittenhouse: It may make sense to default this to `false` until we're confident that this doesn't have unintended regressions, considering that we already have some BBR overrides in place

antoniovicente: Could go either way on this one.  Note that the API to override this config is relatively awkward and plumbing of this config through the system is only available on one of the proxies; we also want this fix to the other proxies in relatively short order.

quiche/src/recovery/gcongestion/bbr2.rs
```diff
@@ -492,6 +507,10 @@ impl BBRv2 {
         }
     }
 
+    pub fn set_time_sent_to_now(&self) -> bool {
```
LPardue: it this a getter that is has a  `set` prefiex? Seems weird.

quiche/src/recovery/gcongestion/bbr2.rs
LPardue: no strong opinion from me, seem configurable either way

quiche/src/recovery/gcongestion/bbr2.rs
antoniovicente: Good catch.  Renamed.

I think this is why I went for the two names for the same thing earlier in the PR.

PR #2313, opened 2026-01-16
release: quiche, tokio-quiche, h3i

toidiu:


ghedo:
Since https://github.com/cloudflare/quiche/pull/2303 was merged in the meantime I think we need to do this all again...

PR #2314, opened 2026-01-20
release: quiche, tokio-quiche, h3i, buffer-pool

toidiu:


antoniovicente:
I think the order of the release commits may want tweaking; tokio-quiche has a dev dependency on h3i.

PR #2318, opened 2026-01-20
release: datagram-socket, tokio-quiche.

toidiu:
This picks up the changes in buffer-pool crate which was recently released.

https://github.com/cloudflare/quiche/pull/2317

### DIFF
--- a/tokio-quiche/Cargo.toml
+++ b/tokio-quiche/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tokio-quiche"
-version = "0.14.0"
+version = "0.14.1"
 description = "Asynchronous wrapper around quiche"
 repository = { workspace = true }
 edition = { workspace = true }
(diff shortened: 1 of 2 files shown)


tokio-quiche/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tokio-quiche"
-version = "0.14.0"
+version = "0.14.1"
```
antoniovicente, line 3: you may as well call it 0.15.0

0.14.0 is broken, so there's nothing for 0.14.1 to be compatible with.

inamdarminaz (approved):

tokio-quiche/Cargo.toml
toidiu: cargo will use the latest patch unless somone is pinning to the very specific version so this seems ok.

PR #2319, opened 2026-01-22
Tweak BBR default params so the implementation behaves closer to the RFC draft

antoniovicente:
The defaults enabled several old experiments that we have found to diverge from the algorithm described in draft-ietf-ccwg-bbr-04. Analysis of the code, simulation results and prod experiments suggest that these experiments hurt congestion control behavior including resulting in incorrect BW estimates, erratic behavior under high loss conditions, excess buffering, and other problems.

We are likely to remove these internal config knobs in the near future. Do reach out to us if you depend on the behavior of any of these settings.

### DIFF
--- a/quiche/src/recovery/gcongestion/bbr2.rs
+++ b/quiche/src/recovery/gcongestion/bbr2.rs
@@ -167,7 +167,9 @@ struct Params {
     /// STARTUP.
     decrease_startup_pacing_at_end_of_round: bool,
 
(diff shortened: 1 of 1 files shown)


ghedo:
> LGTM I wonder if we could use `#[deprecated]` on some of the fields so users can notice the options are going to be removed.

The BBR params API is "internal" so we don't need to go through deprecation, we should just drop the params we don't need to experiment with anymore. This should clean-up the code at least a bit...

PR #2322, opened 2026-01-25
FFI: Update CC algorithms to be up to date with bda63e0

dries-c:


quiche/include/quiche.h
```diff
@@ -234,8 +234,7 @@ void quiche_config_set_initial_congestion_window_packets(quiche_config *config,
 enum quiche_cc_algorithm {
     QUICHE_CC_RENO = 0,
     QUICHE_CC_CUBIC = 1,
-    QUICHE_CC_BBR = 2,
-    QUICHE_CC_BBR2 = 3,
```
antoniovicente, line 238: It would be a good idea to prefer `quiche_config_set_cc_algorithm_name` over this enum.

Lucas: thoughts about removing this enum and `quiche_config_set_cc_algorithm`?

dries-c: I'm not a big fan of quiche_config_set_cc_algorithm_name because it changes its meaning. For example, "bbr" used to be "bbr v1" but is now "bbr2_gcongestion". bbr2, bbr & bbr2_gcongestion yield the same result, + it doesn't provide compile-time errors in FFI.

dries-c: Little bump

antoniovicente: Sounds good.

Let me get your PR rebased and merged.

rgacogne:
Given that this is an incompatible API change, are you planning to reflect that in the next version number?

PR #2325, opened 2026-01-27
tokio-quiche: Distinguish RETRY SCID from connection SCID

TheJokr:
tokio-quiche can perform stateless address validation via retry packets. If this is enabled, we trust the client to reflect the SCID from the Retry packet back to us as the Initial's DCID. This is a flawed assumption that we don't validate, so we either need to start validating it (see #2256) or treat the retry SCID as untrusted like we do when address validation is disabled.

I personally prefer this solution as its more permissive, so I went ahead and implemented the quiche changes I suggested in https://github.com/cloudflare/quiche/pull/2256#issuecomment-3563907591.

### DIFF
--- a/tokio-quiche/src/quic/router/acceptor.rs
+++ b/tokio-quiche/src/quic/router/acceptor.rs
@@ -34,6 +34,7 @@ use datagram_socket::DatagramSocketSendExt;
 use datagram_socket::MAX_DATAGRAM_SIZE;
 use quiche::ConnectionId;
 use quiche::Header;
+use quiche::RetryConnectionIds;
 use quiche::Type as PacketType;
 use task_killswitch::spawn_with_killswitch;
 
@@ -88,31 +89,29 @@ where
     }
 
     fn accept_conn(
-        &mut self, incoming: Incoming, scid: ConnectionId<'static>,
-        original_dcid: Option<&ConnectionId>,
-        pending_cid: Option<ConnectionId<'static>>,
-        quiche_config: &mut quiche::Config,
+        &mut self, incoming: Incoming, retry_cids: Option<RetryConnectionIds>,
+        pending_cid: ConnectionId<'static>, quiche_config: &mut quiche::Config,
     ) -> io::Result<Option<NewConnection>> {
         let handshake_start_time = Instant::now();
+        let scid = self.new_connection_id();
 
-        #[cfg(feature = "zero-copy")]
-        let mut conn = quiche::accept_with_buf_factory(
-            &scid,
-            original_dcid,
-            incoming.local_addr,
-            incoming.peer_addr,
-            quiche_config,
-        )
-        .into_io()?;
-
-        #[cfg(not(feature = "zero-copy"))]
-        let mut conn = quiche::accept(
-            &scid,
-            original_dcid,
-            incoming.local_addr,
-            incoming.peer_addr,
-            quiche_config,
-        )
+        let mut conn = if let Some(retry_cids) = retry_cids {
+            quiche::accept_with_retry(
+                &scid,
+                retry_cids,
+                incoming.local_addr,
+                incoming.peer_addr,
+                quiche_config,
+            )
+        } else {
+            quiche::accept_with_buf_factory(
+                &scid,
+                None,
+                incoming.local_addr,
+                incoming.peer_addr,
+                quiche_config,
+            )
+        }
         .into_io()?;
 
         if let Some(qlog_dir) = &self.config.qlog_dir {
@@ -135,7 +134,7 @@ where
         Ok(Some(NewConnection {
             conn,
             handshake_start_time,
-            pending_cid,
+            pending_cid: Some(pending_cid),
             initial_pkt: Some(incoming),
         }))
     }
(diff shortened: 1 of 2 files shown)


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -351,10 +351,8 @@ where
         self.conns.insert(scid, &conn);
 
         // Add the client-generated "pending" connection ID to the map as well.
```
ghedo:
```suggestion
        // Adds the client-generated "pending" connection ID to the map as well.
        //
```

quiche/src/lib.rs
```diff
@@ -1766,18 +1800,29 @@ impl Default for QlogInfo {
     }
 }
 
+struct RetryConnectionIds<'a> {
+    /// The DCID of the very first packet sent by the client.
```
LPardue: nit: there is no requirement this is the very first packet. 

Per https://datatracker.ietf.org/doc/html/rfc9000#section-7.3, it's probably more correct to say something like
```suggestion
    /// The DCID of the first Initial packet received by the server, which triggered a Retry packet.
```

quiche/src/lib.rs
```diff
@@ -1766,18 +1800,29 @@ impl Default for QlogInfo {
...
+    original_destination_cid: &'a ConnectionId<'a>,
+    /// The SCID of the stateless retry packet sent by the server.
```
LPardue: nit: "stateless retry" is not a formal term in RFC 9000 (even if elements of Retry can be stateless). Better to use the formal term

```suggestion
    /// The SCID of the Retry packet sent by the server.
```

quiche/src/lib.rs
```diff
@@ -1802,7 +1847,7 @@ impl<F: BufFactory> Connection<F> {
         );
 
         // If we did stateless retry assume the peer's address is verified.
```
LPardue: Not changed in this OG PR but if we take my naming nit above, should try to apply that consistently 

```suggestion
        // If we did Retry assume the peer's address is verified.
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+        original_destination_cid: odcid,
+        retry_source_cid: scid,
+    });
+
+    Connection::new(scid, retry_cids, local, peer, config, true)
+}
 
-    Ok(conn)
+/// Creates a new server-side connection after the client completed a stateless
+/// retry.
```
```suggestion
/// Creates a new server-side connection after the client responds to a Retry packet.
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+///
+/// To generate a stateless retry packet in the first place, use the [`retry()`]
```
```suggestion
/// To generate a Retry packet in the first place, use the [`retry()`]
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+/// function.
+///
+/// The `scid` parameter represents the server's source connection ID, which can
+/// be freshly generated after the retry succeeded. `odcid` is the original
```
```suggestion
/// be freshly generated after the application has successfully verified the Retry. `odcid` is the original
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+/// destination connection ID used by the client before the stateless retry.
```
```suggestion
/// destination connection ID used by the client that triggered the Retry packet.
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+/// `rscid` is the source connection ID used by the server for the stateless
+/// retry packet, which is allowed to be different from `scid`.
+///
+/// `rscid` should be equivalent to the DCID of the client's Initial packet
+/// which is carrying the retry token. However, as that field is
```
```suggestion
/// which is carrying the Retry token. However, as that field is
```

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+/// client-controlled, it should not be used for the `scid` parameter without
+/// validation.
```
LPardue: what is meant by validation here?

quiche/src/lib.rs
```diff
@@ -1528,9 +1527,44 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+#[inline]
+pub fn accept_with_retry<F: BufFactory>(
+    scid: &ConnectionId, odcid: &ConnectionId, rscid: &ConnectionId,
+    local: SocketAddr, peer: SocketAddr, config: &mut Config,
+) -> Result<Connection<F>> {
+    let retry_ids = RetryConnectionIds {
```
LPardue: I wonder if it would be clearer to push this struct up into the API and make the caller responsible for constructing it?

tokio-quiche/src/quic/router/acceptor.rs
```diff
@@ -88,31 +88,31 @@ where
     }
 
     fn accept_conn(
-        &mut self, incoming: Incoming, scid: ConnectionId<'static>,
-        original_dcid: Option<&ConnectionId>,
-        pending_cid: Option<ConnectionId<'static>>,
-        quiche_config: &mut quiche::Config,
+        &mut self, incoming: Incoming,
+        retry_cids: Option<(ConnectionId, ConnectionId)>,
```
LPardue: I guess the extension of my earlier question is, should we push the use of `RetryConnectionIds` up here too?

quiche/src/lib.rs
TheJokr: What I did in my original PR (#2256), i.e., implementing some scheme to assert that the Initial's DCID was generated by the server.

LPardue: THanks for the clarificaion, how about we tweak the sentence a little for clarity then? See suggestion

quiche/src/lib.rs
```diff
@@ -1526,9 +1525,47 @@ pub fn accept_with_buf_factory<F: BufFactory>(
...
+///
+/// The `scid` parameter represents the server's source connection ID, which can
+/// be freshly generated after the application has successfully verified the
+/// Retry. `retry_cids` is used to tie the new connection to the Initial + Retry
+/// exchange that preceded the connection's creation.
+///
+/// The `retry_source_cid` should be equivalent to the DCID of the client's
+/// Initial packet which is carrying the Retry token. However, as that field is
+/// client-controlled, it should not be used for the `scid` parameter without
+/// validation.
```
```suggestion
/// The DCID of the client Initial packet is inherently untrusted data. It is
/// safe to use the DCID in the `retry_source_cid` field of the
/// `RetryConnectionIds` provided to this function. However, using the DCID
/// for  the `scid` parameter carries risks. Applications are advised to
/// implement their own DCID validation steps before accepting the connection.
```

TheJokr: That brings my point across much better, thanks!

evanrittenhouse (approved):

PR #2328, opened 2026-01-28
BBR: Check for app-limited at the beginning of the packet recv / send loop

antoniovicente:
The new logic matches the app-limited detection algorithm described in
the BBR RFC draft, which involves checking for the previous state of the
connection before timeout and ack processing.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -573,6 +573,7 @@ pub struct Config {
     grease: bool,
 
     cc_algorithm: CongestionControlAlgorithm,
+    enable_bbr_app_limited_fix: bool,
     custom_bbr_params: Option<BbrParams>,
     initial_congestion_window_packets: usize,
     enable_relaxed_loss_threshold: bool,
@@ -651,6 +652,7 @@ impl Config {
             application_protos: Vec::new(),
             grease: true,
             cc_algorithm: CongestionControlAlgorithm::CUBIC,
+            enable_bbr_app_limited_fix: false,
             custom_bbr_params: None,
             initial_congestion_window_packets:
                 DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS,
@@ -1063,6 +1065,15 @@ impl Config {
     /// The default value is `CongestionControlAlgorithm::CUBIC`.
     pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
         self.cc_algorithm = algo;
+    }
+
+    /// Enable a rework of how the BBR congestion control implementation
+    /// computes app-limited.  This config parameter will be removed after we
+    /// build confidence on the new implementation or decide to roll it back.
+    ///
+    /// Defaults to `false`.
+    pub fn set_enable_bbr_app_limited_fix(&mut self, value: bool) {
+        self.enable_bbr_app_limited_fix = value;
     }
 
     /// Sets custom BBR settings.
@@ -2231,6 +2242,32 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Enable a rework of how the BBR congestion control implementation
+    /// computes app-limited.  This config parameter will be removed after we
+    /// build confidence on the new implementation or decide to roll it back.
+    ///
+    /// Currently this only applies if cc_algorithm is
+    /// `CongestionControlAlgorithm::Bbr2Gcongestion`.
+    ///
+    /// This function can only be called inside one of BoringSSL's handshake
+    /// callbacks, before any packet has been sent. Calling this function any
+    /// other time will have no effect.
+    ///
+    /// See [`Config::set_enable_bbr_app_limited_fix()`].
+    ///
+    /// [`Config::set_enable_bbr_app_limited_fix()`]: struct.Config.html#method.set_enable_bbr_app_limited_fix
+    #[cfg(feature = "boringssl-boring-crate")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "boringssl-boring-crate")))]
+    pub fn set_enable_bbr_app_limited_fix_in_handshake(
+        ssl: &mut boring::ssl::SslRef, value: bool,
+    ) -> Result<()> {
+        let ex_data = tls::ExData::from_ssl_ref(ssl).ok_or(Error::TlsFail)?;
+
+        ex_data.recovery_config.enable_bbr_app_limited_fix = value;
+
+        Ok(())
+    }
+
     /// Sets the congestion control algorithm used by string.
     ///
     /// This function can only be called inside one of BoringSSL's handshake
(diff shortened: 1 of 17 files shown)


quiche/src/lib.rs
```diff
@@ -2494,6 +2531,51 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Returns true if at least 1 stream has headers or body data to write.
+    pub fn has_flushable_stream(&mut self) -> bool {
```
antoniovicente: Oops, self shouldn't be `mut`

estcarisimo:
This branch creates add `enable_bbr_app_limited_fix` as a flag to detect the status of the data flow coming for the app side. This is propagated to various places including the connection establishment.

The code reads well and it is generally easy to follow for a reader familiar with the app limited issue. 

As this change propagate signals to various parts of QUIC mechanism, requires some lengthy comment blocks: ([example](https://github.com/cloudflare/quiche/pull/2328/changes/44f9c5b53c73506342b81bf9467be8f8ec6d2265#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R2539))
My recommendation is to add documentation (e.g., md) along with that comment to reduce its length and perhaps add a visual element

The [app_limited.rs](tokio-quiche/tests/integration_tests/app_limited.rs) is a very good test case. As it requires many lines of code it is a bit hard to keep all in cache while you read the code. Another minor point is what if `enable_bbr_app_limited_fix=false`. I am curious about it.

My last comment is that `next_release_time_in_past` as a function name was a bit confusing at the beginning but I then understood what it meant.

antoniovicente:
> This branch creates add `enable_bbr_app_limited_fix` as a flag to detect the status of the data flow coming for the app side. This is propagated to various places including the connection establishment.
> 
> The code reads well and it is generally easy to follow for a reader familiar with the app limited issue.
> 
> As this change propagate signals to various parts of QUIC mechanism, requires some lengthy comment blocks: ([example](https://github.com/cloudflare/quiche/pull/2328/changes/44f9c5b53c73506342b81bf9467be8f8ec6d2265#diff-2ac678416f8d7ea462a714681b63acedcbbb7f2c99e537ce61b7667b3bf4ac41R2539)) My recommendation is to add documentation (e.g., md) along with that comment to reduce its length and perhaps add a visual element

Hmm, maybe.  I tend to prefer docs in line with code unless I'm trying to describe usage of some tool or larger scale design descriptions.  I'll get back to you.

> The [app_limited.rs](tokio-quiche/tests/integration_tests/app_limited.rs) is a very good test case. As it requires many lines of code it is a bit hard to keep all in cache while you read the code. Another minor point is what if `enable_bbr_app_limited_fix=false`. I am curious about it.

I parametrized the test so it runs against CUBIC and BBR with enable_bbr_app_limited_fix=false

The short answer is that BBR incorrectly exits startup when this flag is not set.  It would be interesting to also add some test cases for BBR existing startup due to bandwidth plateau but unfortunately I haven't managed to open source the traffic shaper utility that allows testing that sort of thing.  I do hope to get to that in the coming weeks.

> My last comment is that `next_release_time_in_past` as a function name was a bit confusing at the beginning but I then understood what it meant.

I renamed this method and added an unit test for it.  See if that looks better to you.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1124,6 +1131,49 @@ impl RecoveryOps for GRecovery {
...
+    fn bbr_check_if_app_limited(
+        &mut self, had_flushable_stream_before_poll: bool, now: &Instant,
+    ) {
+        if !self.enable_bbr_app_limited_fix {
+            return;
+        }
+
+        // Implements CheckIfApplicationLimited from the BBR RFC.
+        //
+        // had_flushable_stream_before_poll is used to check for `NoUnsentData()`
```
LPardue: I think we might have discussed some of this before but what about queued datagrams?

LPardue: I.e, how delivery_rate_check_if_app_limited checks in 
https://github.com/cloudflare/quiche/blob/404ce836975c4846add21e56ada55121e5bcde01/quiche/src/lib.rs#L8284

antoniovicente: `|| !self.dgram_send_queue.is_empty()` could be added to `has_flushable_stream`

And `has_flushable_stream` should be renamed to something like `has_data_to_send`

LPardue: Cool. I have no strong opinion on how it's done just wanted to check if it should be. Sounds like we have a rough plan

antoniovicente: Regarding the comparison vs cwnd_available in delivery_rate_check_if_app_limited, the BBR draft doesn't seem concerned about cwnd_available, just wherever or not there's less than 1 packet's worth of data left in the transport send buffer.  I think that's closer to `self.streams.has_flushable() || !self.dgram_send_queue.is_empty()`

Quote:
The transport send buffer has less than one C.SMSS of unsent data available to send.

https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-04.html#name-tracking-application-limite

antoniovicente: Updated `has_flushable_data`

quiche/src/lib.rs
```diff
@@ -2494,6 +2531,52 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// This hook is used to implement the "Detecting application-limited
+    /// phases" logic described in the BBR RFC draft which is described at:
+    /// https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-04.html#name-detecting-application-limit
+    ///
+    /// This function must be invoked at the beginning of each iteration of the
+    /// connection work loop, before receiving packets from the network and
+    /// processing ACKs/timeouts.
+    ///
+    /// The expected work loop order is:
```
ghedo, line 2549: TBH I don't particularly like that every application now has to change how it uses quiche to get correct behaviour... this API also seems kind of leaky and exposes internal architecture details that applications shouldn't really care about.

Could we detect the start automatically? Say, have an internal `was_flushable_at_loop_stat: Option<bool>` parameter that is initialized early in `recv()` if it's `None`, and then reset to `None` on the first `send()` call?

antoniovicente: This seems very magical.  If there is truly a guarantee that the event loop will involve a call to on_timeout, followed by other stuff maybe we can hide the new call inside on_timeout.  But the problem is the transitions from not having flushable data to having flushable data.  I need to know the value of that call before wait_for_data_or_handshake changes it.

The location of this whole event loop feels somewhat wrong.  Things would be simpler if this logic could be moved to quiche;  there is strict order of operations and things will go wrong if you get the order wrong.

antoniovicente: Actually, I think that there is no guarantee that on_timeout or recv will be called in each loop iteration.  Recv will only happen in cases where at least 1 packet has arrived.

I think this means that the first call into quiche could be on_timeout(), recv() or send_on_path().

Similarly there's no great signaling for the end of the loop iteration.  We could try to look at differences in Instant::now, but the lack of a now argument to on_timeout(), recv() or send_on_path() makes that difficult; you can't tell based on time if two consecutive send_on_path calls happen on the same or different iterations of the event loop.

Is there something I'm missing?

antoniovicente: It would be good to gather information from this new experiment while we figure out the API issues.  Could we consider marking some of the new APIs internal and proceed with this PR mostly as-is?

Personally I would like to see a `now: Instant` argument to on_timeout(), recv() or send_on_path().  But I would also like to see some simplifications to the quiche API by moving more of the work loop to quiche.  The current API is just too low level and prone to change.

antoniovicente: Moving to draft while I work on alternative API.

antoniovicente: See https://github.com/cloudflare/quiche/pull/2453

antoniovicente:
Closing in favor of https://github.com/cloudflare/quiche/pull/2453

PR #2329, opened 2026-01-28
lib: add stream_discard function

LPardue:
QUIC requires that stream data is presented to receivers as a contiguous byte
stream. The existing stream_recv() function allows applications to read
contiguous bytes from receive streams, copying data into an application-provided
slice.

Applications may have cases where they want to skip over data in a receive
stream without using it. stream_recv() could be used in these cases, at the cost
of copying.

This change adds a stream_discard function that acts similarly to stream_recv.
Callers indicate a maximum size of bytes to discard, and up to that many will be
discarded without a copy. The return values of this function are identical to
stream_recv, meaning that existing business logic and stream state management
can remain the same. Application can use either function interchangeably,
depending on their needs.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -426,6 +426,7 @@ use crate::recovery::OnLossDetectionTimeoutOutcome;
 use crate::recovery::RecoveryOps;
 use crate::recovery::ReleaseDecision;
 
+use crate::stream::RecvAction;
 use crate::stream::StreamPriorityKey;
 
 /// The current QUIC wire version.
@@ -5167,7 +5168,7 @@ impl<F: BufFactory> Connection<F> {
     /// is returned as a tuple, or [`Done`] if there is no data to read.
     ///
     /// Reading data from a stream may trigger queueing of control messages
-    /// (e.g. MAX_STREAM_DATA). [`send()`] should be called after reading.
+    /// (e.g. MAX_STREAM_DATA). [`send()`] should be called afterwards.
     ///
     /// [`Done`]: enum.Error.html#variant.Done
     /// [`send()`]: struct.Connection.html#method.send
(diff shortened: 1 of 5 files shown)


quiche/src/stream/recv_buf.rs
```diff
@@ -199,18 +199,37 @@ impl RecvBuf {
         Ok(())
     }
 
-    /// Writes data from the receive buffer into the given output buffer.
+    pub fn emit(&mut self, out: &mut [u8]) -> Result<(usize, bool)> {
```
LPardue, line 213: Oops, forgot to document these 2

quiche/src/stream/recv_buf.rs
```diff
@@ -388,19 +409,48 @@ impl RecvBuf {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rstest::rstest;
 
-    #[test]
-    fn empty_read() {
-        let mut recv = RecvBuf::new(u64::MAX, DEFAULT_STREAM_WINDOW);
-        assert_eq!(recv.len, 0);
+    // Helper function for testing either buffer emit or discard
```
jannes: this could use a better comment on what this really asserts on

jannes:
i just wonder whether it's necessary to parameterize all those tests on emit/discard instead of some isolated discard tests

jannes (approved):

LPardue:
That's a good question. Many of the tests are intended to stress stream states or exercise flow control/congestion congestion as a result of taking data from streams. There's 60+ tests that were changed as part of this PR, I can't confidently tell if if its safe to not test the change if I don't test it. Future refactors may also cause unexpected breakages. 

Given how critical streams are, and how many edge cases (or bugs) we've had to accommodate for, I don't really want to take risks.

quiche/src/stream/recv_buf.rs
```diff
@@ -388,19 +429,64 @@ impl RecvBuf {
...
+    // performed on `result_len` to ensure the number of bytes read meets the
+    // caller expectations.
+    //
+    // The `is_fin` parameter relates to the buffer's finished status. An assert
+    // is performed on it to ensure the status meet the caller expectations.
+    //
+    // The `test_bytes` parameter carries an optional slice of bytes. Is set, an
+    // assert is performed against the bytes that were read out of the buffer,
+    // to ensure caller expectations are met.
+    fn t_emit_discard(
```
ghedo:
```suggestion
    fn assert_emit_discard(
```

quiche/src/stream/recv_buf.rs
```diff
@@ -388,19 +429,64 @@ impl RecvBuf {
...
+            (read, fin)
+        };
+        assert_eq!(read, result_len);
+        assert_eq!(is_fin, fin);
+    }
 
-    #[test]
-    fn empty_read() {
+    // Helper function for testing buffer status for either emit or discard.
+    fn t_emit_discard_done(recv: &mut RecvBuf, emit: bool) {
```
ghedo:
```suggestion
    fn assert_emit_discard_done(recv: &mut RecvBuf, emit: bool) {
```

quiche/src/lib.rs
```diff
@@ -1225,6 +1225,11 @@ pub enum TxBufferTrackingState {
     Inconsistent,
 }
 
+enum StreamRecvAction<'a> {
```
ghedo: Nit: We could maybe just push this down into `stream::recv_buf` instead, and pass it directly to `emit_or_discard()`.

ghedo (approved):

PR #2330, opened 2026-01-28
tests: reduce testing for aliased CC algo names

LPardue:
The algorithm names "bbr2" and "bbr2_gcongestion" alias to the same value
CongestionControlAlgorithm::Bbr2Gcongestion.

This change removes a redundant algorithm name from paramaterized tests.


ghedo:
Can we remove the bbr2_gcongestion one instead? I think eventually we'll want to drop that name anyway...

LPardue:
Either one works for me, I'll defer to @antoniovicente though

PR #2331, opened 2026-01-28
Refactor transport parameters code into separate module file

PSUdaemon:
As discussed with @LPardue and @ghedo this is a mechanical move of the transport parameters code into it's own module. Starting small with an easy one and we can go from there. This should be a 1:1 code move, so if anything is found to be changed, that should be considered a bug.

This change results in a ~6.2% reduction in the size of `lib.rs`.


quiche/src/transport_params.rs
```diff
@@ -0,0 +1,626 @@
+// Copyright (C) 2018-2025, Cloudflare, Inc.
```
antoniovicente, line 1: I verified that the change is a pure move of the code.

Thank you for this refactor, I absolutely agree that the size of `lib.rs` is problematic.  It is good to consider ways to improve code structure.

antoniovicente (approved):

ghedo:
Merged, thanks @PSUdaemon!

PR #2332, opened 2026-01-29
qlog: align ErrorSpace with qlog spec

KasenX:
This updates qlog `ErrorSpace` enum variants and call sites to match the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-quic-events.html) (all versions) which uses `"transport"` and `"application"` instead of `"transport_error"` and `"application_error"`.

### DIFF
--- a/h3i/src/prompts/h3/errors.rs
+++ b/h3i/src/prompts/h3/errors.rs
@@ -84,12 +84,12 @@ pub const APPLICATION: &str = "application";
 pub fn prompt_transport_or_app_error() -> InquireResult<(ErrorSpace, u64)> {
     let trans_or_app = prompt_transport_or_app()?;
     let space = if trans_or_app == TRANSPORT {
-        ErrorSpace::TransportError
+        ErrorSpace::Transport
     } else {
-        ErrorSpace::ApplicationError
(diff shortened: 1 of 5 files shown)


LPardue:
Thanks for the PR. This looks like a good change but I'd like to schedule it to come after https://github.com/cloudflare/quiche/pull/2168/changes which should happen soon

KasenX:
Nice! I am looking forward to [PR #2168](https://github.com/cloudflare/quiche/pull/2168). I’m currently implementing qlog for Nginx and found myself in the awkward position of having to use an outdated draft version because the qvis tool doesn't support the newer ones. However, I just discovered the qlog-dancer tool recently and using it instead of qvis will potentially allow me to implement the latest draft versions.

LPardue:
For avoidance if doubt, we're going to purposefully break backwards compat in the next major qlog update PR, and this PR can do similar.

LPardue:
There's a minor conflict after I landed the big schema change. Should be easy enough to fixup though

KasenX:
Done:)

LPardue (approved):

PR #2333, opened 2026-01-29
tokio-quiche: only close connection when H3Event receiver drops if no active streams/flows

inamdarminaz:
**tokio-quiche: only close connection when H3Event receiver drops if no active streams/flows**

Applications built on top of H3Driver process events via the H3Event
receiver. When they drop it, tokio-quiche should close the connection
promptly instead of spinning until idle timeout.

This change detects when the H3Event receiver is dropped (via
h3_event_sender.closed()) and closes the connection with NoError once
both stream_map and flow_map are empty.

The close is deferred when streams or flows are still active to support
tunnel scenarios (CONNECT-UDP/MASQUE) where the controller is
intentionally dropped after setup.

Close logic is extracted into close_if_idle() and called from
cleanup_stream, dgram_ready (after flow removal), and wait_for_data.

Additionally, datagram flow creation in process_available_dgrams and
headers_received is now gated on extended_connect_enabled to prevent
phantom flow entries when extended CONNECT is disabled.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -362,6 +362,9 @@ pub struct H3Driver<H: DriverHooks> {
     /// Tracks whether we have forwarded the HTTP/3 SETTINGS frame
     /// to the [H3Controller] once.
     settings_received_and_forwarded: bool,
+    /// Tracks whether the H3 event receiver has been dropped.
+    /// Used to avoid busy-looping on `h3_event_sender.closed()`.
+    h3_event_receiver_dropped: bool,
 }
 
 impl<H: DriverHooks> H3Driver<H> {
@@ -397,6 +400,7 @@ impl<H: DriverHooks> H3Driver<H> {
                 waiting_streams: FuturesUnordered::new(),
 
                 settings_received_and_forwarded: false,
+                h3_event_receiver_dropped: false,
             },
             H3Controller {
                 cmd_sender,
@@ -920,6 +924,7 @@ impl<H: DriverHooks> H3Driver<H> {
                         },
                     )?;
                     self.flow_map.remove(&flow_id);
+                    self.close_if_idle(qconn);
                     break;
                 },
                 Ok(_) => unreachable!("Flows can't send frame of other types"),
@@ -993,7 +998,21 @@ impl<H: DriverHooks> H3Driver<H> {
                 .send(H3Event::StreamClosed { stream_id }.into());
         }
 
+        self.close_if_idle(qconn);
+
         Ok(())
+    }
+
+    /// Closes the connection with `NoError` if the H3 event receiver
+    /// has been dropped and there are no active streams or flows.
+    fn close_if_idle(&self, qconn: &mut QuicheConnection) {
+        if self.h3_event_receiver_dropped &&
+            self.stream_map.is_empty() &&
+            self.flow_map.is_empty()
+        {
+            let _ =
+                qconn.close(true, quiche::h3::WireErrorCode::NoError as u64, &[]);
+        }
     }
 
     /// Shuts down the indicated HTTP/3 stream by sending frames and cleaning
@@ -1082,9 +1101,13 @@ impl<H: DriverHooks> H3Driver<H> {
     ) -> H3ConnectionResult<()> {
         loop {
             match datagram::receive_h3_dgram(qconn) {
-                Ok((flow_id, dgram)) => {
+                Ok((flow_id, dgram))
+                    if !qconn.is_server() ||
+                        self.hooks.extended_connect_enabled() =>
+                {
                     self.get_or_insert_flow(flow_id)?.send_best_effort(dgram);
                 },
+                Ok(_) => {},
                 Err(quiche::Error::Done) => return Ok(()),
                 Err(err) => return Err(H3ConnectionError::from(err)),
             }
@@ -1324,6 +1347,11 @@ impl<H: DriverHooks> ApplicationOverQuic for H3Driver<H> {
             Some(dgram) = self.dgram_recv.recv() => self.dgram_ready(qconn, dgram),
             Some(cmd) = self.cmd_recv.recv() => H::conn_command(self, qconn, cmd),
             r = self.hooks.wait_for_action(qconn), if H::has_wait_action(self) => r,
+            _ = self.h3_event_sender.closed(), if !self.h3_event_receiver_dropped => {
+                self.h3_event_receiver_dropped = true;
+                self.close_if_idle(qconn);
+                Ok(())
+            }
         }?;
 
         // Make sure controller is not starved, but also not prioritized in the
--- a/tokio-quiche/src/http3/driver/server.rs
+++ b/tokio-quiche/src/http3/driver/server.rs
@@ -164,6 +164,9 @@ pub struct ServerHooks {
     /// Handle to the post-accept timeout entry. If present, the server must
     /// receive a HEADERS frame before this timeout.
     post_accept_timeout: Option<TimeoutKey>,
+    /// Whether the extended CONNECT protocol is enabled. When disabled,
+    /// skip DATAGRAM flow creation for `:protocol` requests.
+    extended_connect_enabled: bool,
 }
 
 impl ServerHooks {
@@ -191,11 +194,19 @@ impl ServerHooks {
         let (mut stream_ctx, send, recv) =
             StreamCtx::new(stream_id, STREAM_CAPACITY);
 
-        if let Some(flow_id) =
-            datagram::extract_quarter_stream_id(stream_id, &headers)
-        {
-            let _ = driver.get_or_insert_flow(flow_id)?;
-            stream_ctx.associated_dgram_flow_id = Some(flow_id);
+        // When Extended CONNECT is enabled, extract the datagram flow ID
+        // from the request headers and register it in flow_map. This
+        // allows incoming DATAGRAMs to be routed to the correct stream.
+        // Note: flow_map entries are considered active work, so the
+        // connection will not be closed in cleanup_stream() while
+        // flows remain.
+        if driver.hooks.extended_connect_enabled() {
+            if let Some(flow_id) =
+                datagram::extract_quarter_stream_id(stream_id, &headers)
+            {
+                let _ = driver.get_or_insert_flow(flow_id)?;
+                stream_ctx.associated_dgram_flow_id = Some(flow_id);
+            }
         }
 
         let latest_priority_update: Option<RawPriorityValue> = driver
@@ -253,7 +264,12 @@ impl DriverHooks for ServerHooks {
             settings_enforcer: settings.into(),
             requests: 0,
             post_accept_timeout: None,
+            extended_connect_enabled: settings.enable_extended_connect,
         }
+    }
+
+    fn extended_connect_enabled(&self) -> bool {
+        self.extended_connect_enabled
     }
 
     fn conn_established(
--- a/tokio-quiche/src/http3/driver/hooks.rs
+++ b/tokio-quiche/src/http3/driver/hooks.rs
@@ -83,6 +83,12 @@ pub trait DriverHooks: Sized + Send + 'static {
         cmd: Self::Command,
     ) -> H3ConnectionResult<()>;
 
+    /// Whether the extended CONNECT protocol is enabled. Used to gate
+    /// datagram flow creation.
+    fn extended_connect_enabled(&self) -> bool {
+        false
+    }
+
     /// Determines whether the hook's `wait_for_action` future will be polled
     /// as part of `ApplicationOverQuic::wait_for_data`. Defaults to `false` and
     /// must be overridden if `wait_for_action` is overridden.
--- a/tokio-quiche/src/http3/driver/tests.rs
+++ b/tokio-quiche/src/http3/driver/tests.rs
@@ -1816,4 +1816,385 @@ mod server_side_driver {
         assert_eq!(helper.peer_client_recv_body_vec(0, 1024), Ok(vec![3; 10]));
         assert_eq!(helper.peer_client_poll(), Ok((0, h3::Event::Finished)));
     }
+
+    // Test that dropping the H3 event receiver when there are no active streams
+    // and datagram flows causes the connection to close immediately.
+    #[test]
+    fn h3_controller_drop_closes_connection_when_maps_empty() {
+        let mut helper = DriverTestHelper::<ServerHooks>::new().unwrap();
+        helper.complete_handshake().unwrap();
+
+        assert!(helper.driver.stream_map.is_empty());
+        assert!(helper.driver.flow_map.is_empty());
+        assert!(helper.pipe.server.local_error().is_none());
+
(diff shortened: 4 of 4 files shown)


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -741,7 +741,22 @@ where
...
-                    log::error!("Incoming packet router encountered recvmsg error"; "error" => e);
+                    // Check if this is a fatal error that means the socket is
+                    // closed.
+                    let is_fatal = matches!(
+                        e.kind(),
+                        io::ErrorKind::BrokenPipe |
+                            io::ErrorKind::ConnectionReset |
+                            io::ErrorKind::NotConnected |
+                            io::ErrorKind::UnexpectedEof
+                    ) || e.raw_os_error() == Some(9); // EBADF
```
schien: can we replace the magic number with the libc constant?
https://docs.rs/libc/latest/libc/constant.EBADF.html

or nix constant
https://docs.rs/nix/latest/nix/type.Error.html#variant.EBADF

ghedo:
TBH it's not entirely clear to me what this is fixing by reading the PR description, could you maybe provide some more context/details? It also seems like this could use a test to show the expected behaviour.

Also, it'd be good to put the PR description in the commit message body as well for posterity.

evanrittenhouse:
> TBH it's not entirely clear to me what this is fixing by reading the PR description, could you maybe provide some more context/details? It also seems like this could use a test to show the expected behaviour.
> 
> Also, it'd be good to put the PR description in the commit message body as well for posterity.

Agree on a test here, this fixes fairly subtle behavior

tokio-quiche/src/quic/router/mod.rs
inamdarminaz: thanks, updated.

inamdarminaz:
> TBH it's not entirely clear to me what this is fixing by reading the PR description, could you maybe provide some more context/details? It also seems like this could use a test to show the expected behaviour.
> 
> Also, it'd be good to put the PR description in the commit message body as well for posterity.

> > TBH it's not entirely clear to me what this is fixing by reading the PR description, could you maybe provide some more context/details? It also seems like this could use a test to show the expected behaviour.
> > Also, it'd be good to put the PR description in the commit message body as well for posterity.
> 
> Agree on a test here, this fixes fairly subtle behavior

Thankyou both, I've updated the PR with tests & commit message

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -164,6 +164,9 @@ pub struct ServerHooks {
     /// Handle to the post-accept timeout entry. If present, the server must
     /// receive a HEADERS frame before this timeout.
     post_accept_timeout: Option<TimeoutKey>,
+    /// Whether the extended CONNECT protocol is enabled. When disabled,
+    /// skip DATAGRAM flow creation for `:protocol` requests.
+    extended_connect_enabled: bool,
```
toidiu, line 169: This seems unrelated to the receiver drop. Is there a reason we need it part of this PR?

Is it included for convenience or is needed for correctness?

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -988,6 +992,16 @@ impl<H: DriverHooks> H3Driver<H> {
                 .send(H3Event::StreamClosed { stream_id }.into());
         }
 
+        // If the event receiver was already dropped and all work is done, close
+        // the connection now.
+        if self.h3_event_receiver_dropped &&
+            self.stream_map.is_empty() &&
+            self.flow_map.is_empty()
+        {
```
toidiu, line 1012: can you extract this into a separate well named variable:

```
let close_connection = self.h3_event_receiver_dropped && self.stream_map.is_empty() ...
```

Also instead of the variable `h3_event_receiver_dropped`, can you simply call `self.h3_event_sender.closed()` which is the source of truth?

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -191,11 +194,13 @@ impl ServerHooks {
...
-        {
-            let _ = driver.get_or_insert_flow(flow_id)?;
-            stream_ctx.associated_dgram_flow_id = Some(flow_id);
+        if driver.hooks.extended_connect_enabled {
+            if let Some(flow_id) =
+                datagram::extract_quarter_stream_id(stream_id, &headers)
+            {
+                let _ = driver.get_or_insert_flow(flow_id)?;
+                stream_ctx.associated_dgram_flow_id = Some(flow_id);
+            }
```
toidiu: can you add a comment what this block is doing and possibly how it relates to the close functionality in this PR

tokio-quiche/src/http3/driver/server.rs
toidiu: Looking at the tests.. it seems related.

tokio-quiche/src/http3/driver/server.rs
inamdarminaz: updated with a comment

tokio-quiche/src/http3/driver/mod.rs
inamdarminaz: yes, updated now

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -362,6 +362,9 @@ pub struct H3Driver<H: DriverHooks> {
     /// Tracks whether we have forwarded the HTTP/3 SETTINGS frame
     /// to the [H3Controller] once.
     settings_received_and_forwarded: bool,
+    /// Tracks whether the H3 event receiver has been dropped.
+    /// Used to avoid busy-looping on `h3_event_sender.closed()`.
+    h3_event_receiver_dropped: bool,
```
toidiu, line 367: since you are using `h3_event_sender.is_closed()`, I dont think you need this anymore

toidiu (approved):

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -988,6 +992,16 @@ impl<H: DriverHooks> H3Driver<H> {
                 .send(H3Event::StreamClosed { stream_id }.into());
         }
 
+        // If the event receiver was already dropped and all work is done,
+        // close the connection now.
+        let should_close_connection = self.h3_event_receiver_dropped &&
+            self.stream_map.is_empty() &&
+            self.flow_map.is_empty();
```
ghedo: Given that this checks the flows map, I guess the same logic needs to be applied when a flow is dropped too? i.e. in `dgram_ready()` on `FlowShutdown`.

Either way I think we should have a test for this case too. Actually, maybe two tests:
* Drop event receiver with an open stream, and then shutdown the stream, and verify that the connection is closed.
* Drop event receiver with an open flow, and then shutdown the flow, and verify that the connection is closed.

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -191,11 +194,19 @@ impl ServerHooks {
...
-        {
-            let _ = driver.get_or_insert_flow(flow_id)?;
-            stream_ctx.associated_dgram_flow_id = Some(flow_id);
+        // When Extended CONNECT is enabled, extract the datagram flow ID
+        // from the request headers and register it in flow_map. This
+        // allows incoming DATAGRAMs to be routed to the correct stream.
+        // Note: flow_map entries are considered active work, so the
+        // connection will not be closed in cleanup_stream() while
+        // flows remain.
+        if driver.hooks.extended_connect_enabled {
```
ghedo: So, the new hook option is basically a breaking change, right? Because existing applications need to consider whether to enable this or now.

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -761,7 +761,22 @@ where
                 },
 
                 Poll::Ready(Err(e)) => {
-                    log::error!("Incoming packet router encountered recvmsg error"; "error" => e);
+                    // Check if this is a fatal error that means the socket is
```
ghedo: This change seems unrelated?

ghedo (changes requested):

tokio-quiche/src/http3/driver/server.rs
ghedo: Actually, re-reading this, not necessarily a change if the app already configures `enable_extended_connect` but not sure if an app can break if it doesn't enable that setting, and somehow still depends on this behaviour?

tokio-quiche/src/http3/driver/mod.rs
inamdarminaz: cleanup_stream (called from shutdown_stream inside dgram_ready) already handles this.  Added both the tests

tokio-quiche/src/quic/router/mod.rs
inamdarminaz: yea, this was related to a logging problem. Unrelated to this PR so removed it

tokio-quiche/Cargo.toml
```diff
@@ -56,6 +56,7 @@ foundations = { workspace = true, default-features = false, features = [
 futures = { workspace = true }
 futures-util = { workspace = true }
 ipnetwork = { workspace = true }
+libc = { workspace = true }
```
ghedo: This is now unnecessary as well.

tokio-quiche/src/http3/driver/mod.rs
ghedo: I don't think it does, because AFAICT the entry is removed from the flow map only _after_ `shutdown_stream()` is called https://github.com/cloudflare/quiche/blob/0c4aa65cc436f638cb1cbbd7421e5e4028fac913/tokio-quiche/src/http3/driver/mod.rs#L917

For example, this test seems to fail:
```patch
diff --git a/tokio-quiche/src/http3/driver/tests.rs b/tokio-quiche/src/http3/driver/tests.rs
index c6d0fb7f..386d289a 100644
--- a/tokio-quiche/src/http3/driver/tests.rs
+++ b/tokio-quiche/src/http3/driver/tests.rs
@@ -1888,4 +1888,100 @@ mod server_side_driver {
             "should close with H3 NoError"
         );
     }
+
+    /// Drop the event receiver with an autonomous flow open (a flow
+    /// with no associated stream), then shut down the flow. Verify
+    /// the connection closes.
+    #[test]
+    fn controller_drop_then_autonomous_flow_shutdown_triggers_connection_close() {
+        let mut config = default_quiche_config();
+        config.enable_dgram(true, 100, 100);
+
+        let mut helper =
+            DriverTestHelper::<ServerHooks>::with_pipe_and_http3_settings(
+                quiche::test_utils::Pipe::with_config_and_buf(&mut config)
+                    .unwrap(),
+                Http3Settings {
+                    enable_extended_connect: true,
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+        helper.complete_handshake().unwrap();
+        helper.advance_and_run_loop().unwrap();
+
+        // Peer sends an H3 datagram for a flow_id with no associated
+        // CONNECT-UDP stream. Wire format: varint(flow_id) || payload.
+        let flow_id: u64 = 8;
+        let payload: &[u8] = b"hi";
+        let prefix_len = octets::varint_len(flow_id);
+        let mut wire = vec![0u8; prefix_len + payload.len()];
+        {
+            let mut enc = octets::OctetsMut::with_slice(&mut wire);
+            enc.put_varint(flow_id).unwrap();
+            enc.put_bytes(payload).unwrap();
+        }
+        helper.pipe.client.dgram_send(&wire).unwrap();
+
+        // process_available_dgrams creates an autonomous flow and
+        // emits NewFlow.
+        helper.advance_and_run_loop().unwrap();
+
+        let flow_send = assert_matches!(
+            helper.driver_recv_server_event().unwrap(),
+            ServerH3Event::Core(H3Event::NewFlow { send, .. }) => send
+        );
+        let flow_sender = flow_send.get_ref().unwrap().clone();
+
+        assert!(helper.driver.stream_map.is_empty());
+        assert_eq!(helper.driver.flow_map.len(), 1);
+
+        // Drop the controller. closed() fires, sets the flag, but
+        // flow_map is non-empty so no close yet.
+        drop(helper.controller);
+        tokio::task::unconstrained(
+            helper.driver.wait_for_data(&mut helper.pipe.server),
+        )
+        .now_or_never();
+        assert!(helper.pipe.server.local_error().is_none());
+
+        // Send FlowShutdown via the per-flow channel.
+        flow_sender
+            .try_send(OutboundFrame::FlowShutdown {
+                flow_id,
+                stream_id: flow_id,
+            })
+            .unwrap();
+
+        // wait_for_data picks up the frame and calls dgram_ready.
+        // shutdown_stream returns early (no stream), so the close
+        // gate in cleanup_stream is never invoked.
+        tokio::task::unconstrained(
+            helper.driver.wait_for_data(&mut helper.pipe.server),
+        )
+        .now_or_never();
+        drop(flow_sender);
+        drop(flow_send);
+
+        assert!(helper.driver.stream_map.is_empty());
+        assert!(helper.driver.flow_map.is_empty());
+
+        // closed() is gated off; no other arm can drive a close.
+        tokio::task::unconstrained(
+            helper.driver.wait_for_data(&mut helper.pipe.server),
+        )
+        .now_or_never();
+
+        let local_error = helper
+            .pipe
+            .server
+            .local_error()
+            .expect("connection should be closing");
+        assert!(local_error.is_app, "should be application-level close");
+        assert_eq!(
+            local_error.error_code,
+            h3::WireErrorCode::NoError as u64,
+            "should close with H3 NoError"
+        );
+    }
 }
```

So I think this logic needs to be applied after the `self.flow_map.remove()` call.

In any case it'd be good to refactor this into its own method (e.g. `close_if_idle(&self, qconn: &mut QuicheConnection)` or something like that).

tokio-quiche/src/http3/driver/server.rs
```diff
@@ -191,11 +194,19 @@ impl ServerHooks {
...
+            if let Some(flow_id) =
+                datagram::extract_quarter_stream_id(stream_id, &headers)
+            {
+                let _ = driver.get_or_insert_flow(flow_id)?;
```
ghedo, line 207: I noticed that `process_available_dgrams()` also calls `get_or_insert_flow()` unconditionally, it seems like that also needs to be wrapped in `if driver.hooks.extended_connect_enabled {`?

ghedo (changes requested):

tokio-quiche/Cargo.toml
inamdarminaz: yup, leftover. Removed

tokio-quiche/src/http3/driver/server.rs
inamdarminaz: great catch, added the check but only for the server side, it looks client needs to always process datagrams. Added a test `process_dgram_skips_flow_creation_when_extended_connect_disabled`

inamdarminaz: Referencing https://github.com/cloudflare/quiche/pull/2333#discussion_r3146435205 here, as I cannot reply there. You're right the close check is never reached for flows with no stream. Refactored close logic into its method and added close_if_idle() after flow_map.remove()

LPardue (approved):

ghedo (approved):

PR #2338, opened 2026-01-30
metric: track how often the flush operation is aborted due to the fut…

toidiu:
playground example demonstrating usage: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=53b991dd80a5935ecdce5e8fbb3ca6db

### DIFF
--- a/tokio-quiche/src/metrics/mod.rs
+++ b/tokio-quiche/src/metrics/mod.rs
@@ -58,6 +58,10 @@ pub trait Metrics: Send + Sync + Clone + Unpin + 'static {
     /// WouldBlock and are retried in a loop.
     fn send_to_wouldblock_duration_s(&self) -> TimeHistogram;
 
+    /// Number of mid-handshake flush operations that were skipped due to future
+    /// cancellation.
+    fn skipped_mid_handshake_flush_count(&self) -> Counter;
+
     /// Number of QUIC packets received where the CID could not be verified.
     fn invalid_cid_packet_count(&self, reason: crate::BoxError) -> Counter;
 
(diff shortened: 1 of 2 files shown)


tokio-quiche/src/metrics/mod.rs
```diff
@@ -58,6 +58,10 @@ pub trait Metrics: Send + Sync + Clone + Unpin + 'static {
     /// WouldBlock and are retried in a loop.
     fn send_to_wouldblock_duration_s(&self) -> TimeHistogram;
 
+    /// Number of flush operations that were incomplete due to the future being
```
evanrittenhouse:
```suggestion
    /// Number of mid-handshake flush operations that were skipped due to future cancellation.
```

Same for below

tokio-quiche/src/metrics/mod.rs
```diff
@@ -58,6 +58,10 @@ pub trait Metrics: Send + Sync + Clone + Unpin + 'static {
...
+    /// canceled.
+    fn incomplete_flush_operation_count(&self) -> Counter;
```
evanrittenhouse:
```suggestion
    fn skipped_mid_handshake_flush_count(&self) -> Counter;
```

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -945,3 +955,32 @@ fn min_of_some<T: Ord>(v1: Option<T>, v2: Option<T>) -> Option<T> {
...
+struct TrackCompleteToken<M: Metrics, F: Fn(&M)> {
+    complete: AtomicBool,
+    metrics: M,
+    f: F,
+}
+
+impl<M: Metrics, F: Fn(&M)> TrackCompleteToken<M, F> {
+    fn new(metrics: M, f: F) -> Self {
+        Self {
+            complete: AtomicBool::from(false),
```
evanrittenhouse: Does this need to be atomic?

Also, this is currently only used for incrementing one counter. It may be clearer to just pass in the counter and increment it on drop rather than register a callback.

toidiu: I think we should be fine without the atomic here since we create a new token per flush attempt and operations on the token should be serialized.

The code is a bit generic but also not too complicated (but possibly too much for a single use atm)

PR #2343, opened 2026-02-03
quiche : add H3Controller::shutdown_stream() API

dineshadhi:
**Summary**

Add `H3Controller::shutdown_stream()` API to explicitly shut down HTTP/3 streams in specific directions by sending `RESET_STREAM` and/or `STOP_SENDING` frames.

**Changes**
- Add StreamShutdown enum with three variants:
  - Read { error_code } - sends STOP_SENDING
  - Write { error_code } - sends RESET_STREAM
  - Both { read_error_code, write_error_code } - sends both frames
- Add H3Controller::shutdown_stream(stream_id, shutdown) public method
- Refactor internal stream management into two functions:
  - cleanup_stream - internal state cleanup only (no frames sent)
  - shutdown_stream - sends frames and calls cleanup_stream
- Add tests for all three StreamShutdown variants

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -50,6 +50,7 @@ use foundations::telemetry::log;
 use futures::FutureExt;
 use futures_util::stream::FuturesUnordered;
 use quiche::h3;
+use quiche::h3::WireErrorCode;
 use tokio::select;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::error::TryRecvError;
@@ -544,7 +545,7 @@ impl<H: DriverHooks> H3Driver<H> {
         match status {
             StreamStatus::Done { close } => {
                 if close {
-                    return self.finish_stream(qconn, stream_id, None, None);
+                    return self.cleanup_stream(qconn, stream_id);
                 }
 
                 // The QUIC stream is finished, manually invoke `process_h3_fin`
@@ -566,7 +567,7 @@ impl<H: DriverHooks> H3Driver<H> {
                     .send(H3Event::ResetStream { stream_id }.into())
                     .map_err(|_| H3ConnectionError::ControllerWentAway)?;
                 if ctx.both_directions_done() {
-                    return self.finish_stream(qconn, stream_id, None, None);
+                    return self.cleanup_stream(qconn, stream_id);
                 }
             },
             StreamStatus::Blocked => {
@@ -653,7 +654,7 @@ impl<H: DriverHooks> H3Driver<H> {
                         .send(H3Event::ResetStream { stream_id }.into())
                         .map_err(|_| H3ConnectionError::ControllerWentAway)?;
                     if ctx.both_directions_done() {
-                        return self.finish_stream(qconn, stream_id, None, None);
+                        return self.cleanup_stream(qconn, stream_id);
                     }
                 }
 
@@ -889,11 +890,13 @@ impl<H: DriverHooks> H3Driver<H> {
                     let _ = datagram::send_h3_dgram(qconn, flow_id, dgram);
                 },
                 Ok(OutboundFrame::FlowShutdown { flow_id, stream_id }) => {
-                    self.finish_stream(
+                    self.shutdown_stream(
                         qconn,
                         stream_id,
-                        Some(quiche::h3::WireErrorCode::NoError as u64),
-                        Some(quiche::h3::WireErrorCode::NoError as u64),
+                        StreamShutdown::Both {
+                            read_error_code: WireErrorCode::NoError as u64,
+                            write_error_code: WireErrorCode::NoError as u64,
+                        },
                     )?;
                     self.flow_map.remove(&flow_id);
                     break;
@@ -925,29 +928,17 @@ impl<H: DriverHooks> H3Driver<H> {
         H3ConnectionError::H3(h3::Error::TransportError(quiche::Error::TlsFail))
     }
 
-    /// Removes a stream from the stream map if it exists. Also optionally sends
-    /// `RESET` or `STOP_SENDING` frames if `write` or `read` is set to an
-    /// error code, respectively.
-    fn finish_stream(
+    /// Cleans up internal state for the indicated HTTP/3 stream.
+    ///
+    /// This function removes the stream from the stream map, closes any pending
+    /// futures, removes associated DATAGRAM flows, and sends a
+    /// [`H3Event::StreamClosed`] event (for servers).
+    fn cleanup_stream(
         &mut self, qconn: &mut QuicheConnection, stream_id: u64,
-        read: Option<u64>, write: Option<u64>,
     ) -> H3ConnectionResult<()> {
         let Some(stream_ctx) = self.stream_map.remove(&stream_id) else {
             return Ok(());
         };
-
-        let audit_stats = &stream_ctx.audit_stats;
-
-        if let Some(err) = read {
-            audit_stats.set_sent_stop_sending_error_code(err as _);
-            let _ = qconn.stream_shutdown(stream_id, quiche::Shutdown::Read, err);
-        }
-
-        if let Some(err) = write {
-            audit_stats.set_sent_reset_stream_error_code(err as _);
-            let _ =
-                qconn.stream_shutdown(stream_id, quiche::Shutdown::Write, err);
-        }
 
         // Find if the stream also has any pending futures associated with it
         for pending in self.waiting_streams.iter_mut() {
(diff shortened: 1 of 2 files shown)


LPardue:
IIUC this PR is basically a wrapper around tokio-quiche's finish_stream method.

I think now is the time we should take the opportunity to fix some of the confusing naming issues we have. Tokio-quiche is loose with its terminology already; finish_stream cleans up local tokio-quiche state and optionally instructs quiche to shutdown a stream in the read and/or write directions, which may result in STOP_SENDING or RESET_STREAM frame. finish_stream doesn't actually ever do anything like send a FIN flag.

Calling this new function send_reset_stream is confusing because it might not ever actually reset the stream by QUICs definition of a reset.

Instead, we should probably change finish_stream to a new name (a small bikeshed), and change it to take an enum (like the new one you're adding here) with a similar name to the eenamed function. Then the new H3Command uses an identical name and enum, and effectively just becomes a pass through call.

dineshadhi:
@LPardue - I renamed the `finish_stream` with `shutdown_stream`. It complements the already existing `quiche::Connection::stream_shutdown` API. 

I also renamed the enum to `ShutdownStream` and added `None` variant for local cleanups.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -925,28 +939,57 @@ impl<H: DriverHooks> H3Driver<H> {
...
-    /// error code, respectively.
-    fn finish_stream(
+    /// Shuts down a stream in the specified direction(s) and removes it from
+    /// the stream map.
+    ///
+    /// This function always removes the stream from local state (stream map,
+    /// pending futures, DATAGRAM channels). Depending on the `shutdown`
+    /// parameter, it may also send `STOP_SENDING` (read direction) and/or
+    /// `RESET_STREAM` (write direction) frames to the peer.
+    fn shutdown_stream(
```
LPardue: Some minor suggestion for clarity. I think it would be easier to delegate the exact details to the enum definition

```suggestion
    /// Shuts down the indicated HTTP/3 stream.
    ///
    /// This function always tidies away tokio-quiche state related to the stream
    /// (e.g., stream map, pending futures, DATAGRAM channels).
    /// The `shutdown` parameter can be used to request that QUIC STOP_SENDING and/or
    /// RESET_STREAM frames also be sent.
    fn shutdown_stream(
```

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1288,6 +1354,44 @@ pub enum H3Command {
...
+/// neither.
+///
+/// Note: Despite its name, "shutdown" here refers to signaling the peer about
+/// stream termination, not sending a FIN flag. `STOP_SENDING` asks the peer to
+/// stop sending data, while `RESET_STREAM` abruptly terminates the write side.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum StreamShutdown {
+    /// Don't send any shutdown frames; just clean up local state.
+    #[default]
+    None,
```
LPardue: I don't think we should have this, seems like a footgun. Yeah, that means we can't derive a default but I don't think we should - applications need too make the correct decisions.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1288,6 +1354,44 @@ pub enum H3Command {
...
+    /// Shut down only the read direction (sends `STOP_SENDING` frame with the
+    /// given error code).
+    Read(u64),
+    /// Shut down only the write direction (sends `RESET_STREAM` frame with the
+    /// given error code).
+    Write(u64),
+    /// Shut down both directions (sends both `STOP_SENDING` and `RESET_STREAM`
+    /// frames). The first value is the read error code, the second is the
+    /// write error code.
+    Both(u64, u64),
```
LPardue: Let's use a structure with read and write members rather than rely on ordering that needs documenting.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1360,4 +1464,28 @@ impl<H: DriverHooks> H3Controller<H> {
...
+
+    /// Creates an [`H3Command`] sender for the paired [H3Driver].
+    pub fn h3_cmd_sender(&self) -> RequestSender<H::Command, H3Command> {
+        RequestSender {
+            sender: self.cmd_sender.clone(),
+            _r: Default::default(),
+        }
+    }
+
+    /// Shuts down a stream in the specified direction(s) and removes it from
```
LPardue, line 1459: This comment should also be updated in line with the other.

LPardue (commented):
Thank for applying previous suggestions. A few more for clarity

tokio-quiche/src/http3/driver/mod.rs
dineshadhi: I had the struct with explicit names. But it turned out to be verbose while initializing, this seemed concise. 

I see the ordering may cause confusion. I'll change this back.

tokio-quiche/src/http3/driver/mod.rs
dineshadhi: Having `StreamShutdown` argument has an Option would be a good approach I guess, instead of having a `None` variant.

dineshadhi: I just made `StreamShutdown` optional in the `stream_shutdown` function signature. This could help us remove the `None` variant in the enum.

LPardue: I think that has the same problem, it makes cleaning up the stream state too easy to opt out of sending something on the wire, which can eventually cause the QUIC connection to deadlock.

Its complex and people find this stuff hard to rationalise about, so by forcing one of the options that can generate a frame, we make them think.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -925,30 +929,18 @@ impl<H: DriverHooks> H3Driver<H> {
...
-    /// error code, respectively.
-    fn finish_stream(
+    /// Cleans up internal state for the indicated HTTP/3 stream.
+    ///
+    /// This function removes the stream from the stream map, closes any pending
+    /// futures, removes associated DATAGRAM flows, and sends a `StreamClosed`
+    /// event (for servers).
+    fn cleanup_stream(
         &mut self, qconn: &mut QuicheConnection, stream_id: u64,
-        read: Option<u64>, write: Option<u64>,
```
dineshadhi, line 933: @LPardue - The original implementation had optional read and write error codes. If they are empty, this function simply performs an internal cleanup and sends no frame. Function calls with None arguments were used in multiple places to clean up the internal states.

So, the original `finish_stream` was used to both clean up the internal states and then to send frames.

I refactored this into two functions `cleanup_stream` and `shutdown_stream`.

**cleanup_stream** - Cleans up internal states without sending any frames.
**shutdown_stream** - Explicity requires `StreamShutdown`. It sends the frames based on that directive and calls `cleanup_stream`.

tokio-quiche/src/http3/driver/mod.rs
dineshadhi: I am resolving this thread since I opened a new one below with follow up comment.

jannes:
lgtm, let's have @LPardue  do a final pass

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -984,6 +976,60 @@ impl<H: DriverHooks> H3Driver<H> {
...
+                );
+                audit_stats
+                    .set_sent_reset_stream_error_code(write_error_code as _);
+                let _ = qconn.stream_shutdown(
+                    stream_id,
+                    quiche::Shutdown::Write,
+                    write_error_code,
+                );
+            },
+        }
```
jannes, line 1027: this could be shortened, but not sure if it actually is worth it because it is very clear as is.

```
        let (read_err_code, write_err_code) = match shutdown {
            StreamShutdown::Read { error_code } => (Some(error_code), None),
            StreamShutdown::Write { error_code } => (None, Some(error_code)),
            StreamShutdown::Both {
                read_error_code,
                write_error_code,
            } => (Some(write_error_code), Some(read_error_code)),
        };

        if let Some(error_code) = read_err_code {
            audit_stats.set_sent_stop_sending_error_code(error_code as _);
            let _ = qconn.stream_shutdown(
                stream_id,
                quiche::Shutdown::Read,
                error_code,
            );
        }
        if let Some(error_code) = write_err_code {
            audit_stats.set_sent_reset_stream_error_code(error_code as _);
            let _ = qconn.stream_shutdown(
                stream_id,
                quiche::Shutdown::Write,
                error_code,
            );
        }
```

dineshadhi: I experimented with a tuple match style in a previous commit (now squashed), but it felt a bit dense and harder to parse at a glance. 

The current implementation is much more straightforwardm. It lays out the logic linearly, which makes it easier to follow. 

I suggest we stick with the current version for better readability.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -889,11 +889,15 @@ impl<H: DriverHooks> H3Driver<H> {
...
                 },
                 Ok(OutboundFrame::FlowShutdown { flow_id, stream_id }) => {
-                    self.finish_stream(
+                    self.shutdown_stream(
                         qconn,
                         stream_id,
-                        Some(quiche::h3::WireErrorCode::NoError as u64),
-                        Some(quiche::h3::WireErrorCode::NoError as u64),
+                        StreamShutdown::Both {
+                            read_error_code: quiche::h3::WireErrorCode::NoError
```
LPardue: nit: lets `use` WireErrorCode to save a few chars

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -925,30 +929,18 @@ impl<H: DriverHooks> H3Driver<H> {
...
     }
 
-    /// Removes a stream from the stream map if it exists. Also optionally sends
-    /// `RESET` or `STOP_SENDING` frames if `write` or `read` is set to an
-    /// error code, respectively.
-    fn finish_stream(
+    /// Cleans up internal state for the indicated HTTP/3 stream.
+    ///
+    /// This function removes the stream from the stream map, closes any pending
+    /// futures, removes associated DATAGRAM flows, and sends a `StreamClosed`
```
LPardue: What is a `StreamClosed` event? Can you add a doc link?

tokio-quiche/src/http3/driver/mod.rs
LPardue: This one works for me. I kind hate tuple magic sometimes

tokio-quiche/src/http3/driver/mod.rs
dineshadhi: It is `H3Event::StreamClosed`. I linked with the docs now.

tokio-quiche/src/http3/driver/mod.rs
dineshadhi: Done!

PR #2346, opened 2026-02-04
Implement MAX_PROBES for PMTUD probe failure tolerance

cf-rhett:
## Summary

Implement MAX_PROBES (=3) consecutive failure requirement before treating a probe size as an MTU limitation. This prevents false positives from random packet loss on lossy networks.

Per RFC 8899 Section 5.1.2, some probe loss is expected while searching, and loss of a single probe is not an indication of a PMTU problem.

### DIFF
--- a/quiche/src/pmtud.rs
+++ b/quiche/src/pmtud.rs
@@ -1,6 +1,30 @@
-/// Contains the logic to implement PMTUD. Given a maximum supported MTU,
-/// finds the PMTU between the given max and [`MIN_CLIENT_INITIAL_LEN`].
-use crate::MIN_CLIENT_INITIAL_LEN;
+//! Path MTU Discovery ([RFC 8899] DPLPMTUD).
+//!
+//! Discovers the path MTU using loss-based inference: probe packets are sent
+//! and their acknowledgment (or lack thereof) determines path capacity.
+//!
+//! # Algorithm
+//!
+//! Optimistic binary search between [`MIN_PLPMTU`] (1200) and max supported
+//! MTU:
+//! 1. Probe at max MTU
+//! 2. On max_probes consecutive failures, record as smallest failed size
+//! 3. Binary search between largest success and smallest failure
+//! 4. Complete when difference ≤ 1 byte
+//!
+//! A successful probe at any point resets the failure counter and updates
+//! the largest known working size.
+//!
+//! [RFC 8899]: https://datatracker.ietf.org/doc/html/rfc8899
+
+/// Maximum number of probe attempts before treating a size as failed.
+/// https://datatracker.ietf.org/doc/html/rfc8899#section-5.1.2
+pub(crate) const MAX_PROBES_DEFAULT: u8 = 3;
+
+/// Min Packetization Layer Path MTU (PLPMTU).
+/// https://datatracker.ietf.org/doc/html/rfc8899#section-5.1.2
+/// For QUIC, this is 1200 bytes per https://datatracker.ietf.org/doc/html/rfc9000#section-14.1
+const MIN_PLPMTU: usize = crate::MIN_CLIENT_INITIAL_LEN;
 
 #[derive(Default)]
 pub struct Pmtud {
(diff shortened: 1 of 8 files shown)


quiche/src/pmtud.rs
```diff
@@ -1,6 +1,30 @@
...
+//! 4. Complete when difference ≤ 1 byte
+//!
+//! A successful probe at any point resets the failure counter and updates
+//! the largest known working size.
+//!
+//! [RFC 8899]: https://datatracker.ietf.org/doc/html/rfc8899
+
+/// Maximum number of probe attempts before treating a size as failed.
+/// https://datatracker.ietf.org/doc/html/rfc8899#section-5.1.3
+const MAX_PROBES: u8 = 3;
```
LPardue: Since we're adding stuff here, I think it would be better to define this as MAX_PROBES_DEFAULT and allow configuration of the max_probes value in the Pmtu struct (i.e. pass it in the new()`. 

This would more clearly show we're supporting the RFC default behaviour, while allowing future flexibility.

quiche/src/pmtud.rs
```diff
@@ -1,6 +1,30 @@
...
+
+/// Base Packetization Layer Path MTU (PLPMTU).
+/// https://datatracker.ietf.org/doc/html/rfc8899#section-5.1.2
+/// For QUIC, this is 1200 bytes per https://datatracker.ietf.org/doc/html/rfc9000#section-14.1
+const MIN_PLPMTU: usize = 1200;
```
LPardue: seems like we should still rely on the crate' size definition here

```suggestion
const MIN_PLPMTU: usize = crate::MIN_CLIENT_INITIAL_LEN;
```

quiche/src/pmtud.rs
```diff
@@ -141,18 +166,29 @@ impl Pmtud {
...
-
-        // Check if we have one instance of a failed probe so that a min
-        // comparison can be made otherwise if this is the first failed
-        // probe just record it
-        self.smallest_failed_probe_size = self
-            .smallest_failed_probe_size
-            .map_or(Some(probe_size), |existing_size| {
-                Some(std::cmp::min(probe_size, existing_size))
-            });
+        let probe_size = std::cmp::max(probe_size, MIN_PLPMTU);
```
LPardue, line 187: The old comment seemed useful TBH

cf-rhett: Agreed and restored. Just got lost in the refactor

quiche/src/pmtud.rs
cf-rhett: I went and made this configurable back to TQ. Let me know what you think here

quiche/src/pmtud.rs
cf-rhett: Done

tokio-quiche/src/settings/quic.rs
```diff
@@ -177,6 +177,13 @@ pub struct QuicSettings {
     /// Defaults to `false`.
     pub discover_path_mtu: bool,
 
+    /// Configures the maximum number of PMTUD probe attempts before treating
+    /// a size as failed.
+    ///
+    /// Defaults to 3. If 0 is passed, the default value is used.
```
LPardue: nit: The public docs are a great place to link to the RFC (you have that internally so just copy that here)

quiche/src/pmtud.rs
LPardue: noice

tokio-quiche/src/settings/quic.rs
cf-rhett: Done! I wasn't sure where to link and where to not

quiche/src/pmtud.rs
```diff
@@ -192,15 +263,21 @@ mod tests {
...
+        let pmtud = Pmtud::new(1350, 1);
         assert_eq!(pmtud.get_current_mtu(), 1200);
         assert_eq!(pmtud.get_probe_size(), 1350);
         assert!(pmtud.should_probe());
     }
 
+    #[test]
+    fn pmtud_max_probes_zero_uses_default() {
+        let pmtud = Pmtud::new(1500, 0);
+        assert_eq!(pmtud.max_probes, MAX_PROBES_DEFAULT);
```
antoniovicente, line 275: Also need a test for max probes set to the provided value.

```
        let pmtud = Pmtud::new(1500, 5);
        assert_eq!(pmtud.max_probes, 5);
        assert_ne!(pmtud.max_probes, MAX_PROBES_DEFAULT);
```

quiche/src/pmtud.rs
```diff
@@ -127,6 +167,8 @@ impl Pmtud {
 
     /// Records a successful probe and returns the largest successful probe size
     pub fn successful_probe(&mut self, probe_size: usize) -> Option<usize> {
+        self.probe_failure_count = 0;
```
antoniovicente, line 170: Do we have enough testing for `probe_failure_count` being set back to 0 on success or when reducing the probe packet size?

I'm thinking a setup where the first probe always fails, but second succeeds.

Make sure that convergence to the correct MTU of 1400 is achieved if starting from 1500 with a max retry of 2 or 3.

cf-rhett: Added!

quiche/src/pmtud.rs
cf-rhett: Added!

PR #2348, opened 2026-02-04
qlog: add ex_data field to MetricsUpdated for extension data

estcarisimo:
Add an ex_data field to MetricsUpdated that allows non-standard metrics (like delivery_rate, send_rate, ack_rate) to be serialized as flattened JSON fields. This enables extensibility without modifying the core qlog schema.

### DIFF
--- a/qlog/src/testing/event_tests.rs
+++ b/qlog/src/testing/event_tests.rs
@@ -173,7 +173,7 @@ fn packet_header() {
 #[test]
 fn metrics_updated_with_ex_data() {
     // Test that ex_data fields are flattened into the same object
-    let mut ex_data = ExData::from([(
+    let ex_data = ExData::from([(
         "delivery_rate".to_string(),
         serde_json::json!(DELIVERY_RATE),
     )]);
@@ -202,7 +202,7 @@ fn metrics_updated_ex_data_collision() {
     // With serde's preserve_order feature and ex_data at the top of the
     // struct, standard fields are serialized last and take precedence.
 
-    let mut ex_data = ExData::from([(
+    let ex_data = ExData::from([(
         "min_rtt".to_string(),
         serde_json::json!(COLLISION_VALUE),
     )]);


qlog/src/events/quic.rs
```diff
@@ -772,6 +773,10 @@ pub struct MetricsUpdated {
     pub packets_in_flight: Option<u64>,
 
     pub pacing_rate: Option<u64>,
+
+    /// Extension data for non-standard fields (flattened into data object)
```
LPardue:
```suggestion
    /// Extension data for non-standard fields. `flatten` causes these fields to
    /// be serialized into the `data` field of a qlog event. On deserialize,
    /// unknown fields are collected into `ex_data`.
```

qlog/src/events/quic.rs
```diff
@@ -815,6 +820,15 @@ mod tests {
...
     use crate::testing::*;
 
+    // Test constants for MetricsUpdated tests
+    const TEST_MIN_RTT: f32 = 10.0;
+    const TEST_SMOOTHED_RTT: f32 = 15.0;
+    const TEST_CONGESTION_WINDOW: u64 = 12000;
+    const TEST_PACING_RATE: u64 = 500000;
+    const TEST_DELIVERY_RATE: u64 = 1000000;
+    const TEST_SEND_RATE: u64 = 500000;
+    const TEST_COLLISION_VALUE: f32 = 999.0;
```
LPardue: nit: since this is the `tests` module, we can drop the `TEST_` prefix

qlog/src/testing/event_tests.rs
```diff
@@ -142,3 +144,143 @@ fn packet_sent_event_some_frames() {
...
+}"#;
+
+    assert_eq!(serde_json::to_string_pretty(&pkt_hdr).unwrap(), log_string);
+}
+
+#[test]
+fn metrics_updated_with_ex_data() {
+    // Test that ex_data fields are flattened into the same object
+    let mut ex_data = ExData::new();
+    ex_data.insert(
```
LPardue: nit Testing two fields doesn't seem much more advantageous that just testing one. Plus, since the ex_data doesn't need mutability other than to do inserts we can instead do somethign like

```
let ex_data = ExData::from([
    ("delivery_rate".to_string(), serde_json::json!(DELIVERY_RATE)),
]);
```

or send_rate, but just pick one.

qlog/src/testing/event_tests.rs
```diff
@@ -142,3 +144,140 @@ fn packet_sent_event_some_frames() {
...
+#[test]
+fn metrics_updated_ex_data_collision() {
+    // Test collision: same field set via struct AND ex_data
+    // serde(flatten) behavior: ex_data (flattened map) takes precedence
+    // over struct fields that serialize later in the output.
+    //
+    // NOTE: This test documents serde's silent overwrite behavior.
+    // serde does NOT error or panic on duplicate keys - it just overwrites.
+    // This test exists to warn users: avoid ex_data keys matching struct
+    // fields!
```
LPardue: Users don't read tests 😄 

If we need to warn them, that's something we need to put in the public API. A short sentence next to the ex_data field itself should suffice.

Then we can remove this paragraph since it seems to duplicate what is described above already.

qlog/src/testing/event_tests.rs
```diff
@@ -142,3 +144,140 @@ fn packet_sent_event_some_frames() {
...
+    let metrics = MetricsUpdated {
+        min_rtt: Some(MIN_RTT), // struct field value
+        ex_data,                // ex_data also has min_rtt
+        ..Default::default()
+    };
+
+    let json = serde_json::to_value(&metrics).unwrap();
+
+    // WARNING: ex_data wins in collision - avoid using ex_data keys
+    // that match existing struct field names!
```
LPardue: Can remove this too

qlog/src/testing/event_tests.rs
```diff
@@ -142,3 +144,140 @@ fn packet_sent_event_some_frames() {
...
+}
+
+#[test]
+fn metrics_updated_round_trip() {
+    // Test serialization -> deserialization round-trip
+    let mut ex_data = ExData::new();
+    ex_data.insert(
+        "delivery_rate".to_string(),
+        serde_json::json!(DELIVERY_RATE),
+    );
```
LPardue:
```suggestion
    let ex_data = ExData::from([
        ("delivery_rate".to_string(), serde_json::json!(DELIVERY_RATE)),
    ]);
```

qlog/src/testing/event_tests.rs
estcarisimo: :)

qlog/src/testing/event_tests.rs
LPardue: It seems this suggestion didn't apply or was wiped, same with the other one.

LPardue: Seems github keeps swallowing these statements, so bumping

estcarisimo: oh my bad, I had some back and forth with `cargo +nightly fmt` dealing with that. I think it will pass it.

LPardue: Let me rephrase the ask. I can't predict the fomatting requirements while reviewing MRs. So please make the functional change I proposed (initialize the ex_data directly here and in the other instance, instead of creating a mutable one) and apply whatever formatting change nightly requires.

PR #2349, opened 2026-02-04
Refactor qlog crate testing

LPardue:
While it is typical to include in-line unit tests in rust files, qlog is a _bit_ weird. There's a form of inheritance hierachy eith th abstract base event and then concrete event data. Furthermore there are extension points all over the plac. Tests can be verbose, especially if checking serialized outputs or doing roundtripping.

Furthermore, the amount of testing is quite small at the moment and getting more coverage would be a good thing.

In order to make things a bit clearer, this PR is a pure refactor that moves testing related things out of qlog/lib.rs and into more-targetted files. New tests can then be added under this structure, helping to keep the event definition files lighterweight.

- **qlog: separate out testing module**
- **qlog: separate out tests in purpose-focused files**


qlog/src/testing/event_tests.rs
```diff
@@ -0,0 +1,144 @@
...
+      "version": "1",
+      "scil": 8,
+      "dcil": 8,
+      "scid": "7e37e4dcc6682da8",
+      "dcid": "36ce104eee50101c"
+    },
+    "raw": {
+      "length": 1251,
+      "payload_length": 1224
+    },
```
estcarisimo, line 91: If I looked at this correctly, this part is (nearly) identical to the prior test. As we are streamlining the code, is it worth deduplicating this part? There may be a counterargument that having all within the tests help the readability

qlog/src/testing/trace_tests.rs
```diff
@@ -0,0 +1,156 @@
...
+    let log_string = r#"{
+  "vantage_point": {
+    "type": "server"
+  },
+  "title": "Quiche qlog trace",
+  "description": "Quiche qlog trace description",
+  "configuration": {
+    "time_offset": 0.0
+  }
+}"#;
```
estcarisimo, line 71: Same comment here, is it worth deduplicating `log_string`

qlog/src/testing/event_tests.rs
```diff
@@ -0,0 +1,144 @@
...
+      }
+    ]
+  }
+}"#;
+
+    let pkt_hdr = make_pkt_hdr(PacketType::Initial);
+
+    let frames = vec![
+        QuicFrame::Padding {
+            payload_length: 1234,
```
estcarisimo, line 115: I am a big fan of constants, can we replace numbers by constants wherever is possible? Another potential optimization is avoid having the same number 1234 here and in the `log_string`

qlog/src/testing/event_tests.rs
LPardue: Since this is a restructure PR, I'd prefer to avoid making any functional changes.

That can always be done as a followup.

qlog/src/testing/event_tests.rs
LPardue: They need to be the same value, since the test asserts they are the same value.

estcarisimo: Yep, I was wondering both the log and the QuicFrame::Padding can use the same constant. Just minor cosmetic comment biased by my preference for constant

estcarisimo: But I believe the spirit of not making changes in this PR prevails

qlog/src/testing/event_tests.rs
LPardue: To answer the question though: yes its duplicative but that's inherent in the nature of  the test. This is not about roundtripping but checking the output of the event serialization matching expectation

There's probably ways to improve this stuff but we can deal with that in future.

PR #2351, opened 2026-02-09
Allow updates to the recovery config after sending ACK-only packets.

antoniovicente:
This allows in-handshake config overrides to work reliably with multi-packet handshakes.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -7452,20 +7452,12 @@ impl<F: BufFactory> Connection<F> {
             Ok(_) => (),
 
             Err(Error::Done) => {
-                // Apply in-handshake configuration from callbacks before any
-                // packet with inflight data have been sent. The congestion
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -7453,8 +7453,21 @@ impl<F: BufFactory> Connection<F> {
 
             Err(Error::Done) => {
                 // Apply in-handshake configuration from callbacks before any
-                // packet has been sent.
-                if self.sent_count == 0 {
+                // packet with inflight data have been sent. The congestion
+                // controller doesn't track anything interesting until inflight
+                // data is sent. Handshake ACKs may be sent prior to arrival of
+                // the full client hello, but the send of ACK only packets
```
LPardue: nit

```suggestion
                // the full ClientHello, but the send of ACK only packets
```

PR #2353, opened 2026-02-12
Adding qlog support for lost packets and bytes through ex_data

estcarisimo:


qlog/src/testing/event_tests.rs
```diff
@@ -274,3 +274,114 @@ fn metrics_updated_no_ex_data() {
...
 }
+
+// Test constants for loss metrics
+const LOST_PACKETS: u64 = 42;
+const LOST_BYTES: u64 = 52500;
+const LOST_PACKETS_DELTA: u64 = 3;
+const LOST_BYTES_DELTA: u64 = 3750;
+
+#[test]
+fn metrics_updated_with_loss_ex_data() {
```
antoniovicente: I don't think we need these kinds of tests since we're not doing any other changes to qlog proper.

quiche/src/recovery/mod.rs
```diff
@@ -587,6 +589,35 @@ impl QlogMetrics {
...
+            if let Some(val) = latest.lost_packets {
+                let delta = val - self.lost_packets.unwrap_or(0);
+                self.lost_packets = latest.lost_packets;
+                emit_event = true;
+                ex_data.insert(
+                    "cf_lost_packets".to_string(),
+                    serde_json::json!(val),
+                );
+                ex_data.insert(
+                    "cf_lost_packets_delta".to_string(),
```
antoniovicente: Should we consider only including the deltas for now?

I guess both provide value, no objection to including both.

estcarisimo: Yeah, deltas provide much more value right now. At the risk of overpopulating the QLOG i am inclined towards adding both now

qlog/src/testing/event_tests.rs
estcarisimo: I will remove the tests

quiche/src/recovery/mod.rs
LPardue: Consider representing these as a structure defined in quiche rather than two independent fields e.g.

```
struct CfLostPackets {
  total: u64,
  delta: Option<u64>,
}
```
(activitity for the reader, to apply the needed serde decorations)
More formally defining our custom event object model will help maintenance in the longer run.

antoniovicente: I think the idea behind your suggestion is that we could do something like:

```
struct TotalAndDelta {
  total: u64,
  delta: u64,
}

ex_data.insert(
  "cf_lost_packets".to_string(),
  serde_json::json!(TotalAndDelta {total, delta}),
);
```

LPardue: I'd take it a bit further and do something like I proposed in https://github.com/cloudflare/quiche/pull/2355. That way, you'd use CustomCfQlogField to wrap any/all of these fields, with the value being generic but needing to implement a Serialize trait

estcarisimo: @LPardue this is amazing. A lot of Rust magic especially with all those Traits. Merged and adding into my pipeline

quiche/src/recovery/mod.rs
LPardue: In theory, you could tweak these to also use a CustomQlogField. It would be more verbose now, but in future might help. Feel free to punt on that though

LPardue (approved):

quiche/src/recovery/mod.rs
antoniovicente: Would be good for a followup.

quiche/src/recovery/mod.rs
```diff
@@ -587,6 +629,38 @@ impl QlogMetrics {
...
+                        total: latest.lost_packets,
+                        delta: Some(val - self.lost_packets.unwrap_or(0)),
+                    });
+                self.lost_packets = latest.lost_packets;
+                emit_event = true;
+
+                ex_data.insert(
+                    lost_packets.name().to_string(),
+                    lost_packets.as_json_value(),
+                );
```
antoniovicente: Could be simplified by introducing a wrapper around ex_data that provides:

```
fn insert(&mut self, field: CustomCfQlogField) {
  self.ex_data.insert(field.name().to_string(),
                                  lost_packets.as_json_value());
}
```

Even better would be to be able to do:
```
wrapped_ex_data.insert("cf_lost_packets", TotalAndDelta {
                        total: latest.lost_packets,
                        delta: Some(val - self.lost_packets.unwrap_or(0)),
                      });

By having insert take a T: Serialize as the second argument and using the CustomCfQlogField framework introduced in this PR.
```

quiche/src/recovery/mod.rs
```diff
@@ -601,63 +620,47 @@ impl QlogMetrics {
...
                 self.lost_packets = latest.lost_packets;
                 emit_event = true;
-
-                ex_data.insert(
-                    lost_packets.name().to_string(),
-                    lost_packets.as_json_value(),
-                );
+                ex_data.insert("cf_lost_packets", TotalAndDelta {
+                    total: latest.lost_packets,
+                    delta: Some(val - self.lost_packets.unwrap_or(0)),
```
antoniovicente, line 652: You introduced a bug here.

self.lost_packets is updated on line 648 above, so `val - self.lost_packets` is always 0.

Please move the ex_data.insert before updating `self.lost_packets`

quiche/src/recovery/mod.rs
```diff
@@ -601,63 +620,47 @@ impl QlogMetrics {
...
                 self.lost_bytes = latest.lost_bytes;
                 emit_event = true;
-                ex_data.insert(
-                    lost_bytes.name().to_string(),
-                    lost_bytes.as_json_value(),
-                );
+                ex_data.insert("cf_lost_bytes", TotalAndDelta {
+                    total: latest.lost_bytes,
+                    delta: Some(val - self.lost_bytes.unwrap_or(0)),
+                });
```
antoniovicente, line 663: See previous comment.

quiche/src/recovery/mod.rs
estcarisimo: done

quiche/src/recovery/mod.rs
estcarisimo: repaired

PR #2354, opened 2026-02-13
fix: MAX_DATA frames were not retransmitted when lost

gregor-cf:
The retransmission for MAX_DATA frames was broken. When a MAX_DATA frame was lost, we set `self.almost_full = true`. In `send_single()` we there is also a check for `flow_control.max_data() < flow_control.max_data_next()`. If the application has not read from the recv buffer, this check is false and the MAX_DATA will not be sent. We've seen this issue happen in real traffic.

I've removed `almost_full` and instead do the check for `should_update_max_data()` directly. That seems a lot cleaner than than setting the flag in several places. And I've added a flag to force-send a max-data frame.

### DIFF
--- a/quiche/src/flowcontrol.rs
+++ b/quiche/src/flowcontrol.rs
@@ -127,7 +127,8 @@ impl FlowControl {
     /// the current window.
     pub fn ensure_window_lower_bound(&mut self, min_window: u64) {
         if min_window > self.window {
-            self.window = min_window;
+            // ... we still need to clamp to `max_window`
+            self.window = std::cmp::min(min_window, self.max_window);
         }
     }
 }
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1324,7 +1324,7 @@ where
     flow_control: flowcontrol::FlowControl,
 
     /// Whether we send MAX_DATA frame.
-    almost_full: bool,
+    should_send_max_data: bool,
 
     /// Number of stream data bytes that can be buffered.
     tx_cap: usize,
@@ -2011,7 +2011,7 @@ impl<F: BufFactory> Connection<F> {
                 cmp::min(max_rx_data / 2 * 3, DEFAULT_CONNECTION_WINDOW),
                 config.max_connection_window,
             ),
-            almost_full: false,
+            should_send_max_data: false,
 
             tx_cap: 0,
             tx_cap_factor: config.tx_cap_factor,
@@ -4094,7 +4094,7 @@ impl<F: BufFactory> Connection<F> {
                     },
 
                     frame::Frame::MaxData { .. } => {
-                        self.almost_full = true;
+                        self.should_send_max_data = true;
                     },
 
                     frame::Frame::NewConnectionId { seq_num, .. } => {
@@ -4525,8 +4525,13 @@ impl<F: BufFactory> Connection<F> {
                     },
                 };
 
-                // Autotune the stream window size.
-                stream.recv.autotune_window(now, path.recovery.rtt());
+                // Autotune the stream window size, but only if this is not a
+                // retransmission (on a retransmit the stream will be in
+                // `self.streams.almost_full()` but it's `almost_full()`
+                // method returns false.
+                if stream.recv.almost_full() {
+                    stream.recv.autotune_window(now, path.recovery.rtt());
+                }
 
                 let frame = frame::Frame::MaxStreamData {
                     stream_id,
@@ -4548,26 +4553,26 @@ impl<F: BufFactory> Connection<F> {
                     flow_control.ensure_window_lower_bound(
                         (recv_win as f64 * CONNECTION_WINDOW_FACTOR) as u64,
                     );
-
-                    // Also send MAX_DATA when MAX_STREAM_DATA is sent, to avoid a
-                    // potential race condition.
-                    self.almost_full = true;
                 }
             }
 
             // Create MAX_DATA frame as needed.
-            if self.almost_full &&
+            if flow_control.should_update_max_data() &&
                 flow_control.max_data() < flow_control.max_data_next()
             {
-                // Autotune the connection window size.
+                // Autotune the connection window size. We only tune the window
+                // if we are sending an "organic" update, not on retransmits.
                 flow_control.autotune_window(now, path.recovery.rtt());
+                self.should_send_max_data = true;
+            }
 
+            if self.should_send_max_data {
                 let frame = frame::Frame::MaxData {
                     max: flow_control.max_data_next(),
                 };
 
                 if push_frame_to_pkt!(b, frames, frame, left) {
-                    self.almost_full = false;
+                    self.should_send_max_data = false;
 
                     // Commits the new max_rx_data limit.
                     flow_control.update_max_data(now);
@@ -5479,10 +5484,6 @@ impl<F: BufFactory> Connection<F> {
             q.add_event_data_with_instant(ev_data, now).ok();
         });
 
-        if self.should_update_max_data() {
-            self.almost_full = true;
-        }
-
         if priority_key.incremental && readable {
             // Shuffle the incremental stream to the back of the queue.
             self.streams.remove_readable(&priority_key);
@@ -5849,9 +5850,6 @@ impl<F: BufFactory> Connection<F> {
             Shutdown::Read => {
                 let consumed = stream.recv.shutdown()?;
                 self.flow_control.add_consumed(consumed);
-                if self.flow_control.should_update_max_data() {
-                    self.almost_full = true;
-                }
 
                 if !stream.recv.is_fin() {
                     self.streams.insert_stopped(stream_id, err);
@@ -7770,7 +7768,8 @@ impl<F: BufFactory> Connection<F> {
         let send_path = self.paths.get(send_pid)?;
         if (self.is_established() || self.is_in_early_data()) &&
             (self.should_send_handshake_done() ||
-                self.almost_full ||
+                self.flow_control.should_update_max_data() ||
+                self.should_send_max_data ||
                 self.blocked_limit.is_some() ||
                 self.dgram_send_queue.has_pending() ||
                 self.local_error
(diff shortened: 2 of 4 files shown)


quiche/src/lib.rs
```diff
@@ -4451,26 +4451,25 @@ impl<F: BufFactory> Connection<F> {
...
                 }
             }
 
             // Create MAX_DATA frame as needed.
-            if self.almost_full &&
+            if flow_control.should_update_max_data() &&
                 flow_control.max_data() < flow_control.max_data_next()
             {
                 // Autotune the connection window size.
                 flow_control.autotune_window(now, path.recovery.rtt());
```
gregor-cf, line 4565: Note: we should only tune, if it's an "organic" window update, not a retransmit.

quiche/src/lib.rs
```diff
@@ -4451,26 +4451,25 @@ impl<F: BufFactory> Connection<F> {
...
-
-                    // Also send MAX_DATA when MAX_STREAM_DATA is sent, to avoid a
-                    // potential race condition.
-                    self.almost_full = true;
                 }
             }
 
             // Create MAX_DATA frame as needed.
-            if self.almost_full &&
+            if flow_control.should_update_max_data() &&
```
gregor-cf, line 4560: because of multiple mut borrows, I need to use `flow_control.should_update_max_data()` and not `self.should_update_max_data()`. And with that `self.should_update_max_data()` becomes unused, so it's being removed too.

LPardue:
Seems like we didn't have a test to catch such behaviour we think is wrong. So need a test to prevent future regressions

gregor-cf:
> Seems like we didn't have a test to catch such behaviour we think is wrong. So need a test to prevent future regressions

Indeed. The problem is that `test_utils` has no infrastructure for testing retransmits as far as I could tell :-(. The only test  I could find in a cursory search that deals with retransmit is `early_retransmit()`, which only checks that stream data is send in PTO probes but no tests that check the loss signal from the `recovery`. So we'd probably need to add the plumbing for that to test_utils and then add tests for all frame types that need to be retransmitted.

LPardue:
Take a look  at early_retransmit test, it might provide a template for how to do it for max_data without a full-blown utility change

gregor-cf:
> Take a look at early_retransmit test, it might provide a template for how to do it for max_data without a full-blown utility change

I've added a test.

quiche/src/flowcontrol.rs
```diff
@@ -59,7 +59,7 @@ impl FlowControl {
         Self {
             max_data,
 
-            window,
+            window: std::cmp::min(window, max_window),
```
LPardue: Was this the precursor to https://github.com/cloudflare/quiche/pull/2356/changes?

If so, I'd like to keep this PR focused on just the retransmission aspect

quiche/src/lib.rs
LPardue: Is this a meant as a todo for the PR, or an observation? If its an observation it would probably be useful to capture as a code comment

quiche/src/tests.rs
```diff
@@ -7006,6 +7005,87 @@ fn early_retransmit(
...
+    // flow control on client is unchanged
+    assert_eq!(pipe.client.max_tx_data, 30);
+    assert_eq!(pipe.client.stream_capacity(0), Ok(0));
+    // the server sent MAX_DATA (even though it got lost, so it has updated
+    // its max_rx_data
+    assert_eq!(pipe.server.max_rx_data(), 75);
+
+    // Wait until PTO expires. Since the RTT is very low, wait a bit more.
+    // We assume that this is long enough for recovery to mark the previous
+    // packet as lost.
```
LPardue: Rather than rely on PTO, you might consider triggering a packet loss per the new `trigger_ack_based_loss()` function I needed to add in https://github.com/cloudflare/quiche/pull/2358/changes#diff-03a51114186c5434909f1302ba3f0296f3b8a044e16c53a78622ccd62fae9705R565

If PTO if preferable, I have another non-public commit somewhere to refactor the various instances of the sleeps into a util function

quiche/src/tests.rs
```diff
@@ -7006,6 +7005,87 @@ fn early_retransmit(
...
+    pipe.server.on_timeout();
+    pipe.advance().unwrap();
+    // connection level flow-control is updated
+    assert_eq!(pipe.client.max_tx_data, 75);
+    // ... and so is stream-level
+    assert_eq!(pipe.client.stream_capacity(0), Ok(30));
+
+    // only stream_data and crypto frames increment retrans
+    // TODO: is this correct?
+    assert_eq!(pipe.client.stats().retrans, 0);
```
LPardue, line 7328: I think this is a good observation. The way retransmit stats work today is focused on reliable stream data - both crypto and QUIC streams. It's kind of broken in the context of comparing to TCP due to the difference in retransmission behaviour. 

I would leave them as. If we have a need for observability of retransmitted control frames, then that would probably best be done as a separate counter. I'm not sure its super interesting though.

quiche/src/test_utils.rs
```diff
@@ -36,7 +36,7 @@ pub struct Pipe {
 }
 
 impl Pipe {
-    pub fn new(cc_algorithm_name: &str) -> Result<Pipe> {
+    pub fn default_config(cc_algorithm_name: &str) -> Result<Config> {
```
LPardue, line 39: Nice!

quiche/src/tests.rs
LPardue: @gregor-cf  I merged the other PR so if you want to use the new function its there ready

quiche/src/lib.rs
gregor-cf: was just and FYI. added comment in code

quiche/src/flowcontrol.rs
gregor-cf: no, that was separate. But it makes sense to add it to 2356.

quiche/src/lib.rs
```diff
@@ -4534,26 +4451,25 @@ impl<F: BufFactory> Connection<F> {
...
                 flow_control.max_data() < flow_control.max_data_next()
```
antoniovicente, line 4561: This PR changes the behavior of auto tuning slightly since max_data() is being updated in cases where `flow_control.autotune_window` is not called.  I think the solution is to use a copy of `flow_control.max_data()` that is used exclusively by auto tuning which tracks what was the byte offset when auto tuning was last performed.

quiche/src/tests.rs
gregor-cf: Nice. Actually found another bug once I switched to it. (STREAM_MAX_DATA was autotuning on retransmits and inflating the stream window....)

quiche/src/lib.rs
```diff
@@ -4548,26 +4553,26 @@ impl<F: BufFactory> Connection<F> {
                     flow_control.ensure_window_lower_bound(
                         (recv_win as f64 * CONNECTION_WINDOW_FACTOR) as u64,
                     );
-
-                    // Also send MAX_DATA when MAX_STREAM_DATA is sent, to avoid a
-                    // potential race condition.
-                    self.almost_full = true;
```
antoniovicente, line 4554: Can you comment on the removal of `self.almost_full = true` when sending MAX_STREAM_DATA?

I think that the issue is that this shouldn't be necessary now that MaxData frames are correctly retransmitted.

quiche/src/lib.rs
gregor-cf: I think auto-tuning was incorrect before this PR. We wcould call auto-tune back-to-back (e.g., during a retransmission where the app did consume data) and auto-tune would *just* look at the timestamps and happily double the window again during the retransmission. Note, the auto-tune logic does *not* look at max_data, it only checks the last_update time.  

I think this behavior is "more" correct. On a retransmit we call `update_max_data()` to opportunistically send a larger MAX_DATA than in the original, lost frame (if `consumed` has increased)` and reset the clock for auto-tuning updates. That seems reasonable, since we're giving the peer the largest possible window immediately.  In the worst-case for auto-tuning, the window doubling is delayed one RTT.

antoniovicente: I think my worry was that every time that we detect MaxData lost, we may not rerun auto tuning for a while.  thinking about this more, it likely doesn't matter.

quiche/src/flowcontrol.rs
```diff
@@ -127,7 +127,8 @@ impl FlowControl {
     /// the current window.
     pub fn ensure_window_lower_bound(&mut self, min_window: u64) {
         if min_window > self.window {
-            self.window = min_window;
+            // ... we still need to clamp to `max_window`
+            self.window = std::cmp::min(min_window, self.max_window);
```
antoniovicente, line 131: This change needs a test.  Consider at least adding:

```
--- a/quiche/src/flowcontrol.rs
+++ b/quiche/src/flowcontrol.rs
@@ -223,5 +223,9 @@ mod tests {
         // Window changed to the new value.
         fc.ensure_window_lower_bound(w * 2);
         assert_eq!(fc.window(), 40);
+
+        // max_window is respected.
+        fc.ensure_window_lower_bound(1000);
+        assert_eq!(fc.window(), 100);
     }
 }
```

quiche/src/lib.rs
gregor-cf: That said, it might be worthwhile revisiting the auto tuning algorithm but I think that's unrelated to the PR

quiche/src/flowcontrol.rs
antoniovicente: I see you're adding this exact test in https://github.com/cloudflare/quiche/pull/2356

quiche/src/lib.rs
antoniovicente: I think I agree, we should look at auto tuning in followups.

quiche/src/lib.rs
gregor-cf: I don't think it's necessary. The logic to send MAX_DATA frame should work just on its own and any window increase from line 4549 above will be taken into account in `should_update_max_data()`

Setting `almost_full` here, can also interfere with auto-tuning. Either by growing the window overly aggressively (old logic) since (a) the auto-tuning algorithm only checks timestamps, not bytes covered and (b) we potentially just increased the window in like 4549 and auto-tune might double it on top of that. If we don't run tuning (this PR) when we set `almost_full` / `should_send_max_data` flag, we might prevent tuning by sending MAX_DATA updates from this path.

PR #2355, opened 2026-02-13
Example of using a custom generic structure to wrap extension fields

LPardue:
The generic type should probably be defined outside of the recovery module, but this was just an example to help further the discussions

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -37,6 +37,8 @@ use crate::Result;
 
 #[cfg(feature = "qlog")]
 use qlog::events::EventData;
+#[cfg(feature = "qlog")]
+use serde::Serialize;
 
 use smallvec::SmallVec;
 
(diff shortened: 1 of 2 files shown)


quiche/src/recovery/mod.rs
```diff
@@ -591,29 +631,32 @@ impl QlogMetrics {
...
-                let delta = val - self.lost_packets.unwrap_or(0);
+                let lost_packets =
+                    CustomQlogField::new("cf_lost_packets", TotalAndDelta {
+                        total: latest.lost_packets,
+                        delta: Some(val - self.lost_packets.unwrap_or(0)),
+                    });
                 self.lost_packets = latest.lost_packets;
                 emit_event = true;
+
                 ex_data.insert(
```
LPardue, line 642: You could potentially further change `ex_data` to be more than a type def, and have a method that takes a CustomQlogField to simplify all of the call sites that want to insert

quiche/src/recovery/mod.rs
```diff
@@ -485,6 +487,44 @@ struct QlogMetrics {
...
+    fn name(&self) -> &'static str;
+    fn as_json_value(&self) -> serde_json::Value;
+}
+
+#[cfg(feature = "qlog")]
+#[serde_with::skip_serializing_none]
+#[derive(Serialize)]
+struct TotalAndDelta {
+    total: Option<u64>,
+    delta: Option<u64>,
```
LPardue, line 501: These could be values rather than options, just using it to demonstrate the power of relying on the Serialize properties

PR #2356, opened 2026-02-13
fix pathological MAX_DATA/STREAM_MAX_DATA update behavior

gregor-cf:
The flow control logic (both stream and connection) checks how much
flow control budget is still available and if it falls to less than
half the flow control window, MAX_DATA is updated. But quiche clamps
the initial window to small, constant values (32KB, 48KB).

This leads to pathological behavior when larger initial_max_data
values are used. E.g., consider 1000KB initial_max_data, 32KB window:
The sender is allowed to send up to 1000KB, the receiver will only update
MAX_DATA once less than 16KB(!!) are available and the increment is limited
to the window size, so new MAX_DATA will be 1016KB. And the sender might go
from being able to have 100's of KB in-flight to only have 32KB in-flight.
On long-fat-pipes this can collapse throughput until auto-tuning has grown
the window over many RTTs.

This PR makes the initial window configurable: use the current default or set the initial 
window to the same value as initial_max_data. Once the new setting has been verified 
to work well we can remove the config flag and always use the new value. 
After all, `initial_max_data` *is* the window / amount of data the app is willing to accept.
The new behavior can be enable either via `struct Config` or during the SSL handshake.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -605,6 +605,10 @@ pub struct Config {
     track_unknown_transport_params: Option<usize>,
 
     initial_rtt: Duration,
+
+    /// When true, uses the initial max data (for connection
+    /// and stream) as the initial flow control window.
+    use_initial_max_data_as_flow_control_win: bool,
 }
 
 // See https://quicwg.org/base-drafts/rfc9000.html#section-15
@@ -683,6 +687,8 @@ impl Config {
 
             track_unknown_transport_params: None,
             initial_rtt: DEFAULT_INITIAL_RTT,
+
+            use_initial_max_data_as_flow_control_win: false,
         })
     }
 
@@ -1243,6 +1249,20 @@ impl Config {
     /// The default is that the feature is disabled.
     pub fn enable_track_unknown_transport_parameters(&mut self, size: usize) {
         self.track_unknown_transport_params = Some(size);
+    }
+
+    /// Sets whether the initial max data value should be used as the initial
+    /// flow control window.
+    ///
+    /// If set to true, the initial flow control window for streams and the
+    /// connection itself will be set to the initial max data value for streams
+    /// and the connection respectively. If false, the window is set to the
+    /// minimum of initial max data and `DEFAULT_STREAM_WINDOW` or
+    /// `DEFAULT_CONNECTION_WINDOW`
+    ///
+    /// The default is false.
+    pub fn set_use_initial_max_data_as_flow_control_win(&mut self, v: bool) {
+        self.use_initial_max_data_as_flow_control_win = v;
     }
 }
 
@@ -2033,6 +2053,12 @@ impl<F: BufFactory> Connection<F> {
             reset_token,
         );
 
+        let initial_flow_control_window =
+            if config.use_initial_max_data_as_flow_control_win {
+                max_rx_data
+            } else {
+                cmp::min(max_rx_data / 2 * 3, DEFAULT_CONNECTION_WINDOW)
+            };
         let mut conn = Connection {
             version: config.version,
 
@@ -2091,7 +2117,7 @@ impl<F: BufFactory> Connection<F> {
             rx_data: 0,
             flow_control: flowcontrol::FlowControl::new(
                 max_rx_data,
-                cmp::min(max_rx_data / 2 * 3, DEFAULT_CONNECTION_WINDOW),
+                initial_flow_control_window,
                 config.max_connection_window,
             ),
             should_send_max_data: false,
@@ -2205,6 +2231,9 @@ impl<F: BufFactory> Connection<F> {
 
             max_amplification_factor: config.max_amplification_factor,
         };
+        conn.streams.set_use_initial_max_data_as_flow_control_win(
+            config.use_initial_max_data_as_flow_control_win,
+        );
 
         if let Some(retry_cids) = retry_cids {
             conn.local_transport_params
@@ -7998,7 +8027,7 @@ impl<F: BufFactory> Connection<F> {
     /// The connection-level flow control window will only be changed if it
     /// hasn't been auto tuned yet. For streams: only newly created streams
     /// receive the new setting.
-    pub(crate) fn enable_use_initial_max_data_as_flow_control_win(&mut self) {
+    fn enable_use_initial_max_data_as_flow_control_win(&mut self) {
         self.flow_control.set_window_if_not_tuned_yet(
             self.local_transport_params.initial_max_data,
         );
--- a/quiche/src/stream/mod.rs
+++ b/quiche/src/stream/mod.rs
@@ -200,7 +200,7 @@ pub struct StreamMap<F: BufFactory = DefaultBufFactory> {
     max_stream_window: u64,
 
     /// When `true`, the initial flow control window will be set to the
-    /// `max_rx_data`, if false, it will be set to
+    /// `max_rx_data`, if false, it will be set to `DEFAULT_STREAM_WINDOW`
     use_initial_max_data_as_flow_control_win: bool,
 }
 
@@ -722,7 +722,7 @@ impl<F: BufFactory> StreamMap<F> {
     }
 
     /// When `true`, the initial flow control window will be set to the
-    /// `max_rx_data`, if false, it will be set to
+    /// `max_rx_data`, if false, it will be set to `DEFAULT_STREAM_WINDOW`
     pub(crate) fn set_use_initial_max_data_as_flow_control_win(
         &mut self, v: bool,
     ) {
--- a/quiche/src/stream/recv_buf.rs
+++ b/quiche/src/stream/recv_buf.rs
@@ -677,7 +677,8 @@ mod tests {
 
     #[test]
     fn split_read_incremental_buf() {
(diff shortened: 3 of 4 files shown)


quiche/src/lib.rs
```diff
@@ -4534,26 +4451,25 @@ impl<F: BufFactory> Connection<F> {
...
-            if self.almost_full &&
+            if flow_control.should_update_max_data() &&
                 flow_control.max_data() < flow_control.max_data_next()
             {
                 // Autotune the connection window size.
                 flow_control.autotune_window(now, path.recovery.rtt());
+                self.should_send_max_data = true;
+            }
 
+            if self.should_send_max_data {
```
antoniovicente, line 4817: Could we have a separate PR for just the fix to MAX_DATA frame retransmission?

quiche/src/flowcontrol.rs
```diff
@@ -59,7 +59,7 @@ impl FlowControl {
         Self {
             max_data,
 
-            window,
+            window: std::cmp::min(window, max_window),
```
antoniovicente, line 62: Would be good to have some tests.

Ideally we'ld apply this min as part of config validation?

quiche/src/lib.rs
```diff
@@ -1932,7 +1929,7 @@ impl<F: BufFactory> Connection<F> {
             rx_data: 0,
             flow_control: flowcontrol::FlowControl::new(
                 max_rx_data,
-                cmp::min(max_rx_data / 2 * 3, DEFAULT_CONNECTION_WINDOW),
```
antoniovicente, line 2094: I think this relates to the auto tuning behavior and tries to minimize internal buffering in some way.  I think the authority on this topic is likely Alessandro, but unfortunately he's OOO this week.

We likely want to add some further testing for this behavior and possibly even make it configurable althought I agree the behavior is strange and unnecessarily slows down new streams.

gregor-cf: I'm wondering if it was supposed to be cmp::max.... 
But changing it to cmp::max would break any test dealing with flow_control.

gregor-cf:
rebased since #2354 is merged.

gregor-cf:
@ghedo any thoughts on this PR?

quiche/src/stream/recv_buf.rs
```diff
@@ -78,9 +76,7 @@ impl RecvBuf {
     pub fn new(max_data: u64, max_window: u64) -> RecvBuf {
         RecvBuf {
             flow_control: flowcontrol::FlowControl::new(
-                max_data,
-                cmp::min(max_data, DEFAULT_STREAM_WINDOW),
-                max_window,
+                max_data, max_data, max_window,
```
antoniovicente: Could we add a test to this file that verifies that flow_control window (the second argument to FlowControl::new) is set to the expected max_data value?

quiche/src/tests.rs
```diff
@@ -3375,17 +3380,9 @@ fn stream_shutdown_read_after_fin(
...
-    let mut dummy = buf[..len].to_vec();
-    let frames =
-        test_utils::decode_pkt(&mut pipe.client, &mut dummy[..len]).unwrap();
-    for f in frames {
-        assert!(!matches!(f, frame::Frame::StopSending { .. }));
-    }
-    assert_eq!(pipe.client_recv(&mut buf[..len]), Ok(len));
+    // Server does NOT send STOP_SENDING frame, since it has already received a
+    // FIN from the client.
+    assert_eq!(pipe.server.send(&mut buf), Err(Error::Done));
```
antoniovicente, line 3840: Change in test behavior seems odd.  Do we know what changed?

quiche/src/stream/recv_buf.rs
```diff
@@ -78,9 +76,7 @@ impl RecvBuf {
     pub fn new(max_data: u64, max_window: u64) -> RecvBuf {
         RecvBuf {
             flow_control: flowcontrol::FlowControl::new(
-                max_data,
-                cmp::min(max_data, DEFAULT_STREAM_WINDOW),
```
antoniovicente, line 82: Can we add a boolean config to enable/disable these changes?  Default: disabled for now.  We could consider flipping the default in the future.

quiche/src/tests.rs
gregor-cf: yes, previously the connection level initial window as 1.5*max_data, which ended up being 45. (initial max data = 30).
We consumed 12 bytes, so available window was max_data - consumed = 18. We send a flow control update if available < window/2 so 18 < 22 ==> send update.

Now, the initial window equals max_data, so 30. We consume 15 bytes now (that change wasn't strictly necessary). So now we have 30 - 15 < 15, which is false ==> no update is sent.

quiche/src/tests.rs
```diff
@@ -3579,8 +3576,8 @@ fn stream_shutdown_write_update_max_data(
     assert_eq!(pipe.server.rx_data, 30);
     assert_eq!(pipe.server.flow_control.consumed(), 30);
     assert_eq!(pipe.client.tx_data, 30);
-    // default window is 1.5 * initial_max_data, so 45
-    assert_eq!(pipe.client.tx_cap, 45);
```
gregor-cf, line 4023: note that the previous code here tested `tx_cap` that was wrong. Should have tested `max_tx_data` which the new code does

quiche/src/stream/recv_buf.rs
gregor-cf: the code has changed

quiche/src/stream/mod.rs
```diff
@@ -195,6 +198,10 @@ pub struct StreamMap<F: BufFactory = DefaultBufFactory> {
 
     /// The maximum size of a stream window.
     max_stream_window: u64,
+
+    /// When `true`, the initial flow control window will be set to the
+    /// `max_rx_data`, if false, it will be set to
```
antoniovicente: incomplete sentence.

quiche/src/stream/mod.rs
```diff
@@ -707,6 +720,14 @@ impl<F: BufFactory> StreamMap<F> {
     pub fn len(&self) -> usize {
         self.streams.len()
     }
+
+    /// When `true`, the initial flow control window will be set to the
+    /// `max_rx_data`, if false, it will be set to
```
antoniovicente: incomplete sentence.

quiche/src/lib.rs
```diff
@@ -2743,6 +2743,25 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// See [`Connection::enable_use_initial_max_data_as_flow_control_win()`].
+    #[cfg(feature = "boringssl-boring-crate")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "boringssl-boring-crate")))]
+    pub fn set_use_initial_max_data_as_flow_control_win_in_handshake(
+        ssl: &mut boring::ssl::SslRef,
+    ) -> Result<()> {
+        let ex_data = tls::ExData::from_ssl_ref(ssl).ok_or(Error::TlsFail)?;
+
+        ex_data.use_initial_max_data_as_flow_control_win = true;
```
antoniovicente, line 2800: I think we also need set_use_initial_max_data_as_flow_control_win so this behavior can be configured without the help of the handshake hook.  Not all implementations are expected to use the handshake hook.

Cases that enable via the Config object, should have a call to enable_use_initial_max_data_as_flow_control_win in the Connection constructor.

quiche/src/stream/recv_buf.rs
```diff
@@ -75,11 +73,11 @@ pub struct RecvBuf {
 
 impl RecvBuf {
     /// Creates a new receive buffer.
-    pub fn new(max_data: u64, max_window: u64) -> RecvBuf {
+    pub fn new(max_data: u64, initial_window: u64, max_window: u64) -> RecvBuf {
         RecvBuf {
             flow_control: flowcontrol::FlowControl::new(
                 max_data,
-                cmp::min(max_data, DEFAULT_STREAM_WINDOW),
+                initial_window,
```
antoniovicente, line 80: I would have considered plumbing the bool to decide if using `max_data` or `cmp::min(max_data, DEFAULT_STREAM_WINDOW)` but this is fine.

quiche/src/lib.rs
```diff
@@ -1244,6 +1250,12 @@ impl Config {
     pub fn enable_track_unknown_transport_parameters(&mut self, size: usize) {
         self.track_unknown_transport_params = Some(size);
     }
+
+    /// Use the value of the intial max_data / initial stream max_data setting
```
LPardue: nit: the doc cmment should match the style of others for consistency, including stating what the default is

quiche/src/lib.rs
```diff
@@ -7930,6 +7976,19 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Use the value of the intial max_data / initial stream max_data setting
+    /// as the initial flow control window for the connection and streams.
+    /// The connection-level flow control window will only be changed if it
+    /// hasn't been auto tuned yet. For streams: only newly created streams
+    /// receive the new setting.
+    pub(crate) fn enable_use_initial_max_data_as_flow_control_win(&mut self) {
```
LPardue: not clear to me why this is a pub function

PR #2358, opened 2026-02-16
lib: Ensure RESET_STREAM frames are retransmitted even after a stream is collected

LPardue:
Depending on the sequence of events, a quiche stream might be collected from the
internal map by the time a sender detects packet loss that would trigger retransmission.

This fix removes the check on the stream collection when dealing with lost packets,
which is safe because the use of insert_reset() controls only state related to reset
streams, which is consumed later in send_single() to prepare a RESET_STREAM frame.

In order to make testing retransmission issues like this easier, a
trigger_ack_based_loss() function is added to test_util.rs. This is required because
PTO does not trigger loss detection and retransmission of all frame types.

### DIFF
--- a/quiche/src/test_utils.rs
+++ b/quiche/src/test_utils.rs
@@ -554,3 +554,47 @@ pub fn stream_recv_discard(
         conn.stream_recv(stream_id, &mut buf)
     }
 }
+
+/// Triggers ACK-based loss detection for packets sent by `sender` before this
+/// call.
+///
+/// This works by sending multiple PING packets from the sender and having the
+/// receiver ACK them. Since loss detection uses a packet threshold, this
+/// function sends as many packets as needed to ensure any previously
+/// unacknowledged packets from the sender are detected as lost.
+#[cfg(test)]
+pub fn trigger_ack_based_loss(
+    sender: &mut Connection, receiver: &mut Connection,
+) {
+    let mut buf = [0; 65535];
+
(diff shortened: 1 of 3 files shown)


quiche/src/test_utils.rs
```diff
@@ -554,3 +554,39 @@ pub fn stream_recv_discard(
...
+/// receiver ACK them. Since loss detection uses a packet threshold of 3, this
+/// sends 3 packets to ensure any previously unacknowledged packets from the
+/// sender are detected as lost.
+pub fn trigger_ack_based_loss(
+    sender: &mut Connection, receiver: &mut Connection,
+) {
+    let mut buf = [0; 65535];
+
+    // Loss detection uses a packet threshold of 3, so send and deliver
+    // 3 packets to trigger loss detection for any earlier unacked packets.
```
jannes: this is referring to `INITIAL_PACKET_THRESHOLD`? should the constant be used directly here?

jannes (approved):

quiche/src/test_utils.rs
LPardue: Good idea

LPardue: Actually, each path has a packet threshold variable, so I chose to use that rather than a hard-coded value. This will help avoid unintended failures if ever someone wants to trigger loss and the count is greater than 3.

PR #2359, opened 2026-02-16
lib: Ensure STOP_SENDING frames are retransmitted only when needed

LPardue:
Previously, there was no handling for STOP_SENDING frames in lost packets. This change allows for retranmission of STOP_SENDING if the stream is still active and the read side is not done. If the read side is done, for example because of receivng a FIN or RESET_STREAM after sending a STOP_SENDING that is lost, skip retransmission since the frame can't do anything useful.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4066,6 +4066,21 @@ impl<F: BufFactory> Connection<F> {
                             .insert_reset(stream_id, error_code, final_size);
                     },
 
+                    frame::Frame::StopSending {
+                        stream_id,
+                        error_code,
+                    } =>
+                    // We only need to retransmit the STOP_SENDING frame if
+                    // the stream is still active and not FIN'd. Even if the
+                    // packet was lost, if the application has the final
+                    // size at this point there is no need to retransmit.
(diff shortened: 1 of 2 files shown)


quiche/src/lib.rs
```diff
@@ -4066,6 +4066,21 @@ impl<F: BufFactory> Connection<F> {
                             .insert_reset(stream_id, error_code, final_size);
                     },
 
+                    frame::Frame::StopSending {
+                        stream_id,
+                        error_code,
+                    } =>
+                    // We only need to retransmit the STOP_SENDING frame if
```
antoniovicente, line 4073: Unconditional retransmission would work too and wouldn't make much of a difference since this case is relatively unlikely.

Your version seems more correct but also more complicated.

antoniovicente:
Waiting for https://github.com/cloudflare/quiche/pull/2358 merge and rebase, but I think looks good.

quiche/src/lib.rs
LPardue: Discussed offline: we'll go ahead with the more complicated version for now

PR #2361, opened 2026-02-20
lib: Ensure MAX_STREAMS frames are retransmitted

LPardue:


dineshadhi (approved):

antoniovicente:
Title said MAX_DATA.  This is about MAX_STREAMS

quiche/src/lib.rs
antoniovicente, line 4491: There is a bug here.

If the frame can't be pushed to the packet, should_retransmit_max_streams is lost.

We need to save should_retransmit_max_streams in the connection so next send attempt can retransmit it.

antoniovicente, line 4466: I think this check can result in most of the retransmits with the current logic to be lost.

antoniovicente: Nevermind, I misinterpreted what Type::Short means.

PR #2364, opened 2026-02-23
lib: Ensure bidi stream limit consistency when TLS handshake callback is used

LPardue:
This changes ensures the internal bidi stream limit used by the local
endpoint is consistent with the advertised value. Without this, MAX_STREAM
updates don't behave as anticipated when the value has been customized
via TLS handshake callback.


LPardue:
I think the fine, if someone changes thresholds that's quite a serious function change and tests should be updated accordingly.

PR #2365, opened 2026-02-23
feat: emit path and cid specific invalid state error.

toidiu:
**feat: emit path and cid specific invalid state error.**
quiche currently has a general InvalidState error which is emitted for
various reasons. This PR adds InvalidPathState and InvalidCidState
errors which provide a more specific error reason and can help with
reasoning about quiche behavior.


toidiu:
@LPardue pointed that even though they shouldn't, applications might depend on the InvalidState error code. So it seems best to avoid this change, especially since this is only for observability.

PR #2366, opened 2026-02-23
fix: gracefully handle PMTU probe errors.

toidiu:
PMTU probes are not critical for QUIC functionality so it is better to handle errors from this module more gracefully.

The diff is large due to the tab shift but the only important change is:

`let active_path = self.paths.get_active_mut()?;`
to 
`if let Ok(active_path) = self.paths.get_active_mut()`

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4339,8 +4339,6 @@ impl<F: BufFactory> Connection<F> {
 
         let mut challenge_data = None;
 
-        let active_path = self.paths.get_active_mut()?;
-
         if pkt_type == Type::Short {
             // Create PMTUD probe.
             //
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -4353,64 +4345,68 @@ impl<F: BufFactory> Connection<F> {
...
             // is confirmed, to avoid interfering with the handshake
             // (e.g. due to the anti-amplification limits).
-            let should_probe_pmtu = active_path.should_send_pmtu_probe(
-                self.handshake_confirmed,
-                self.handshake_completed,
-                out_len,
-                is_closing,
-                frames.is_empty(),
-            );
+            if let Ok(active_path) = self.paths.get_active_mut() {
```
LPardue, line 4354: There's a lot of nesting/indentation - can we extract all this PMTU stuff in this block to a function? (the borrow checker might barf but would be good to see if we can start breaking send_single apart)

toidiu: Yea I tried that but there were a bunch of local variables it was using from send_single and it started getting messy.

Lemme follow thru and see what the final change looks like.

schien: can we do `let should_probe_pmtu = self.paths.get_active_mut().is_ok() && active_path.should_send_pmtu_probe(...)` to reduce the need of extra indentation?

toidiu: The pmtu change feels like the wrong place to refactor send_single.

I can impl @schien recommendation if you think the indentation is a big issue, but I personally like `if let Some(..)` pattern vs the `paths.active_mut().is_ok() + paths.get_active()` because `if let` is error proof (accidentally removing is_ok check).

---

It should happen in a few steps so its easy to reason about and review.
- Isolate: here is a change https://github.com/cloudflare/quiche/pull/2367 to isolate send APIs into a separate file send.rs.
- Breakup: there is lots of global state and &mut refs that make the borrow checker sad so this will be non-trivial and we need to be careful to not break things

LPardue: I can live with the indentation if we plan to address things later

LPardue:
I'd like @cf-rhett to take a look at this PR too

quiche/src/lib.rs
```diff
@@ -4353,64 +4345,68 @@ impl<F: BufFactory> Connection<F> {
...
-                    };
+                        match left.checked_sub(overhead) {
+                            Some(v) => left = v,
 
-                    if push_frame_to_pkt!(b, frames, frame, left) {
-                        let frame = frame::Frame::Ping {
-                            mtu_probe: Some(probe_size),
+                            None => {
+                                // We can't send more because there isn't enough
+                                // space available
```
cf-rhett: nit -- these are split weirdly now

```suggestion
                                // space available in the output buffer.
```

quiche/src/lib.rs
```diff
@@ -4353,64 +4345,68 @@ impl<F: BufFactory> Connection<F> {
...
+                                // in the output buffer.
+                                //
+                                // This usually happens when we try to send a new
+                                // packet but failed
+                                // because cwnd is almost full.
+                                //
+                                // In such case app_limited is set to false here
+                                // to
```
cf-rhett: Same here for splitting comments

quiche/src/lib.rs
cf-rhett, line 4336: When is this failing? I don't have objections but would be interested in some additional context

quiche/src/lib.rs
cf-rhett, line 4354: I would be interested in a log for when PMTUD should occur but there's no active path just to see if I see this for users

toidiu: With the amount of traffic logs are not really feasible. Metrics are the way to go but we will need to expose this error up to TQ where we record the metric.

quiche/src/lib.rs
toidiu: We can get an error if we dont have an active path. The active path is only used for PMTU functionality but that is not critical to a functional connection so we should not close the connection here. Also I moved the access inside the Short packet closure so we only get the active path when we use it.

cf-rhett: Understood -- LGTM

quiche/src/lib.rs
cf-rhett: Metrics work for me -- thank you!

cf-rhett:
Everything looks good to me. I don't have official approval power but Approved from my side

quiche/src/lib.rs
schien: +1, we can refactor this code separately.

jrouviere:
> The diff is large due to the tab shift

protip: you can disable whitespace changes in the diff settings:
https://github.com/cloudflare/quiche/pull/2366/changes?w=1

PR #2367, opened 2026-02-24
refactor: isolate connection send APIs.

toidiu:
quiche/lib.rs is huge and impl Connection makes up a large part of the code. In particular the send and recv are critical functions and also massive.

This PR isolates the send functions into their own file so we can better reason about it and start to break it up. We can do something similar with the recv functions.


LPardue:
Thanks for trying to tackle it, my concern is such a big change can be disruptive to git blame history and open PRs, since Send is so fundamental to a lot

An alternative is to move other things out of this file first. @PSUdaemon had some ideas about tackling it that way

jannes:
> Thanks for trying to tackle it, my concern is such a big change can be disruptive to git blame history and open PRs

IMHO, it's worth it. PRs can be easily refactored and there is no way around affecting Git blame history.

LPardue:
There is no immediate benefit to moving all the code out as a block, yet there is immediate pain. 

Hence, I'd rather tackle the harder task of breaking apart the large functions so we can get immediate benefit and minimal pain.

antoniovicente:
I think I'm with Lucas.  Just splitting the class implementation across multiple files does not address the problems in the current implementation.

toidiu:
Agreed. Git history is important especially for these critical fns. I am going to close this and we should pursue breaking up the fns first.

PR #2370, opened 2026-02-25
lib: Fix CUBIC epoch shift calculation during application-limited periods

estcarisimo:
The previous logic measured idle time from the last packet send, Track last_ack_time to approximate when bytes_in_flight actually transitioned to zero. In the prior approach if the cwnd was too small, when packet trains were acked, bytes in flight drop to zero triggering a recovery phase when the behavior was actually normal, rathen a connection stalled or idle

### DIFF
--- a/quiche/src/recovery/congestion/cubic.rs
+++ b/quiche/src/recovery/congestion/cubic.rs
@@ -882,6 +882,9 @@ mod tests {
         sender.lose_n_packets(1, size, None);
         let post_loss_cwnd = sender.congestion_window;
         assert_eq!(post_loss_cwnd, (initial_cwnd as f64 * BETA_CUBIC) as usize);
+        let initial_recovery_start_time =
+            sender.congestion_recovery_start_time.unwrap();
+        assert_eq!(initial_recovery_start_time, sender.time);
 
         // ACK remaining in-flight packets to exit recovery.
         sender.ack_n_packets(sender.initial_congestion_window_packets - 1, size);
@@ -894,8 +897,6 @@ mod tests {
         let packets_per_burst = cmp::max(1, sender.congestion_window / size);
 
         for _ in 0..20 {
-            sender.advance_time(rtt);
-
             for _ in 0..packets_per_burst {
                 sender.send_packet(size);
             }
@@ -912,6 +913,13 @@ mod tests {
             sender.congestion_window,
             post_loss_cwnd,
         );
+        // Recovery start hasn't changed.
+        assert_eq!(
+            sender.congestion_recovery_start_time.unwrap() -
+                initial_recovery_start_time,
+            Duration::ZERO,
+            "epoch should not have shifted without idle gap",
+        );
     }
 
     #[test]
@@ -920,6 +928,7 @@ mod tests {
         // shift the epoch so that cwnd growth resumes from where it
         // left off rather than jumping.
         let mut sender = test_sender();
+        let test_start = sender.time;
         let size = sender.max_datagram_size;
         let rtt = Duration::from_millis(100);
 
@@ -930,6 +939,14 @@ mod tests {
         sender.update_rtt(rtt);
         sender.advance_time(rtt);
         sender.lose_n_packets(1, size, None);
+
+        let initial_recovery_start =
+            sender.congestion_recovery_start_time.unwrap();
+        assert_eq!(
+            initial_recovery_start,
+            test_start + rtt,
+            "Recovery start is set",
+        );
 
         let post_loss_cwnd = sender.congestion_window;
 
@@ -961,6 +978,16 @@ mod tests {
             recovery_start >
                 sender.cubic_state.last_sent_time.unwrap() - idle_duration,
             "epoch should have been shifted forward by idle period",
+        );
+        assert_eq!(
+            recovery_start - sender.cubic_state.last_sent_time.unwrap(),
+            Duration::ZERO,
+            "epoch should have been shifted forward by idle period",
+        );
+        assert_eq!(
+            recovery_start - initial_recovery_start,
+            idle_duration,
+            "Recovery start should have moved forward by the idle duration",
         );
     }
 


quiche/src/recovery/congestion/cubic.rs
```diff
@@ -179,6 +190,8 @@ fn on_packets_acked(
     r: &mut Congestion, bytes_in_flight: usize, packets: &mut Vec<Acked>,
     now: Instant, rtt_stats: &RttStats,
 ) {
+    r.cubic_state.last_ack_time = Some(now);
```
antoniovicente, line 200: Could you try to add some tests for the behavior changes in this PR?

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -150,14 +156,19 @@ fn on_init(_r: &mut Congestion) {}
...
     let cubic = &mut r.cubic_state;
 
     if let Some(last_sent_time) = cubic.last_sent_time {
         if bytes_in_flight == 0 {
-            let delta = now - last_sent_time;
+            // Measure idle from when bytes_in_flight hits 0 (approximate last ACK time), not from last send.
+            // When cwnd is small, bif transiently hits 0 between ACK
+            // processing and sending. Using last_sent_time would inflate
+            // the delta by a full RTT, incorrectly shifting the epoch.
+            let idle_start = cubic.last_ack_time.unwrap_or(last_sent_time);
```
antoniovicente: I think we may want a `max(last_sent_time, last_ack_time)`

quiche/src/recovery/congestion/cubic.rs
antoniovicente, line 193: I think there may be some additional issues in cases where sent_bytes == 0

We probably don't want to update `congestion_recovery_start_time` and `cubic.last_sent_time` in cases when `sent_bytes == 0`

quiche/src/recovery/congestion/cubic.rs
antoniovicente, line 200: I think we should think about the meaning of last_ack_time and what we're trying to capture.

Should we be updating this in cases where `bytes_in_flight` was already 0?  Or in cases where the ACK received was redundant and did not decrease `bytes_in_flight`?

quiche/src/recovery/congestion/cubic.rs
antoniovicente: I think `sent_bytes == 0` when sending ACK only packets.

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -151,13 +157,30 @@ fn on_init(_r: &mut Congestion) {}
...
+    if sent_bytes == 0 {
+        reno::on_packet_sent(r, sent_bytes, bytes_in_flight, now);
+        return;
+    }
+
     // See https://github.com/torvalds/linux/commit/30927520dbae297182990bb21d08762bcc35ce1d
     // First transmit when no packets in flight
     let cubic = &mut r.cubic_state;
 
     if let Some(last_sent_time) = cubic.last_sent_time {
```
antoniovicente: I don't think it makes sense to consider cubic.last_sent_time first.
I think what we're looking for is closer to this (plus your code comments):

```
if bytes_in_flight == 0 {
  if let Some(recovery_start_time) = r.congestion_recovery_start_time {
    let idle_start = cmp::max(cubic.last_ack_time, cubic.last_sent_time);
    if let Some(idle_start) = idle_start && idle_start < now {
      let delta = now - idle_start;
      r.congestion_recovery_start_time = Some(recovery_start_time + delta);
    }
  }
}
```

quiche/src/recovery/congestion/cubic.rs
antoniovicente: Actually, I think that at least for the purposes of this PR I think that updating last_ack_time unconditionally is fine.  Resolving this comment.

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +863,155 @@ mod tests {
             prev_cwnd as f64 * (1.0 + BETA_CUBIC) / 2.0
         );
     }
+
+    #[test]
+    fn cubic_perpetual_recovery_trap() {
+        // Reproduces the bug where cwnd stays pinned at minimum after a
+        // loss event because the idle-time epoch shift pushes
+        // congestion_recovery_start_time into the future on every
+        // send→ACK→send cycle when bif transiently hits 0.
```
antoniovicente: `→` caught my attention since I'm not used to seeing non-ascii characters in source code.  Don't know if the team has an opinion on this.  I at least checked that emacs does seem happy to render this unicode character.

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +863,155 @@ mod tests {
...
+            sender.congestion_window,
+            post_loss_cwnd,
+        );
+
+        // Verify recovery_start_time was shifted forward (i.e., the
+        // idle shift logic ran).
+        if let Some(recovery_start) = sender.congestion_recovery_start_time {
+            assert!(
+                recovery_start >
+                    sender.cubic_state.last_sent_time.unwrap() - idle_duration,
```
antoniovicente: Is it possible to make this an equality check?

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +863,155 @@ mod tests {
...
+        assert!(
+            sender.congestion_window >= post_loss_cwnd,
+            "cwnd should not have shrunk after idle: {} < {}",
+            sender.congestion_window,
+            post_loss_cwnd,
+        );
+
+        // Verify recovery_start_time was shifted forward (i.e., the
+        // idle shift logic ran).
+        if let Some(recovery_start) = sender.congestion_recovery_start_time {
```
antoniovicente: I think we have an expectation of `sender.congestion_recovery_start_time != None`

I would recommend an `unwrap()` instead of `if let` or `assert_eq!(sender.congestion_recovery_start_time, Some(value))`

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +863,155 @@ mod tests {
...
+        for _ in 0..packets_per_burst {
+            sender.send_packet(size);
+        }
+        sender.advance_time(rtt);
+        sender.ack_n_packets(packets_per_burst, size);
+
+        // After one RTT of sending post-idle, cwnd should grow but
+        // not have jumped to an unreasonably large value.
+        assert!(
+            sender.congestion_window >= post_loss_cwnd,
```
antoniovicente: Possible to make this an assert_eq?

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -151,13 +157,30 @@ fn on_init(_r: &mut Congestion) {}
 fn on_packet_sent(
     r: &mut Congestion, sent_bytes: usize, bytes_in_flight: usize, now: Instant,
 ) {
+    // Don't adjust epoch or track send time for non-data packets
+    // (e.g. ACKs). These have in_flight=true but size=0.
+    // Skip all the following logic for these packets.
+    if sent_bytes == 0 {
+        reno::on_packet_sent(r, sent_bytes, bytes_in_flight, now);
+        return;
+    }
```
antoniovicente, line 167: You could try simplify the code by moving the call to `reno::on_packet_sent` to the top of this method

Something like:

```
--- a/quiche/src/recovery/congestion/cubic.rs
+++ b/quiche/src/recovery/congestion/cubic.rs
@@ -157,11 +157,12 @@ fn on_init(_r: &mut Congestion) {}
 fn on_packet_sent(
     r: &mut Congestion, sent_bytes: usize, bytes_in_flight: usize, now: Instant,
 ) {
+    reno::on_packet_sent(r, sent_bytes, bytes_in_flight, now);
+
     // Don't adjust epoch or track send time for non-data packets
     // (e.g. ACKs). These have in_flight=true but size=0.
     // Skip all the following logic for these packets.
     if sent_bytes == 0 {
-        reno::on_packet_sent(r, sent_bytes, bytes_in_flight, now);
         return;
     }
 
@@ -194,8 +195,6 @@ fn on_packet_sent(
     }
 
     cubic.last_sent_time = Some(now);
-
-    reno::on_packet_sent(r, sent_bytes, bytes_in_flight, now);
 }

```

quiche/src/recovery/congestion/cubic.rs
estcarisimo: I am not sure if we can established a strict boundary. We just want to check whether the perpetual loop is broken, I guess

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +858,150 @@ mod tests {
...
+        // After one RTT of sending post-idle, cwnd should not have
+        // shrunk.
+        assert_eq!(sender.congestion_window, post_loss_cwnd);
+
+        // Verify recovery_start_time was shifted forward (i.e., the
+        // idle shift logic ran).
+        let recovery_start = sender.congestion_recovery_start_time.unwrap();
+        assert!(
+            recovery_start >
+                sender.cubic_state.last_sent_time.unwrap() - idle_duration,
```
antoniovicente: Seems like in this case
`recovery_start == sender.cubic_state.last_sent_time.unwrap()`

Is that expected?

Should that assertion also be added to the test?

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +858,150 @@ mod tests {
...
+        }
+
+        // cwnd must have grown beyond the post-loss value.
+        assert!(
+            sender.congestion_window > post_loss_cwnd,
+            "cwnd stuck at {} (post-loss {}): perpetual recovery trap",
+            sender.congestion_window,
+            post_loss_cwnd,
+        );
+    }
```
antoniovicente, line 923: Consider the following changes to tighten this test further.

I think we want to exercise the case where time between ACK and next send is 0.  We could also exercise the case of an extra delay between ACK and send.

```
--- a/quiche/src/recovery/congestion/cubic.rs
+++ b/quiche/src/recovery/congestion/cubic.rs
@@ -882,6 +882,8 @@ mod tests {
         sender.lose_n_packets(1, size, None);
         let post_loss_cwnd = sender.congestion_window;
         assert_eq!(post_loss_cwnd, (initial_cwnd as f64 * BETA_CUBIC) as usize);
+        let initial_recovery_start_time = sender.congestion_recovery_start_time.unwrap();
+        assert_eq!(initial_recovery_start_time, sender.time);
 
         // ACK remaining in-flight packets to exit recovery.
         sender.ack_n_packets(sender.initial_congestion_window_packets - 1, size);
@@ -894,8 +896,6 @@ mod tests {
         let packets_per_burst = cmp::max(1, sender.congestion_window / size);
 
         for _ in 0..20 {
-            sender.advance_time(rtt);
-
             for _ in 0..packets_per_burst {
                 sender.send_packet(size);
             }
@@ -912,6 +912,12 @@ mod tests {
             sender.congestion_window,
             post_loss_cwnd,
         );
+        // Recovery start hasn't changed.
+        assert_eq!(
+            sender.congestion_recovery_start_time.unwrap() - initial_recovery_start_time,
+            Duration::ZERO,
+            "epoch should have been shifted forward by idle period",
+        );
     }

```

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -838,4 +858,150 @@ mod tests {
...
+
+        // Verify recovery_start_time was shifted forward (i.e., the
+        // idle shift logic ran).
+        let recovery_start = sender.congestion_recovery_start_time.unwrap();
+        assert!(
+            recovery_start >
+                sender.cubic_state.last_sent_time.unwrap() - idle_duration,
+            "epoch should have been shifted forward by idle period",
+        );
+    }
```
antoniovicente, line 992: Consider the following extra assertions for this test:

```
         // shift the epoch so that cwnd growth resumes from where it
         // left off rather than jumping.
         let mut sender = test_sender();
+        let test_start = sender.time;
         let size = sender.max_datagram_size;
         let rtt = Duration::from_millis(100);
 
@@ -931,6 +938,13 @@ mod tests {
         sender.advance_time(rtt);
         sender.lose_n_packets(1, size, None);
 
+        let initial_recovery_start = sender.congestion_recovery_start_time.unwrap();
+        assert_eq!(
+            initial_recovery_start,
+            test_start + rtt,
+            "Recovery start is set",
+        );
+
         let post_loss_cwnd = sender.congestion_window;
 
         // ACK remaining to exit recovery.
@@ -962,6 +976,17 @@ mod tests {
                 sender.cubic_state.last_sent_time.unwrap() - idle_duration,
             "epoch should have been shifted forward by idle period",
         );
+        assert_eq!(
+            recovery_start -
+                (sender.cubic_state.last_sent_time.unwrap()),
+            Duration::ZERO,
+            "epoch should have been shifted forward by idle period",
+        );
+        assert_eq!(
+            recovery_start - initial_recovery_start,
+            idle_duration,
+            "Recovery start should have moved forward by the idle duration",
+        );
     }
```

quiche/src/recovery/congestion/cubic.rs
```diff
@@ -151,18 +157,32 @@ fn on_init(_r: &mut Congestion) {}
...
-            // Shift epoch start to keep cwnd growth to cubic curve.
-            if let Some(recovery_start_time) = r.congestion_recovery_start_time {
-                if delta.as_nanos() > 0 {
+    if bytes_in_flight == 0 {
+        if let Some(recovery_start_time) = r.congestion_recovery_start_time {
+            // Measure idle from the most recent activity: either the
+            // last ACK (approximating when bif hit 0) or the last data
+            // send, whichever is later. Using last_sent_time alone
+            // would inflate the delta by a full RTT when cwnd is small
+            // and bif transiently hits 0 between ACK and send.
```
evanrittenhouse, line 180: nit: what's bif?

estcarisimo: bytes in flight :)

PR #2372, opened 2026-02-27
tokio-quiche: set sent error codes in audit stats

evanrittenhouse:
The new test fails without the changes to IoWorker. We add a QuicAuditStats receiver so that future tests can access them if need be.


jannes:
> lgtm

same

jannes:
when we extend the test util functions to start server further, we should make a new type to return instead of growing the tuple.

PR #2374, opened 2026-02-28
tokio-quiche: keep early-data connections in handshake stage until established

dknecht:
### Motivation
- Prevent a resource-exhaustion/DoS scenario where 0-RTT (early-data) causes the connection to leave the handshake stage and thereby bypass handshake timeout enforcement.

### Description
- Change `Handshake::on_flush` in `tokio-quiche/src/quic/io/connection_stage.rs` to transition to the `RunningApplication` stage only when `qconn.is_established()` is true and stop treating `qconn.is_in_early_data()` as sufficient for the transition, preserving handshake-timeout checks during early data.

### Testing
- Ran `cargo test -p tokio-quiche --no-run`, which completed successfully and produced the test binaries (compile-only run succeeded).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a1fa16e35c8323a54dd14cb4fa387c)

### DIFF
--- a/tokio-quiche/src/quic/io/connection_stage.rs
+++ b/tokio-quiche/src/quic/io/connection_stage.rs
@@ -122,9 +122,9 @@ impl ConnectionStage for Handshake {
         &mut self, qconn: &mut QuicheConnection,
         _ctx: &mut ConnectionStageContext<A>,
     ) -> ControlFlow<QuicResult<()>> {
-        // Transition to RunningApplication if we have 1-RTT keys (handshake is
-        // complete) or if we have 0-RTT keys (in early data).
-        if qconn.is_established() || qconn.is_in_early_data() {
+        // Transition to RunningApplication only after the QUIC handshake is
+        // complete (1-RTT keys are available).
+        if qconn.is_established() {
             ControlFlow::Break(Ok(()))
         } else {
             ControlFlow::Continue(())


toidiu:
**TQ: Handshake timeout**
> preserving handshake-timeout checks during early data

I think this issue is referring to the [Handshake timeout](https://github.com/cloudflare/quiche/blob/d6644f193a70099719e66957b90bc7a5835cd1af/tokio-quiche/src/quic/connection/mod.rs#L607-L608) which is cleared via [check_handshake_timeout_expired](https://github.com/cloudflare/quiche/blob/d6644f193a70099719e66957b90bc7a5835cd1af/tokio-quiche/src/quic/io/connection_stage.rs#L104)

**TQ: Post Accept timeout**
TQ also has [post_accept_timeout](https://github.com/cloudflare/quiche/blob/d6644f193a70099719e66957b90bc7a5835cd1af/tokio-quiche/src/http3/settings.rs#L57-L60) is used as the timeout after establishing a connection and receiving the first headers: [code](https://github.com/cloudflare/quiche/blob/d6644f193a70099719e66957b90bc7a5835cd1af/tokio-quiche/src/http3/driver/server.rs#L257-L278).

**nginx: client_header_timeout**
This [comment](https://trac.nginx.org/nginx/ticket/2235#comment:1) suggests that nginx uses [client_header_timeout](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_header_timeout) as a proxy for the handshake timeout so that would be cleared when we process the early data request. This combines TQ handshake and post_accept timeout behavior.

---

**Analysis**
Clearing the handshake timeout seems reasonable since at that point the Server can complete the handshake without client involvement.

LPardue:
The proposed PR breaks the 0-RTT feature, as indicated by the failing CI tests. A different approach is required, so closing.

PR #2404, opened 2026-03-16
Fix cargo doc warnings

wutchzone:
When running `cargo doc --workspace --all-features --no-deps`  it prints out plenty of warnings about invalid hyperlinks. This series fixes them all.

Maybe we should add documentation generation to the pipeline so we do not have broken links in the documentation?

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1745,12 +1745,13 @@ pub fn connect_with_buffer_factory<F: BufFactory>(
 
 /// Creates a new client-side connection, with a custom buffer generation
 /// method using the given dcid initially.
-/// Be aware the RFC places requirements for unpredictability and length
-/// on the client DCID field.
-/// [`RFC9000`]:  https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
+/// Be aware the [RFC9000, Section 7.2.3] places requirements for
+/// unpredictability and length on the client DCID field.
 ///
 /// The buffers generated can be anything that can be drereferenced as a byte
 /// slice. See [`connect`] and [`BufFactory`] for more info.
+///
+/// [RFC9000, Section 7.2.3]: https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
 #[cfg(feature = "custom-client-dcid")]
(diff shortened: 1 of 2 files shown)


antoniovicente:
Adding docs checks to CI does seem like a good call.

antoniovicente (approved):

quiche/src/lib.rs
```diff
@@ -1745,12 +1745,13 @@ pub fn connect_with_buffer_factory<F: BufFactory>(
...
 /// method using the given dcid initially.
-/// Be aware the RFC places requirements for unpredictability and length
+/// Be aware the [`RFC9000`] places requirements for unpredictability and length
 /// on the client DCID field.
-/// [`RFC9000`]:  https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
 ///
 /// The buffers generated can be anything that can be drereferenced as a byte
 /// slice. See [`connect`] and [`BufFactory`] for more info.
+///
+/// [`RFC9000`]:  https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3
```
```suggestion
```

quiche/src/lib.rs
```diff
@@ -1745,12 +1745,13 @@ pub fn connect_with_buffer_factory<F: BufFactory>(
 
 /// Creates a new client-side connection, with a custom buffer generation
 /// method using the given dcid initially.
-/// Be aware the RFC places requirements for unpredictability and length
+/// Be aware the [`RFC9000`] places requirements for unpredictability and length
 /// on the client DCID field.
```
LPardue: This results in odd grammar, better to replace the whole block with

```
/// Creates a new client-side connection, with a custom buffer generation method
/// using the given dcid initially. Be aware that [RFC 9000, Section
/// 7.2.3](https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-3) places
/// requirements for unpredictability and length on the client DCID field.
```

quiche/src/lib.rs
wutchzone: I agree with mentioning the section part, but why do you think inlining the hyperlink into the text, compared to this fragment specifier method, is better? IMO, when inlined, it is much harder to read in an IDE.

Also, the `connect_with_dcid` has been commented in a similar fashion.

PR #2405, opened 2026-03-16
qlog-dancer: minimize work on pending plot

LPardue:


qlog-dancer/src/plots/pending.rs
```diff
@@ -81,7 +81,10 @@ pub fn plot_pending(
 
     let mut series = vec![];
 
-    while x < x_max {
+    let total_reqs = ds.http_requests.values().len();
+
+    // Don't need to do any work if there are no requests.
+    while x < x_max && total_reqs > 0 {
```
antoniovicente, line 87: How common is it to have no requests?

antoniovicente (approved):

qlog-dancer/src/plots/pending.rs
LPardue: So it depends a lot on the context of the qlog being parsed, some will not log HTTP events, so even though there is L4 traffic, there's nothing to hook on to for this plot. 

I can't recall exactly why I added this check, but it had snuck into the qlog refactor where it really didn't belong. No urgency on this PR.

estcarisimo (approved):

PR #2409, opened 2026-03-20
Move on_timeout handling to the main part of the work loop

antoniovicente:
This change simplifies the event loop and removes the need to bias the select! to avoid starvation by serializing all the event loop actions and clarifying the start and end of the work loop iteration.

Split from https://github.com/cloudflare/quiche/pull/2328

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -225,6 +225,16 @@ where
         loop {
             let now = Instant::now();
 
+            if let Some(deadline) = current_deadline {
+                if deadline <= now {
+                    qconn.on_timeout();
+
+                    self.write_state.next_release_time = None;
+                    current_deadline = None;
+                    sleep.as_mut().reset((now + DEFAULT_SLEEP).into());
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -313,19 +323,8 @@ where
 
             select! {
                 biased;
```
evanrittenhouse: Now that we don't rely on the sleep arm to modify quiche's connection state, should we remove this bias? It should no longer matter which arm resolves the `select!`

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -225,6 +225,16 @@ where
         loop {
             let now = Instant::now();
 
+            if let Some(deadline) = current_deadline {
```
evanrittenhouse: nit:

```
if let Some(deadline) = current_deadline.filter(|d| d <= now)
```

could reduce the extra block. We could also use `take_if` but it would make `current_deadline = None` more implicit, so I'm fine leaving it as-is.

Also could be worth a comment explaining why this is helpful

PR #2410, opened 2026-03-20
Add EventLoopIteration to Connection methods: send_on_path, recv and on_timeout

antoniovicente:
This refactor is in preparation for https://github.com/cloudflare/quiche/pull/2328

### DIFF
--- a/quiche/src/ffi.rs
+++ b/quiche/src/ffi.rs
@@ -831,8 +831,34 @@ impl From<&RecvInfo<'_>> for crate::RecvInfo {
 }
 
 #[no_mangle]
+pub extern "C" fn quiche_event_loop_iteration_new() -> *mut EventLoopIteration {
+    Box::into_raw(Box::new(EventLoopIteration::new()))
+}
+
+#[no_mangle]
+pub extern "C" fn quiche_event_loop_iteration_free(
+    iteration: *mut EventLoopIteration,
+) {
+    drop(unsafe { Box::from_raw(iteration) });
+}
+
+#[no_mangle]
 pub extern "C" fn quiche_conn_recv(
     conn: &mut Connection, buf: *mut u8, buf_len: size_t, info: &RecvInfo,
+) -> ssize_t {
+    quiche_conn_recv_with_iteration(
+        conn,
+        buf,
+        buf_len,
+        info,
+        &EventLoopIteration::new(),
+    )
+}
+
+#[no_mangle]
+pub extern "C" fn quiche_conn_recv_with_iteration(
+    conn: &mut Connection, buf: *mut u8, buf_len: size_t, info: &RecvInfo,
+    iteration: &EventLoopIteration,
 ) -> ssize_t {
     if buf_len > <ssize_t>::MAX as usize {
         panic!("The provided buffer is too large");
@@ -840,7 +866,7 @@ pub extern "C" fn quiche_conn_recv(
 
     let buf = unsafe { slice::from_raw_parts_mut(buf, buf_len) };
 
-    match conn.recv(buf, info.into()) {
+    match conn.recv(iteration, buf, info.into()) {
         Ok(v) => v as ssize_t,
 
         Err(e) => e.to_c(),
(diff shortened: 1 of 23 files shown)


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -737,10 +747,11 @@ where
     /// once one of the `gather_data_from_quiche_conn()` calls writes to the
     /// send buffer, we signal to the caller which has to take care of flushing
     async fn wait_for_quiche(
-        &mut self, qconn: &mut QuicheConnection, buffer: &mut [u8],
+        &mut self, iteration: &EventLoopIteration, qconn: &mut QuicheConnection,
+        buffer: &mut [u8],
     ) -> QuicResult<()> {
         std::future::poll_fn(|_| {
-            match self.gather_data_from_quiche_conn(qconn, buffer) {
+            match self.gather_data_from_quiche_conn(iteration, qconn, buffer) {
```
antoniovicente: TODO: We may need a new iteration on every poll so we can have a fresh now

quiche/src/ffi.rs
```diff
@@ -830,17 +830,43 @@ impl From<&RecvInfo<'_>> for crate::RecvInfo {
     }
 }
 
+#[no_mangle]
```
LPardue, line 833: general comment: this seems like a sizeable amount of code to add to the FFI and it doesn't update the C examples to call it. I'd suggest punting on C entirely for now

quiche/src/lib.rs
```diff
@@ -1936,6 +1937,37 @@ struct QlogInfo {
...
+/// [`on_timeout`]. This allows the methods to share a single timestamp for
+/// the iteration rather than each calling [`Instant::now`] independently.
+///
+/// [`recv`]: Connection::recv
+/// [`send`]: Connection::send
+/// [`send_on_path`]: Connection::send_on_path
+/// [`on_timeout`]: Connection::on_timeout
+pub struct EventLoopIteration {
+    /// The time at which this event loop iteration started.
+    now: Instant,
```
LPardue: nit: if the field represents the start, just call it that

```suggestion
    start: Instant,
```

quiche/src/lib.rs
```diff
@@ -1936,6 +1937,37 @@ struct QlogInfo {
...
+impl EventLoopIteration {
+    /// Creates a new [`EventLoopIteration`], capturing the current time.
+    pub fn new() -> Self {
+        Self {
+            now: Instant::now(),
+        }
+    }
+
+    /// Returns the time at which this event loop iteration started.
+    pub fn now(&self) -> Instant {
```
LPardue: same here, it strikes me as odd that now would not return a now, that seems to conflate things

```suggestion
    pub fn start(&self) -> Instant {
```

quiche/src/ffi.rs
antoniovicente: I think that the right thing would be to update examples to use the new API and fully delete the old API in a future PR.

quiche/src/lib.rs
```diff
@@ -2785,7 +2817,9 @@ impl<F: BufFactory> Connection<F> {
     /// }
     /// # Ok::<(), quiche::Error>(())
     /// ```
-    pub fn recv(&mut self, buf: &mut [u8], info: RecvInfo) -> Result<usize> {
+    pub fn recv(
+        &mut self, iteration: &EventLoopIteration, buf: &mut [u8], info: RecvInfo,
```
LPardue, line 2821: I think the changes to extant methods are too disruptive TBH, as it pushes work on all downstreams and risks them making the wrong decision. If its safe to call them with a new EventLoopIteration each time, that should just be the default internal behaviour. 

IOW like with the FFI proposal, I'd suggest adding variants that have a suffix of with_iteration, _with_ev_loop, or something. Then the old methods can just construct a EventLoopIteration to pass onto the new ones

antoniovicente: It is not correct to pass a new EventLoopIteration each time.  Currently it mostly works, but https://github.com/cloudflare/quiche/pull/2328 require this information and it does need to be correct.

Creating a new EventLoopIteration each time will result in BBR not working correctly; app-limited will be computed incorrectly some of the time.

quiche/src/lib.rs
antoniovicente: Excellent suggestion.  ty.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -223,7 +224,18 @@ where
         tokio::pin!(sleep);
 
         loop {
-            let now = Instant::now();
+            let iteration = EventLoopIteration::new();
```
LPardue, line 227: I'm struggling to follow this change a bit since I'm less familiar with TQ guts. It would seem that if ever we broke the loop at https://github.com/cloudflare/quiche/pull/2410/changes#diff-77397d0e30b86551f31a2fc4fc5e2adc5fb62fd3091c5e8a35b950bd43fc7c1cR345, then the quiche timeout wouldn't be called. Is that ok?

quiche/src/lib.rs
LPardue: Right but requiring all apps to make a change for a congestion control algo they might not be interested in seems pretty extreme. 

As this PR highlights, we have dozens of tests that do not care about the app-limited behaviour one way or another

antoniovicente: If the solution is to have different API requirements for BBR and not, I think we could do that.  But I'ld like us to make a decision.

We could do something to require that the hook introduced in https://github.com/cloudflare/quiche/pull/2328 be called if the code wants to use BBR, or we could try forcing a wider API change to force users of quiche to reconsider the correctness of their integrations.

Ideally we'ld have a simpler API to the event loop which provides hooks to consume read packets, emit packets and handling of timeouts using a single call, instead of the current API which forces the app to do timeout handling, recv and send in the right order.

Issue #2411, opened 2026-03-20
CryptoFail due to Invalid Nonce, caused by Quiche AEAD algorithm selection


somethingcorrosive:
As part of the contributing guidelines I am creating this issue to start the conversation before pushing a PR with the fix, which I will include the fix in this issue as well.

I also looked for other conversations, I found this one from 2020, with no response: https://github.com/cloudflare/quiche/issues/524

In our QUIC stack, connections would complete handshake but then fail on early 1-RTT sends because quiche’s BoringSSL backend selected TLS13-specific AES-GCM AEAD entrypoints (EVP_aead_aes_*_gcm_tls13), which on the resolved boring backend rejected that runtime nonce/counter usage with INVALID_NONCE; this surfaced from quiche as CryptoFail, and in tokio-quiche appeared as send_on_path failure followed by local transport close code 1, causing reconnect churn. Switching quiche’s AES-GCM selection to the generic AEAD entrypoints (EVP_aead_aes_*_gcm) resolved the nonce failure and restored stable connections.

Findings:

- In tokio-quiche, that failure is observed in the worker send path at qconn.send_on_path(...), then it closes with transport InternalError (1) if no local error is already set , see source here:

https://github.com/cloudflare/quiche/blob/b30f9e76c32332aa35377dcb00f556626d47a841/tokio-quiche/src/quic/io/worker.rs#L562-L614

- The send path calls into quiche packet encryption (encrypt_pkt -> seal_with_u64_counter): 

https://github.com/cloudflare/quiche/blob/b30f9e76c32332aa35377dcb00f556626d47a841/quiche/src/packet.rs#L711-L725

- In quiche’s BoringSSL backend, AEAD algorithm selection uses TLS13-specific AES-GCM entrypoints:

https://github.com/cloudflare/quiche/blob/b30f9e76c32332aa35377dcb00f556626d47a841/quiche/src/crypto/boringssl.rs#L23-L31

- The actual crypto failure is emitted from the seal call (EVP_AEAD_CTX_seal_scatter) which maps to CryptoFail on non-success: 
 https://github.com/cloudflare/quiche/blob/b30f9e76c32332aa35377dcb00f556626d47a841/quiche/src/crypto/boringssl.rs#L96-L140

I have tested this patch locally and it works, however, it could have some potential impacts on decisions later. 

The fix is as follows:

[...]

Issue #2108, opened 2025-07-16
Compatibility issues with quic-go server


nathanejohnson:
I've been seeing issues connecting to a quic-go server from a quiche based client.  This works fine with 0.22.0 based quiche, but no newer published versions will work properly.  This commit seems to be the culprit:

https://github.com/cloudflare/quiche/commit/b17904e522777746b2a6b12667d85283e1df368c

Below are some trace logs.  Initial connection works, handshake succeeds, but then sometime later I get a CryptoFail error from quiche::send_on_path method.

```
[TRACE mio::poll] registering event source with poller: token=Token(0), interests=READABLE
[TRACE quiche::tls] ccc6ccf2d1124ff11c5bebe6075e027a395e32a1 write message lvl=Initial len=261
[TRACE quiche] ccc6ccf2d1124ff11c5bebe6075e027a395e32a1 tx pkt Initial version=1 dcid=286c9160ef2b9748421b2723d2df60fd scid=ccc6ccf2d1124ff11c5bebe6075e027a395e32a1 len=265 pn=0 
[TRACE quiche] ccc6ccf2d1124ff11c5bebe6075e027a395e32a1 tx frm CRYPTO off=0 len=261
[...]

PR #2412, opened 2026-03-21
Addressing nonce mismatch and cryptofail with quiche/boring

somethingcorrosive:
Fix as mentioned in the Github issue.

https://github.com/cloudflare/quiche/issues/2411


ghedo:
Thanks for the report and patch. I am closing this because switching from `EVP_aead_aes_*_gcm_tls13()` to the generic `EVP_aead_aes_*_gcm()` functions is not an acceptable fix. BoringSSL treats the generic external-IV AES-GCM service as not FIPS-approved, while the TLS-specific variants are approved. The generic functions also bypass the nonce-order validation that detected the problem.

I audited the key lifecycle and did not find an uncovered production path. Commit 7d686e13a primes keys created during key updates, and 987dc6691 primes Initial keys recreated after Retry or Version Negotiation. Fresh Initial keys and Retry integrity keys begin by sealing packet number zero.

The symptoms closely match #2108. That issue was fixed by 33cf86f2a, which makes packet sealing mutable so the optimizer cannot treat BoringSSL context state as unchanged. A user on #2411 also confirmed that updating past that commit resolved their CryptoFail. Current master additionally uses BoringSSL-owned `EVP_AEAD_CTX` allocation through cdfb65eda, avoiding copied context layout and move-related risks.

This branch contains 33cf86f2a through its base but has kept the generic-AEAD change applied across rebases, so it does not isolate whether the existing fix resolves the reporters environment. If this remains reproducible on current master with this patch removed, please open a new issue with a minimal reproducer, exact quiche and BoringSSL revisions, target platform, build profile and panic strategy, and the BoringSSL error and packet-number trace.

PR #2416, opened 2026-03-25
Add connection-level timing metrics and key update tracking to Stats

devin-cog:
## Summary

Add new fields to the public `Stats` struct for production observability:

- **`handshake_duration: Option<Duration>`** — time from connection creation to handshake completion, `None` if handshake hasn't completed
- **`key_update_count: u64`** — number of 1-RTT key updates
- **`peer_max_idle_timeout: Option<Duration>`** — the peer's max idle timeout from transport params
- **`connection_duration: Duration`** — wall-clock time since connection creation

### Implementation details

- Track `created_at: Instant` and `handshake_completed_at: Option<Instant>` on `Connection`, initialized in `with_tls()` and set when handshake completes
- Track `key_update_count: u64` on `Connection`, incremented when a peer-initiated key update is verified
- Populate new `Stats` fields in `stats()` method
- Update `Debug` impl for `Stats` to include new fields
- Expose via FFI: new fields in the C `Stats` struct plus standalone query functions (`quiche_conn_stats_handshake_duration_ms`, `quiche_conn_stats_key_update_count`, `quiche_conn_stats_peer_max_idle_timeout_ms`, `quiche_conn_stats_connection_duration_ms`)
- Add unit tests verifying pre/post handshake behavior

### Verification

- `cargo build` succeeds
- `cargo +nightly fmt -- --check` clean
- `cargo clippy --features=boringssl-vendored --workspace -- -D warnings` clean
- All 963 quiche tests pass (including 4 new tests)

## Review & Testing Checklist for Human
- [ ] Verify `handshake_completed_at` is set at the right place — it's set right after `self.handshake.is_completed()` returns true
- [ ] Verify `key_update_count` increment location — placed after key update is verified (after successful decryption with new keys), per RFC 9001
- [ ] Check FFI return type conventions — using `i64` with `-1` sentinel for optional durations; confirm this matches downstream C consumer expectations
- [ ] Run the full test suite in your CI environment to verify no regressions

### Notes
- The only pre-existing test failure is `integration_tests::test_so_mark_receive_data` in tokio-quiche, which requires special socket permissions and is unrelated to this change
- `peer_max_idle_timeout` returns `None` when the peer's transport params haven't been parsed yet or when `max_idle_timeout` is 0 (which per RFC 9000 means no timeout)

### DIFF
--- a/quiche/src/ffi.rs
+++ b/quiche/src/ffi.rs
@@ -1346,6 +1346,10 @@ pub struct Stats {
     path_challenge_rx_count: u64,
     bytes_in_flight_duration_msec: u64,
     tx_buffered_inconsistent: bool,
+    handshake_duration_ms: i64,
+    key_update_count: u64,
+    peer_max_idle_timeout_ms: i64,
+    connection_duration_ms: u64,
 }
 
 pub struct TransportParams {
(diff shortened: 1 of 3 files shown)


quiche/src/lib.rs
```diff
@@ -9259,6 +9307,20 @@ impl std::fmt::Debug for Stats {
...
 
+        write!(
+            f,
+            " key_update_count={} connection_duration={:?}",
+            self.key_update_count, self.connection_duration,
+        )?;
+
+        if let Some(hs) = self.handshake_duration {
+            write!(f, " handshake_duration={:?}", hs)?;
+        }
```
antoniovicente, line 9318: Also add `key_update_count` ?

quiche/src/lib.rs
```diff
@@ -7638,6 +7655,20 @@ impl<F: BufFactory> Connection<F> {
...
+            peer_max_idle_timeout: if self.parsed_peer_transport_params &&
+                self.peer_transport_params.max_idle_timeout > 0
+            {
+                Some(Duration::from_millis(
+                    self.peer_transport_params.max_idle_timeout,
+                ))
+            } else {
+                None
+            },
+            connection_duration: self.created_at.elapsed(),
```
antoniovicente, line 7671: I would have considered providing the creation time since that's also an important piece of info when logging info on a connection.  The caller can compute duration on it's own based on that.

I could see there being some challenges with logging wanting to use SystemTime instead of Monotonic time.

quiche/src/lib.rs
```diff
@@ -7638,6 +7655,20 @@ impl<F: BufFactory> Connection<F> {
             path_challenge_rx_count: self.path_challenge_rx_count,
             bytes_in_flight_duration: self.bytes_in_flight_duration(),
             tx_buffered_state: self.tx_buffered_state,
+            handshake_duration: self
+                .handshake_completed_at
+                .map(|completed| completed.duration_since(self.created_at)),
```
antoniovicente, line 7660: Should the connection store the `handshake_duration` instead of `handshake_completed_at`?

quiche/src/ffi.rs
```diff
@@ -1396,6 +1400,50 @@ pub extern "C" fn quiche_conn_stats(conn: &Connection, out: &mut Stats) {
...
+    out.key_update_count = stats.key_update_count;
+    out.peer_max_idle_timeout_ms = stats
+        .peer_max_idle_timeout
+        .map_or(-1, |d| d.as_millis() as i64);
+    out.connection_duration_ms = stats.connection_duration.as_millis() as u64;
+}
+
+/// Returns the handshake duration in milliseconds, or -1 if the
+/// handshake has not yet completed.
+#[no_mangle]
```
antoniovicente, line 1415: Are the extra ffi functions below needed despite the data being available via `quiche_conn_stats` above?

antoniovicente:
Thanks for the stats improvements.  See comments above.

PR #2441, opened 2026-04-16
chore: address new clippy lints

LPardue:


tokio-quiche/src/quic/router/mod.rs
```diff
@@ -501,6 +501,7 @@ where
 
                         for cmsg in cmsgs {
                             match cmsg {
+                                #[allow(clippy::collapsible_match)]
```
LPardue, line 504: I've ignored this one because the match has a catchall and collapsing the check would change the semantic of the catchall in a low value / high risk way

antoniovicente: Looking at the cases in this PR makes me think that the collapsible_match clippy is actively harmful and we should consider opt-out of it. Perhaps we should disable this check globally.

quiche/src/recovery/bandwidth.rs
```diff
@@ -145,6 +145,7 @@ impl Bandwidth {
     }
 
     pub fn transfer_time(&self, bytes: usize) -> Duration {
+        #[allow(clippy::manual_checked_ops)]
```
LPardue, line 148: I've ignored this one because I think the current logic is clearer than 

```
let foo = (bytes as u64 * 8 * NUM_NANOS_PER_SECOND).checked_div(self.bits_per_second).unwrap_or_default();
Duration::from_nanos(foo)
```

antoniovicente (approved):

tokio-quiche/src/quic/router/mod.rs
LPardue: I tend to agree, especially because code reviewers might entirely miss the actual effect of such changes

antoniovicente: We could add this to Cargo.toml:

```
[workspace.lints.clippy]
collapsible_match = allow
```

jannes (approved):

tokio-quiche/src/quic/router/mod.rs
jrouviere: Agreed, not sure that lint really helps readability.
The second example (`qlog-dancer/src/datastore.rs`) seems better, but I think it's purely because the indentation got reduced.

jrouviere (approved):

tokio-quiche/src/quic/router/mod.rs
LPardue: ok, I'm going to try the global disable and see if it works

LPardue: I ended up doing it on an individual crate basis. If more crates start to see problems we can consider a workspace-wide setting.

PR #2443, opened 2026-04-20
tokio-quiche: drain coalesced QUIC packets on recv

babasha:
quiche::Connection::recv only processes one QUIC packet per call, but a single UDP datagram can contain several coalesced packets (a server commonly coalesces Initial + Handshake, and loopback MTU=65536 makes this common in local tests). The existing call sites in ClientConnector and IoWorker invoked recv once per buffer, so any packet coalesced after the first was silently discarded -- handshakes stall waiting for a Handshake-level frame that was already delivered.

Introduce a recv_coalesced helper at each call site that loops over conn.recv until it returns Done, advancing the buffer by the number of bytes quiche consumed.

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -617,12 +617,15 @@ where
             to: pkt.local_addr,
         };
 
+        // A single UDP datagram may contain multiple coalesced QUIC packets
+        // (e.g. Initial + Handshake). `quiche::recv` only processes one packet
+        // per call, so loop until it signals Done to drain the whole datagram.
         if let Some(gro) = pkt.gro {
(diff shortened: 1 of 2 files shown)


LPardue:
quiche's `recv()` handles coalesced packets, as noted in https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L2819. Therefore, I don't understand what problem is being addressed

antoniovicente:
> quiche's `recv()` handles coalesced packets, as noted in https://github.com/cloudflare/quiche/blob/master/quiche/src/lib.rs#L2819. Therefore, I don't understand what problem is being addressed

Some evidence of it being handled correctly: If you look at the `recv()` implementation, you'll notice that the 2 cases that return `Ok()` should return a number of bytes equal to `buf.len()` since `done` should equal `len` when `left == 0`

PR #2444, opened 2026-04-22
tokio-quiche: opt-in QLOG compression (gzip, zstd)

estcarisimo:
## Summary

Adds two opt-in Cargo features, `qlog-gzip` and `qlog-zstd`, that wrap `tokio-quiche`'s qlog writer in a streaming gzip or zstd encoder and emit `.sqlog.gz` / `.sqlog.zst` respectively. A new `QuicSettings::qlog_compression` field selects the algorithm (`QlogCompression::{None, Gzip, Zstd}`, default `None`). When `None` is selected — or when a feature is not compiled in — the writer stays on the pre-patch `BufWriter<File>` path and emits byte-identical `.sqlog` output.

Both features may be enabled together: the choice is made per connection at runtime via the config field.

Motivation: QUIC / HTTP/3 benchmarking and observability pipelines produce large qlog traces per run, many of which ship off-host for offline analysis. Using simulation in a controlled environment on ~10 real traces shows zstd-1 reaching ~8.0% of raw at ~34 ms/connection, vs gzip-6 at ~8.1% / ~374 ms/connection — same ratio, ~11× cheaper CPU.

## What's in the change

- **New public enum `tokio_quiche::settings::QlogCompression`** with `None` (always available, default), `Gzip` (requires `qlog-gzip`), and `Zstd` (requires `qlog-zstd`). Uses the foundations `#[settings]` macro so it deserializes from YAML/JSON like the rest of the config surface; derives `Copy + Eq + PartialEq` so it can be passed by value into the internal `Config`.
- **New field** `QuicSettings::qlog_compression: QlogCompression` with `#[serde(default)]`. Existing config files without this field continue to deserialize unchanged.
- **Dedicated internal module `tokio_quiche::qlog`** (`src/qlog.rs`) owns the enum, `make_qlog_writer`, and the zstd drop-wrapper. This keeps the qlog plumbing in one place and leaves room for future additions (configurable compression level, custom filename format, output-buffer tuning) without bloating `quic/mod.rs`. The public path `tokio_quiche::settings::QlogCompression` is preserved as a re-export.
- **`make_qlog_writer` returns `Box<dyn Write + Send + Sync>`** with three arms:
  - `None` → `.sqlog` + `BufWriter<File>` (byte-identical to the pre-patch default).
  - `Gzip` → `.sqlog.gz` + `flate2::write::GzEncoder<BufWriter<File>>`.
  - `Zstd` → `.sqlog.zst` + `zstd::Encoder<BufWriter<File>>` wrapped in a local `ZstdFinishOnDrop<W>` so the zstd frame trailer is flushed when the qlog streamer drops. `zstd`'s own `AutoFinishEncoder` is `!Sync` (holds a user closure), while `quiche::Connection::set_qlog` requires `Send + Sync`; the local wrapper preserves those bounds.
- **Defence-in-depth** path-traversal validator at the top of `make_qlog_writer` rejects `id` values that are empty or contain `/`, `\`, `\0`, or `..`. Today the `id` is the `Debug` form of the source CID (lowercase hex), so the validator only fires on genuinely malformed inputs — a future `Debug` change or a caller passing a differently-formatted identifier cannot escape the configured qlog directory. Rejections return `io::ErrorKind::InvalidInput`; existing `if let Ok(writer)` call sites degrade gracefully by skipping qlog setup for that connection.
- **Memory invariant**: no `Vec<u8>` accumulator anywhere in the qlog path. Every encoder wraps `BufWriter<File>` directly; chunks flush at the 8 KiB buffer boundary regardless of connection lifetime.
- **Dependencies**: `flate2` is optional with `default-features = false, features = ["rust_backend"]`, so `qlog-gzip` adds no C dependency. `zstd` is optional and links against `zstd-sys` (C); opt-in and documented in both `Cargo.toml` and the enum docstring.

## Testing

`tokio-quiche/tests/integration_tests/qlog_compression.rs` drives each `QlogCompression` variant end-to-end through a real QUIC + HTTP/3 handshake and asserts per variant:

- Emitted file has the expected extension (`.sqlog`, `.sqlog.gz`, `.sqlog.zst`).
- File starts with the expected magic bytes (gzip `1f 8b`, zstd `28 b5 2f fd`).
- Decompressed content round-trips through `qlog::reader::QlogSeqReader`, yielding a header whose `serialization_format == "JSON-SEQ"` and at least one event. For zstd this implicitly validates that `ZstdFinishOnDrop::drop` fired — a truncated frame would fail to decode or parse.

An additional test `qlog_compression_both_features_coexist_at_runtime` (gated on `all(feature = "qlog-gzip", feature = "qlog-zstd")`) drives two connections in the same process, one configured for `Gzip` and one for `Zstd`, and round-trips both emitted files through the matching decoders — proving the two features coexist without conflict.

All feature combinations build:

```
cargo check -p tokio-quiche
cargo check -p tokio-quiche --features qlog-gzip
cargo check -p tokio-quiche --features qlog-zstd
cargo check -p tokio-quiche --features qlog-gzip,qlog-zstd
```

Integration suite runs clean under each of the above (`cargo test -p tokio-quiche --test main -- --test-threads=1`) with the new `qlog_compression` tests added per-feature.

## Notes for reviewers

- 10 files changed, +601 / −16. If your diff looks larger, check the target branch.
- `qlog` and `tempfile` are added as `[dev-dependencies]` only (for the `QlogSeqReader` round-trip helper and test tmpdirs).
- The `zstd` level is hardcoded to 3 (the zstd default) in `make_qlog_writer`. Exposing it via a config field is a natural follow-up if benchmarking shows a different default is warranted.

### DIFF
--- a/qlog/src/reader.rs
+++ b/qlog/src/reader.rs
@@ -24,7 +24,14 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+
 use crate::QlogSeq;
+use crate::SQLOG_EXT;
+use crate::SQLOG_GZ_EXT;
+use crate::SQLOG_ZST_EXT;
 
 /// Represents the format of the read event.
 #[allow(clippy::large_enum_variant)]
@@ -64,6 +71,92 @@ impl<'a> QlogSeqReader<'a> {
 
             Err(e) => Err(e.into()),
         }
+    }
+
+    /// Convenience constructor that opens `path` and picks a streaming
+    /// decoder based on the file's compound extension:
+    ///
+    /// * `*.sqlog` -> raw JSON-SEQ (always available).
+    /// * `*.sqlog.gz` -> gzip via `flate2` (requires the `gzip` feature).
+    /// * `*.sqlog.zst` -> zstd via `zstd` (requires the `zstd` feature).
+    ///
+    /// The dispatch tests the *compound* suffix (`.sqlog.gz`,
+    /// `.sqlog.zst`, `.sqlog`) on the full filename, in that order.
+    /// This rejects bare `.gz` or `.zst` names that do not also carry
+    /// the `.sqlog` segment, and avoids the trap where
+    /// [`Path::extension`] strips only the last component (which
+    /// would silently accept `something.tar.gz`).
+    ///
+    /// Unknown extensions, or compressed extensions whose matching
+    /// feature is not enabled, return an [`std::io::ErrorKind::Unsupported`]
+    /// error with a message pointing at the feature that is needed.
+    ///
+    /// This is the intended single entry point for reading a qlog
+    /// file regardless of compression.
+    pub fn with_file(
+        path: impl AsRef<Path>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let path = path.as_ref();
+        let file = File::open(path)?;
+        let name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
+
+        // Order matters: `.sqlog.gz` and `.sqlog.zst` must be tested
+        // before `.sqlog` (every compressed suffix also ends with
+        // `.sqlog` if you only look at the last extension).
+        //
+        // Each compound suffix is matched exactly once. The
+        // `#[cfg(feature = "...")]` lives inside the `if` body so the
+        // disabled-feature path can return a helpful error rather
+        // than falling through to the unknown-extension branch.
+        if name.ends_with(SQLOG_GZ_EXT) {
+            #[cfg(feature = "gzip")]
+            {
+                let reader: Box<dyn std::io::BufRead + Send + Sync> =
+                    Box::new(BufReader::new(flate2::read::GzDecoder::new(file)));
+                return Self::new(reader);
+            }
+            #[cfg(not(feature = "gzip"))]
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Unsupported,
+                format!(
+                    "qlog file {name:?} requires the `gzip` feature on \
+                     the qlog crate to decode"
+                ),
+            )
+            .into());
+        }
+        if name.ends_with(SQLOG_ZST_EXT) {
+            #[cfg(feature = "zstd")]
+            {
+                let decoder = zstd::Decoder::new(file)?;
+                let reader: Box<dyn std::io::BufRead + Send + Sync> =
+                    Box::new(BufReader::new(decoder));
+                return Self::new(reader);
+            }
+            #[cfg(not(feature = "zstd"))]
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Unsupported,
+                format!(
+                    "qlog file {name:?} requires the `zstd` feature on \
+                     the qlog crate to decode"
+                ),
+            )
+            .into());
+        }
+        if name.ends_with(SQLOG_EXT) {
+            let reader: Box<dyn std::io::BufRead + Send + Sync> =
+                Box::new(BufReader::new(file));
+            return Self::new(reader);
+        }
+
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Unsupported,
+            format!(
+                "qlog file {name:?} does not match a known qlog \
+                 extension ({SQLOG_EXT}, {SQLOG_GZ_EXT}, {SQLOG_ZST_EXT})"
+            ),
+        )
+        .into())
     }
 
     fn read_record(
--- a/qlog/src/writer.rs
+++ b/qlog/src/writer.rs
@@ -1,0 +1,245 @@
+// Copyright (C) 2026, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//! QLOG file writer plumbing -- compression-aware companion to
+//! [`crate::reader`].
+//!
+//! This module owns the writer-side surface for emitting JSON-SEQ
+//! qlog streams, optionally wrapped in a streaming compressor. It is
+//! a deliberate exception to the qlog crate's otherwise pure-data
+//! posture: the writer helpers, the compression enum, and the
+//! filename / extension conventions live here so any consumer
+//! (tokio-quiche, [`crate::reader::QlogSeqReader::with_file`],
+//! external tooling) shares a single canonical writer story.
+//!
+//! Compression is opt-in via Cargo features:
+//! * `gzip` pulls in [`flate2`].
+//! * `zstd` pulls in the [`zstd`] crate (C dependency via `zstd-sys`).
+//!
+//! The [`QlogCompression`] enum's `Gzip` and `Zstd` variants are
+//! compile-time gated on their respective features, so a build that
+//! disables one of them cannot construct the unsupported variant.
+//!
+//! Bytes flow `producer -> [compressor] -> W` where `W` is any
+//! `Write + Send + Sync + 'static`. No buffering is added here: both
+//! `flate2` and `zstd` emit output in large chunks (DEFLATE blocks /
+//! zstd frames), so an extra `BufWriter` is redundant; callers are
+//! free to add one if they need to.
+
+use std::fs::File;
+use std::io;
+use std::io::Write;
+use std::path::Path;
+
+// Extension constants live at the crate root so they can be shared
+// between [`crate::writer`] and [`crate::reader`] without either
+// module depending on the other. Re-exported here for ergonomic
+// access via `qlog::writer::SQLOG_EXT` etc.
+pub use crate::SQLOG_EXT;
+pub use crate::SQLOG_GZ_EXT;
+pub use crate::SQLOG_ZST_EXT;
+
+/// Boxed `Write` returned by [`make_qlog_writer`] /
+/// [`make_qlog_writer_from_path`].
+///
+/// Producers (e.g. `quiche::Connection::set_qlog`) typically require
+/// `Send + Sync`; the boxed form keeps the writer object-safe across
+/// the compression-vs-no-compression branches.
+pub type QlogFileWriter = Box<dyn Write + Send + Sync>;
+
+/// Compression algorithm applied to QLOG output streams.
+///
+/// `None` is always available. `Gzip` and `Zstd` are compile-time
+/// gated on the `gzip` and `zstd` Cargo features respectively; a
+/// build that disables one of those features cannot reference the
+/// corresponding variant.
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum QlogCompression {
+    /// No compression. Emit raw `.sqlog` files.
+    #[default]
+    None,
+    /// Gzip streaming compression (DEFLATE + gzip framing). Emits
+    /// `.sqlog.gz` files. Requires the `gzip` Cargo feature.
+    #[cfg(feature = "gzip")]
+    Gzip,
+    /// Zstd streaming compression. Emits `.sqlog.zst` files. Requires
+    /// the `zstd` Cargo feature.
+    #[cfg(feature = "zstd")]
+    Zstd,
+}
+
+#[cfg(feature = "foundations")]
+impl foundations::settings::Settings for QlogCompression {}
+
+/// Return the qlog filename (not including the directory) for a
+/// stream whose identifier is `id`, with the suffix matching
+/// `compression`: `<id>.sqlog`, `<id>.sqlog.gz`, or `<id>.sqlog.zst`.
+pub fn qlog_file_name(id: &str, compression: QlogCompression) -> String {
+    match compression {
+        QlogCompression::None => format!("{id}{SQLOG_EXT}"),
+        #[cfg(feature = "gzip")]
+        QlogCompression::Gzip => format!("{id}{SQLOG_GZ_EXT}"),
+        #[cfg(feature = "zstd")]
+        QlogCompression::Zstd => format!("{id}{SQLOG_ZST_EXT}"),
+    }
+}
+
+/// Wrap `inner` in the streaming encoder selected by `compression`
+/// and return a boxed `Write` that a qlog producer (e.g. quiche via
+/// `set_qlog`) writes into.
+///
+/// The generic `W` bound lets production call sites pass a
+/// `std::fs::File` while tests can pass an in-process buffer (e.g.
+/// `Vec<u8>`). For the common case of "open a file at this path and
+/// pick the compressor by extension" use
+/// [`make_qlog_writer_from_path`].
+///
+/// No buffering is added here.
+pub fn make_qlog_writer<W>(
+    inner: W, compression: QlogCompression,
+) -> io::Result<QlogFileWriter>
+where
+    W: Write + Send + Sync + 'static,
+{
+    match compression {
+        QlogCompression::None => Ok(Box::new(inner)),
+        #[cfg(feature = "gzip")]
+        QlogCompression::Gzip => {
+            let encoder = flate2::write::GzEncoder::new(
+                inner,
+                flate2::Compression::default(),
+            );
+            Ok(Box::new(encoder))
+        },
+        #[cfg(feature = "zstd")]
+        QlogCompression::Zstd => {
+            // Level 3 is the zstd default: a balanced point on the
+            // ratio-vs-speed curve.
+            let encoder = zstd::Encoder::new(inner, 3)?;
+            Ok(Box::new(ZstdFinishOnDrop {
+                encoder: Some(encoder),
+            }))
+        },
+    }
+}
+
+/// Convenience function to create a File at `path` with a writer
+/// based on `compression`.
+///
+/// Equivalent to manually creating a File and passing it to
+/// [`make_qlog_writer`].
+pub fn make_qlog_writer_from_path<P: AsRef<Path>>(
+    path: P, compression: QlogCompression,
+) -> io::Result<QlogFileWriter> {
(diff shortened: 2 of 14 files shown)


qlog/src/writer.rs
```diff
@@ -0,0 +1,213 @@
...
+/// build.
+#[settings]
+#[derive(Copy, Eq, PartialEq)]
+pub enum QlogCompression {
+    /// No compression. Emit raw `.sqlog` files.
+    #[default]
+    None,
+    /// Gzip streaming compression (DEFLATE + gzip framing). Emits
+    /// `.sqlog.gz` files. Requires the `qlog-gzip` Cargo feature.
+    Gzip,
```
gregor-cf, line 95: I'd gate this variant on `#[cfg(feature = "qlog-gzip")]` (and similarly for Zstd). That way it's a compile-time error if a user tries to use a compression algo that's not enabled. 
This will also work smoothly when the config is deserialized by `foundations` (foundations has two modes. the user can decide if unknown variants are deserialized as the default option or if they fail deserialization).

qlog/src/writer.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    let mut path = std::path::PathBuf::from(dir);
+
+    match compression {
+        QlogCompression::None => {
+            path.push(format!("{id}.sqlog"));
+            let file = std::fs::File::create(&path)?;
+            Ok(Box::new(std::io::BufWriter::new(file)))
+        },
+        QlogCompression::Gzip => {
```
gregor-cf, line 138: if you add `cfg` guards to the variants as I suggested above, you can move the `#[cfg(feature = "qlog-gzip")]` to the match-arm directly.

tokio-quiche/tests/integration_tests/qlog_compression.rs
```diff
@@ -0,0 +1,319 @@
...
+/// Decode an emitted qlog file through the appropriate streaming
+/// decoder (picked by file extension), then parse the decoded byte
+/// stream as JSON-SEQ via [`QlogSeqReader`]. Returns the parsed
+/// [`qlog::QlogSeq`] header plus every [`Event`] the reader yields.
+///
+/// This is the end-to-end "does this file round-trip through the qlog
+/// parser" check: it goes further than the magic-byte / `0x1e`
+/// assertions by proving the emitted bytes form a syntactically valid
+/// JSON-SEQ qlog stream.
+fn decode_and_parse_qlog(path: &Path) -> (qlog::QlogSeq, Vec<Event>) {
```
LPardue: Now I look at it, this probably makes sense to put in the qlog crate itself. That way, any reading app (qlog-dancer, tests etc) can benefit.

For example, we could add a companion to QlogSeqReader::new called `with_file` that takes `path` and returns `-> Result<Self, Box<dyn std::error::Error>> `. Then here, you'd call that and do the result checkign and collection.

tokio-quiche/src/qlog.rs
```diff
@@ -0,0 +1,213 @@
...
+            path.push(format!("{id}.sqlog"));
+            let file = std::fs::File::create(&path)?;
+            Ok(Box::new(std::io::BufWriter::new(file)))
+        },
+        QlogCompression::Gzip => {
+            #[cfg(feature = "qlog-gzip")]
+            {
+                path.push(format!("{id}.sqlog.gz"));
+                let file = std::fs::File::create(&path)?;
+                let buffered = std::io::BufWriter::new(file);
```
gregor-cf: do we need a BufWriter here or does flate2 already write data in buffered chunks? If the latter, we shouldn't add another buffer.

tokio-quiche/tests/integration_tests/qlog_compression.rs
```diff
@@ -0,0 +1,319 @@
...
+/// teardown on localhost typically takes <10 ms; 200 ms is generous and
+/// keeps the tests fast.
+const DRAIN_DELAY: Duration = Duration::from_millis(200);
+
+/// Locate the single sqlog-ish file under `qlog_dir`, regardless of its
+/// exact name (the server picks the filename from the connection's
+/// source CID) or extension. Asserts exactly one file is present so a
+/// stray leftover from a prior run surfaces as a failure rather than a
+/// silent pick-one.
+fn find_qlog(qlog_dir: &Path) -> PathBuf {
```
LPardue: It seems we could sidestep some of this test busywork and simplify things with a little bit of effort. For example, If we added a `make_qlog_writer_for_test` under a `cfg test`, we have flexibility to target any writer. That could allow us to pipe the logs directly into a resource the test creates itself, such as some memmaped buffer.

Such an approach avoids having to go and fish for an actual file that is being created by TQ.

tokio-quiche/Cargo.toml
```diff
@@ -78,6 +96,15 @@ tokio-util = { workspace = true, features = [
 ] }
 url = { workspace = true }
 
+# Optional: gzip backend for `qlog-gzip`. The `rust_backend` feature selects
+# the pure-Rust `miniz_oxide` implementation, avoiding any C dependency.
+flate2 = { version = "1", default-features = false, features = [
+  "rust_backend",
```
gregor-cf: I think we should use the default features here. The `rust_backend` is the default anyways. this makes it clearer that apps using tokio-quiche can select their own flate2 backend and that backend will take precedence.

tokio-quiche/src/qlog.rs
```diff
@@ -0,0 +1,213 @@
...
+                let file = std::fs::File::create(&path)?;
+                Ok(Box::new(std::io::BufWriter::new(file)))
+            }
+        },
+        QlogCompression::Zstd => {
+            #[cfg(feature = "qlog-zstd")]
+            {
+                path.push(format!("{id}.sqlog.zst"));
+                let file = std::fs::File::create(&path)?;
+                let buffered = std::io::BufWriter::new(file);
```
gregor-cf: ditto about potential double-buffering

tokio-quiche/Cargo.toml
```diff
@@ -40,6 +40,24 @@ capture_keylogs = []
...
 tokio-task-metrics = []
 
+# Enable opt-in gzip compression for QLOG output files. When enabled,
+# `QuicSettings::qlog_compression = QlogCompression::Gzip` wraps the qlog
+# writer in a streaming gzip encoder and emits `.sqlog.gz` files. Pulls in
+# `flate2` with its pure-Rust backend (no C dependency).
+#
+# May be combined with `qlog-zstd`; the algorithm is selected per
+# connection at runtime via `QuicSettings::qlog_compression`.
+qlog-gzip = ["dep:flate2"]
```
gregor-cf: I would make `qlog-gzip` and `qlog-zstd` default features.

gregor-cf (commented):
LGTM. This is awesome! 

CI currently doesn't execute the tests with compression. So we should either make `qlog-gzip` and `qlog-zsdt` default features of tokio-quiche, or if not then we should run the tests with the features enabled in CI

tokio-quiche/src/qlog.rs
```diff
@@ -0,0 +1,213 @@
...
+        id.contains('/') ||
+        id.contains('\\') ||
+        id.contains('\0') ||
+        id.contains("..")
+    {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            "qlog connection id contains disallowed characters",
+        ));
+    }
```
LPardue: This seems superfluous, the `File::create` call will fail if the path is bad and return an error anyway

tokio-quiche/Cargo.toml
LPardue: How about default enabled if quiche qlog feature is enabled?

tokio-quiche/tests/integration_tests/qlog_compression.rs
estcarisimo: Sounds good!

estcarisimo:
## Re @gregor-cf top-level comment

Resolved by making qlog-gzip and qlog-zstd default features. The existing CI workflow (cargo test --workspace --features=async,ffi,qlog) now runs the compression paths automatically; no extra matrix entry needed. Verified locally: the qlog_compression_gzip_*, qlog_compression_zstd_*, and qlog_compression_both_features_* tests all run under the default-features configuration.

qlog/src/writer.rs
estcarisimo: Done. The Gzip and Zstd variants are now #[cfg(feature = "…")]

qlog/src/writer.rs
estcarisimo: Done.

tokio-quiche/src/qlog.rs
estcarisimo: Good observation. Dropped the BufWriter. flate2 already writes in chunks large enough that an extra 8 KB buffer in front of it just adds overhead.

tokio-quiche/src/qlog.rs
estcarisimo: Same as the gzip arm. dropped the BufWriter. zstd::Encoder buffers input internally up to a full frame (~128 KB at default level) before writing, so an outer 8 KB buffer was redundant.

tokio-quiche/Cargo.toml
estcarisimo: Done. flate2 is now { version = "1", optional = true } — no feature override. rust_backend stays the default in flate2's own Cargo.toml, and downstream consumers can now pick a different backend (e.g. zlib-ng) via their own flate2 dep.

tokio-quiche/Cargo.toml
estcarisimo: Both qlog-gzip and qlog-zstd are in the default feature list now. Consumers who want a leaner binary (no C dep on zstd-sys, for instance) can set default-features = false and opt features back in explicitly.

tokio-quiche/src/qlog.rs
estcarisimo: Agreed, removed. id comes from format!("{:?}", scid) (lowercase hex today), and File::create already surfaces bad paths on its own — the validator was silencing a Semgrep tainted-path finding rather than defending against a real threat.

tokio-quiche/tests/integration_tests/qlog_compression.rs
estcarisimo: Done. Added qlog::reader::QlogSeqReader::with_file(path) that opens the file, wraps it in BufReader, and delegates to QlogSeqReader::new. Kept the qlog crate format-agnostic — with_file only handles raw .sqlog; for compressed files the caller opens the matching streaming decoder and calls QlogSeqReader::new directly.

tokio-quiche/tests/integration_tests/qlog_compression.rs
```diff
@@ -0,0 +1,201 @@
...
+    let ext = path
+        .extension()
+        .and_then(|s| s.to_str())
+        .expect("qlog file has an extension");
+
+    let mut seq_reader = match ext {
+        "sqlog" => QlogSeqReader::with_file(path).expect("open .sqlog"),
+        #[cfg(feature = "qlog-gzip")]
+        "gz" => {
+            let file = std::fs::File::open(path).expect("open .sqlog.gz");
```
LPardue: Sorry, in my previous comment I meant all the suffix detection should be done in the qlog crate too, that makes it way more reusable

tokio-quiche/tests/integration_tests/qlog_compression.rs
antoniovicente: I think that Lucas is suggesting moving `tokio-quiche/src/qlog.rs` and some of the integration tests to the qlog crate.

estcarisimo: true, will move it right now. It is a small move and big win

gregor-cf:
LGTM. Looks like github won't let me resolve my comments,  but consider them resolved.

tokio-quiche/tests/integration_tests/qlog_compression.rs
estcarisimo: Adjusted it. Please let me know if the current version aligns with your feedback.

qlog/src/compression.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Wrap `inner` in the streaming encoder selected by `compression`
+/// and return a boxed `Write` that a qlog producer (e.g. quiche via
+/// `set_qlog`) writes into.
+///
+/// The generic `W` bound lets production call sites pass a
+/// `std::fs::File` while tests can pass an in-process buffer (e.g.
+/// `Vec<u8>`). No buffering is added here.
+pub fn make_qlog_writer<W>(
+    inner: W, compression: QlogCompression,
+) -> std::io::Result<Box<dyn std::io::Write + Send + Sync>>
```
LPardue: These seem too low level for a public API. We can see in the TQ example its got to manually construct the path, then a file in order, to covert that into a writer via make_qlog_writer. It not hard but it is tedious.

What I would recommend is instead to change this module to `writer.rs`. This can mirror a bit how reader.rs works. It should define something like a `struct QlogFileWriter` and a `new(id: &str, compression: QlogCompression)` and returns  `std::io::Result<Self>`. The struct can hold the inner writer and provide an accessor  for it, which can be used to pass that it to `conn.set_qlog()` e.g.

qlog/src/reader.rs
```diff
@@ -66,6 +70,52 @@ impl<'a> QlogSeqReader<'a> {
...
+        path: impl AsRef<Path>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let path = path.as_ref();
+        let file = File::open(path)?;
+        let ext = path.extension().and_then(|s| s.to_str());
+
+        let reader: Box<dyn std::io::BufRead + Send + Sync> = match ext {
+            Some("sqlog") | None => Box::new(BufReader::new(file)),
+            #[cfg(feature = "gzip")]
+            Some("gz") =>
```
LPardue: These magic strings like `gz` should be string constants defined in the qlog module and shared across reader and my proposed writer

qlog/src/compression.rs
LPardue: note, you can keep these functions for private usage

tokio-quiche/src/settings/quic.rs
```diff
@@ -152,6 +154,16 @@ pub struct QuicSettings {
...
     pub qlog_dir: Option<String>,
 
+    /// Compression applied to QLOG output files.
+    ///
+    /// Defaults to [`QlogCompression::None`], preserving the historical
+    /// behavior of emitting raw `.sqlog` files. The `Gzip` and `Zstd`
+    /// variants require the `qlog-gzip` and `qlog-zstd` Cargo features
+    /// (both enabled by default); builds that disable those features
+    /// cannot reference the corresponding variant.
+    #[serde(default)]
```
LPardue, line 164: I don't know if you need this default declatation or not. It doesn't seem to fit the pattern of other settings. Perhaps @antoniovicente can advise.

qlog/src/reader.rs
LPardue: Also, I've just realised the check is buggy compared to the documentation. It's not teting the compound extension and you risk some typo like `sqlogs.tar.gz` beng passed into the reader. The Path API is not great for this kind of thing, you might need to convert the path to a string and checking `ends_with`

qlog/src/compression.rs
gregor-cf: IMHO the function is useful to have as a `pub fn` if you have a `std::io::Write` that's not based on a file. So it'd keep it `pub`. 

nit: I don't know if we need a `struct QlogFileWriter`, but I think a type alias +  convenience free function would be nice: 
```
type QlogFileWriter = Box<dyn std::io::Write + Send + Sync>;
pub fn make_qlog_writer_from_path<P: AsRef<Path>>(path: P, compression: QlogCompression) -> Result<QlogFileWriter>
```

LPardue: sure something like that works too. The writer mdule still makes more sense to me than `compression`

gregor-cf: > The writer mdule still makes more sense to me than compression

+1
missed that part.

qlog/src/reader.rs
```diff
@@ -66,6 +75,92 @@ impl<'a> QlogSeqReader<'a> {
...
+        if name.ends_with(SQLOG_EXT) {
+            let reader: Box<dyn std::io::BufRead + Send + Sync> =
+                Box::new(BufReader::new(file));
+            return Self::new(reader);
+        }
+
+        // Disabled-feature compressed extensions: surface a helpful
+        // error rather than the silent fallback the previous
+        // implementation provided.
+        if name.ends_with(".sqlog.gz") {
```
antoniovicente: Could be merged into `if name.ends_with(SQLOG_GZ_EXT) {` above if the feature check was moved inside the if.

Similarly with `if name.ends_with(SQLOG_ZST_EXT) {`

qlog/tests/writer_roundtrip.rs
```diff
@@ -0,0 +1,245 @@
...
+#[cfg(feature = "gzip")]
+#[test]
+fn roundtrip_gzip() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let path = emit_one_event(QlogCompression::Gzip, dir.path());
+    assert!(
+        path.to_str().unwrap().ends_with(".sqlog.gz"),
+        "expected .sqlog.gz extension, got {path:?}"
+    );
+    assert_roundtrip(&path);
```
antoniovicente, line 188: I wonder if we could run some check on the file to make sure that the contents is compressed.  Perhaps a file magic number check?

I think that the first 2 characters for uncompressed qlog are `0x1e 0x7b` (record separator followed by {)

gzip starts with `1F 8B 08`
zstd starts with `28 b5 2F FD`

tokio-quiche/src/settings/quic.rs
antoniovicente: I don't know if it is needed.  It could be related to use of `#[default]` in the QlogCompression enum.

I would consider changing this line to something like:

`#[serde(default = "QuicSettings::default_no_compression")]`

As a way to succinctly document the default.  I think you could remove `#[default]` from the enum after doing that.

antoniovicente (approved):

qlog/src/reader.rs
estcarisimo: Done

qlog/tests/writer_roundtrip.rs
estcarisimo: Good idea! Really like it. Added in the latest push

qlog/src/reader.rs
estcarisimo: Confirmed bug. Fixed

qlog/src/compression.rs
estcarisimo: Various changes were made here. Please let me know what you think

qlog/src/writer.rs
```diff
@@ -0,0 +1,265 @@
...
+/// (`format!("{id}{SQLOG_EXT}")`) and matched as a path suffix
+/// (`name.ends_with(SQLOG_EXT)`). Shared between the writer and
+/// [`crate::reader::QlogSeqReader::with_file`].
+pub const SQLOG_EXT: &str = ".sqlog";
+
+/// File extension for a gzip-compressed qlog stream.
+pub const SQLOG_GZ_EXT: &str = ".sqlog.gz";
+
+/// File extension for a zstd-compressed qlog stream.
+pub const SQLOG_ZST_EXT: &str = ".sqlog.zst";
```
LPardue: These should probably go in qlog/lib.rs since they are shared across modules.

qlog/src/writer.rs
```diff
@@ -0,0 +1,265 @@
...
+
+#[cfg(feature = "foundations")]
+impl foundations::settings::Settings for QlogCompression {}
+
+/// Return the qlog filename (not including the directory) that
+/// matches `compression` for a stream whose identifier is `id`.
+///
+/// Kept alongside [`make_qlog_writer`] so the filename convention
+/// lives in one place: `<id>.sqlog`, `<id>.sqlog.gz`, or
+/// `<id>.sqlog.zst`.
```
LPardue: This doc comment can be a lot more concise.

As a rule of thumb, anything in a public API should not leak implementation detail such as "Kept alongside [`make_qlog_writer`] "

qlog/src/writer.rs
```diff
@@ -0,0 +1,265 @@
...
+/// function does not append an extension. Pair with
+/// [`qlog_file_name`] when constructing the path:
+///
+/// ```ignore
+/// let dir = std::path::Path::new("/var/log/qlog");
+/// let id = "abc123";
+/// let path = dir.join(qlog::writer::qlog_file_name(id, compression));
+/// let writer =
+///     qlog::writer::make_qlog_writer_from_path(&path, compression)?;
+/// ```
```
LPardue: Can be simplified 

```suggestion
/// Convenience function to create a File at `path` with a writer based on
/// `compression`.
///
/// Equivalent to manually creating a File and passing it to 
/// `make_qlog_writer()`
```

qlog/src/writer.rs
estcarisimo: Done!

qlog/src/writer.rs
estcarisimo: Trimmed

qlog/src/writer.rs
estcarisimo: Changed

qlog/src/writer.rs
```diff
@@ -0,0 +1,245 @@
...
+use std::io::Write;
+use std::path::Path;
+
+// Extension constants live at the crate root so they can be shared
+// between [`crate::writer`] and [`crate::reader`] without either
+// module depending on the other. Re-exported here for ergonomic
+// access via `qlog::writer::SQLOG_EXT` etc.
+pub use crate::SQLOG_EXT;
+pub use crate::SQLOG_GZ_EXT;
+pub use crate::SQLOG_ZST_EXT;
```
LPardue: I don't think t helps to rexport anything, there's not example of that being used in this PR for instance

```suggestion

use crate::SQLOG_EXT;
use crate::SQLOG_GZ_EXT;
use crate::SQLOG_ZST_EXT;
```

LPardue (approved):

PR #2446, opened 2026-04-23
Upgrade boring to 5.2

cjpatton:
Bumps the `boring` workspace dependency from 4.3 to 5.2 and pins the `quiche/deps/boringssl` submodule to match `boring-sys` 5.2. Updates test expectations and CI accordingly (see commit messages for details).
Reviewers are encouraged to review these commit-by-commit. The changes were driven mostly by CI feedback.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -68,12 +68,15 @@ jobs:
     # duplicate builds for PRs created from internal branches.
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     # `quiche-fuzz` calls `RAND_reset_for_fuzzing`, which BoringSSL only
-    # exports when `BORINGSSL_UNSAFE_DETERMINISTIC_MODE` is defined.
-    # `boring-sys` doesn't expose a feature for that, but cmake-rs honors
-    # `CFLAGS`/`CXXFLAGS`, so inject the defines there.
+    # exports when built with `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`.
+    # BoringSSL 5.x consolidated the older
+    # `BORINGSSL_UNSAFE_{DETERMINISTIC,FUZZER}_MODE` defines into this
+    # single switch. `boring-sys` doesn't expose a feature for it, but
+    # cmake-rs (via cc-rs) honors `CFLAGS`/`CXXFLAGS`, so inject the
+    # define there.
     env:
-      CFLAGS: "-DBORINGSSL_UNSAFE_DETERMINISTIC_MODE -DBORINGSSL_UNSAFE_FUZZER_MODE"
-      CXXFLAGS: "-DBORINGSSL_UNSAFE_DETERMINISTIC_MODE -DBORINGSSL_UNSAFE_FUZZER_MODE"
+      CFLAGS: "-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
+      CXXFLAGS: "-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -7,7 +7,7 @@ permissions:
   pull-requests: write
 
 env:
-  FEATURES: "async,ffi,qlog"
+  FEATURES: "async,ffi,qlog,rpk"
   RUSTFLAGS: "-D warnings"
   RUSTDOCFLAGS: "--cfg docsrs"
   RUSTTOOLCHAIN: "nightly"
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -130,11 +130,21 @@ jobs:
       matrix:
         target: ["x86_64-pc-windows-msvc", "i686-pc-windows-msvc", "x86_64-pc-windows-gnu", "i686-pc-windows-gnu"]
         include:
+          # i686 (32-bit) BoringSSL builds require SSE2 explicitly. The
+          # toolchain file in BoringSSL sets `-msse2`, but cmake-rs
+          # passes `-DCMAKE_C_FLAGS=...` on the command line which
+          # overrides that, so we re-add it here.
+          - target: "i686-pc-windows-msvc"
+            cxxflags: "-msse2"
           - target: "i686-pc-windows-gnu"
-            cflags: "-mlong-double-64" # bindgen detects that Rust doesn't support 80-bit long double
+            # `-mlong-double-64`: bindgen detects that Rust doesn't
+            # support 80-bit long double.
+            # `-msse2`: same reason as i686-pc-windows-msvc above.
+            cflags: "-mlong-double-64 -msse2"
+            cxxflags: "-mlong-double-64 -msse2"
     env:
       CFLAGS: ${{ matrix.cflags }}
-      CXXFLAGS: ${{ matrix.cflags }}
+      CXXFLAGS: ${{ matrix.cxxflags }}
       BINDGEN_EXTRA_CLANG_ARGS: ${{ matrix.cflags }}
     # Only run on "pull_request" event for external PRs. This is to avoid
     # duplicate builds for PRs created from internal branches.
@@ -149,18 +159,55 @@ jobs:
           toolchain: ${{ env.RUSTTOOLCHAIN }}
           targets: ${{ matrix.target }}
 
+      # The windows-2022 image preinstalls MSYS2 (gcc 14.2.0) at
+      # `C:\msys64` but doesn't put it on PATH. For x86_64-pc-windows-gnu
+      # we rely on these preinstalled tools and just point the C/C++
+      # toolchain at them, matching `boring`'s own CI workflow.
       - name: Set up MinGW for 64 bit
         if: matrix.target == 'x86_64-pc-windows-gnu'
-        uses: bwoodsend/setup-winlibs-action@v1.10
-        with:
-          tag: 12.2.0-16.0.0-10.0.0-msvcrt-r5
+        shell: bash
+        run: |
+          echo >> "$GITHUB_ENV" CC=gcc
+          echo >> "$GITHUB_ENV" CXX=g++
+          echo >> "$GITHUB_ENV" 'C_INCLUDE_PATH=C:\msys64\usr\include'
+          echo >> "$GITHUB_ENV" 'CPLUS_INCLUDE_PATH=C:\msys64\usr\include'
+          echo >> "$GITHUB_ENV" 'LIBRARY_PATH=C:\msys64\usr\lib'
 
+      # For i686 we install a fresh 32-bit MSYS2 toolchain (and a
+      # 32-bit cmake from MSYS2, since the system cmake is 64-bit and
+      # picks up the wrong `gcc`). Mirrors `boring`'s CI workflow.
       - name: Set up MinGW for 32 bit
         if: matrix.target == 'i686-pc-windows-gnu'
-        uses: bwoodsend/setup-winlibs-action@v1.10
+        uses: msys2/setup-msys2@v2
+        id: msys2
         with:
-          architecture: i686
-          tag: 12.2.0-16.0.0-10.0.0-msvcrt-r5
+          msystem: MINGW32
+          path-type: inherit
+          install: >-
+            mingw-w64-i686-gcc
+            mingw-w64-i686-cmake
+
+      - name: Set up 32-bit MSYS2 env vars
+        if: matrix.target == 'i686-pc-windows-gnu'
+        shell: bash
+        run: |
+          MSYS_ROOT='${{ steps.msys2.outputs.msys2-location }}'
+          test -d "$MSYS_ROOT\\mingw32\\bin"
+          echo >> $GITHUB_PATH "$MSYS_ROOT\\mingw32\\bin"
+          echo >> $GITHUB_PATH "$MSYS_ROOT\\usr\\bin"
+          echo >> $GITHUB_ENV CC="$MSYS_ROOT\\mingw32\\bin\\gcc"
+          echo >> $GITHUB_ENV CXX="$MSYS_ROOT\\mingw32\\bin\\g++"
+          echo >> $GITHUB_ENV AR="$MSYS_ROOT\\mingw32\\bin\\ar"
+          echo >> $GITHUB_ENV CFLAGS="$CFLAGS -I$MSYS_ROOT\\mingw32\\include"
+          echo >> $GITHUB_ENV CXXFLAGS="$CXXFLAGS -I$MSYS_ROOT\\mingw32\\include"
+          echo >> $GITHUB_ENV BINDGEN_EXTRA_CLANG_ARGS="$BINDGEN_EXTRA_CLANG_ARGS -I$MSYS_ROOT\\mingw32\\include"
+          echo >> $GITHUB_ENV LIBRARY_PATH="$MSYS_ROOT\\mingw32\\lib"
+          echo >> $GITHUB_ENV LDFLAGS="-L$MSYS_ROOT\\mingw32\\lib"
+          # Force cmake-rs to use the MinGW Makefiles generator instead
+          # of MSYS Makefiles; cmake-rs's default for windows-gnu picks
+          # the latter, which then can't find `cl.exe` (cmake's default
+          # Windows compiler) and fails configuration.
+          echo >> $GITHUB_ENV "CMAKE_GENERATOR=MinGW Makefiles"
 
       - name: Install dependencies
         uses: crazy-max/ghaction-chocolatey@v3
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -7,7 +7,7 @@ permissions:
   pull-requests: write
 
 env:
-  DEFAULT_OPTIONS: "--features=async,ffi,qlog --workspace"
+  DEFAULT_OPTIONS: "--features=async,ffi,qlog,rpk --workspace"
   # Used by `quiche_multiarch`, which can't link `boring` for foreign targets.
   NO_BORING_OPTIONS: "--features=ffi,qlog --workspace --exclude h3i --exclude tokio-quiche"
   RUSTFLAGS: "-D warnings"
@@ -20,6 +20,10 @@ concurrency:
 jobs:
   quiche:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        boring_version: ["latest", "4"]
     # Only run on "pull_request" event for external PRs. This is to avoid
     # duplicate builds for PRs created from internal branches.
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
@@ -37,6 +41,20 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install libexpat1-dev libfreetype6-dev libfontconfig1-dev
+
+      - name: Maybe pin boring to MSRV (4.19.0)
+        if: matrix.boring_version == '4'
+        run: |
+          # Pin boring to the minimum supported 4.x. cargo's resolver
+          # normally picks the highest match (5.x); generate the
+          # lockfile first, then downgrade to the floor of the
+          # workspace dep range (`>= 4.19, < 6`) via `--precise`.
+          # Pinning to the floor (rather than the latest 4.x) catches
+          # accidental use of APIs added in later 4.x point releases.
+          # `quiche/src/build.rs` and `tokio-quiche/build.rs` then read
+          # the lockfile and flip `cfg(boring_v4)` on.
+          cargo generate-lockfile
+          cargo update -p boring --precise 4.19.0
 
       - name: Unused dependency check
         uses: bnjbvr/cargo-machete@main
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -52,7 +52,7 @@ jobs:
           # Pinning to the floor (rather than the latest 4.x) catches
           # accidental use of APIs added in later 4.x point releases.
           # `quiche/src/build.rs` and `tokio-quiche/build.rs` then read
-          # the lockfile and flip `cfg(boring_v4)` on.
+          # the lockfile and leave `cfg(boring_v5)` off (the default).
           cargo generate-lockfile
           cargo update -p boring --precise 4.19.0
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ mem_forget = "deny"
 unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(capture_keylogs)',
   'cfg(test_invalid_len_compilation_fail)',
+  'cfg(boring_v4)',
 ] }
 
 [workspace.metadata.release]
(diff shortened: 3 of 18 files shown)


quiche/src/tests.rs
```diff
@@ -6425,7 +6443,7 @@ fn client_rst_stream_while_bytes_in_flight(
         } else if cfg!(feature = "openssl") {
             Ok(13964)
         } else {
-            Ok(13878)
+            Ok(15030)
```
cjpatton: Reviewer note: Size difference is due to the server sending back an X25519MLKEM768 key share:

```
15030 - 13878 = 1152 = 1120 /* MLKEM768 ciphertext + X25519 key share */ + 32 /* ? */
```

I'm not sure what the extra 32 bytes are for. Could be QUIC specific.

quiche/src/tests.rs
```diff
@@ -6450,7 +6468,7 @@ fn client_rst_stream_while_bytes_in_flight(
             if cfg!(feature = "openssl") {
                 27928
             } else {
-                27756
+                30060
```
cjpatton: I'm not sure what's being counted here. A full handshake would add at least 2336 to the handshake: 1184+32 for the client key share and 1088+32 for the server key share.

quiche/src/tests.rs
cjpatton: Note: the byte count doesn't change for OpenSSL because the current version (3.1) doesn't have PQ support. We could upgrade OpenSSL to 3.5 and enable PQ for it.

.github/workflows/stable.yml
cjpatton, line 1: Reviewer note: compare to boring's CI setup for windows:
https://github.com/cloudflare/boring/blob/ac3469371c4da185da51647a54687edc3260be48/.github/workflows/ci.yml#L286

quiche/src/tests.rs
```diff
@@ -6793,16 +6793,22 @@ fn validate_peer_sent_ack_range(
...
-    // to the classical case, which is why these counts are one higher than
-    // they would be for a single-Initial ClientHello.
-    let expected_max_active_pkt_sent = 4;
+    // keyshares enabled by our BoringSSL build) the handshake exchanges
+    // one extra packet per side compared to the classical case, so these
+    // counts are one higher than they would be for a single-Initial
+    // ClientHello. The `openssl` build doesn't use PQ keyshares.
+    let extra = if cfg!(feature = "openssl") { 0 } else { 1 };
+
+    let expected_max_active_pkt_sent = 3 + extra;
```
cjpatton: Reviewer note: An alternative might be to bump OpenSSL 3.1 -> 3.5, which has support for PQ. I'm not sure if it's enabled by default.

quiche/src/build.rs
```diff
@@ -215,19 +232,166 @@ fn target_dir_path() -> std::path::PathBuf {
...
+            std::fs::copy(&src_path, &dst_path).unwrap_or_else(|e| {
+                panic!(
+                    "copy {} -> {}: {e}",
+                    src_path.display(),
+                    dst_path.display()
+                )
+            });
+        }
+    }
+}
```
cjpatton: AI: Claude preferred to hand roll this function rather than use a crate. It's possible we can use `cpydir` here instead.

quiche/src/build.rs
```diff
@@ -215,19 +232,166 @@ fn target_dir_path() -> std::path::PathBuf {
...
+        );
+    }
+
+    let out = std::process::Command::new("git")
+        .arg("apply")
+        .arg("--verbose")
+        .arg(&patch)
+        .current_dir(&patched)
+        .output()
+        .expect("running `git apply` for boring-pq.patch");
```
cjpatton: AI: Claude preferred to use `git` to apply the patch here. In the `boring` crate, we just use `apply`:
https://github.com/cloudflare/boring/blob/ac3469371c4da185da51647a54687edc3260be48/boring-sys/build/main.rs#L483

I don't have a strong preference. Ideally we use as much of `boring-sys`'s build system as we can, but ultimately we still need to copy-paste it here.

Cross.toml
cjpatton, line 1: AI: This strikes me as a bit hacky and hard to maintain. Claude explored several options for getting cross compiling to work. From the perspective of the build, the least invasive we looked at was emulating the target with QEMU and calling `cargo` instead of `cross` to run it. However, many of the tests are sensitive to runtime: tests encode a microseconds measurement in a QUIC packet; since varint is used for the encoding, the packet size varies depending on the actual measurement. QEMU emulation seems to slow things down enough that the packet size actually changes.

quiche/src/tests.rs
cjpatton: Question for reviewers: Do we need a way to disable PQ by default? This PR uses set_curves_list() to control curve preferences in tests; we could also expose it in the public API if we wanted users to be able to disable. Otherwise, with the `boring` backend, it's already possible to disable via the `SSL_CTX` and `SSL` builder APIs.

quiche/patches/boring-pq.patch
```diff
@@ -0,0 +1,706 @@
+From cb5689e091f515fc8a42ceaff08d702333e505ed Mon Sep 17 00:00:00 2001
```
ghedo: TBH I don't think duplicating the boring patch here is the way forward.

Incidentally we've been working on removing the vendored BoringSSL from this repo and default to boring crate in https://github.com/cloudflare/quiche/pull/2080 which is currently blocked due to some arch-specific build failures which https://github.com/cloudflare/quiche/pull/2142 was meant to solve, but that in turn is blocked on test failures that the new change causes 🙈 

Unfortunately I gaven't had time to spend on that in a while, but if it's something you have cycle for I think getting that completed would significantly simplify this PR as well.

quiche/src/lib.rs
```diff
@@ -759,6 +759,16 @@ impl Config {
         self.tls_ctx.load_verify_locations_from_directory(dir)
     }
 
+    /// Configures the TLS curve preference list (colon-separated, e.g.
+    /// `"X25519:P-256:P-384"`). Corresponds to `SSL_CTX_set1_curves_list`.
+    ///
+    /// Only used from [`test_utils`] helpers; gated on the same cfg so the
```
ghedo: This should just be a public API IMO, it seems like it would be useful in general.

Cross.toml
ghedo: Also see https://github.com/cloudflare/quiche/pull/2142, per https://github.com/cloudflare/quiche/pull/2446/changes#r3152758201 I'd rather get that other PR working so that we can drop the vendored BoringSSL completely.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ publish = false
 [workspace.dependencies]
 anyhow = { version = "1" }
 assert_matches = { version = "1" }
-boring = { version = "4.3" }
+boring = { version = "5.1.0" }
```
ghedo: I deally we'd allow both v4 and v5 (e.g. `version = ">=4.21,<6"`) to make migration easier, and avoid blocking any application still on v4 from updating quiche until the full chain of dependencies is updated too.

I guess for RPK we'd need to provide both implementations based on which boring version is used, which I think would be fine in the short term, any other missing APIs?

ghedo (changes requested):

Cargo.toml
cjpatton: Forcing downstream applications to bump to boring 5 before than can take this change to quiche seems like a way easier path. Otherwise we'll end up with a lot of hacks in cruft in quiche that will make maintenance even more complex than it already is.

quiche/src/tests.rs
Lekensteyn: OpenSSL 3.5 sends X25519MLKEM768 and X25519 keyshares by default.

Its preference order is:

    ?*X25519MLKEM768 / ?*X25519:?secp256r1 / ?X448:?secp384r1:?secp521r1 / ?ffdhe2048:?ffdhe3072

Cargo.toml
Lekensteyn: What would dual-support of boring v4+v5 in quiche look like? In my experience, maintaining support for both the old and new version makes it more likely for downstream users to adopt new version over time, even if it would require more initial work upfront from upstream maintainers.

quiche/patches/boring-pq.patch
cjpatton: It looks like #2142 is meant to solve a similar problem we encountered here. It looks like you're using QEMU there. We gave this a try, and what we found is that emulation slows things down so much that it actually causes tests to flake. See commit messages for details:
https://github.com/cloudflare/quiche/pull/2446/changes/a27e8ac3f9b8e5f7f3a552813b1abdda8bff68fb

What we _could_ do is adjust the test expectations to account for varint's sensitivity to measured timings.

I'd be happy to drop vendored boringSSL as part of this PR.

quiche/src/lib.rs
cjpatton: Done: https://github.com/cloudflare/quiche/pull/2446/commits/492f8ab9422594e7938d4185a6248783c2231fc2

quiche/src/tests.rs
cjpatton: We've added set_curves_list() to the public API: https://github.com/cloudflare/quiche/commit/492f8ab9422594e7938d4185a6248783c2231fc2

quiche/patches/boring-pq.patch
cjpatton: I've tacked on a commit that drops the vendored boringSSL: 974059a02413dacdf6c20845316f535d89625000

Cross.toml
cjpatton: Resolving this since it's a duplicate of https://github.com/cloudflare/quiche/pull/2446#discussion_r3152758201

quiche/patches/boring-pq.patch
cjpatton: Note: I don't have write permission for this repo, so I can't push to #2080.

.github/workflows/stable.yml
```diff
@@ -93,11 +94,21 @@ jobs:
       #   run: cargo package --verbose --workspace --exclude=quiche_apps --allow-dirty
 
       - name: Run cargo clippy
```
antoniovicente, line 77: Duplicated CI step name
Should the name mention boring or the next clippy run mention ssl?

.github/workflows/stable.yml
```diff
@@ -93,11 +94,21 @@ jobs:
...
 
+      - name: Run cargo clippy
+        if: ${{ matrix.tls-feature != 'boringssl-boring-crate' }}
+        run: cargo clippy --no-default-features --features=${{ matrix.tls-feature }} ${{ env.NO_BORING_OPTIONS }} -- -D warnings
+
       - name: Run cargo clippy on examples
+        if: ${{ matrix.tls-feature == 'boringssl-boring-crate' }}
         run: cargo clippy --examples --features=${{ matrix.tls-feature }} ${{ env.DEFAULT_OPTIONS }} -- -D warnings
 
+      - name: Run cargo clippy on examples
```
antoniovicente, line 80: Similar comment about duplicated CI step names

quiche/src/tests.rs
antoniovicente: Larger handshake with PQ would move more bytes back and forth and increase the cwnd.  This test was for a fix to tx_buffered where we were not updating this value after requests were cancelled by the client.

Cargo.toml
antoniovicente: I'm guessing that moving away from the submodule is a possible pre-req.

Could support for boring 5.1 be built as a feature?

cjpatton: @antoniovicente do you mean a feature for quiche? I think it would be possible, but it we would have to add a lot of exceptions throughout the code and build infra. Off the top of my head:

1. The RPK api has changed from 4 to 5. Shouldn't be too hard to support both.
2. Many tests are sensitive to packet sizes and counts. These differ between 4 and 5 because PQ is enabled by default in the latter. We would either (a) set the test expectations based on which version of boring is used or (b) disable PQ in all tests that are sensitive to this. There are perhaps 10s of such test, so neither option would be too bad.
3. The prerequisites for building BoringSSL have changed from 4 to 5. It's perhaps the case that the prerequisites for 5 are sufficient for 4, but I haven't tested this. I'm not an expert here, but I'm worried about the maintenance burden this could add to quiche.

I can give this a try if we really really need it. My intuition is that this would add a lot of complexity that won't be worth it in the end, but I'm open to being wrong.

.github/workflows/stable.yml
cjpatton: Good catch! I went with boring here. Likewise for "run cargo clippy on examples" below.

Cargo.toml
antoniovicente: I was just asking how to implement support for multiple versions of boring suggested by others.

Seems like part of the problem is that boring is a leaky abstraction.

quiche/patches/boring-pq.patch
antoniovicente: Could you create a clone PR based on #2080 or suggest a merge to the branch used by #2080?

ghedo: I created https://github.com/cloudflare/quiche/pull/2457 with the cross image change, https://github.com/cloudflare/quiche/pull/2458 to remove the OpenSSL backend, and updated https://github.com/cloudflare/quiche/pull/2080 (which now passes CI).

It'd be better to get those merged first, since this PR is quite large, but I think we'll still want something like https://github.com/cloudflare/quiche/pull/2446#discussion_r3171196179 as well.

ghedo: These are now merged, so this PR needs rebase (a few commits can be dropped as well).

quiche/src/tests.rs
cjpatton: the openssl backend has been removed on master, so this is no longer relevant.

quiche/src/build.rs
cjpatton: The vendored-boringssl feature has been removed on master, so this is no longer relevant.

Cargo.toml
cjpatton: > Seems like part of the problem is that boring is a leaky abstraction.

Yup, that's far. The build system is subject to change as upstream BoringSSL makes changes to their build; That's part of the reason why internally we use a pre-compiled BoringSSL. But if you're consuming `boring` with its default feature set, then you may have to update your build dependencies from time to time.

quiche/patches/boring-pq.patch
cjpatton: Thanks @ghedo! I've rebased this PR and was able to drop a few commits relevant to them. I tacked on a couple of minor fixes needed for boring 5 here: https://github.com/cloudflare/quiche/pull/2446/commits/7843a7196c77ca628cdabc1b949093fc57690dae

quiche/src/tests.rs
```diff
@@ -636,7 +636,9 @@ fn handshake_0rtt(
 ) {
     let mut buf = [0; 65535];
 
-    let mut config = Config::new(PROTOCOL_VERSION).unwrap();
+    // Test assumes the server transitions to early-data state after a
+    // single `server_recv`, which requires a single-Initial ClientHello.
+    let mut config = test_utils::config_no_pq(PROTOCOL_VERSION).unwrap();
```
antoniovicente, line 664: It would be good to add 0-RTT tests that operate on post-quantum handshakes.  How difficult would it be to parametrize this test to cover both cases?

antoniovicente (approved):

Cargo.toml
antoniovicente: @ghedo tells me that backwards compat with boring 4 is a hard requirement.  We need smoother migration to boring 5.

You two should talk.

cjpatton: We've already talked, and I'm exploring options.

.github/workflows/nightly.yml
```diff
@@ -7,7 +7,7 @@ permissions:
   pull-requests: write
 
 env:
-  FEATURES: "async,ffi,qlog"
+  FEATURES: "async,ffi,qlog,rpk"
```
cjpatton, line 10: There are differences between boring 4 and 5 in how RPK is implemented, so make sure we're exercising this feature in the build. Likewise for stable.

quiche/src/tests.rs
```diff
@@ -32,6 +32,28 @@ use crate::Header;
...
+        #[cfg(boring_v4)]
+        {
+            $b4
+        }
+        #[cfg(not(boring_v4))]
+        {
+            $b5
+        }
+    }};
+}
```
cjpatton, line 56: It would be nice if we could avoid this by enabling PQ by default in boring 4.x, but unfortunately we don't control the feature flags with which boring is configured.

Cargo.toml
cjpatton: Alrighty, I believe I got this working! 05c6bd35f0e086e2119fd4c5a3c6836597645b85

The "boring5" feature idea didn't pan out: boring-sys doesn't allow you to have two versions of itself in the same dependency tree. This makes sense since the version of BoringSSL between each might be different, and my understanding is that this would provide the linker with multiple implementations of the same symbol.

@ghedo's initial idea ended up working out. The strategy is to detect which version of boring is in the Cargo.lock at build time and emit `cfg(boring_v4)` if boring 4 is detected. This allows us to handle to handle the RPK and PQ enablement differences at compile time.

The script for detecting the boring version is a little hacky since it's just parsing the Cargo.lock. I've tested this with OXY and it seems to work just fine. The draft MR is linked from the internal ticket.

cjpatton: I've added a few inline comments for things for reviewers to double check.

quiche/src/tests.rs
cjpatton: This test is really hard coded around Client Hello fitting in the initial packet. We could write a different test that exercises 0-RTT, but it would work differently. Want me to add that?

cjpatton:
Rebased, ready for review once more.

quiche/src/build.rs
```diff
@@ -30,6 +30,83 @@ Cflags: -I${{includedir}}
...
 
+/// Returns true if cargo resolved `boring` to a 4.x version.
+///
+/// Walks up from `OUT_DIR` looking for a `Cargo.lock`, then scans it
+/// for the `boring` package. We use `Cargo.lock` rather than shelling
+/// out to `cargo metadata` because (a) the lockfile is guaranteed to
+/// exist at this point in the build, (b) parsing it is cheap and has
+/// no extra dependencies, and (c) it avoids re-entering cargo from a
+/// build script.
+fn detect_boring_v4() -> bool {
```
ghedo: Have you considered using `build-info-build` / `build-info` for this? https://docs.rs/build-info-build/latest/build_info_build/struct.BuildScriptOptions.html#method.collect_runtime_dependencies

Seems like we could limit this to `build-dependency` with no default features (to avoid the git dependency).

ghedo: Actually nvm, I don't think this would work since it just looks at cargo metadata.

ghedo: Another idea 😅  Could we check `boring_sys::BORINGSSL_API_VERSION` during build? It looks like v4 and v5 use different versions:
https://docs.rs/boring-sys/4.22.0/boring_sys/constant.BORINGSSL_API_VERSION.html
https://docs.rs/boring-sys/5.1.0/boring_sys/constant.BORINGSSL_API_VERSION.html

and since I don't think we'll ever update BoringSSL in v4 again `BORINGSSL_API_VERSION <= 21`.

Would require adding `boring-sys` as build dependency as well (optional in quiche and gated on `boringssl-boring-crate`), but not entirely sure it would work properly...

quiche/src/build.rs
```diff
@@ -44,7 +121,18 @@ fn target_dir_path() -> std::path::PathBuf {
...
+    // Emit `cfg(boring_v4)` if boring version 4.x is detected. This is used to
+    // pick which APIs to expect and to guide test expectations. (Larger post
+    // quantum key shares are enabled by default in boring 5.x but not boring
+    // 4.x.)
+    //
+    // The cfg is always registered (even when the backend feature is
+    // off) so rustc doesn't warn about unknown cfg names.
+    println!("cargo::rustc-check-cfg=cfg(boring_v4)");
     if cfg!(feature = "boringssl-boring-crate") {
+        if detect_boring_v4() {
```
ghedo: To avoid this being a breaking change, I think it'd be better to assume v4 by default, and check for v5. quiche is also used as C library so there isn't necessarily a Cargo.lock to check.

quiche/src/build.rs
cjpatton: What about differences in the `boring` API itself? E.g., in v4 the RPK API is different than in v5. While it's true that BORINGSSL_API_VERSION is sufficient for detecting differences between v4 and v5, this may not be true for v6, or even v5.2.x. It seems like detecting the crate version is strictly more robust, since it always has the information that we need.

quiche/src/build.rs
cjpatton: Done: eff1f9db818f5ec05eb00413b0799e048fdcd49e

cjpatton:
Rebased.

quiche/src/build.rs
ghedo: TBH I don't think we'd need something like this for future updates. v4 -> v5 is somewhat special because it's such a big jump, but once internal users are updated we can drop the v4 support, and for future v6 or whatever I think we could just do a clean cut rather than support multiple versions again.

cjpatton: I feel like we've had a similar conversation about v3 to v4 :)

In any case, what is the argument for using `BORINGSSL_API_VERSION` to detect the version rather than parsing the `Cargo.lock`?

cjpatton: Wayyyy nicer solution, proposed by @kornelski:
https://github.com/cloudflare/boring/pull/506

cjpatton: Thinking about this more, there is a gap: The new env var doesn't appear in 5.1, so for this to work we'll have to cut a release of 5.2 first. It also requires adding boring-sys to the workspace dependencies, which isn't great but not terrible.

We could either:
1. Merge as-is (quiche is compatible with 5.1)
2. Release boring 5.2 and then update this PR

ryanglobus-A:
Would love to see this land. Is approval/merging in the new couple weeks realistic, or should we plan a workaround for now? Just trying to gauge timing. Thanks!

cjpatton:
> Would love to see this land. Is approval/merging in the new couple weeks realistic, or should we plan a workaround for now? Just trying to gauge timing. Thanks!

I'm just waiting for boring 5.2 to land in crates.io, then I will ping the team to get this merged :)

ryanglobus-A:
> > Would love to see this land. Is approval/merging in the new couple weeks realistic, or should we plan a workaround for now? Just trying to gauge timing. Thanks!
> 
> I'm just waiting for boring 5.2 to land in crates.io, then I will ping the team to get this merged :)

Awesome, thanks!

quiche/src/tests.rs
```diff
@@ -6463,7 +6504,7 @@ fn client_rst_stream_while_bytes_in_flight(
         if cc_algorithm_name == "cubic" {
             Ok(12000)
         } else {
-            Ok(13878)
+            Ok(by_boring!(b4: 13878, b5: 15030))
```
dingxiangfei2009, line 6508: Hi `quiche`! I am genuinely curious about how these numbers were derived and whether there is a genuine concern so that we want to put assertions on the size of sent bytes.

Is this about asserting a certain property of congestion control? In that case, wouldn't this number be sensitive to how many bytes goes into the TLS records? I am afraid that this may vary depending on how protocol defaults are configured across different library versions. cc @davidben

dingxiangfei2009: cc @divverent who filed #2553.

cjpatton: The difference here is because PQ key shares are enabled by default in `boring` version 5:
15030 - 13878 = 1152 = 1088 (MLKEM768 ciphertext) + 32 (X25519 key share) + a few extra bytes for TLS negotiation.

davidben: Is it actually worth asserting on them? Over at Google, we tend to call these sorts of things "change detector" tests. They're generally a lot of toil and not very useful. In a monorepo, they're already tedious. Across different projects versioned separately, they're real disruptive.

Ideally assertions would only trip in response to some bug, not just because _something_ changed.

cjpatton: I don't disagree, but reverting the existing tests is beyond the scope of this PR.

davidben: Fair enough! A suggestion for some follow-up changes. :-)

quiche/src/build.rs
cjpatton: Took Option 2. We should be good to go here.

quiche/src/tests.rs
cjpatton: Thanks for the feedback guys! Tracking changes to test here: #2579.

mhrobinson:
@cjpatton, with 5.2 in crates.io now, is this good to merge?

normanmaurer:
We also need this on the netty side as the currently used BoringSSL version is too low for us. Without this change we can't upgrade and get the required security fixes.

cjpatton:
Hi gang, I was out on PTO for two weeks, sorry for the radio silence! I'll rebase and follow up with @ghedo for approval

ghedo (approved):
Squashed and re-ordered some of the commits to make future bisects work, otherwise LGTM.

ghedo:
Regarding the missing "quiche" CI build, this is expected as the name changed. I'll update the repo configuration once this is ready to be merged.

jrouviere (approved):

ghedo:
Merged :tada:, thanks @cjpatton for the work (and sorry it took so long)!

cjpatton:
yw!

normanmaurer:
Thanks a lot <3

PR #2449, opened 2026-04-25
packet: never emit 1-byte truncated packet numbers

### LINKED ISSUE #2448: pkt_num_len can return 1 byte, leading to AEAD decryption failures under packet reorder
### Summary

`quiche/src/packet.rs::pkt_num_len` will choose a 1-byte truncated packet number when fewer than 128 packets are unacked. RFC 9000 §17.1 permits this, but it is unsafe in practice: if the receiver observes more than 128 packets reordered, the entire valid range of a 1-byte truncation collapses, and the decoded full packet number lands on the wrong candidate. The AEAD nonce is derived from the full packet number, so the receiver fails to decrypt an otherwise-good packet.

### Source

[`quiche/src/packet.rs::pkt_num_len`](https://github.com/cloudflare/quiche/blob/master/quiche/src/packet.rs#L569-L575):

```rust
pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
    let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
    let min_bits = u64::BITS - num_unacked.leading_zeros() + 1;
    min_bits.div_ceil(8) as usize     // returns 1 when num_unacked < 128
}
```

The reference Go implementation, [quic-go](https://github.com/quic-go/quic-go), refuses 1-byte truncation for exactly this reason. From [`internal/protocol/packet_number.go#L41-L42`](https://github.com/quic-go/quic-go/blob/master/internal/protocol/packet_number.go#L41-L42):

> // it never chooses a PacketNumberLen of 1 byte, since this is too short under certain circumstances

### Field observation

Hit in production at https://viatrail.app — Cloudflare R2-served HTTP/3 tile fetches stalling on mobile. qlog shows clusters of `packet_dropped` with `trigger=payload_decrypt_error` on 1-RTT packets, with decoded PN offset from actual PN by exactly 256·*k* — the precise signature of 1-byte truncation ambiguity.

### Reproducer

https://github.com/dave/quiche-bug — Docker-based server (stock or patched via `--build-arg APPLY_PATCH=0|1`) plus Go HTTP/3 client using quic-go.

| Server | AEAD failures (90s, 8 workers) |
|---|---|
| Stock | **6** |
| With `pkt_num_len.max(2)` | **0** |

### Proposed fix

Floor `pkt_num_len` at 2 bytes (1-line change). PR to follow.

Cost: 1 byte per outgoing 1-RTT packet during the brief window where `num_unacked < 128`. quic-go has shipped this trade-off since 2017.

dave:
Fixes #2448

### What

Floor [`pkt_num_len`](https://github.com/cloudflare/quiche/blob/master/quiche/src/packet.rs#L569-L575) at 2 bytes. Previously returned 1 when `num_unacked < 128`.

```diff
 pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
     let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
     // computes ceil of num_unacked.log2() + 1
     let min_bits = u64::BITS - num_unacked.leading_zeros() + 1;
-    min_bits.div_ceil(8) as usize
+    (min_bits.div_ceil(8) as usize).max(2)
 }
```

(commit also expands the inline comment to explain the floor, and updates the existing `pkt_num_encode_decode` test, which asserted the 1-byte case.)

### Why

1-byte truncation is unambiguous only with ≤128-packet reorder windows. Any path that produces >128-packet reordering — common on mobile data, behind some VPN egress, anywhere pacing meets aggregator hardware — collapses the decode space, AEAD nonce is wrong, decryption fails on the otherwise-correct ciphertext.

quic-go refuses for the same reason — [`PacketNumberLengthForHeader`](https://github.com/quic-go/quic-go/blob/master/internal/protocol/packet_number.go#L41-L42).

### Reproducer

https://github.com/dave/quiche-bug — same hardware, same client, 90s:

| Server | AEAD failures |
|---|---|
| Stock | 6 |
| Patched (this PR) | 0 |

### Cost

1 extra byte per 1-RTT packet during the brief window where `num_unacked < 128`. Steady-state high-throughput connections see zero impact; post-ACK-burst conditions see at most 1 byte per packet for a small number of packets.

quic-go has shipped this floor since 2017 with no field regressions.

### Test changes

`pkt_num_encode_decode` previously asserted `pkt_num_len(0, 0) == 1` with a roundtrip loop branched on the result. Updated: asserts `== 2`, roundtrip loop simplified to test 2 bytes uniformly.

### DIFF
--- a/quiche/src/packet.rs
+++ b/quiche/src/packet.rs
@@ -570,8 +570,15 @@ pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
     let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
     // computes ceil of num_unacked.log2() + 1
     let min_bits = u64::BITS - num_unacked.leading_zeros() + 1;
-    // get the num len in bytes
-    min_bits.div_ceil(8) as usize
+    // get the num len in bytes; floor at 2 to avoid the 1-byte
+    // truncation ambiguity that breaks AEAD decryption when the
+    // receiver sees more than 128 packets reordered (the entire
+    // valid range of a 1-byte truncation collapses, so the decoder
+    // cannot recover the full packet number unambiguously). The
+    // reference Go implementation, quic-go, refuses 1-byte truncation
+    // for the same reason — see PacketNumberLengthForHeader in
+    // quic-go/internal/protocol/packet_number.go.
+    (min_bits.div_ceil(8) as usize).max(2)
 }
 
 pub fn decrypt_hdr(
@@ -1356,8 +1363,11 @@ mod tests {
 
     #[test]
     fn pkt_num_encode_decode() {
+        // pkt_num_len floors at 2: the function never emits a 1-byte
+        // truncated packet number, so the smallest unacked-distance
+        // encoding still uses 2 bytes.
         let num_len = pkt_num_len(0, 0);
-        assert_eq!(num_len, 1);
+        assert_eq!(num_len, 2);
         let pn = decode_pkt_num(0xa82f30ea, 0x9b32, 2);
         assert_eq!(pn, 0xa82f9b32);
         let mut d = [0; 10];
@@ -1380,19 +1390,14 @@ mod tests {
         let hdr_num = u64::from(b.get_u24().unwrap());
         let pn = decode_pkt_num(0xace9fa, hdr_num, num_len);
         assert_eq!(pn, 0xace9fe);
-        // roundtrip
+        // roundtrip — every result is at least 2 bytes
         let base = 0xdeadbeef;
         for i in 1..255 {
             let pn = base + i;
             let num_len = pkt_num_len(pn, base);
-            if num_len == 1 {
-                let decoded = decode_pkt_num(base, pn & 0xff, num_len);
-                assert_eq!(decoded, pn);
-            } else {
-                assert_eq!(num_len, 2);
-                let decoded = decode_pkt_num(base, pn & 0xffff, num_len);
-                assert_eq!(decoded, pn);
-            }
+            assert_eq!(num_len, 2);
+            let decoded = decode_pkt_num(base, pn & 0xffff, num_len);
+            assert_eq!(decoded, pn);
         }
     }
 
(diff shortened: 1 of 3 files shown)


dave:
@ghedo Sorry — I just realised there were failing tests on this. I've rebased onto master and fixed them (green locally now). Could you re-approve the CI run when you get a chance?

quiche/src/packet.rs
```diff
@@ -1380,19 +1390,14 @@ mod tests {
...
-            if num_len == 1 {
-                let decoded = decode_pkt_num(base, pn & 0xff, num_len);
-                assert_eq!(decoded, pn);
-            } else {
-                assert_eq!(num_len, 2);
-                let decoded = decode_pkt_num(base, pn & 0xffff, num_len);
-                assert_eq!(decoded, pn);
-            }
+            assert_eq!(num_len, 2);
+            let decoded = decode_pkt_num(base, pn & 0xffff, num_len);
```
ghedo, line 1399: Please preserve an explicit one-byte decode case. This change only prevents emitting one-byte packet numbers; peers may still legally send them. The removed branch was the only test covering `decode_pkt_num(..., 1)`. A single hard-coded assertion before this loop would retain that coverage.

quiche/src/packet.rs
```diff
@@ -570,8 +570,15 @@ pub fn pkt_num_len(pn: u64, largest_acked: u64) -> usize {
     let num_unacked: u64 = pn.saturating_sub(largest_acked) + 1;
     // computes ceil of num_unacked.log2() + 1
     let min_bits = u64::BITS - num_unacked.leading_zeros() + 1;
-    // get the num len in bytes
-    min_bits.div_ceil(8) as usize
+    // get the num len in bytes; floor at 2 to avoid the 1-byte
```
ghedo, line 573: This comment is disproportionate to the implementation and duplicates the PR description. Could it be reduced to the local rationale, for example: `Avoid one-byte packet numbers, which become ambiguous after deep reordering.`

ghedo (changes requested):
Thanks for the detailed report, reproducer, and focused patch. Please rebase this branch onto the current `master` and resolve the test expectation conflicts. While doing so, please preserve the one-byte receive/decode coverage called out inline. The comment-length note is a small cleanup that would keep the rationale local and concise.

PR #2453, opened 2026-04-28
BBR: Check for app-limited when entering a new iteration of the send loop

antoniovicente:
The new logic matches the app-limited detection algorithm described in
the BBR RFC draft, which involves checking for the previous state of the
connection before timeout and ack processing.

This change detects the end of the current iteration of the send loop by
looking at the result of send_single and notifying the congestion controller
when send stopped early with Err::Done due to either there being no data
left to send or cwnd being fully utilized.

By automatically detecting the end of the current iteration we can avoid 
making changes to the quiche Connection's public API.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -573,6 +573,7 @@ pub struct Config {
     grease: bool,
 
     cc_algorithm: CongestionControlAlgorithm,
+    enable_bbr_app_limited_fix: bool,
     custom_bbr_params: Option<BbrParams>,
     initial_congestion_window_packets: usize,
     enable_relaxed_loss_threshold: bool,
@@ -658,6 +659,7 @@ impl Config {
             application_protos: Vec::new(),
             grease: true,
             cc_algorithm: CongestionControlAlgorithm::CUBIC,
+            enable_bbr_app_limited_fix: false,
             custom_bbr_params: None,
             initial_congestion_window_packets:
                 DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS,
@@ -1084,6 +1086,15 @@ impl Config {
     /// The default value is `CongestionControlAlgorithm::CUBIC`.
     pub fn set_cc_algorithm(&mut self, algo: CongestionControlAlgorithm) {
         self.cc_algorithm = algo;
+    }
+
+    /// Enable a rework of how the BBR congestion control implementation
+    /// computes app-limited.  This config parameter will be removed after we
+    /// build confidence on the new implementation or decide to roll it back.
+    ///
+    /// Defaults to `false`.
+    pub fn set_enable_bbr_app_limited_fix(&mut self, value: bool) {
+        self.enable_bbr_app_limited_fix = value;
     }
 
     /// Sets custom BBR settings.
@@ -2509,6 +2520,32 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Enable a rework of how the BBR congestion control implementation
+    /// computes app-limited.  This config parameter will be removed after we
+    /// build confidence on the new implementation or decide to roll it back.
+    ///
+    /// Currently this only applies if cc_algorithm is
+    /// `CongestionControlAlgorithm::Bbr2Gcongestion`.
+    ///
+    /// This function can only be called inside one of BoringSSL's handshake
+    /// callbacks, before any packet has been sent. Calling this function any
+    /// other time will have no effect.
+    ///
+    /// See [`Config::set_enable_bbr_app_limited_fix()`].
+    ///
+    /// [`Config::set_enable_bbr_app_limited_fix()`]: struct.Config.html#method.set_enable_bbr_app_limited_fix
+    #[cfg(feature = "boringssl-boring-crate")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "boringssl-boring-crate")))]
+    pub fn set_enable_bbr_app_limited_fix_in_handshake(
+        ssl: &mut boring::ssl::SslRef, value: bool,
+    ) -> Result<()> {
+        let ex_data = tls::ExData::from_ssl_ref(ssl).ok_or(Error::TlsFail)?;
+
+        ex_data.recovery_config.enable_bbr_app_limited_fix = value;
+
+        Ok(())
+    }
+
     /// Sets the congestion control algorithm used by string.
     ///
     /// This function can only be called inside one of BoringSSL's handshake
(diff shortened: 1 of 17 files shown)


quiche/src/lib.rs
```diff
@@ -2810,6 +2847,12 @@ impl<F: BufFactory> Connection<F> {
         Ok(())
     }
 
+    /// Returns true if at least 1 stream has headers or body data to
+    /// write, or there are items in the DATAGRAM send queue.
```
LPardue, line 2851: nit: QUIC treats stream bytes as opaque data. It doesn't matter what the bytes represent at the transport layer.

```suggestion
    /// Returns true if at least 1 stream has data to
    /// write, or there are items in the DATAGRAM send queue.
```

quiche/src/recovery/mod.rs
```diff
@@ -319,6 +321,17 @@ pub trait RecoveryOps {
...
+    // Called if the send loop stopped early because send_single
+    // returned Err::Done; this indicates that it is time to yield
+    // because either there is no cwnd remaining or there is no data
+    // left to send.
+    fn send_stopped_early(&mut self, has_flushable_data: bool);
+
+    // Allow the BBR implementation to compute if the congestion
+    // controller is app-limited, before processing timeouts, ACKs or
+    // generating packets to send.
+    fn bbr_check_if_app_limited(&mut self, now: &Instant);
```
LPardue, line 334: nit: its not 100% clear to me what this does at a high level. The &mut indicates it has a side effect, is that an important detail? I.e. should you say something like "if the check determines it is app-limited then the controller marks itself as app-limited" ?

quiche/src/tests.rs
```diff
@@ -6236,9 +6236,11 @@ fn stream_data_blocked_unblocked_flow_control(
 fn app_limited_true(
     #[values("cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
     #[values(true, false)] discard: bool,
+    #[values(false, true)] enable_bbr_app_limited_fix: bool,
```
LPardue, line 6239: nit: for consistency with the prior line the order can be flipped. I don't really care much though


```suggestion
    #[values(true, false)] enable_bbr_app_limited_fix: bool,
```

quiche/src/recovery/mod.rs
```diff
@@ -319,6 +321,17 @@ pub trait RecoveryOps {
     fn get_next_release_time(&self) -> ReleaseDecision;
 
     fn gcongestion_enabled(&self) -> bool;
+
+    // Called if the send loop stopped early because send_single
+    // returned Err::Done; this indicates that it is time to yield
+    // because either there is no cwnd remaining or there is no data
+    // left to send.
+    fn send_stopped_early(&mut self, has_flushable_data: bool);
```
LPardue, line 329: nit: perhaps worth documenting what `has_flushable_data` represents here, its just flushable application data right?

LPardue (approved):
LGTM modulo some nits.

I didn't go into all the BBR low-level details though

estcarisimo (approved):

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1134,6 +1144,58 @@ impl RecoveryOps for GRecovery {
...
+        // The check for C.pending_transmissions == 0 is done by
+        // checking !pacer_has_pending_transmissions by comparing the
+        // pacer's next release time vs 'now'.
+        //
+        // The case of Immediate next_release_time results in the work_loop
+        // iterating until min(cwnd, tx buffer) is exhausted.  If we decide to
+        // exit the send loop earlier we should mark the connection in some way so
+        // the next call to bbr_check_if_app_limited does not mark the
+        // connection app-limited after yielding when C.pending_transmissions > 0.
+        if self.cwnd() > self.bytes_in_flight.get() + frame::MAX_STREAM_OVERHEAD &&
```
gregor-cf, line 1183: I think you need to actually check for `self.cwnd() > bytes_in_flight + max_udp_payload_len`
For datagrams the under-utilization can be up to a full packet. For streams,  you still need to account at least for `packet_header + MAX_STREAM_OVERHEAD`

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1134,6 +1144,58 @@ impl RecoveryOps for GRecovery {
...
+            !pacer_has_pending_transmissions(
```
gregor-cf, line 1184: why do we need to check the pacer here? 

And is `now` the right time to use? Should it be the time when `send_stopped_early()` was last called? 
Or rather, should `send_stopped_early()` be: 
```
  fn send_stopped_early(&mut self, has_flushable_data: bool, now: Instant) {
        if !has_flushable_data && !pacer_has_pending_transmissions(next_release_time, now) {
            self.check_for_app_limited_next_iteration = true;
        }
    }
```

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -1207,11 +1269,53 @@ impl std::fmt::Debug for GRecovery {
     }
 }
 
+fn pacer_has_pending_transmissions(
```
gregor-cf, line 1272: nit: I think you can replace this function with `next_release_time.time(now).is_none()`. 
(Or add a more descriptive function to `ReleaseDecision`, e.g., ReleaseDecision::can_release(now)`.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: We're getting into the differences between the BBR RFC and the older/expired delivery rate estimation draft.

https://www.ietf.org/archive/id/draft-cheng-iccrg-delivery-rate-estimation-01.html#section-3.4-3
vs
https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-05.html#section-4.1.2.4-3

I think we do need the precision in this case: calling the connection not congestion limited when remaining cwnd is less than 1 packet can cause problems when the cwnd is close to the 2 packet minimum; sending significantly less than the full cwnd won't result in a measurable BW increase and the connection will exit due to bandwidth plateau in case where it shouldn't.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: This attempts to mimic the `C.pending_transmissions == 0` check from https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-05.html#section-4.1.2.4-3

I think it tries to capture the next time when the connection is expected to be starved of data to send.  The connection could end up app limited if the attempt to resume send is delayed for some reason, which could include the machine being busy doing sends or handling events for other active connections.

And yes, I think the comparison is meant to be against the current time, not the time when the previous packet was sent.

quiche/src/recovery/gcongestion/recovery.rs
gregor-cf: (I'm not a BBR expert by any means, so take with a grain of salt).
I don't think we really capture NoUnsentData, because `send_single()` returning Error::Done could be because there is no unsent data, or there is something we wanted to send, but we are out of cwnd. I think we need to distinguish these two cases to get a reliable app_limited signal (*). 

> I think we do need the precision in this case: calling the connection not congestion limited when remaining cwnd is less than 1 packet can cause problems when the cwnd is close to the 2 packet minimum;

The check I'm proposing is "less than 1 packet remaining". 
I'm wondering if the 2 packet minimum will actually work well for MASQUE, since might only be able to send full-sized packets or nothing. Maybe we need a 3 or packet minimum. 

(*) in quiche we also need to be careful in what order we call `send()`, `recv()`, and the function to write to / read from streams. If these are called in the "wrong" order the send buffer could be empty in every iteration.

quiche/src/recovery/gcongestion/recovery.rs
gregor-cf: sounds good. (Looks like I can start a comment thread, by I can't resolve one :-/)

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: There being a required ordering between "the function to write to / read from streams" and `send()` is a problem.

And I think you're right in that there is an ordering requirement: recv, generate data, send

I think the current ordering is: recv, send, generate data!

gregor-cf: we do `recv, generate, send` in work_loop (`self.conn_stage.on_read(did_recv, qconn, ctx)?;` generates data). But I haven't thought through all possible interleavings that are possible with different branches of the `tokio::select` call finishing and/or timeout handling.... 

IMHO the fact that we can process a ton of packets (or writes) at once without interleaving network IO and app IO is also a potential concern.

Issue #2452, opened 2026-04-27
Control frames not marked as lost on old path after migration


birneee:
After migration, the server may have unacknowledged packets on the old path carrying control frames that are never retransmitted.

Neither retransmission mechanism covers these frames:

**ACK-based** loss detection does not fire on the old path. 
On the old path, `newly_acked` is always empty so `on_ack_received` returns early without calling `detect_lost_packets`.

**PTO-based** retransmission skips control frames. 
The PTO filter only selects packets where `has_data` is true, so control frames are excluded.

Affected frames are: `RESET_STREAM`, `STOP_SENDING`, `MAX_STREAM_DATA`, `MAX_DATA`, `MAX_STREAMS`, `RETIRE_CONNECTION_ID`, `NEW_CONNECTION_ID`

This can lead to **indefinite stalls**.

I will share my unit tests soon.

shahoxo:
Same failure mode here, but with STREAM data rather than control frames.

Reproduces on 0.29.3 (built from the tag) on Linux and Windows, and on 0.22.0 on macOS — identical result on all three. Server side, C FFI, passive migration only.

After the peer's address changes, data that was in flight on the old path is never retransmitted. One PTO probe's worth gets through and the rest is stranded, so the receiving stream keeps a permanent hole while the connection itself looks healthy.

Repro (server → client, through a UDP relay I control):

1. connect, do one round trip so nothing is outstanding
2. write 20 messages (~214 B each) to a unidirectional stream, flushing after each one, so 20 separate packets go on the wire
3. close the relay's server-facing socket and open a new one, dropping all 20 — this is what a CGNAT rebinding does
4. stay silent for 500 ms
5. the client sends one packet from the new address, triggering the migration

20 messages written, 2 delivered, on every platform and both versions. PathEvent::PeerMigrated fires and ACKs keep flowing; the sender sees nothing wrong.

Server-side packet trace (times in ms):

168.45-168.71  packet_sent pn=3..22   stream off=29,243,...,4095   20 packets, all lost
232.01         packet_sent pn=23      ack + stream off=29
369.87         packet_sent pn=24      ack + stream off=29
623.34         packet_sent pn=26      ack + stream off=29
672.81         packet_received pn=3   stream          <- client's first packet from the new address
672.84         packet_sent pn=28      path_challenge
695.23         packet_sent pn=29      ack + path_challenge + stream off=4309
1127.65        packet_sent pn=31      stream off=29 len=467   <- this one reaches the client
[...]

PR #2454, opened 2026-04-28
Fix control frames not marked as lost on old path after migration (#2452)

birneee:
unit tests for https://github.com/cloudflare/quiche/issues/2452


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -832,55 +832,53 @@ impl RecoveryOps for GRecovery {
...
-            prior_in_flight,
-            self.bytes_in_flight.get(),
-            now,
-            &self.newly_acked,
-            &self.lost_reuse,
-            self.epochs[epoch].least_unacked(),
-            &self.rtt_stats,
-            &mut self.recovery_stats,
-        );
+        if !self.newly_acked.is_empty() {
```
ghedo, line 867: When an ACK only covers packets sent on another path, `newly_acked` is empty but `detect_and_remove_lost_packets()` can populate `lost_reuse`. This guard then prevents BBR2 from processing those losses, including removing packet state from its bandwidth sampler. The timeout path explicitly calls `on_congestion_event()` with an empty ACK list for this reason. Please invoke it when either ACKs or losses exist, while keeping only the PTO reset conditional on `newly_acked`.

quiche/src/tests.rs
```diff
@@ -12926,3 +12926,337 @@ fn server_qlog() {
         panic!("expected Qlog event");
     }
 }
+
+/// Tests that frames are correctly retransmitted across paths
+fn do_after_migration_receive_retransmit(
```
ghedo, line 12931: This helper supports 10 frame tests across two recovery implementations and both migration values, producing 40 cases and 336 added lines. Existing tests already cover same-path retransmission for these frame types. Could this be reduced to an extension of `validate_ack_range_on_ack_received` for the empty-local-ACK case plus one representative end-to-end migration test using a control-only frame such as `RESET_STREAM`?

quiche/src/tests.rs
```diff
@@ -12926,3 +12926,337 @@ fn server_qlog() {
...
+#[rstest]
+fn after_migration_receive_retransmit_stream_data(
+    #[values("cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
+    #[values(true, false)] migrate: bool,
+) {
+    do_after_migration_receive_retransmit(
+        cc_algorithm_name,
+        migrate,
+        |_| {},
+        |pipe| assert_eq!(pipe.server.stream_send(1, b"hello", true), Ok(5)),
```
ghedo, line 13055: A single STREAM packet is already eligible for PTO because `has_data` is true, so this does not exercise the reported stream-data starvation. That failure occurs when more than `MAX_PTO_PROBES_COUNT` packets are outstanding and every PTO selects the same oldest packets. Please either remove this redundant case or send more than two separate packets and verify that data from the final packet is recovered.

ghedo (changes requested):
The underlying migration bug is valid and the general recovery change is appropriate. However, the gcongestion path currently drops loss-only congestion notifications, and the regression suite is substantially broader than needed. Please address the loss accounting and reduce the tests to focused coverage.

PR #2455, opened 2026-05-04
Fix leaked StreamCtx for bodyless H3 requests

csujedihy:
When the client sends a request without a body, the FIN is sent by `h3::Connection::send_request()` before the driver creates the corresponding stream context, `StreamCtx`. Thus, when the context is initialized, the fin bit (`fin_or_reset_sent`) will stay `false` permanently, which prevents the context from being removed because later on, `both_directions_done` returns false and `cleanup_stream` will not be called. So, these stream contexts will be leaked in the stream map.

This led to OOM in a simple test app that cycles short-lived H3 requests without a body.


antoniovicente (approved):

tokio-quiche/src/http3/driver/client.rs
```diff
@@ -214,6 +214,13 @@ impl ClientHooks {
         let (mut stream_ctx, send, recv) =
             StreamCtx::new(stream_id, STREAM_CAPACITY);
 
+        if body_finished {
+            // `send_request()` already sent FIN for bodyless requests, so
+            // mark the send side complete and drop the request-body receiver.
+            stream_ctx.fin_or_reset_sent = true;
+            stream_ctx.recv = None;
```
ghedo, line 223: I think this should also call `set_sent_stream_fin()` like `H3Driver::on_fin_sent()` does, so something like:

```suggestion
            stream_ctx.recv = None;
            stream_ctx.audit_stats.set_sent_stream_fin(StreamClosureKind::Explicit);
```

tokio-quiche/src/http3/driver/client.rs
```diff
@@ -214,6 +214,13 @@ impl ClientHooks {
         let (mut stream_ctx, send, recv) =
             StreamCtx::new(stream_id, STREAM_CAPACITY);
 
+        if body_finished {
+            // `send_request()` already sent FIN for bodyless requests, so
+            // mark the send side complete and drop the request-body receiver.
```
ghedo:
```suggestion
            // mark the send side complete and drop the outbound-frame receiver
            // since no body will be sent.
```

ghedo (changes requested):
Looks good, just added a couple suggestions.

ghedo (approved):

PR #2459, opened 2026-05-05
tokio-quiche: fix goaway handling in h3 driver

jannes:
receiving GOAWAY frame is not a connection error, the connection should
be kept intact and acknowledged streams should continue to be processed.

introduces new H3Event Goaway variant that applications need to process
to correctly tear down unacknowledged streams and prevent new streams
from being opened.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -141,8 +141,6 @@ pub enum H3ConnectionError {
     ControllerWentAway,
     /// Other error at the connection, but not stream level.
     H3(h3::Error),
-    /// Received a GOAWAY frame from the peer.
-    GoAway,
     /// Received data for a stream that was closed or never opened.
     NonexistentStream,
     /// The server's post-accept timeout was hit.
@@ -169,7 +167,6 @@ impl fmt::Display for H3ConnectionError {
         let s: &dyn fmt::Display = match self {
             Self::ControllerWentAway => &"controller went away",
             Self::H3(e) => e,
-            Self::GoAway => &"goaway",
             Self::NonexistentStream => &"nonexistent stream",
             Self::PostAcceptTimeout => &"post accept timeout hit",
         };
@@ -261,6 +258,9 @@ pub enum H3Event {
     /// don't result from RST_STREAM frames, unlike the
     /// [`H3Event::ResetStream`] variant.
     StreamClosed { stream_id: u64 },
+    /// A GOAWAY frame was received from the peer containing `id`,
+    /// as described in https://datatracker.ietf.org/doc/html/rfc9114#section-5.2.
+    GoAway { id: u64 },
 }
 
 impl H3Event {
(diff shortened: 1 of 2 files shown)


jannes:
not sure if letting the app do the processing is the right approach. alternatively the h3 driver could reject new streams after receiving goaway and somehow inform application about unacknowledged existing streams that are open but not processed by peer

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -261,6 +258,12 @@ pub enum H3Event {
     /// don't result from RST_STREAM frames, unlike the
     /// [`H3Event::ResetStream`] variant.
     StreamClosed { stream_id: u64 },
+    /// A GOAWAY frame was received from the peer. The `id` is the stream
+    /// ID up to which the peer has processed (or will process). Streams
+    /// with IDs below `id` complete normally; streams at or above `id`
+    /// were not processed and may be retried on a new connection. No new
+    /// requests should be initiated on this connection.
```
LPardue: Technically, a client-sent GOAWAY references a push ID not a stream ID, and the interactions are all about server push. 

Rather than try to summarize the RFC, you're probably better off just saying

```suggestion
    /// A GOAWAY frame was received from the peer containing `id`, 
    /// as described in https://datatracker.ietf.org/doc/html/rfc9114#section-5.2.
```

tokio-quiche/src/http3/driver/mod.rs
LPardue, line 634: nit: technically this is not solely `stream_id` any more

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -562,6 +563,85 @@ mod client_side_driver {
...
         assert_eq!(audit_stats.downstream_bytes_sent(), 0);
     }
+
+    /// Verify that a GOAWAY received mid-stream does not kill in-flight
+    /// request streams. The server sends GOAWAY with an ID above the
+    /// active stream, indicating the stream was accepted. The client
+    /// should continue receiving the remaining response body and
+    /// complete the stream normally.
+    #[test]
+    fn client_goaway_during_streaming_response() {
```
LPardue: Probably want a test for the client -> server GOAWAY too. Since we don't support suerver push, we don't need to exercise that fully, instead just a test to cover that hte connection doesn't immediately terminate on reception of a GOAWAY

LPardue:
> not sure if letting the app do the processing is the right approach. alternatively the h3 driver could reject new streams after receiving goaway and somehow inform application about unacknowledged existing streams that are open but not processed by peer

GOAWAY is inherently racy. Applying any opinion in this layer might surprise apps. I recommend letting them deal with it.

tokio-quiche/src/http3/driver/mod.rs
jannes: yea good point, adjusted

tokio-quiche/src/http3/driver/mod.rs
jannes: it wasn't stream_id before the change though. doc comment of https://github.com/cloudflare/quiche/blob/80c5cfe1f845926d02f64a46b6f44c8ccd7606d5/quiche/src/h3/mod.rs#L2058 does accurately describe the semantics of the ID. 

do you suggest renaming the argument in `process_read_event` and adding a link to the poll doc comment in its own doc comment?

tokio-quiche/src/http3/driver/tests.rs
jannes: done

ghedo (approved):

jrouviere (approved):

tokio-quiche/src/http3/driver/mod.rs
jannes: i can do that in a follow up MR :)

PR #2460, opened 2026-05-05
quiche: Add inline variant to ConnectionId

TheJokr:
I recommend reviewing this PR commit-by-commit.

The existing `ConnectionId` struct is 24 bytes in size, which is
sufficient to store any QUIC v1 connection ID inline. Taking advantage
of this allows us to avoid allocations when storing and moving CIDs
around throughout quiche and tokio-quiche.

The implementation is heavily inspired by `CidOwned` in tokio-quiche,
which this new implementation replaces. This includes optimizations to
make hashing and comparisons more efficient by interpreting the entire
inline storage as a u64 array to enable fewer comparisons.

The public API for `ConnectionId` stays exactly the same, except that `ConnectionId::from_vec`
can't be `const` anymore. I had to raise the MSRV to 1.87 to enable the rest
of the functions to stay `const`.

### DIFF
--- a/quiche/src/packet.rs
+++ b/quiche/src/packet.rs
@@ -182,38 +182,195 @@ impl Type {
     }
 }
 
+/// Special ranged u8 type for [`ConnectionIdInline`].
+/// Restricting the range of values allows the compiler to use free values as
+/// niches for enum size optimization. Additionally, range checks during
+/// indexing can be optimized away.
+#[derive(Clone, Copy, bytemuck::NoUninit)]
+#[repr(u8)]
+enum CidInlineLen {
+    Z   = 0,
+    P1  = 1,
+    P2  = 2,
+    P3  = 3,
+    P4  = 4,
+    P5  = 5,
+    P6  = 6,
+    P7  = 7,
+    P8  = 8,
+    P9  = 9,
+    P10 = 10,
+    P11 = 11,
+    P12 = 12,
+    P13 = 13,
+    P14 = 14,
+    P15 = 15,
+    P16 = 16,
+    P17 = 17,
+    P18 = 18,
+    P19 = 19,
+    P20 = 20,
+    P21 = 21,
+    P22 = 22,
+    P23 = 23,
+}
+
+impl CidInlineLen {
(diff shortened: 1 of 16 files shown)


gregor-cf:
Nice. I would actually get rid of ConnectionIdInline enum and instead just a SmallVec that stores up to 20 bytes (if we want to support CID of unbounded length). If we just need up to 20bytes, we can just an inline storage.
Having a lifetime on the ConnectionId is a pain in a couple of places.

quiche/src/packet.rs
```diff
@@ -182,38 +182,195 @@ impl Type {
...
+}
+
+// Use `repr(C)` to control struct layout, and `align(8)` to ensure alignment
+// for access as u64 array.
+#[derive(Clone, Copy, bytemuck::NoUninit)]
+#[repr(C, align(8))]
+struct ConnectionIdInline {
+    storage: [u8; 23],
+    len: CidInlineLen,
+}
```
gregor-cf, line 308: Using a 23 byte inline array means that ConnectionIdInline can't use niche optimization for enums, so e.g., an `Option<ConnectionIdLine>` (or ConnectionIdInner) will actually use 32 bytes. Instead, if you limit the inline storage it to 22 bytes, Option<ConnectionIdInline> should take just 24 bytes. (You'll want to double check with the compiler though :-) )

TheJokr: Ha, that's why we have CidInlineLen! CidInlineLen provides the niches, which is why `ConnectionIdInner` is only 24 bytes in total.

The same applies to `Option<ConnectionIdInline>` and even `Option<ConnectionId>`.

gregor-cf: Ah, yes. It would have helped if I read the all the changes :-)

Instead of the match, you can also use `NonZeroU8`. And encode it as `len + 1`. (i.e., non_zero_len==1 represents an actual length of 0). Implication is you can't represent a length of 255. (Or more generally create a `NonMaxU8` struct that uses NonZero internally ¯\_(ツ)_/¯

PR #2464, opened 2026-05-09
recovery: clear loss timer when no PTO timeout is calculated

mhyumike:
Currently, in `set_loss_detection_timer`, the loss timer is only updated if `pto_time_and_space` returns `Some(timeout)`. If it returns `None`, the timer is left unchanged.

`pto_time_and_space` can return `None` when there are packets in flight in the Application space (e.g., 0-RTT data on the client, or 1-RTT data on the server before handshake confirmation) but no packets in flight in Initial or Handshake spaces, and the handshake is not yet confirmed.

If the timer had already expired when this state is reached, leaving it unchanged causes the connection to repeatedly see an expired timer and immediately trigger `on_loss_detection_timeout`. This creates a tight infinite loop where `pto_count` rapidly increases until it overflows, crashing the process with "attempt to multiply with overflow".

This commit restores the behavior prior to commit bc952e8e09210d5b242640e3e6e56245cd780ded ("recovery: implement the loss detection timer as a struct") by explicitly clearing the loss timer when `pto_time_and_space` returns `None`.

A unit test `pto_overflow_reproduction` is added to reproduce this scenario on a client by simulating 0-RTT data in flight while the Initial space becomes empty before handshake confirmation.

### DIFF
--- a/quiche/src/recovery/mod.rs
+++ b/quiche/src/recovery/mod.rs
@@ -2761,6 +2761,101 @@ mod tests {
             );
         }
     }
+    #[test]
+    fn pto_overflow_reproduction() {
+        let cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        // Use LegacyRecovery (which is part of the default Recovery)
+        let mut r = Recovery::new(&cfg);
+        let now = Instant::now();
+
+        // Scenario: Handshake not completed
+        let handshake_status = HandshakeStatus {
+            has_handshake_keys: true,
+            peer_verified_address: true,
+            completed: false,
+        };
+
+        // 1. Send Initial packet to arm the timer
+        let p_initial = Sent {
+            pkt_num: 0,
+            frames: smallvec::smallvec![],
+            time_sent: now,
+            time_acked: None,
+            time_lost: None,
+            size: 1000,
+            ack_eliciting: true,
+            in_flight: true,
+            delivered: 0,
+            delivered_time: now,
+            first_sent_time: now,
+            is_app_limited: false,
+            tx_in_flight: 0,
+            lost: 0,
+            has_data: false,
+            is_pmtud_probe: false,
+        };
+        r.on_packet_sent(
+            p_initial,
+            packet::Epoch::Initial,
+            handshake_status,
+            now,
+            "",
+        );
+
+        // The timer should now be set for the Initial packet.
+        assert!(r.loss_detection_timer().is_some());
+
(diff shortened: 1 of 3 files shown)


mhyumike:
Hey quiche reviewers,

Could you please help review this patch? This patch addresses an infinite timeout loop regression.

The issue occurs under the following scenario:
1. `conn.timeout()` sees an expired `loss_detection_timer` and returns `Duration::ZERO`.
2. The application immediately calls `conn.on_timeout()`, which triggers `on_loss_detection_timeout()`.
3. `pto_count` is incremented.
4. `set_loss_detection_timer()` is called, but since `pto_time_and_space()` returns `None`, `loss_detection_timer` is not cleared and retains its expired value. 
5. `conn.timeout()` sees the same expired `loss_detection_timer` and returns `Duration::ZERO` again, repeating the loop.

This infinite loop rapidly increments `pto_count` in a tight loop until `2_u32.pow(pto_count)` overflows and crashes the process with a panic ("attempt to multiply with overflow" in `pto_time_and_space`).

This patch restores the behavior prior to commit bc952e8e by explicitly clearing the `loss_timer` when `pto_time_and_space()` returns `None`.

mhyumike:
I ran the tests locally using `cargo test --verbose --all-targets --features=async,ffi,qlog` and all tests passed successfully. 

Would it be possible to trigger a re-run of the failing CI checks?

quiche/src/recovery/mod.rs
```diff
@@ -2756,6 +2756,101 @@ mod tests {
             );
         }
     }
+    #[test]
+    fn pto_overflow_reproduction() {
+        let cfg = Config::new(crate::PROTOCOL_VERSION).unwrap();
+        // Use LegacyRecovery (which is part of the default Recovery)
```
antoniovicente: We should run the test against both recovery implementations.  Please change the previous 4 lines to:

```
#[rstest]
fn pto_overflow_reproduction(
        #[values("reno", "cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
    ) {
```

mhyumike: Done. Thanks for providing the exact code snippet.

antoniovicente (approved):

LPardue:
For posterity, the PTO panic triggered by this is now mitigated by https://github.com/cloudflare/quiche/pull/2465. This fix still looks good though.

LPardue (approved):

LPardue:
Merged, thanks @mhyumike

PR #2465, opened 2026-05-11
Fix panic in exponential backoff logic

elizagamedev:
Previously, exponential backoff was uncapped and would increase exponentially until reaching an arithmetic overflow. This change caps the exponent at 20.

Tested with `cargo test -p quiche`

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -504,6 +504,11 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
+// The upper cap on the exponent when using exponential backoff for probes. With a
+// value of 5, the maximum possible value is 2^5 = 32 times the minimum PTO. This
+// prevents arithmetic overflow.
+const MAX_PTO_EXPONENT: u32 = 5;
+
 /// Ancillary information about incoming packets.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct RecvInfo {
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -504,11 +504,6 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
-// The upper cap on the exponent when using exponential backoff for probes. With a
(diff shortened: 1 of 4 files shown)


quiche/src/recovery/congestion/recovery.rs
```diff
@@ -426,7 +426,7 @@ impl LegacyRecovery {
     fn pto_time_and_space(
         &self, handshake_status: HandshakeStatus, now: Instant,
     ) -> (Option<Instant>, Epoch) {
-        let mut duration = self.pto() * 2_u32.pow(self.pto_count);
+        let mut duration = self.pto() * 2_u32.pow(self.pto_count.min(20));
```
antoniovicente: Would it be possible to add a named constant for the PTO max exponent?  I think that a relevant constant from RFCs is MAX_PROBING_TIMEOUTS which comes up during new path validation and in the idle timeout computation when combined with the negotiated idle timeout.  I think the max exponent here could be capped at a lower value of 3-5 with a suitable justification and links to relevant sections of the QUIC RFCs.

elizagamedev: Done. I couldn't find any evidence in the RFC about a maximum idle timeout, but 32 times the minimum PTO seems like a reasonable limit.

antoniovicente (approved):
Thanks for the fix.

quiche/src/lib.rs
```diff
@@ -504,6 +504,11 @@ const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
 
+// The upper cap on the exponent when using exponential backoff for probes. With a
+// value of 5, the maximum possible value is 2^5 = 32 times the minimum PTO. This
+// prevents arithmetic overflow.
+const MAX_PTO_EXPONENT: u32 = 5;
+
```
antoniovicente: Was wondering if it could be moved to `quiche/src/recovery/mod.rs`

Also, I see a lint failure due to `cargo +nightly fmt` generating a diff.

https://github.com/cloudflare/quiche/actions/runs/25717668563/job/75892474367?pr=2465

Looking at the other CI issues, could be related to needing a rebase.

elizagamedev: Moved and fixed formatting.

LPardue (approved):

LPardue:
Merged, thanks @elizagamedev

PR #2477, opened 2026-05-13
h3: harden request stream frame processing in clients

LPardue:
Previously, the request-stream frame-type validation in
`set_frame_type` was gated on `!self.is_local`, which meant that
on a client (where request streams are locally initiated) the entire
match block was skipped. As a result, a server could send CANCEL_PUSH,
SETTINGS, GOAWAY, MAX_PUSH_ID or PRIORITY_UPDATE frames on a request
stream and the client would silently accept them, in some cases
mutating connection state (peer_settings, peer_goaway_id) or surfacing
spurious events to the application.

RFC 9114 requires these frames to be rejected with H3_FRAME_UNEXPECTED
regardless of which endpoint opened the stream.

This change is focused on hardening illegal-frame-type check gaps that
existed for the client; new tests have been added to exercise them. Other
types of exchanges already have coverage, or are due for a significant
refactor in planned work.

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -6718,6 +6718,127 @@ mod tests {
     }
 
     #[test]
+    /// Tests that a client rejects a SETTINGS frame received on a request
+    /// stream.
+    fn settings_on_request_stream_client() {
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        let (stream, _req) = s.send_request(true).unwrap();
+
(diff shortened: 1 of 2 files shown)


LPardue:
Note that the H3 stream state prcessing logic also has other pending improvements such as https://github.com/cloudflare/quiche/pull/2153. I want to keep this one as focused as possible.

ghedo (approved):

antoniovicente (approved):

PR #2480, opened 2026-05-15
Only poll ApplicationOverQuic wait_for_data after handshake is ready or we're in early data

MaxF12:
tokio-quiche can poll wait_for_data() too early during connection setup. For apps that always report should_act()==true an overly eager wait_for_data() could preempt the handshake which might stall the handshake itself. Now wait_for_data() can only be polled if the 1-RTT handshake is either completed or we are in early-data

### DIFF
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -704,7 +704,9 @@ where
     async fn wait_for_data_or_handshake<A: ApplicationOverQuic>(
         &mut self, qconn: &mut QuicheConnection, quic_application: &mut A,
     ) -> QuicResult<WaitForDataOrHandshakeDirective> {
-        if quic_application.should_act() {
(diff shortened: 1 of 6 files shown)


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -704,7 +704,9 @@ where
     async fn wait_for_data_or_handshake<A: ApplicationOverQuic>(
         &mut self, qconn: &mut QuicheConnection, quic_application: &mut A,
     ) -> QuicResult<WaitForDataOrHandshakeDirective> {
-        if quic_application.should_act() {
+        let connection_ready = qconn.is_established() || qconn.is_in_early_data();
+
+        if connection_ready && quic_application.should_act() {
```
antoniovicente, line 709: This change makes some sense to me.  What tests could we add for this case to ensure we don't regress?

MaxF12: Added regression tests now

PR #2490, opened 2026-05-19
Fix Bandwidth calculation overflow in Bandwidth::to_bytes_per_period and Bandwidth::transfer_time

antoniovicente:
Current calculation can overflow calculating bytes that can be sent when BW and RTTs are very large (~2 sec RTT at 10Gbps). Fixing as a way to avoid the incorrect calculations and to a lesser extent future proofing.

### DIFF
--- a/quiche/src/recovery/bandwidth.rs
+++ b/quiche/src/recovery/bandwidth.rs
@@ -134,6 +134,7 @@ impl Bandwidth {
         Bandwidth::from_kbits_per_second(m_bits_per_second * 1_000)
     }
 
+    /// Returns a sentinel representing infinite bandwidth.
     pub const fn infinite() -> Self {
         Bandwidth {
             bits_per_second: u64::MAX,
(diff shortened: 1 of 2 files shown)


LPardue (approved):

quiche/src/recovery/bandwidth.rs
```diff
@@ -144,21 +145,70 @@ impl Bandwidth {
         Bandwidth { bits_per_second: 0 }
     }
 
-    pub fn transfer_time(&self, bytes: usize) -> Duration {
-        #[allow(clippy::manual_checked_ops)]
+    /// Returns the time to transfer `bytes` at this bandwidth.
+    ///
+    /// Returns `Duration::ZERO` for infinite or zero bandwidth.
```
jannes, line 150: i don't know how this is used, but it doesn't make sense to me that zero bandwidth returns Duration::ZERO? should be infinity? but yea, this behavior already existed before, so shouldn't block this change i guess.

quiche/src/recovery/bandwidth.rs
```diff
@@ -144,21 +145,70 @@ impl Bandwidth {
...
-            NUM_NANOS_PER_SECOND
+        // Handle infinite bandwidth sentinel.
+        if self.bits_per_second == u64::MAX {
+            if time_period != Duration::ZERO {
+                return u64::MAX;
+            } else {
+                return 0;
+            }
+        }
+
```
jannes, line 194: technically this would overestimate if `self.bits_per_second` is exactly `u64::MAX` and not semantically the infinite sentinel + `time_period` is lower than a second but not 0, no?

quiche/src/recovery/bandwidth.rs
```diff
@@ -264,6 +314,48 @@ mod tests {
         assert_eq!(one_kbit_sec.transfer_time(100), Duration::from_millis(800));
     }
 
+    #[test]
+    fn transfer_time_overflow() {
+        // Test that large byte values that would overflow u64 are handled
+        // correctly using u128 arithmetic.
```
jannes: nit: i would put this comment on top of the test

jannes (approved):

quiche/src/recovery/bandwidth.rs
antoniovicente: The use case is the pacer.  Not having BW info should result in sending quickly.

We could change the function definition to return Option in the case of 0 bits per second.  We could try to do it in this PR if you like.

quiche/src/recovery/bandwidth.rs
antoniovicente: It would be good to split the infinite sentinel so it is different from u64::MAX.  Switch over to Option could be a way to do that.

I see an infinite return as an error; perhaps this function should return None in cases where the return would be "infinite"

quiche/src/recovery/bandwidth.rs
antoniovicente: Done

PR #2492, opened 2026-05-21
fix: do not send zero-length non-FIN stream frames

gregor-cf:
While these frames are valid per RFC 9000, it is pointless to send them (and they triggered another bug on the receive side).

The MAX_STREAM_OVERHEAD constant we used to guard the path that writes STREAM frames was too low; STREAM frame headers can be up to 19 bytes. However, using the maximum overhead is a bit iffy as well, since it would prevent us from writing the more common, smaller frames. So instead, we guard on MIN_STREAM_OVERHEAD and add a secondary check to ensure the remaining capacity is enough to fit at least one byte of stream payload.

While doing this, I also noticed that we were removing the stream from the flushable queue when there was not enough space for its header. That could strand the stream permanently and may also account for connections running out of flow-control send credit. Fixed that as well.

The MAX_STREAM_OVERHEAD constant was also used in the recovery on_app_limited() check, so I had to tweak that logic too.

### DIFF
--- a/quiche/src/frame.rs
+++ b/quiche/src/frame.rs
@@ -42,9 +42,12 @@ use qlog::events::quic::QuicFrame;
 #[cfg(feature = "qlog")]
 use qlog::events::quic::StreamType;
 
-pub const MAX_CRYPTO_OVERHEAD: usize = 8;
+/// Maximum CRYPTO frame overhead:
+/// 1-byte type + varint-encoded offset (bounded by `MAX_CRYPTO_STREAM_OFFSET`)
+/// + 2-byte length (we always encode as 2 bytes)
+pub const MAX_CRYPTO_OVERHEAD: usize =
+    1 + octets::varint_len(crate::MAX_CRYPTO_STREAM_OFFSET) + 2;
 pub const MAX_DGRAM_OVERHEAD: usize = 2;
-pub const MAX_STREAM_OVERHEAD: usize = 12;
 pub const MAX_STREAM_SIZE: u64 = 1 << 62;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -499,7 +499,13 @@ const MAX_PROBING_TIMEOUTS: usize = 3;
 const DEFAULT_INITIAL_CONGESTION_WINDOW_PACKETS: usize = 10;
 
 // The maximum data offset that can be stored in a crypto stream.
-const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
+pub(crate) const MAX_CRYPTO_STREAM_OFFSET: u64 = 1 << 16;
+
+// TODO: Removed once https://github.com/cloudflare/quiche/pull/2453 is done.
+// This used to be the value of MAX_STREAM_OVERHEAD (but it's incorrect for
+// that purpose) and is now only used for gcongestion recovery for
+// app-limited tracking. Cf. https://github.com/cloudflare/quiche/pull/2492
+const LEGACY_APP_LIMITED_BYTE_THRESHOLD: usize = 12;
 
 // The send capacity factor.
 const TX_CAP_FACTOR: f64 = 1.0;
@@ -5199,7 +5205,6 @@ impl<F: BufFactory> Connection<F> {
 
         // Create a single STREAM frame for the first stream that is flushable.
         if (pkt_type == Type::Short || pkt_type == Type::ZeroRTT) &&
-            left > frame::MAX_STREAM_OVERHEAD &&
             !is_closing &&
             path.active() &&
             !dgram_emitted
@@ -5239,13 +5244,24 @@ impl<F: BufFactory> Connection<F> {
                     octets::varint_len(stream_off) + // offset
                     2; // length, always encode as 2-byte varint
 
+                // Require at least 1 byte of payload beyond the header
+                // or an empty FIN
+                //
+                // If even the minimum payload (hdr_len + 1 byte (or empty fin)
+                // doesn't fit, stop trying.  We could continue to the next
+                // stream (a lower stream_id or offset might encode in fewer
+                // varint bytes), but we'd need to keep track of  which streams
+                // we've tried to avoid looping forever.
                 let max_len = match left.checked_sub(hdr_len) {
-                    Some(v) => v,
-                    None => {
-                        let priority_key = Arc::clone(&stream.priority_key);
-                        self.streams.remove_flushable(&priority_key);
-
-                        continue;
+                    Some(v) if v > 0 || stream.send.empty_fin_next() => v,
+                    _ => {
+                        if stream.incremental {
+                            // Shuffle the incremental stream to the back of the
+                            // queue (of streams with the same priority).
+                            self.streams.remove_flushable(&priority_key);
+                            self.streams.insert_flushable(&priority_key);
+                        }
+                        break;
                     },
                 };
 
@@ -5255,6 +5271,14 @@ impl<F: BufFactory> Connection<F> {
                 // Write stream data into the packet buffer.
                 let (len, fin) =
                     stream.send.emit(&mut stream_payload.as_mut()[..max_len])?;
+
+                // The stream was flushable and max_len ≥ 1, so emit() must
+                // have produced at least one byte, or set FIN. While a zero
+                // length non-fin STREAM frame is legal, it's silly.
+                debug_assert!(
+                    len > 0 || fin,
+                    "emit() from flushable stream produced 0 bytes without FIN"
+                );
 
                 // Encode the frame's header.
                 //
@@ -5301,10 +5325,9 @@ impl<F: BufFactory> Connection<F> {
 
                 #[cfg(feature = "fuzzing")]
                 // Coalesce STREAM frames when fuzzing.
-                if left > frame::MAX_STREAM_OVERHEAD {
-                    continue;
-                }
+                continue;
 
+                #[cfg(not(feature = "fuzzing"))]
                 break;
             }
         }
@@ -5337,12 +5360,13 @@ impl<F: BufFactory> Connection<F> {
 
         if !has_data &&
             !dgram_emitted &&
-            cwnd_available > frame::MAX_STREAM_OVERHEAD
+            cwnd_available > LEGACY_APP_LIMITED_BYTE_THRESHOLD
         {
             path.recovery.on_app_limited();
         }
 
         if frames.is_empty() {
+            // Used by legacy recovery.
             // When we reach this point we are not able to write more, so set
             // app_limited to false.
             path.recovery.update_app_limited(false);
--- a/quiche/src/stream/send_buf.rs
+++ b/quiche/src/stream/send_buf.rs
@@ -498,6 +498,12 @@ impl<F: BufFactory> SendBuf<F> {
         false
     }
 
+    /// Returns true if the next call to [`Self::emit`] will emit an empty
+    /// FIN.
+    pub fn empty_fin_next(&self) -> bool {
+        self.fin_off == Some(self.off_front())
+    }
+
     /// Returns true if the send-side of the stream is complete.
     ///
     /// This happens when the stream's send final size is known, and the peer
--- a/quiche/src/tests.rs
+++ b/quiche/src/tests.rs
@@ -5497,6 +5497,183 @@ fn stream_zero_length_non_fin(
     assert!(r.next().is_none());
 }
 
+/// Helper for parameterizing tests over different QUIC stream ID varint
+/// sizes.  Each variant yields a client-initiated bidirectional stream ID
+/// whose varint encoding is 1, 2, or 4 bytes respectively.
+#[derive(Clone, Copy, Debug)]
+enum StreamIdSize {
+    /// stream_id 0: 1-byte varint.
+    OneByteVarint,
+    /// stream_id 64: 2-byte varint.
+    TwoByteVarint,
+    /// stream_id 16384: 4-byte varint.
+    FourByteVarint,
+}
+
+impl StreamIdSize {
+    fn stream_id(self) -> u64 {
+        match self {
+            StreamIdSize::OneByteVarint => 0,
+            StreamIdSize::TwoByteVarint => 64,
+            StreamIdSize::FourByteVarint => 16384,
+        }
+    }
+}
+
+#[rstest]
+/// Tests STREAM frame emission with tight packet buffers.
+///
+/// Two scenarios are covered via the `empty_fin` parameter:
+///
+/// - `false`: the stream has pending data but no FIN.  We assert that no
+///   zero-length non-FIN STREAM frame is ever emitted, and that the stream
+///   remains flushable when the buffer is too small even for the header. The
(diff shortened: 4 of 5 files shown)


quiche/src/lib.rs
antoniovicente, line 5313: It would be useful to see if this `push_frame_to_pkt!` can ever fail.  It would indicate that we generated a frame that is larger than `left` despite taking it into account when picking the frame length.

quiche/src/lib.rs
```diff
@@ -5335,14 +5346,19 @@ impl<F: BufFactory> Connection<F> {
             path.recovery.ping_sent(epoch);
         }
 
+        // Mark the path as application-limited only when there truly is no
+        // more data to send.
         if !has_data &&
-            !dgram_emitted &&
```
csujedihy, line 5339: Is this change really needed/

quiche/src/lib.rs
```diff
@@ -5239,13 +5239,16 @@ impl<F: BufFactory> Connection<F> {
...
-                let max_len = match left.checked_sub(hdr_len) {
-                    Some(v) => v,
+                // Require at least 1 byte of payload beyond the header.
+                // If even the minimum payload (hdr_len + 1 byte) doesn't
+                // fit, stop trying.  We could continue to the next stream
+                // (a lower stream_id or offset might encode in fewer varint
+                // bytes), but we'd need a different iterator, since
+                // peek_flushable() returns the same stream on every iteration.
+                let max_len = match left.checked_sub(hdr_len + 1) {
+                    Some(v) => v + 1,
```
antoniovicente: Technically the +1 is only needed in cases where the stream has at least 1 byte to send.  The case of stream FIN with no data could be transmitted with v == 0.

quiche/src/lib.rs
```diff
@@ -5239,13 +5239,16 @@ impl<F: BufFactory> Connection<F> {
...
                     None => {
-                        let priority_key = Arc::clone(&stream.priority_key);
-                        self.streams.remove_flushable(&priority_key);
-
-                        continue;
+                        break;
```
antoniovicente, line 5264: An implication of break instead of continue is that other streams with lower stream id or offset could have written a tiny frame but now they won't.

I need to look into the implications of the `remove_flushable`

The check for `MAX_STREAM_OVERHEAD` in the original code made this branch unlikely to be reached and we could run into problems when we do hit it.

I think with the lower MIN this branch will be easier to test correct.  This is a very good thing.

quiche/src/lib.rs
```diff
@@ -5335,14 +5346,19 @@ impl<F: BufFactory> Connection<F> {
...
         }
 
+        // Mark the path as application-limited only when there truly is no
+        // more data to send.
         if !has_data &&
-            !dgram_emitted &&
-            cwnd_available > frame::MAX_STREAM_OVERHEAD
+            !self.streams.has_flushable() &&
+            !self.dgram_send_queue.has_pending() &&
+            cwnd_available < recovery::APP_LIMITED_MIN_AVAILABLE_WINDOW
```
antoniovicente: I think this should be:
`cwnd_available > recovery::APP_LIMITED_MIN_AVAILABLE_WINDOW`

Intuition is that when app-limited we expect there to be cwnd_available, not the opposite.

quiche/src/lib.rs
```diff
@@ -5335,14 +5346,19 @@ impl<F: BufFactory> Connection<F> {
...
         {
+            // Used by gcongestion recovery.
```
antoniovicente: Ugh, this is a big change.  I need to go finish https://github.com/cloudflare/quiche/pull/2453 ASAP

gregor-cf: I can also leave it as-is for now and hard-code the old 12 byte value here. 
that's probably the safest way to tackle it.

antoniovicente: I would consider doing that to avoid tickling the dragon's tail.

quiche/src/lib.rs
gregor-cf: oops. I'll actually revert the changes to the app-limited check and just hard code `12` (the old value of MAX_STREAM_OVERHEAD) for now. This way we keep the app-limited check as it was and #2453 can be the comprehensive fix.

quiche/src/lib.rs
gregor-cf: FYI, `remove_flushable()` means we don't try to emit from this stream again until it's re-inserted into flushable. That only happens when the app writes more data to the stream. If the app already wrote a FIN, the stream will be stranded forever. 

I was actually wondering if should remove the MIN_STREAM_OVERHEAD guard altogether and just rely on this check here. As soon as a stream has offset>64 we're past the guard anyways, so I don't think it really does much and just adds clutter / complexity.

quiche/src/lib.rs
gregor-cf: FWIW, I have local changes (in preparation for in-band-traceroute stuff) where I'm refactoring that part of send_single into a helper struct. I'd prefer to add it there to minimize merge conflicts :-)

quiche/src/lib.rs
gregor-cf: Yeah, good point. I'll revert it.

quiche/src/tests.rs
```diff
@@ -5442,6 +5442,112 @@ fn stream_zero_length_non_fin(
...
+///
+/// We sweep buffer sizes from 38 to 45.  A Short packet header for a
+/// 16-byte DCID is approximately 35 bytes
+///   (1 flags + 16 dcid + 2 pkt_num + 16 AEAD tag),
+fn do_not_send_empty_stream_frames_without_fin(
+    #[values("cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
+) {
+    let mut config = test_utils::Pipe::default_config(cc_algorithm_name).unwrap();
+    // stream_id 16384 requires a 4-byte varint.
+    let stream_id = 16384_u64;
```
antoniovicente: Consider parametrizing this test so it covers stream ids that can be encoded in `[1, 2, 4, 8]` bytes.  Similarly, would it be at all possible to cover offsets that have varint encoding in `[1, 2, 4, 8]` bytes

quiche/src/lib.rs
antoniovicente: I have no objections to removal of the MIN_STREAM_OVERHEAD guard.

quiche/src/lib.rs
antoniovicente: Comment was a note for future me.

send_single could use a lot of refactor.  The lost frame detection should be moved elsewhere too.

quiche/src/lib.rs
gregor-cf: If we want to continue instead of break, we'd just need to add a different iterator for flushable. I think many of the other `streams.<adjective>` collection already to. 
But with the old MAX_STREAM_OVERHEAD guard, we also skipped many streams that would have fit.

quiche/src/stream/send_buf.rs
```diff
@@ -297,7 +297,9 @@ impl<F: BufFactory> SendBuf<F> {
         // report final_size
         self.emit_off = cmp::max(self.emit_off, next_off);
 
-        Ok((out.len() - out_len, fin))
+        let written = out.len() - out_len;
```
csujedihy: what does this change do?

quiche/src/lib.rs
antoniovicente: I'm fine with the break.

quiche/src/tests.rs
gregor-cf: I thought about changing the offset, but I think that requires more code in the test that it's worth. We'd need to first write to the stream, and then drain all the data out it again and then do the test.

quiche/src/stream/send_buf.rs
antoniovicente: Looks like a functional no-op

quiche/src/frame.rs
```diff
@@ -42,9 +42,17 @@ use qlog::events::quic::QuicFrame;
 #[cfg(feature = "qlog")]
 use qlog::events::quic::StreamType;
 
-pub const MAX_CRYPTO_OVERHEAD: usize = 8;
+/// Maximum CRYPTO frame overhead:
+/// 1-byte type + varint-encoded offset (bounded by `MAX_CRYPTO_STREAM_OFFSET`)
+/// + 2-byte length (we always encode as 2 bytes)
+pub const MAX_CRYPTO_OVERHEAD: usize =
+    1 + octets::varint_len(crate::MAX_CRYPTO_STREAM_OFFSET) + 2;
```
antoniovicente, line 49: Not sure why this one is changing from 8 bytes to 7, but it does seem correct.

quiche/src/lib.rs
csujedihy: I am confused by the `remove_flushable` call. Wouldn't this cause a stream hang?

quiche/src/lib.rs
```diff
@@ -5335,14 +5346,16 @@ impl<F: BufFactory> Connection<F> {
...
 
-        if !has_data &&
-            !dgram_emitted &&
-            cwnd_available > frame::MAX_STREAM_OVERHEAD
-        {
+        // TODO: 12 is the old MAX_STREAM_OVERHEAD value, which was removed in
+        // this commit in favour of MIN_STREAM_OVERHEAD. See
+        // https://github.com/cloudflare/quiche/pull/2453 for more
+        // comprehensive app-limited changes/fixes.
+        if !has_data && !dgram_emitted && cwnd_available > 12 {
```
antoniovicente: I think we discovered that a more correct number is 20.  But I'm fine keeping it as it for consistency.

quiche/src/lib.rs
```diff
@@ -5335,14 +5346,16 @@ impl<F: BufFactory> Connection<F> {
             path.recovery.ping_sent(epoch);
         }
 
-        if !has_data &&
-            !dgram_emitted &&
-            cwnd_available > frame::MAX_STREAM_OVERHEAD
-        {
+        // TODO: 12 is the old MAX_STREAM_OVERHEAD value, which was removed in
```
antoniovicente: Could we add a local constant LEGACY_APP_LIMITED_BYTE_THRESHOLD for this magic number?

quiche/src/frame.rs
csujedihy: I am assuming 8 contains 1 byte crypto data. This will make a behavior difference. We should do this in this PR?

quiche/src/lib.rs
csujedihy: Oh I see, you removed the call to `remove_flushable`. Did we just discover another bug here? :)

gregor-cf:
addressed all the comments

quiche/src/frame.rs
gregor-cf: It was off-by-one before. the check in send_single() is `left > frame::MAX_CRYPTO_OVERHEAD &&` 
So that check does account allow for 1 byte of payload data.

csujedihy: Ok, indeed it was wrong before. Nevermind.

quiche/src/lib.rs
LPardue: I think this change in behaviour needs some more consideration, the break risks a high-priority stream blocking forward progress on others and introducing tail latency. Worst case, a certain combo could completely stall out all other streams.

I think I'd be more comfortable with having a min guard (to prevent looping through the whole set when we really tand no hope of sending anything) and trying to walk the entire flushable queue. We could avoid needing a new fluhable iterator type if we store the removed flushable values and reinsert them after we're done sending. But if its easy enough to add a new iterator that would be ok too.

quiche/src/lib.rs
LPardue: +1 I think we need a little more nuance here

quiche/src/lib.rs
gregor-cf: I reverted all these app-limited changes. Will wait for #2453

quiche/src/lib.rs
gregor-cf: I think this is still a net improvement. The vast majority of cases / streams would have never hit this `continue` at all (only streams with hdr_len > 12 would have, so essentially only a stream_off of 1GB+ would have). The  MAX_STREAM_OVERHEAD guard would have skipped even attempting to fit a frame. Now we at least try to see how much space the header would actually need. In many cases (hdr_len < 12), we can actually send a frame now where we previously couldn't. The `break` here we  mimics the behavior of the old MAX_STREAM_OVERHEAD: if we can't fit the frame, we simply skip the attempt of sending a STREAM until cwnd_available increases again. 

The impact on lower priority streams should be minimal, I think. In the best case, another stream could have fit 
a max 14bytes of payload into a frame before it also has to wait for more available cwnd. 

That said, adding the iterator is pretty straightforward and it's certainly a further improvement.

LPardue: yes the new approach does benefit from supporting smaller stream frames.

I think the old bug had an upside. If you have a huge STREAM frame to send and `left` was between 12 and 20 bytes so you couldn't actually send, the stream would be dropped and since `checked_sub` didn't update `left`, the continue an then emit one frame in the same round, and never be worried about the huge frame again 👿 

Adding an iterator lets us have the best of both worlds

gregor-cf: > That said, adding the iterator is pretty straightforward and it's certainly a further improvement.

It turns out, it isn't as trivial, we'd need to create a Vec/copy of *all* flushable stream_ids. (That's how readable and writable are handled). That seems wasteful since in the general case we only look at the first id. Cases in which it helps are rare (a large stream at the front, but still have additional smaller streams that are flushable), and the benefit it marginal (can send at most 14 more bytes, 1% of a full-sized frame). Plus it adds code complexity. 

So TL;DR: I'd leave it as-is.

quiche/src/stream/send_buf.rs
gregor-cf: oops. that was a leftover. Removed

quiche/src/lib.rs
gregor-cf: :-)

quiche/src/lib.rs
gregor-cf: done

quiche/src/lib.rs
antoniovicente: I would start with:

```
let max_len = match left.checked_sub(hdr_len) {
  Some(v) if v > 0 => v,
```

The case where emit could return bytes=0 and fin=true is an edge case where I think we could still require v>1.

We could add a method to query if emit will generate a 0 length frame with FIN set and tweak the expression above to:

`Some(v) if v > 0 || zero_bytes_with_fin => v,`

quiche/src/lib.rs
antoniovicente: I'm fine with leaving this as-is.

ghedo:
Thanks for the contribution. #2697 consolidates this sender-side fix with the corresponding receive-side flow-control correction and regression tests. Closing this PR in favor of #2697.

gregor-cf:
> Thanks for the contribution. #2697 consolidates this sender-side fix with the corresponding receive-side flow-control correction and regression tests. Closing this PR in favor of #2697.

A couple of things that this PR added are missing from #2697. Opened https://github.com/cloudflare/quiche/pull/2740 with the missing items

PR #2493, opened 2026-05-22
Only add to lost_frames on PTO if there is indication that previous added frames were sent.

antoniovicente:
This avoids scheduling the same frames for retransmission multiple times.

### DIFF
--- a/quiche/src/recovery/congestion/recovery.rs
+++ b/quiche/src/recovery/congestion/recovery.rs
@@ -89,7 +89,12 @@ struct RecoveryEpoch {
     in_flight_count: usize,
 
     acked_frames: Vec<frame::Frame>,
-    lost_frames: Vec<frame::Frame>,
+
+    // Frames scheduled for retransmission due to PTO are tracked
+    // separately so we can check that frames were drained before
+    // generating more PTO probes.
+    lost_frames_ack: Vec<frame::Frame>,
+    lost_frames_pto: Vec<frame::Frame>,
 
     /// The largest packet number sent in the packet number space so far.
     #[cfg(test)]
@@ -247,7 +252,7 @@ impl RecoveryEpoch {
             if unacked.time_sent <= lost_send_time ||
                 largest_acked >= unacked.pkt_num + pkt_thresh
             {
-                self.lost_frames.extend(unacked.frames.drain(..));
+                self.lost_frames_ack.extend(unacked.frames.drain(..));
 
                 unacked.time_lost = Some(now);
 
@@ -320,6 +325,31 @@ impl RecoveryEpoch {
 
             self.sent_packets.pop_front();
         }
+    }
+
+    /// Returns the next lost frame, trying ACK-based lost frames first,
+    /// then PTO-based lost frames.
+    fn next_lost_frame(&mut self) -> Option<frame::Frame> {
+        self.lost_frames_ack
+            .pop()
+            .or_else(|| self.lost_frames_pto.pop())
+    }
+
+    /// Returns true if there are any lost frames (ACK or PTO).
+    fn has_lost_frames(&self) -> bool {
+        !self.lost_frames_ack.is_empty() || !self.lost_frames_pto.is_empty()
+    }
+
+    /// Returns the total count of lost frames (ACK + PTO).
+    #[cfg(test)]
+    fn lost_frames_count(&self) -> usize {
+        self.lost_frames_ack.len() + self.lost_frames_pto.len()
+    }
+
+    /// Clears all lost frames (both ACK and PTO).
+    fn clear_lost_frames(&mut self) {
+        self.lost_frames_ack.clear();
+        self.lost_frames_pto.clear();
     }
 }
 
@@ -557,7 +587,7 @@ impl RecoveryOps for LegacyRecovery {
     }
 
     fn next_lost_frame(&mut self, epoch: Epoch) -> Option<frame::Frame> {
-        self.epochs[epoch].lost_frames.pop()
+        self.epochs[epoch].next_lost_frame()
     }
 
     fn get_largest_acked_on_epoch(&self, epoch: Epoch) -> Option<u64> {
@@ -565,7 +595,7 @@ impl RecoveryOps for LegacyRecovery {
     }
 
     fn has_lost_frames(&self, epoch: Epoch) -> bool {
-        !self.epochs[epoch].lost_frames.is_empty()
+        self.epochs[epoch].has_lost_frames()
     }
 
     fn loss_probes(&self, epoch: Epoch) -> usize {
@@ -575,6 +605,11 @@ impl RecoveryOps for LegacyRecovery {
     #[cfg(test)]
     fn inc_loss_probes(&mut self, epoch: Epoch) {
         self.epochs[epoch].loss_probes += 1;
+    }
+
+    #[cfg(test)]
+    fn lost_frames_count(&self, epoch: Epoch) -> usize {
+        self.epochs[epoch].lost_frames_count()
     }
 
     fn ping_sent(&mut self, epoch: Epoch) {
@@ -768,8 +803,17 @@ impl RecoveryOps for LegacyRecovery {
         epoch.loss_probes =
             cmp::min(self.pto_count as usize, MAX_PTO_PROBES_COUNT);
 
+        let sent_packets_iter_limit = if !epoch.lost_frames_pto.is_empty() {
+            // Skip the search for frames to add to PTO probes if frames
+            // added in a prior PTO haven't been processed yet.
+            0
+        } else {
+            usize::MAX
+        };
+
         let unacked_iter = epoch.sent_packets
-            .iter_mut()
+            .iter()
+            .take(sent_packets_iter_limit)
             // Skip packets that have already been acked or lost, and packets
             // that don't contain either CRYPTO or STREAM frames.
             .filter(|p| p.has_data && p.time_acked.is_none() && p.time_lost.is_none())
@@ -785,7 +829,7 @@ impl RecoveryOps for LegacyRecovery {
         // HANDSHAKE_DONE and MAX_DATA / MAX_STREAM_DATA as well, in addition
         // to CRYPTO and STREAM, if the original packet carried them.
         for unacked in unacked_iter {
-            epoch.lost_frames.extend_from_slice(&unacked.frames);
+            epoch.lost_frames_pto.extend_from_slice(&unacked.frames);
         }
 
         self.set_loss_detection_timer(handshake_status, now);
@@ -814,7 +858,7 @@ impl RecoveryOps for LegacyRecovery {
         self.bytes_in_flight.saturating_subtract(unacked_bytes, now);
 
         epoch.sent_packets.clear();
-        epoch.lost_frames.clear();
+        epoch.clear_lost_frames();
         epoch.acked_frames.clear();
 
         epoch.time_of_last_ack_eliciting_packet = None;
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -106,7 +106,12 @@ struct RecoveryEpoch {
     pkts_in_flight: usize,
 
     acked_frames: VecDeque<frame::Frame>,
-    lost_frames: VecDeque<frame::Frame>,
+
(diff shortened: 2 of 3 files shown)


quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -918,36 +923,41 @@ impl RecoveryOps for GRecovery {
...
-
-        // Retransmit the frames from the oldest sent packets on PTO. However
-        // the packets are not actually declared lost (so there is no effect to
-        // congestion control), we just reschedule the data they carried.
-        //
-        // This will also trigger sending an ACK and retransmitting frames like
-        // HANDSHAKE_DONE and MAX_DATA / MAX_STREAM_DATA as well, in addition
-        // to CRYPTO and STREAM, if the original packet carried them.
-        epoch.lost_frames.extend(unacked_frames.cloned());
+        if epoch.lost_frames.is_empty() {
```
ghedo: If I understand this correctly, I think `lost_frames.is_empty()` is too broad as a PTO dedupe guard, as `lost_frames` can already contain frames queued by normal loss detection that do not re-arm retransmittable data when drained, such as PING or datagrams. In that case, a later PTO skips queuing available stream/crypto data and only sends a probe PING, delaying retransmission by another PTO.

For reference, this test works on master, but not in this branch:
```rust
    #[rstest]
    fn pto_retransmits_data_when_lost_ping_is_pending(
        #[values("reno", "cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
    ) {

        let mut r = Recovery::new(&cfg);

        let now = Instant::now();

        // Packet threshold loss will queue this PING in lost_frames.
        let mut ping = test_utils::helper_packet_sent(0, now, 1000);
        ping.frames = smallvec![frame::Frame::Ping { mtu_probe: None }];
        ping.has_data = false;

        r.on_packet_sent(
            ping,
            packet::Epoch::Application,
            HandshakeStatus::default(),
            now,
            "",
        );

        for pkt_num in 1..4 {
            let mut p = test_utils::helper_packet_sent(pkt_num, now, 1000);
            p.has_data = false;

            r.on_packet_sent(
                p,
                packet::Epoch::Application,
                HandshakeStatus::default(),
                now,
                "",
            );
        }

        // This packet is newer than the largest acked packet and stays in
        // flight until PTO.
        let mut stream = test_utils::helper_packet_sent(4, now, 1000);
        stream.frames = smallvec![frame::Frame::StreamHeader {
            stream_id: 4,
            offset: 0,
            length: 5,
            fin: false,
        }];

        r.on_packet_sent(
            stream,
            packet::Epoch::Application,
            HandshakeStatus::default(),
            now,
            "",
        );

        let mut acked = RangeSet::default();
        acked.insert(1..4);

        // Acking packets 1-3 declares packet 0 lost by packet threshold, but
        // does not consider packet 4 lost.
        assert_eq!(
            r.on_ack_received(
                &acked,
                0,
                packet::Epoch::Application,
                HandshakeStatus::default(),
                now + Duration::from_millis(10),
                None,
                "",
            )
            .unwrap(),
            OnAckReceivedOutcome {
                lost_packets: 1,
                lost_bytes: 1000,
                acked_bytes: 3000,
                spurious_losses: 0,
            }
        );

        let pto = r.loss_detection_timer().unwrap();
        r.on_loss_detection_timeout(HandshakeStatus::default(), pto, "");

        let mut has_ping = false;
        let mut has_stream = false;

        while let Some(lost) = r.next_lost_frame(packet::Epoch::Application) {
            match lost {
                frame::Frame::Ping { mtu_probe: None } => has_ping = true,

                frame::Frame::StreamHeader {
                    stream_id: 4,
                    offset: 0,
                    length: 5,
                    fin: false,
                } => has_stream = true,

                _ => {},
            }
        }

        assert!(has_ping, "loss detection should queue the lost PING frame");
        assert!(
            has_stream,
            "PTO should queue stream data even when lost_frames already contains \
             a non-retransmittable frame"
        );
    }
```

ghedo: I guess you could maybe use `epoch.loss_probes` instead?

gregor-cf: I think the code sequence that this test executes is not what the real code does. If a PTO is pending and lost_frames already contains data, the call to `send_single()` will iterate through all `lost_frames` and marks things for retransmission accordingly. So non-retransmissible frames will be processed. The end effect should be that a lost, retransmittable frame is sent as the PTO probe. 

(If my above statement is wrong, then I think the bug is that we we don't process all the lost_frame when we send a PTO probe. We should).

antoniovicente: I was working on the assumption that the event loop runs in the following order:
- Timeout handling
- Packet receives
- Send which includes lost_frame processing.

The only reason why lost_frames wouldn't be empty is if packet receive / ack processing ran before timeout handling. But we don't have a way to ensure implementations based on quiche, but not tokio-quiche, do the operations in the right order so it's better to not depend on that.

We could have a different Vec for loss probes as you suggest which would make things more clear.  We'ld need to make sure to run through both vectors when doing loss processing in send_single, but that's easy enough to do.

I'll try to get your suggestion implemented.

gregor-cf: Actually, I guess if only non-retransmittable frames have been lost, we would still only send a PING on PTO. IMHO that's enough of an edge case that it's ok (after all, when PTO triggers, the network is either really bad or we are probing for tail loss). 

If that's a concern, we could change `lost_frames.is_empty()` check to iterating through `lost_frames` and see if it contains any retransmittable frames. If it does, we are done. If it does not, we proceed to find and add such a frame from unacked.

gregor-cf: FWIW, the doc strings say that `send()` must be called after timeout handling.

antoniovicente: What is important in this case is that timeouts are run before recv processing.  Send happens after those two.

The bad ordering is
1. recv / ack processing
2. timeouts

antoniovicente: Tracking lost_frames added during PTO in a separate Vec

PTAL

quiche/src/recovery/congestion/recovery.rs
```diff
@@ -89,7 +89,8 @@ struct RecoveryEpoch {
     in_flight_count: usize,
 
     acked_frames: Vec<frame::Frame>,
-    lost_frames: Vec<frame::Frame>,
+    lost_frames_ack: Vec<frame::Frame>,
```
ghedo, line 96: Could use a comment on what these fields do and why there are 2 separate ones.

quiche/src/recovery/congestion/recovery.rs
```diff
@@ -768,24 +780,28 @@ impl RecoveryOps for LegacyRecovery {
...
-        // Retransmit the frames from the oldest sent packets on PTO. However
-        // the packets are not actually declared lost (so there is no effect to
-        // congestion control), we just reschedule the data they carried.
-        //
-        // This will also trigger sending an ACK and retransmitting frames like
-        // HANDSHAKE_DONE and MAX_DATA / MAX_STREAM_DATA as well, in addition
-        // to CRYPTO and STREAM, if the original packet carried them.
-        for unacked in unacked_iter {
-            epoch.lost_frames.extend_from_slice(&unacked.frames);
+        if epoch.lost_frames_pto.is_empty() {
```
ghedo: This could use a comment explaining why lost_frames_pto is used, and not, say, lost_frames_ack.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -156,6 +157,8 @@ impl RecoveryEpoch {
             .sum();
 
         std::mem::take(&mut self.sent_packets);
+        self.lost_frames_ack.clear();
```
ghedo: Should this also clear acked_frames then? Seems like legacy recovery does this too.

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -918,39 +932,46 @@ impl RecoveryOps for GRecovery {
...
+                    } = &p.status
+                    {
+                        Some(frames)
+                    } else {
+                        None
+                    }
+                })
+                .take(epoch.loss_probes)
+                .flatten()
+                .filter(|f| !matches!(f, frame::Frame::DatagramHeader { .. }));
```
LPardue: why does this one have a different set of filters compared to the on in `recovery.rs`?

quiche/src/recovery/congestion/recovery.rs
```diff
@@ -768,24 +780,28 @@ impl RecoveryOps for LegacyRecovery {
...
+                .iter()
+                // Skip packets that have already been acked or lost, and packets
+                // that don't contain either CRYPTO or STREAM frames.
+                .filter(|p| p.has_data && p.time_acked.is_none() && p.time_lost.is_none())
+                // Only return as many packets as the number of probe packets that
+                // will be sent.
+                .take(epoch.loss_probes);
+
+            // Retransmit the frames from the oldest sent packets on PTO. However
+            // the packets are not actually declared lost (so there is no effect
```
ghedo: Seems like some of these comments have been rewrapped unnecessarily... not really a problem, it just makes the diff a bit noisier than striclty necessary.

ghedo (changes requested):

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -918,39 +932,46 @@ impl RecoveryOps for GRecovery {
...
+            // well, in addition to CRYPTO and STREAM, if the original
+            // packet carried them.
+            epoch.lost_frames_pto.extend(unacked_frames.cloned());
+        }
 
-        self.pacer
-            .on_retransmission_timeout(!epoch.lost_frames.is_empty());
+        self.pacer.on_retransmission_timeout(
+            !epoch.lost_frames_ack.is_empty() ||
+                !epoch.lost_frames_pto.is_empty(),
```
LPardue: should this be `has_lost_frames()` ?

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: Because these were the filters set in the quiche mallard branch.  We need to go back and merge the congestion and gcongestion directories.

LPardue: Its not clear to me if the missing filter is an omission or intentional (e.g. because recoery just works differently). If its an omission, does that mean we still risk pushing frames for retransmission that shouldn't be such as DATAGRAM?

antoniovicente: Thinking about it: this is an optimization.

Lost frame progressing in send_single ignores datagram frames.  By filtering earlier we avoid putting items in lost_frames that would be ignored later.

Note that the filtering happens too late.  We're still using up the PTO quota if there are datagram frames in a packet.

The other thing to consider is sent_pkt_has_data.  In the gcongestion branch sent_pkt_has_data is set to true when the packet has datagram frames, but that's not true in the congestion branch.  I don't know which is more correct without thinking about the meaning of has_data and doing some code inspection; but I'm tempted to say that the congestion branch is more correct.

Regardless, making the two implementations consistent is not something we should take on while doing the current bug fix.

LPardue: That's cool, thanks for the explanation, lets punt on this as tangential.

LPardue (approved):
lgtm modulo nit

quiche/src/recovery/congestion/recovery.rs
antoniovicente: Let me know what you think after: https://github.com/cloudflare/quiche/pull/2493/changes/504220b004d484c077277bee9315823c28882fab

I tried to minimize the diff by adding an early `take(0 or usize::MAX)` to stop the iteration early if lost_frames_pto is non-empty.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: Added helper methods to the epoch and changed this line.

quiche/src/recovery/gcongestion/recovery.rs
antoniovicente: Done via `std::mem::take`

ghedo (approved):

PR #2494, opened 2026-05-22
Refactor process_lost_frames into a separate function

antoniovicente:
send_single is already too long.  We should think of processing of lost frames as a different operation

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4132,33 +4132,13 @@ impl<F: BufFactory> Connection<F> {
         Ok((done, info))
     }
 
-    fn send_single(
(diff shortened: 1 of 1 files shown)


quiche/src/lib.rs
```diff
@@ -4209,7 +4189,7 @@ impl<F: BufFactory> Connection<F> {
                                         },
                                     );
 
-                                    q.add_event_data_with_instant(ev_data, now)
+                                    q.add_event_data_with_instant(ev_data, _now)
```
gregor-cf, line 4192: We should remove the underscore IMHO. If clippy complains that `now` is not used when qlog is disabled, I'd just add an annotation to `send_single` to allow unused args.

gregor-cf (approved):
Yes, please!!! This is great.

PR #2500, opened 2026-05-29
qlog,quiche,tokio-quiche: add QlogSink abstraction for custom qlogs

strtok:
Add a QlogSink trait that allows overriding qlog behavior. Previously QlogStreamer was instantiated with a writer object. Instead, it is instantiated with a QlogSink object which allows more flexibility. A QlogWriterSink implementation provides backwards compatibility.

Implementations have new methods for overriding qlog behavior:
  - Connection::set_qlog_sink and set_qlog_sink_with_level allow setting a custom qlog sink. The existing set_qlog and set_qlog_with_level exist for backwards compatibility and call these new functions with a QlogWriterSink object.
  - A new ConnectionHook::create_qlog_sink hook can be implemented to install a sink at connection accept/connect time. If a sink is returned it overrides any qlog_dir behavior.

Add QlogStreamer::should_log(EventType) that allows high volume events to be skipped without construction. Modified both the recv and write path to make use of this.

### DIFF
--- a/qlog/src/sink.rs
+++ b/qlog/src/sink.rs
@@ -1,0 +1,265 @@
+// Copyright (C) 2026, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+// IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::io::Write;
+
+use crate::events;
+use crate::Error;
+use crate::QlogSeq;
+use crate::Result;
+
+/// An event that can be delivered to a [`QlogSink`].
+pub enum QlogEvent {
+    /// A native qlog event.
+    Event(events::Event),
+
+    /// An extended JSON event.
+    JsonEvent(events::JsonEvent),
+}
+
+impl From<events::Event> for QlogEvent {
+    fn from(event: events::Event) -> Self {
+        Self::Event(event)
+    }
+}
+
+impl From<events::JsonEvent> for QlogEvent {
+    fn from(event: events::JsonEvent) -> Self {
+        Self::JsonEvent(event)
+    }
+}
+
+/// A destination for sequential qlog events.
+pub trait QlogSink: Send + Sync {
+    /// Start a qlog stream by writing or otherwise recording the stream header.
+    fn start_log(&mut self, qlog: &QlogSeq) -> Result<()>;
+
+    /// Add a native qlog event to the stream.
+    fn add_event(&mut self, event: events::Event) -> Result<()>;
+
+    /// Add a pretty-printed native qlog event to the stream.
+    fn add_event_pretty(&mut self, event: events::Event) -> Result<()> {
+        self.add_event(event)
+    }
+
+    /// Add an extended JSON qlog event to the stream.
+    fn add_json_event(&mut self, event: events::JsonEvent) -> Result<()>;
+
+    /// Add a pretty-printed extended JSON qlog event to the stream.
+    fn add_json_event_pretty(&mut self, event: events::JsonEvent) -> Result<()> {
+        self.add_json_event(event)
+    }
+
+    /// Finish the stream.
+    fn finish_log(&mut self) -> Result<()>;
+
+    /// Returns whether this sink wants events of `event_type`.
+    fn should_log(&self, _event_type: events::EventType) -> bool {
+        true
+    }
+}
+
+/// A [`QlogSink`] that writes JSON-SEQ qlog records to a [`Write`].
+pub struct QlogWriterSink<W: Write + Send + Sync> {
+    writer: W,
+}
+
+impl<W: Write + Send + Sync> QlogWriterSink<W> {
+    /// Creates a new writer-backed sink.
+    pub fn new(writer: W) -> Self {
+        Self { writer }
+    }
+}
+
+impl<W: Write + Send + Sync> QlogSink for QlogWriterSink<W> {
+    fn start_log(&mut self, qlog: &QlogSeq) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer(&mut self.writer, qlog).map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
+    }
+
+    fn add_event(&mut self, event: events::Event) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer(&mut self.writer, &event)
+            .map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
+    }
+
+    fn add_event_pretty(&mut self, event: events::Event) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer_pretty(&mut self.writer, &event)
+            .map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
+    }
+
+    fn add_json_event(&mut self, event: events::JsonEvent) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer(&mut self.writer, &event)
+            .map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
+    }
+
+    fn add_json_event_pretty(&mut self, event: events::JsonEvent) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer_pretty(&mut self.writer, &event)
+            .map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
+    }
+
+    fn finish_log(&mut self) -> Result<()> {
+        self.writer.flush()?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::io;
+    use std::io::Write;
+    use std::sync::Arc;
+    use std::sync::Mutex;
+
+    use crate::events::quic;
+    use crate::events::quic::QuicEventType;
+    use crate::events::Event;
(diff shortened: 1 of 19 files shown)


meha23 (approved):
strtok:ebremen/qlog

qlog/src/sink.rs
```diff
@@ -0,0 +1,285 @@
...
+/// A destination for sequential qlog events.
+pub trait QlogSink: Send + Sync {
+    /// Start a qlog stream by writing or otherwise recording the stream header.
+    fn start_log(&mut self, qlog: &QlogSeq) -> Result<()>;
+
+    /// Add a native qlog event to the stream.
+    fn add_event(&mut self, event: events::Event) -> Result<()>;
+
+    /// Add a pretty-printed native qlog event to the stream.
+    fn add_event_pretty(&mut self, event: events::Event) -> Result<()> {
```
gregor-cf, line 64: I would remove the `_pretty_ functions. The sink might not actually have a serialization or similar that can do pretty printing. IMHO if pretty printing is desired, then we that should be handled by the sink / writer internally, but it shouldn't be exposed to callers of QlogSink.

But I'll let the folks who orginally added the QlogStreamer chime in.

qlog/src/sink.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///
+    /// This method exists solely to support the pre-existing
+    /// [`QlogStreamer::writer`] public API. It returns `None` for any
+    /// sink that does not wrap a `Box<dyn std::io::Write + Send + Sync>`;
+    /// only [`QlogWriterSink`] overrides it. New sinks should not
+    /// implement this method.
+    ///
+    /// [`QlogStreamer::writer`]: crate::streamer::QlogStreamer::writer
+    #[allow(clippy::borrowed_box)]
+    fn try_writer(&self) -> Option<&Box<dyn Write + Send + Sync>> {
```
gregor-cf: nit: I think you can remove this. The `QlogStreamer::write()` function is only used in tests IIRC. In tests where this is used you can (I think) `downcast_ref()` the QlogSing to a `QlogWriterSink` and then just use the `writer()` fn on QlogWriterSink.

qlog/src/sink.rs
```diff
@@ -0,0 +1,285 @@
...
+        None
+    }
+}
+
+/// A [`QlogSink`] that writes JSON-SEQ qlog records to a [`Write`].
+pub struct QlogWriterSink {
```
gregor-cf: I think you can get rid of the `Box<dyn Write....>` and instead make `QlogWriterSink<W: Write + Send + Sync>` generic. 

(IMHO the main reason `QlogSink` / `QlogStreamer` use Box<dyn...> insteaed of generics, is so that we don't need to have generic type parameters for quiche::Connection. But that's not a concern for QlogWriterSink)

qlog/src/streamer.rs
```diff
@@ -190,7 +208,7 @@ impl QlogStreamer {
 
     /// Writes a serializable to a pretty-printed JSON-SEQ record using
     /// [std::time::Instant::now()].
-    pub fn add_event_now_pretty<E: Serialize + Eventable>(
+    pub fn add_event_now_pretty<E: Into<crate::QlogEvent> + Eventable>(
```
gregor-cf, line 224: nit: I'd remove the `_pretty` functions here too....

qlog/src/sink.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Add an extended JSON qlog event to the stream.
+    fn add_json_event(&mut self, event: events::JsonEvent) -> Result<()>;
+
+    /// Add a pretty-printed extended JSON qlog event to the stream.
+    fn add_json_event_pretty(&mut self, event: events::JsonEvent) -> Result<()> {
+        self.add_json_event(event)
+    }
+
+    /// Finish the stream.
+    fn finish_log(&mut self) -> Result<()>;
```
gregor-cf, line 77: We should clarify that calling `finish_log()` is optional and that implementors of `QlogSink` must be able to cleanly finish their log if the sink is dropped w/o `finish_log()` being called. 
(Or maybe we can get rid of finish_log()? Need to check why it's separate from Drop)

qlog/src/streamer.rs
```diff
@@ -363,28 +392,69 @@ impl QlogStreamer {
...
+    /// underlying [`QlogSink`] must be a [`QlogWriterSink`] for this
+    /// accessor to succeed; otherwise it panics. Custom sinks that do
+    /// not wrap a writer should call [`QlogStreamer::try_writer`]
+    /// instead.
+    ///
+    /// [`QlogSink`]: crate::QlogSink
+    /// [`QlogWriterSink`]: crate::QlogWriterSink
+    /// [`QlogStreamer::try_writer`]: Self::try_writer
     #[allow(clippy::borrowed_box)]
     pub fn writer(&self) -> &Box<dyn std::io::Write + Send + Sync> {
```
gregor-cf: we should remove this function and replace with a `fn sink(&self)` that returns a reference to the QlogSink. It's ok if we break the api.

qlog/src/streamer.rs
```diff
@@ -363,28 +392,69 @@ impl QlogStreamer {
...
+    /// Non-panicking counterpart to [`QlogStreamer::writer`]. Returns
+    /// `None` for streamers constructed via
+    /// [`QlogStreamer::new_with_sink`] with a sink that does not wrap a
+    /// writer.
+    ///
+    /// [`QlogSink`]: crate::QlogSink
+    /// [`QlogStreamer::writer`]: Self::writer
+    /// [`QlogStreamer::new_with_sink`]: Self::new_with_sink
+    #[allow(clippy::borrowed_box)]
+    pub fn try_writer(&self) -> Option<&Box<dyn std::io::Write + Send + Sync>> {
```
gregor-cf: nit: I'd remove this too.

quiche/src/lib.rs
```diff
@@ -2348,6 +2348,45 @@ impl<F: BufFactory> Connection<F> {
     pub fn set_qlog_with_level(
```
gregor-cf, line 2330: can you update the docstring and add a reference/pointer to `set_qlog_sink`

Maybe we should even add a `#[deprecated]` annotation??

gregor-cf (commented):
This is great. I did a quick pass and had some minor comments and nits.

quiche/src/lib.rs
```diff
@@ -2348,6 +2348,45 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// Only qlog events included in the specified `QlogLevel` are written.
+    ///
+    /// This needs to be called as soon as the connection is created, to avoid
+    /// missing some early logs.
+    ///
+    /// [`QlogSink`]: qlog::QlogSink
+    #[cfg(feature = "qlog")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "qlog")))]
+    pub fn set_qlog_sink_with_level(
```
csujedihy, line 2368: Should we support unsetting the sink? I am thinking in the dynamic enable/disable scenario, we'd want all side-effects of qlog sink to go away.

strtok: Make sense. I'm adding an unset_qlog_sink.

qlog/src/sink.rs
strtok: I've removed it and exposed a sink(). The tests actually just create a QlogSink with a shared writer now.

quiche/src/lib.rs
```diff
@@ -5403,14 +5482,20 @@ impl<F: BufFactory> Connection<F> {
...
 
         #[cfg(feature = "qlog")]
-        let mut qlog_frames: Vec<qlog::events::quic::QuicFrame> =
-            Vec::with_capacity(frames.len());
+        let mut qlog_frames = self
+            .qlog
+            .streamer
+            .as_ref()
+            .is_some_and(|q| q.should_log(QLOG_PACKET_TX))
+            .then(|| Vec::with_capacity(frames.len()));
```
LPardue, line 5466: This seems to be a slight variant of qlog_with_type!, consider adding a companion macro that returns bool so you can do

```
let mut qlog_frames = if should_qlog_with_type!(....) {
   Some(Vec::with_capacity(frames.len()) } else { None };

File: quiche/src/lib.rs [hunk 11]
```diff
@@ -5433,7 +5518,7 @@ impl<F: BufFactory> Connection<F> {
                 let ev_data =
                     EventData::QuicPacketSent(qlog::events::quic::PacketSent {
                         header,
-                        frames: Some(qlog_frames),
+                        frames: Some(qlog_frames.unwrap_or_default()),
```
LPardue, line 5497: qlog_frames is an option with this change
```suggestion
                        frames: qlog_frames
```

qlog/src/sink.rs
LPardue: I think it really depends on what we are trying to abstract. 

The qlog spec only defines two serialization formats, JSON and JSON-SEQ, the latter being a streaming format. Other serialization formats can be used, they just aren't specified (yet).

This trait is describing itself as "/// A destination for sequential qlog events." but its unclear to me what that means. qlog events are don't have a strict sequential requirement - the spec says 

> Events in each individual trace SHOULD be logged in strictly ascending timestamp order 

but this is just a recommendation because in practice achieving that in the logger might be difficult.

So some clarification questions:

1. Is the trait intended for "streaming" qlog events at the unit of a single event?
2. Are we intending to want to support other qlog serialization formats? For example, writing out events as CBOR of BSON.
3. Are we intending to want to support transforms on the serialized data. For example, conventional gzip stream on the text output?

if the answer to (2) is yes, then I'd probably agree with Gregor's point - the "pretty" is a facet of the serializer not the qlog event. However, it is useful to have the flexibility for a single JSON-SEQ log to mix and match condensed vs. pretty printed JSON. We do this with h3i today. That suggests you might need a `add_event_with_customization` (or pick a snazzier name) that would allow per-event sink-specific serialization customization.

qlog/src/sink.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+}
+
+/// A destination for sequential qlog events.
+pub trait QlogSink: Send + Sync {
+    /// Start a qlog stream by writing or otherwise recording the stream header.
+    fn start_log(&mut self, qlog: &QlogSeq) -> Result<()>;
+
+    /// Add a native qlog event to the stream.
+    fn add_event(&mut self, event: events::Event) -> Result<()>;
```
LPardue, line 61: It's not clear to me why the sink trait has both add_event() and add_json_event(), when there is a QlogEvent type defined. Wouldn't it be easier to have a single `add_event(&mut self, event: QlogEvent)` method?

qlog/src/sink.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+impl From<events::JsonEvent> for QlogEvent {
+    fn from(event: events::JsonEvent) -> Self {
+        Self::JsonEvent(event)
+    }
+}
+
+/// A destination for sequential qlog events.
+pub trait QlogSink: Send + Sync {
```
LPardue, line 56: This name doesn't seem completely accurate, consider StreamingQlogSink to make it clear what it is (although see my clarification questions below).

qlog/src/sink.rs
```diff
@@ -0,0 +1,265 @@
...
+    fn finish_log(&mut self) -> Result<()>;
+
+    /// Returns whether this sink wants events of `event_type`.
+    fn should_log(&self, _event_type: events::EventType) -> bool {
+        true
+    }
+}
+
+/// A [`QlogSink`] that writes JSON-SEQ qlog records to a [`Write`].
+pub struct QlogWriterSink<W: Write + Send + Sync> {
```
LPardue, line 86: Strictly, per https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#section-11.2 you'd want include `QlogFileSeq` in the name for clarity. Something like `QlogFileSeqWriterSink` etc.

qlog/src/sink.rs
```diff
@@ -0,0 +1,265 @@
...
+        Ok(())
+    }
+
+    fn add_json_event_pretty(&mut self, event: events::JsonEvent) -> Result<()> {
+        self.writer.write_all(b"\x1e")?;
+        serde_json::to_writer_pretty(&mut self.writer, &event)
+            .map_err(|_| Error::Done)?;
+        self.writer.write_all(b"\n")?;
+
+        Ok(())
```
LPardue, line 139: There's tonnes of duplication across these functions. The old serializer code was able to condense it all down into a single write_event  function, I can't see a good reason we can't do the same.

LPardue (changes requested):
Thanks for the PR. 

Overall I'm a big fan of adding more flexibility to what events are included in a serialized qlog, and how it is precisely written out. This makes a good start on it.

That said, its not 100% clear to me on the abstraction model being used. I've asked some clarification questions in-band to try and understand more. 

Its made a bit more complicated because there is a `QlogStreamer` that now seems to wrap a sink and some of the contracts documented at https://docs.rs/qlog/latest/qlog/#streaming-traces-with-json-seq no longer holds with the change. For example, we can't say its always JSON-SEQ if the sink allows it to be something different 😄 The docs should be updated once we have hammered out the design questions a little

There's a lot of changes in a single commit. While its great to see how tokio-quiche can be updated to use the new API. I think it would help to break this work out over multiple commits or probably even multiple PRs. I think that will help to let us iterate on the qlog sink details and resolve those independent of tokio-quiche issues. This also lets us prove the change is backwards compatible with our own stack.

LPardue:
@strtok FYI I rebased the PR to also let CI have a go and find any other latent issues that might be hiding.

PR #2502, opened 2026-06-04
octets: add generic Huffman encoding writer

ghedo:
Move HPACK Huffman output behind a small writer trait so callers can reuse the encoder with sinks other than OctetsMut. Keep the existing OctetsMut API by delegating through the trait implementation and add tests for custom sink output and error propagation.

This makes the encoder useful for streaming or segmented output where building one contiguous temporary buffer would be unnecessary (e.g. for HTTP/2 HEADERS+CONTINUATION frames).


octets/src/lib.rs
```diff
@@ -1179,6 +1222,121 @@ mod tests {
         assert!(b.get_huffman_decoded().is_err());
     }
 
+    #[cfg(feature = "huffman_hpack")]
+    #[test]
+    fn octets_writer_huffman_matches_rfc_vectors() {
```
LPardue: maybe we should have a different test module for huffman? It could make it a bit clearer by reducing the number of cfg flags

octets/src/lib.rs
```diff
@@ -865,6 +846,68 @@ pub fn huffman_encoding_len<const LOWER_CASE: bool>(src: &[u8]) -> Result<usize>
...
+        if pending < 64 {
+            // Have room for the new token.
+            bits |= code << (64 - pending);
+            continue;
+        }
+
+        pending -= 64;
+        // Take only the bits that fit.
+        bits |= code >> pending;
+        write(&bits.to_be_bytes())?;
```
jrouviere, line 901: I'm guessing that the `write` function needs to be retryable safely?

For instance it's not ok to write directly to the network with it, since `huffman_encode_with` might fail after writing some partial data.

octets/src/lib.rs
ghedo: Yeah, good point. I added a commit to move all octets tests into tests/ with a separate huffman file, and then updated the original change to add tests in that too.

octets/src/lib.rs
ghedo: Yeah, basically the same assumption (the underlying sink has enough capacity) applies, so either an error is returned or there needs to be some buffering within the writer itself.

jrouviere: Sounds good. Might be worth mentioning explicitly in the docs, otherwise I feel like that could be a footgun. 
Users might be tempted to write a sink that eagerly sends on the network, but the sink cannot return an error in that case.

ghedo: Documented this on `OctetsWriter::put_huffman_encoded`: encoding is non-atomic, retries require discarding partial output, and irreversible sinks such as network writes must buffer or guarantee success.

jrouviere (approved):

dineshadhi (approved):

PR #2503, opened 2026-06-05
Fix tx_buffered computation so it matches the sum of bytes in stream buffers

antoniovicente:
We found that the cached tx_buffered in the connection drifts from the actual sum of stream send buffers when taking retransmissions and stream cancellations into account.  Also, this cached value was not properly updated as bytes were send, so the stored value was closer to the sum(send buffers) plus inflight.

The new tx_buffered holds the cached number of bytes in send buffers so it can be be compared with cwnd_available when determining if Reno and CUBIC congestion controllers are app-limited.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -1425,9 +1425,6 @@ where
     /// The send capacity factor.
     tx_cap_factor: f64,
 
-    /// Number of bytes buffered in the send buffer.
-    tx_buffered: usize,
-
     /// Tracks the health of tx_buffered.
     tx_buffered_state: TxBufferTrackingState,
 
@@ -1457,7 +1454,7 @@ where
     lost_bytes: u64,
 
     /// Streams map, indexed by stream ID.
-    streams: stream::StreamMap<F>,
+    pub(crate) streams: stream::StreamMap<F>,
 
     /// Peer's original destination connection ID. Used by the client to
     /// validate the server's transport parameter.
@@ -2138,7 +2135,6 @@ impl<F: BufFactory> Connection<F> {
             tx_cap: 0,
             tx_cap_factor: config.tx_cap_factor,
 
-            tx_buffered: 0,
             tx_buffered_state: TxBufferTrackingState::Ok,
 
             tx_data: 0,
@@ -3633,12 +3629,9 @@ impl<F: BufFactory> Connection<F> {
                         length,
                         ..
                     } => {
-                        // Update tx_buffered and emit qlog before checking if the
-                        // stream still exists.  The client does need to ACK
-                        // frames that were received after the client sends a
-                        // ResetStream.
-                        self.tx_buffered =
-                            self.tx_buffered.saturating_sub(length);
+                        // Emit qlog before checking if the stream still exists.
+                        // The client does need to ACK frames that were received
+                        // after the client sends a ResetStream.
 
                         qlog_with_type!(QLOG_DATA_MV, self.qlog, q, {
                             let ev_data = EventData::QuicStreamDataMoved(
@@ -3664,8 +3657,7 @@ impl<F: BufFactory> Connection<F> {
                             None => continue,
                         };
 
-                        stream.send.ack_and_drop(offset, length);
-
+                        let dropped = stream.send.ack_and_drop(offset, length);
                         let priority_key = Arc::clone(&stream.priority_key);
 
                         // Only collect the stream if it is complete and not
@@ -3697,6 +3689,13 @@ impl<F: BufFactory> Connection<F> {
                         if is_complete && !is_readable && !is_writable {
                             let local = stream.local;
                             self.streams.collect(stream_id, local);
+                        }
+
+                        // Update cache to reflect any data that was dropped from
+                        // buffers (e.g., data marked for retransmission but then
+                        // acked before it could be resent).
+                        if dropped > 0 {
+                            self.streams.sub_tx_buffered(dropped);
                         }
                     },
 
@@ -4188,12 +4187,9 @@ impl<F: BufFactory> Connection<F> {
                             Some(v) if !v.send.is_stopped() => v,
 
                             // Data on a closed stream will not be retransmitted
-                            // or acked after it is declared lost, so update
-                            // tx_buffered and qlog.
+                            // or acked after it is declared lost, so just drop
+                            // it.
                             _ => {
-                                self.tx_buffered =
-                                    self.tx_buffered.saturating_sub(length);
-
                                 qlog_with_type!(QLOG_DATA_MV, self.qlog, q, {
                                     let ev_data = EventData::QuicStreamDataMoved(
                                         qlog::events::quic::StreamDataMoved {
@@ -4221,7 +4217,8 @@ impl<F: BufFactory> Connection<F> {
 
                         let empty_fin = length == 0 && fin;
 
-                        stream.send.retransmit(offset, length);
+                        let retransmitted =
+                            stream.send.retransmit(offset, length);
 
                         // If the stream is now flushable push it to the
                         // flushable queue, but only if it wasn't already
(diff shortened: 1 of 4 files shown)


quiche/src/stream/mod.rs
```diff
@@ -202,6 +202,9 @@ pub struct StreamMap<F: BufFactory = DefaultBufFactory> {
     /// When `true`, the initial flow control window will be set to the
     /// `max_rx_data`, if false, it will be set to `DEFAULT_STREAM_WINDOW`
     use_initial_max_data_as_flow_control_win: bool,
+
+    /// Cached total number of bytes buffered across all streams.
+    tx_buffered_cache: usize,
```
LPardue: nit: The var name is ok but both here and elsewhere consider making the comments a litle more precise. IIUC it is the bytes in the send side (not receive) and does not count streams in a collected state.

quiche/src/stream/mod.rs
```diff
@@ -721,6 +724,69 @@ impl<F: BufFactory> StreamMap<F> {
...
+    fn tx_buffered_actual(&self) -> usize {
+        self.streams.values().map(|s| s.send.len() as usize).sum()
+    }
+
+    /// Checks if the cached tx_buffered matches the actual value.
+    /// Returns true if they match, false otherwise.
+    pub(crate) fn tx_buffered_is_consistent(&self) -> bool {
+        let actual: usize =
+            self.streams.values().map(|s| s.send.len() as usize).sum();
+        self.tx_buffered_cache == actual
```
LPardue: nit: this is a dupe of `tx_buffered_actual` so consider dropping the cfg flag that and calling it directly e.g.

```suggestion
        self.tx_buffered_cache == tx_buffered_actual()
```

quiche/src/stream/mod.rs
```diff
@@ -721,6 +724,69 @@ impl<F: BufFactory> StreamMap<F> {
...
+    /// Returns the total number of bytes buffered across all streams.
+    pub(crate) fn tx_buffered(&self) -> usize {
+        self.tx_buffered_cache
+    }
+
+    /// Computes the actual tx_buffered by summing across all streams.
+    /// This is used for debugging to verify the cache is accurate.
+    #[cfg(debug_assertions)]
+    fn tx_buffered_actual(&self) -> usize {
+        self.streams.values().map(|s| s.send.len() as usize).sum()
```
LPardue: nit: micro optimization might be to filter out peer uni streams?

quiche/src/stream/mod.rs
```diff
@@ -721,6 +724,69 @@ impl<F: BufFactory> StreamMap<F> {
...
+
+    /// Verifies that the cached tx_buffered matches the actual value.
+    /// Enabled in debug builds to catch cache inconsistencies early.
+    #[cfg(debug_assertions)]
+    pub(crate) fn check_tx_buffered_cache(&self) {
+        let actual = self.tx_buffered_actual();
+        let cached = self.tx_buffered_cache;
+        if actual != cached {
+            // Print per-stream details to help debug
+            eprintln!("Stream buffer details:");
```
LPardue: nit: why eprintln and not log::error?

quiche/src/stream/mod.rs
```diff
@@ -2204,6 +2270,200 @@ mod tests {
             prioritized_writable.iter().map(|s| s.id).collect();
         assert_eq!(walk_2, vec![0, 0, 4, 4, 8, 8, 12, 12]);
     }
+
+    #[test]
```
LPardue, line 2273: observation: its probably time we schedule moving the stream tests out of the stream/mod.rs too, getting to be > 50% LoC is tests. This can be a follow on

quiche/src/stream/send_buf.rs
```diff
@@ -354,27 +354,39 @@ impl<F: BufFactory> SendBuf<F> {
         }
 
         if let Some(drop) = drop_until {
+            // Calculate the total length of buffers being dropped and subtract
+            // from len.
+            let dropped_len: u64 =
+                (0..=drop).map(|i| self.data[i].len() as u64).sum();
+            self.len = self.len.saturating_sub(dropped_len);
```
LPardue: opinion:  maybe now might be the time to rename `self.len` to something more explicit like `buffered_len` and the related accessor to match.

quiche/src/lib.rs
```diff
@@ -3698,6 +3685,13 @@ impl<F: BufFactory> Connection<F> {
                             let local = stream.local;
                             self.streams.collect(stream_id, local);
                         }
+
+                        // Update cache to reflect any data that was dropped from
+                        // buffers (e.g., data marked for retransmission but then
+                        // acked before it could be resent).
+                        if dropped > 0 {
+                            self.streams.sub_tx_buffered(dropped);
+                        }
```
LPardue, line 3695: nit: the check for > 0 would probably be optimized out for release builds anyway. For debug builds you probably want to avoid pointless work but consider sticking the check inside the debug function instead. That would make the connection-level logic a bit cleaner for folks that don't care too much about the tream specifics

quiche/src/lib.rs
```diff
@@ -6282,14 +6283,19 @@ impl<F: BufFactory> Connection<F> {
...
+
                 let (final_size, unsent) = stream.send.shutdown()?;
 
                 // Claw back some flow control allowance from data that was
                 // buffered but not actually sent before the stream was reset.
                 self.tx_data = self.tx_data.saturating_sub(unsent);
 
-                self.tx_buffered =
-                    self.tx_buffered.saturating_sub(unsent as usize);
+                // Update cache: subtract only the buffered data, not inflight
```
LPardue: nit: there's a few comments that refer to a `cache` but really its a size of a $thing not the thing itself. Maybe use a dfferent or additional word?

LPardue (approved):
Over LGTM, just a bunch of nits or opinions that I trust you to address or reject

quiche/src/stream/mod.rs
antoniovicente: Done

quiche/src/stream/mod.rs
antoniovicente: Switched over to calling it tx_buffered and removed most references to "cache"

quiche/src/stream/mod.rs
antoniovicente: Refactored to printing a list of `(stream_id, bytes_buffered)` as part of the panic message.

print removed.

quiche/src/lib.rs
antoniovicente: `dropped > 0` is doesn't happen very often but I have seen it happen in integration tests.

I'm fine just doing the sub_tx_buffered call unconditionally if that's preferable.

quiche/src/lib.rs
antoniovicente: Changed most of these from cache to tx_buffered.

quiche/src/stream/mod.rs
antoniovicente: Don't know which would be faster since `.len()` is a trivial accessor.  I think it depends on how smart the compiler is at inlining the map and sum accumulator.

quiche/src/stream/send_buf.rs
antoniovicente: Done.

quiche/src/stream/mod.rs
antoniovicente: I'll make a note for a followup

quiche/src/stream/mod.rs
```diff
@@ -721,6 +724,68 @@ impl<F: BufFactory> StreamMap<F> {
...
+            .values()
+            .map(|s| s.send.buffered_bytes() as usize)
+            .sum()
+    }
+
+    /// Checks if the stored tx_buffered matches the actual value.
+    /// Returns true if they match, false otherwise.
+    pub(crate) fn tx_buffered_is_consistent(&self) -> bool {
+        self.tx_buffered == self.tx_buffered_actual()
+    }
```
estcarisimo, line 746: i might lack some background here but when can these numbers be inconsistent?

estcarisimo (approved):
A minor comment to understand it. I already saw the QLOG plots offline and the change looks consistent

quiche/src/stream/mod.rs
antoniovicente: One of my earlier fix attempts missed the `send.ack_and_drop` case, and quiche tests that we had didn't catch the problem.  Download integration tests triggered the debug_assert and that helped me track down the remaining inconsistency.

Hopefully none remain, but if they do we'll detect them with the help of debug checks and stats monitoring.

PR #2507, opened 2026-06-07
Return `InvalidStreamState` from `stream_send()` for collected streams

### LINKED ISSUE #1695: Connection::stream_send() reports Err::Done for finished streams
This case occurs with streams closed with STOP_SENDING, then a fin bit being received too, then the data being read via `Connection::stream_recv()`.

In high-level terms:
```
server opens stream - sends some data
client reads some data and writes data
server reads that data, sends STOP_SENDING, and writes some data, then sends the fin bit in a separate empty message of length 0
client sees stream as readable, then reads the data
client tries to write some data ... gets Err::Done from stream_send().
```

As far as I gathered from the documentation, `Err::Done` is supposed to be returned when no data could be written because of no capacity and `Err::StreamStopped` returned when the stream has been closed by the peer.

In this specific case though, the stream seems to already have been freed by quiche, thus `StreamMap::get_or_create()` returns `Err::Done`, which `Connection::stream_send()` simply forwards.

I would expect an `Err::StreamStopped`, or at least an `Err::InvalidStreamState` (signaling that a stream with that id could not be created, in that case, obviously, because the stream id is not matching what's allowed for the client/server).

Currently I'm working around it with an extra check for `Err::InvalidStreamState` whenever `Err::Done` is reported, but that seems unnecessary/wrong to me, why would that be needed?

In case this is the expected behaviour though, I would like to ask for a note in the documentation to check stream_writable() explicitly if `Err::Done` is returned.

schiemon:
When a stream has been completed and garbage collected (for example after the peer sends STOP_SENDING, the fin bit is received, and the data is drained via stream_recv()), a subsequent stream_send() call hits get_or_create() for an id that is in the collected set, which returns Error::Done. That error was forwarded unchanged, so callers saw Done (normally "no capacity") for a stream that no longer exists.

Map that case to Error::InvalidStreamState instead, matching the behaviour of stream_recv() which already returns InvalidStreamState for a non-existent stream. This lets applications distinguish a lack of send capacity from an invalid stream id without an extra stream_writable() check.

Fixes #1695

### DIFF
--- a/quiche/src/tests.rs
+++ b/quiche/src/tests.rs
@@ -3473,6 +3473,38 @@ fn stop_sending_fin(
 }
 
 #[rstest]
+/// Tests that stream_send() on a collected stream returns InvalidStreamState
+/// rather than Done (issue #1695).
+fn stream_send_after_stop_sending_and_fin(
+    #[values("cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
+) {
+    let mut pipe = test_utils::Pipe::new(cc_algorithm_name).unwrap();
+    assert_eq!(pipe.handshake(), Ok(()));
+
+    // Server opens a bidi stream and sends data with fin.
+    assert_eq!(pipe.server.stream_send(5, b"hello", true), Ok(5));
+
+    // Server also sends STOP_SENDING, indicating it will not read from the
+    // client on this stream.
+    assert_eq!(pipe.server.stream_shutdown(5, Shutdown::Read, 42), Ok(()));
+
+    // Exchange all packets: client receives the STOP_SENDING and STREAM(fin).
+    assert_eq!(pipe.advance(), Ok(()));
+
+    // Client reads the data. The stream is now complete on both sides (recv
+    // finished, send stopped), so quiche collects it.
+    let mut buf = [0; 64];
+    assert_eq!(pipe.client.stream_recv(5, &mut buf), Ok((5, true)));
+
+    // The stream has been collected. stream_send() must return
+    // InvalidStreamState, not Done.
+    assert_eq!(
+        pipe.client.stream_send(5, b"world", false),
+        Err(Error::InvalidStreamState(5)),
+    );
+}
+
+#[rstest]
 /// Tests that resetting a stream restores flow control for unsent data.
 fn stop_sending_unsent_tx_cap(
     #[values("cubic", "bbr2_gcongestion")] cc_algorithm_name: &str,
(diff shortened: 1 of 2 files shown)


quiche/src/tests.rs
```diff
@@ -3472,6 +3472,38 @@ fn stop_sending_fin(
     assert_eq!(iter.next(), None);
 }
 
+#[rstest]
```
1ritagreeson, line 3475: MHC 
MEETING ON IDIVIDUAL CARE AND ASSEMBY INTAKE 
IN ORDER OF ASSIGNED HOUSED UNIT AND DORM QUARTERS

1ritagreeson (approved):

quiche/src/tests.rs
1ritagreeson: RECORD FILE SEGMENT AND MERGER

quiche/src/tests.rs
1ritagreeson, line 11965: CONSULAR READMITANCE ACQUIRED IN ORDEROF FORMULARS AS DESCRIBED IN PROLIFIC POETFOLIO CONTOURANCES.
EACH HABTAT IS OF EQUALD
 DISCERNS.
JUNE 2026

1ritagreeson, line 11975: SER SET AS INSCRIBED

1ritagreeson, line 12032: "PATHS" ALL CLEAR LITTER AND DEBRIS REMOVED

1ritagreeson, line 12047: ESTABLISHED IN ONSET OF PARTICLE REVISIONS ANULIFIED IN COMPLETION IF TASK ON MICROWARE OR GENERAL MAC SUBSIDY.

JUNE 2026

1ritagreeson, line 12151: CREDITS REFURBISHED ,TASKS COMPLETED

JUNE 2026

1ritagreeson, line 12176: ACTUAL PATH UNREVISED IN MERGE OFDOCUMENTATED REFERENCE CALL

1ritagreeson, line 12176: WRAP CONCLUDED

JUNE 2026

1ritagreeson, line 12196: STIMULI CONVERSIVE IN MATTER OF REOCCURENCES  FOREIGN GRIMACE ASTUTES COMPREHENSED MALWARE

1ritagreeson, line 12199: SENIORITY SPECIFIED AS IS;
WARRANTY DISCLOSED

1ritagreeson, line 12211: career building negatiated proper rule,proper method in order of presented substantiated reference to hire ,to dismissal, to origins, set or unset 
These priorities abound in reasoning of forwards and justifiable straitsasmentionedin guide and handbook of legislates or other alumni

1ritagreeson, line 12236: MAX,(MAC);(IPHONE)(IPAD)
A
CONSTRUENT UNALTERED CONSTRINGENTS

JUNE 2026

quiche/src/tests.rs
schiemon: thanks for the input

1ritagreeson (approved):

Issue #2509, opened 2026-06-10
C FFI: quiche_connection_id_iter_next / quiche_conn_retired_scid_next return dangling pointers (use-after-free)


GrapeBaBa:
## Summary

Two C-API functions in `src/ffi.rs` set their out-pointer into a `ConnectionId` that is dropped before the call returns, handing the caller a dangling pointer into a freed heap buffer:

- `quiche_connection_id_iter_next`
- `quiche_conn_retired_scid_next`

Rust users are unaffected (the borrow checker prevents this usage); every C-API consumer of these two functions reads freed memory. Present in 0.28.0 and still on current `master`.

## Details

`ConnectionIdIter::next` returns an **owned clone** of the stored id:

```rust
impl<'a> Iterator for ConnectionIdIter<'a> {
    type Item = ConnectionId<'a>;

    fn next(&mut self) -> Option<Self::Item> {
        let v = self.cids.get(self.index)?;
        self.index += 1;
        Some(v.clone())   // owned ConnectionId (Cow::Owned for ids cloned out of the connection)
    }
}
```

`quiche_connection_id_iter_next` then points `*out` into that clone's buffer:

```rust
pub extern "C" fn quiche_connection_id_iter_next(
    iter: &mut ConnectionIdIter, out: &mut *const u8, out_len: &mut size_t,
) -> bool {
    if let Some(conn_id) = iter.next() {
        let id = conn_id.as_ref();
        *out = id.as_ptr();      // pointer into the clone's Vec
        *out_len = id.len();
        return true;
    }                            // <- conn_id dropped here, Vec freed, *out dangles
    false
}
```

`quiche_conn_retired_scid_next` has the identical defect with the owned id popped from the retired queue:

```rust
pub extern "C" fn quiche_conn_retired_scid_next(
    conn: &mut Connection, out: &mut *const u8, out_len: &mut size_t,
) -> bool {
    match conn.retired_scid_next() {
        None => false,
        Some(conn_id) => {
            let id = conn_id.as_ref();
            *out = id.as_ptr();
            *out_len = id.len();
            true                 // <- conn_id dropped at scope end, *out dangles
        },
    }
}
```

## Observed behaviour (how we found it)

[...]

PR #2510, opened 2026-06-10
ffi: fix dangling pointer in quiche_connection_id_iter_next

GrapeBaBa:
`quiche_connection_id_iter_next()` sets `*out` to a pointer into the `ConnectionId` clone returned by `Iterator::next()`, which is dropped before the function returns — so C callers read freed memory.

Borrow from the iterator's own storage instead; the pointer then stays valid until `quiche_connection_id_iter_free()`, and the header documents that lifetime.

Part of #2509. `quiche_conn_retired_scid_next` has the same problem, but the popped id has no owner to borrow from, so fixing it needs an API decision — left out of this PR.

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -495,7 +495,8 @@ typedef struct quiche_connection_id_iter quiche_connection_id_iter;
 quiche_connection_id_iter *quiche_conn_source_ids(quiche_conn *conn);
 
 // Fetches the next id from the given iterator. Returns false if there are
-// no more elements in the iterator.
+// no more elements in the iterator. `out` stays valid until the iterator
+// is freed.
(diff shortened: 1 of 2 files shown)


ghedo:
Thank you for the PR. We've already been tracking this issue (and a related one) internally and were preparing to publish a security advisory (see https://github.com/cloudflare/quiche/security/advisories/GHSA-mh64-ph39-mrc9).

We've ended up going with our own fixes since they are more comprehensive (https://github.com/cloudflare/quiche/commit/386ad632bef063aea503be1819f5fda503950876 and https://github.com/cloudflare/quiche/commit/386ad632bef063aea503be1819f5fda503950876), but we appreciate your help nonetheless!!

PR #2513, opened 2026-06-18
ci: require quiche release draft

ghedo:
Quiche release PRs are opened before the release tag exists, so check for a GitHub draft release using the version from quiche/Cargo.toml. The workflow is scoped to quiche version bumps and uses contents: write so the token can see draft releases.


dineshadhi (approved):
lgtm

jannes (approved):

PR #2514, opened 2026-06-19
Implement PKTS_BETWEEN_PROBES for PMTUD

jeniksv:
Add configurable spacing between PMTUD probes. When configured, PMTUD waits until the given number of non-probe packets has been sent before allowing the next probe.

This prevents lost probes on idle connections from inflating the in-flight byte count, triggering PTO, and causing further probes. Without spacing, this can create a self-reinforcing loop that escalates the PTO exponent and generates unnecessary traffic.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -4004,7 +4004,7 @@ impl<F: BufFactory> Connection<F> {
 
         let send_path = self.paths.get_mut(send_pid)?;
 
-        // Update max datagram size to allow path MTU discovery probe to be sent.
+        // Increase output size to allow path MTU discovery probe to be sent.
         if let Some(pmtud) = send_path.pmtud.as_mut() {
             if pmtud.should_probe() {
                 let size = if self.handshake_confirmed || self.handshake_completed
@@ -4014,10 +4014,7 @@ impl<F: BufFactory> Connection<F> {
                     pmtud.get_current_mtu()
                 };
 
-                send_path.recovery.pmtud_update_max_datagram_size(size);
-
-                left =
-                    cmp::min(out.len(), send_path.recovery.max_datagram_size());
+                left = cmp::min(out.len(), size);
             }
         }
 
@@ -4296,6 +4293,9 @@ impl<F: BufFactory> Connection<F> {
                             if let Some(pmtud) = p.pmtud.as_mut() {
                                 trace!("pmtud probe dropped: {failed_probe}");
                                 pmtud.failed_probe(failed_probe);
+                                p.recovery.pmtud_update_max_datagram_size(
+                                    pmtud.get_current_mtu(),
+                                );
                             }
                         }
                     },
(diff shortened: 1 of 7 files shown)


quiche/src/lib.rs
```diff
@@ -795,6 +797,14 @@ impl Config {
         self.pmtud_max_probes = max_probes;
     }
 
+    /// Configures the number of non-probe packets that must be sent between
+    /// PMTUD probes.
+    ///
+    /// The default value is `0`.
+    pub fn set_pmtud_pkts_between_probes(&mut self, pkts: usize) {
```
cf-rhett: Why default 0? If lost probes really escalate into a self-reinforcing PTO loop, opt-in-off means nobody's protected unless they know to flip it. Is 0 just conservatism, or is the loop rare enough in practice that opt-in is the right call?

quiche/src/pmtud.rs
```diff
@@ -75,18 +88,28 @@ impl Pmtud {
...
-    /// found, and that the minimum supported MTU has not been reached.
+    /// found, that the minimum supported MTU has not been reached, and that
+    /// enough non-probe packets have been sent since the last probe.
     pub fn should_probe(&self) -> bool {
         !self.in_flight &&
             self.pmtu.is_none() &&
-            self.smallest_failed_probe_size != Some(MIN_PLPMTU)
+            self.smallest_failed_probe_size != Some(MIN_PLPMTU) &&
+            self.pkts_since_last_probe
+                .is_none_or(|n| n >= self.pkts_between_probes)
```
cf-rhett: More design question than objection: RFC 8899 spaces probes by *time* (≥1 RTT floor, `PROBE_TIMER`) or the idle-inhibit `MAY` in §4.3 ("no app data since previous probe") — not a packet count. And §3.7 is basically this bug: "Loss of a probe packet SHOULD NOT be treated as an indication of congestion and SHOULD NOT trigger a congestion control reaction." So the spec's own remedy is to keep lost probes out of the CC/PTO reaction rather than throttling probe cadence. Was a packet count chosen over either of those on purpose? It isn't non-conformant (§3 only mandates the ≥1-RTT floor), it just feels like it's damping the symptom while probe-loss-as-data-loss stays intact.

jeniksv: Packet count was not chosen over either of those. The intent is to add an additional configurable condition on top of the existing PMTUD probing logic, not to replace the RFC 8899 time-based spacing or idle-inhibit behavior.

This idea is taken from Google QUICHE, where packets_between_probes exists: QuicConnectionMtuDiscoverer uses a packet-count threshold, starting from a base value and doubling it after each MTU probe attempt before scheduling the next probe.

So this should be read as an extra configurable cadence guard, not as the primary RFC 8899 remedy for probe-loss handling.

quiche/src/lib.rs
jeniksv: The default was chosen to preserve the existing behavior of the library.

Use case that motivated this change is a connection that sends all application data early, then stays open mostly idle until it is closed because of a long idle timeout. At that point, sending additional PMTUD probes is usually not useful, because the connection is unlikely to send more data. In that situation, continued probing can just keep exercising the PTO/loss path without providing much benefit.

PR #2516, opened 2026-06-22
quiche: always use initial_max_data as flow control window

LPardue:
The initial flow control window for connections and streams is now
always set to initial_max_data / initial_max_stream_data respectively,
instead of being clamped to small defaults (48 KiB / 32 KiB).

The Config and Connection setter methods are preserved as deprecated
no-ops to avoid a hard API break. The C FFI function is similarly
kept as a no-op. The tokio-quiche QuicSettings field remains but is
ignored. All internal plumbing (Config field, StreamMap field, ExData
field, handshake callback path) is removed.

### DIFF
--- a/quiche/src/stream/mod.rs
+++ b/quiche/src/stream/mod.rs
@@ -46,9 +46,6 @@ use crate::Result;
 
 const DEFAULT_URGENCY: u8 = 127;
 
-/// The default size of the receiver stream flow control window.
-pub(crate) const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
-
 /// The maximum size of the receiver stream flow control window.
 pub const MAX_STREAM_WINDOW: u64 = 16 * 1024 * 1024;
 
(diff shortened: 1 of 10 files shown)


antoniovicente (approved):
Thanks for the cleanup.

jannes (approved):

quiche/src/stream/mod.rs
```diff
@@ -47,7 +47,8 @@ use crate::Result;
 const DEFAULT_URGENCY: u8 = 127;
 
 /// The default size of the receiver stream flow control window.
-pub(crate) const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
+#[cfg(test)]
+const DEFAULT_STREAM_WINDOW: u64 = 32 * 1024;
```
ghedo: Nit: could maybe just move this into `mod tests` instead of gating it here.

LPardue: done

PR #2517, opened 2026-06-24
cid: improve DCID retirement handling

LPardue:
The `dcids` ConnectionIdVecDeque is not guaranteed to be sorted, so
plainly applying `self.dcids.inner.partition_point(...)` can lead to
incorrect result. While we could introduce such a guarantee, it would
require spending more cycles when handling out-of-order sequence numbers.

Instead, this change checks all elements in O(n) time. While it is
possible to optimize further to take advantage of an already-sorted
list, the list is expected to be short. The additional complexity was
determined to unnecessary at this time.

This also fixes a bug where path IDs were not added to
`retired_path_ids` after `retired.insert()` returned an error.

### DIFF
--- a/quiche/src/cid.rs
+++ b/quiche/src/cid.rs
@@ -197,6 +197,25 @@ impl BoundedNonEmptyConnectionIdVecDeque {
             .position(|e| e.seq == seq)
             .and_then(|index| self.inner.remove(index)))
     }
+
+    /// Removes all elements in the collection with sequence numbers less than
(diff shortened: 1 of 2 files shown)


antoniovicente (approved):

dineshadhi (approved):
lgtm

ghedo:
Nit: `quiche: ...` -> `cid: ...` in the commit message :)

PR #2520, opened 2026-06-25
qlog: Add RESET_STREAM_AT

martinthomson:
This minimally adds the definition of the frame from https://quicwg.org/reliable-stream-reset/draft-ietf-quic-reliable-stream-reset.html

I couldn't see any tests for this stuff, so I'm assuming that this is all that is needed.

### DIFF
--- a/qlog/src/events/quic.rs
+++ b/qlog/src/events/quic.rs
@@ -479,6 +479,7 @@ pub enum QuicFrameTypeName {
     Ping,
     Ack,
     ResetStream,
+    ResetStreamAt,
     StopSending,
     Crypto,
     NewToken,
@@ -534,6 +535,16 @@ pub enum QuicFrame {
         error: ApplicationError,
         error_code: Option<u64>,
         final_size: u64,
+
+        raw: Option<Box<RawInfo>>,
+    },
+
+    ResetStreamAt {
+        stream_id: u64,
+        error: ApplicationError,
+        error_code: Option<u64>,
+        final_size: u64,
+        reliable_size: u64,
 
         raw: Option<Box<RawInfo>>,
     },
--- a/qlog/src/events/quic.rs
+++ b/qlog/src/events/quic.rs
@@ -539,6 +539,7 @@ pub enum QuicFrame {
         raw: Option<Box<RawInfo>>,
     },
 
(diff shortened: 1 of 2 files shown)


qlog/src/events/quic.rs
```diff
@@ -538,6 +539,16 @@ pub enum QuicFrame {
         raw: Option<Box<RawInfo>>,
     },
 
+    ResetStreamAt {
+        stream_id: u64,
+        error: ApplicationError,
+        error_code: Option<u64>,
```
LPardue, line 546: This is inconsistent with the definition in the editor's copy https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-8.13.5 where the field is `error_code_bytes`.

It smells like the spec is using an inconsistent field name with https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-8.13.4, rather than your PR being wrong. However, we should ensure the spec is updated before we cut draft 13 (that would introduce reset_stream_at)

LPardue:
Tests?! 

image

qlog/src/events/quic.rs
LPardue: This is fixed in latest drafts, the PR is correct.

LPardue (approved):

qlog/src/events/quic.rs
```diff
@@ -538,6 +539,16 @@ pub enum QuicFrame {
         raw: Option<Box<RawInfo>>,
     },
 
+    ResetStreamAt {
```
antoniovicente, line 543: This could be a comment about the draft more than the PR, but I had the question about what this message represents.  It seems like the difference between this and ResetStreamFrame is the addition of `reliable_size`

I eventually ran into draft-ietf-quic-reliable-stream-reset which has more info on the definition of the frame.

I would find it useful to have a link to the relevant draft version in a comment in case we need to check consistency of these structs vs newer drafts as they are published.

antoniovicente (approved):
Change looks good.  Thanks for the PR!

qlog/src/events/quic.rs
martinthomson:
```suggestion
    /// See https://datatracker.ietf.org/doc/html/draft-ietf-quic-reliable-stream-reset/#section-4
    ResetStreamAt {
```

PR #2531, opened 2026-07-02
FFI fix - fix missing set for max_rtt in PathStats

Magistone:
While trying to acquire statistics for the state of the connection I have noticed that `max_rtt` was always 0, and larger than `min_rtt`. After digging through, I have noticed that the FFI function `quiche_conn_path_stats` was missing a setter for the `max_rtt` field resulting in max_rtt always being 0 when used via FFI.

This commit fixes the aforementioned issue


antoniovicente (approved):

ghedo (approved):
Thanks! LGTM.

PR #2536, opened 2026-07-09
fix: remove MAX_PTO_EXPONENT

jannes:
Use saturating exponentiation instead which also prevents arithmetic overflow in a simpler manner.

### DIFF
--- a/quiche/src/recovery/gcongestion/recovery.rs
+++ b/quiche/src/recovery/gcongestion/recovery.rs
@@ -38,7 +38,6 @@ use crate::recovery::INITIAL_PACKET_THRESHOLD;
 use crate::recovery::INITIAL_TIME_THRESHOLD;
 use crate::recovery::MAX_OUTSTANDING_NON_ACK_ELICITING;
 use crate::recovery::MAX_PACKET_THRESHOLD;
-use crate::recovery::MAX_PTO_EXPONENT;
 use crate::recovery::MAX_PTO_PROBES_COUNT;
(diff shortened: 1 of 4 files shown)


jannes:
grrrr unreated CI failures

TheJokr (approved):

jannes:
> grrrr unreated CI failures

another MR that's fixing them: https://github.com/cloudflare/quiche/pull/2535

quiche/src/recovery/gcongestion/recovery.rs
```diff
@@ -614,8 +613,7 @@ impl GRecovery {
     fn pto_time_and_space(
         &self, handshake_status: HandshakeStatus, now: Instant,
     ) -> (Option<Instant>, packet::Epoch) {
-        let mut duration =
-            self.pto() * 2_u32.pow(self.pto_count.min(MAX_PTO_EXPONENT));
+        let mut duration = self.pto() * 2_u32.saturating_pow(self.pto_count);
```
antoniovicente, line 616: This fix is incomplete.  We need saturing_mul  between pto and the result of the saturating_pow

antoniovicente: I see the error in my reasoning now.  The pto and max_ack_delay are relatively small numbers, not anywhere close to seconds since the unix epoch.  Multiplying by a large u32 number can't result in u64 seconds overflow inside the Duration object.

antoniovicente (approved):

LPardue (approved):

PR #2537, opened 2026-07-09
qlog july id parity update

LPardue:
This PR updates the qlog crate and its dependents to have parity with the following drafts:

* https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html
* https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html
* https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html

It is composed of several commits to aid review. Suggest squashing them all into one commit on merge

- **qlog: add Event tuple/system_info fields and SystemInformation**
- **qlog: drop CommonFields.protocol_types, make reference_time optional**
- **qlog: loglevel info/debug/verbose carry only a required message**
- **qlog: split frames_processed into frames_created/frames_parsed**
- **qlog: align ConnectionClosedTrigger with quic-events-13**
- **qlog: rename connection_id_updated owner to required initiator**
- **qlog: fix packet_received trigger to keys_available**
- **qlog: add duplicate and key_unavailable packet_dropped triggers**
- **qlog: represent CryptoError as a crypto_error_0x1XX string**
- **qlog: correct HTTP/3 ApplicationError string values**
- **qlog: add HTTP/3 ORIGIN frame**
- **qlog: move HTTP/3 frame length into frame raw info**
- **qlog: format mark_closed connection close mapping**
- **qlog: drop datagram field from HTTP/3 datagram events**
- **qlog: add HTTP Capsule Protocol event schema**
- **qlog: bump event schema URIs to latest drafts**
- **qlog: use distinct VantagePointFlow type**


quiche/src/lib.rs
```diff
@@ -9191,7 +9191,7 @@ impl<F: BufFactory> Connection<F> {
                     error_code: None,
                     internal_code: None,
                     reason: Some("Failed to establish connection".to_string()),
-                    trigger: Some(qlog::events::quic::ConnectionClosedTrigger::HandshakeTimeout)
+                    trigger: Some(qlog::events::quic::ConnectionClosedTrigger::IdleTimeout)
```
antoniovicente, line 9194: hmm, I would have thought it would be useful to distinguish between a timeout experienced during the handshake vs post handshake.  I assume that there are ways to tell if handshake is completed, in which case the distinction doesn't seem as important.

I do agree that this is consistent with the more recent version of the draft.

qlog/src/events/mod.rs
```diff
@@ -683,7 +683,7 @@ pub enum ApplicationError {
     HttpGeneralProtocolError,
     HttpInternalError,
     HttpRequestCancelled,
-    HttpIncompleteRequest,
+    HttpRequestIncomplete,
```
antoniovicente, line 686: The order of these doesn't seem to match the order in the draft which makes it harder to verify that the two lists are consistent.

antoniovicente (approved):
Sorry for the delays in review.  Checked each commit vs the drafts for correctness at the best of my ability, but I can't verify that some recent change to drafts was missed as part of this change.

This does seem like an important change to get the definitions and implementation more consistent with the drafts.  It does involve some API changes, but I assume that app code will be updated in a timeline fashion.

PR #2540, opened 2026-07-15
quiche: gate stopped-stream collection on is_stopped(), not writable linkage

vendemiat:
quiche: gate stopped-stream collection on is_stopped(), not writable linkage
recv_single()'s ack-processing arms for STREAM-data and RESET_STREAM
decided whether to collect a completed stream using
`priority_key.writable.is_linked() && stream.send.is_stopped()`. The
intent was to defer collecting a peer-stopped stream so the application
can still observe `StreamStopped` on its next write attempt, rather than
having the stream silently garbage-collected out from under it.

Gating on `writable.is_linked()` is wrong: the writable queue is only a
delivery mechanism for `stream_writable_next()`, which unlinks whatever
stream it returns on every poll. An IoWorker calls `stream_writable_next()`
via `process_writes()` on each loop iteration, immediately after
`process_reads()` and before any network round trip. So a stream stopped
by a STOP_SENDING frame is unlinked again long before the RESET_STREAM
ack that runs these arms arrives, leaving `is_linked()` false and causing
the stream to be collected prematurely. The application's next write then
sees a generic `Done`/`InvalidStreamState` instead of `StreamStopped`,
silently losing the peer's cancellation signal.

Gate on `stream.send.is_stopped()` alone. A stopped stream is deferred
until the application observes `StreamStopped`, independent of the
transient writable-queue link state.

Adds a quiche-core regression test that reproduces the ordering directly
via Pipe: poll stream_writable_next() to unlink the stopped stream, then
deliver the RESET_STREAM ack, and check the next stream_send() returns
StreamStopped instead of Done.

Also adds two tokio-quiche H3 driver regression tests that reproduce the
same failure through the driver: one via the ordinary advance-and-run
loop, and one that hand-stages the real IoWorker ordering with hop-by-hop
packet delivery (emit_flight/process_flight interleaved with direct
process_reads()/process_writes() calls).

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3624,35 +3624,39 @@ impl<F: BufFactory> Connection<F> {
                         };
 
                         let dropped = stream.send.ack_and_drop(offset, length);
-                        let priority_key = Arc::clone(&stream.priority_key);
 
                         // Only collect the stream if it is complete and not
-                        // readable or writable.
+                        // readable or stopped.
                         //
                         // If it is readable, it will get collected when
                         // stream_recv() is next used.
                         //
-                        // If it is writable, it might mean that the stream
-                        // has been stopped by the peer (i.e. a STOP_SENDING
-                        // frame is received), in which case before collecting
-                        // the stream we will need to propagate the
+                        // If the stream has been stopped by the peer (i.e. a
+                        // STOP_SENDING frame is received), before collecting
+                        // the stream we need to propagate the
                         // `StreamStopped` error to the application. It will
                         // instead get collected when one of stream_capacity(),
                         // stream_writable(), stream_send(), ... is next called.
                         //
                         // Note that we can't use `is_writable()` here because
                         // it returns false if the stream is stopped. Instead,
-                        // since the stream is marked as writable when a
-                        // STOP_SENDING frame is received, we check the writable
-                        // queue directly instead.
-                        let is_writable = priority_key.writable.is_linked() &&
-                            // Ensure that the stream is actually stopped.
-                            stream.send.is_stopped();
-
+                        // we check `stream.send.is_stopped()` directly. We
+                        // can't check the writable queue instead:
+                        // `stream_writable_next()` unlinks the stream from
+                        // the writable queue on its very next poll, which
+                        // happens every `IoWorker` iteration before any
+                        // network round trip completes. By the time an ack
+                        // for a stopped stream arrives, the stream is
+                        // already unlinked from the writable queue
+                        // regardless of whether the app observed it as
+                        // writable.
                         let is_complete = stream.is_complete();
                         let is_readable = stream.is_readable();
 
-                        if is_complete && !is_readable && !is_writable {
+                        if is_complete &&
+                            !is_readable &&
+                            !stream.send.is_stopped()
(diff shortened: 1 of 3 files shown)


vendemiat:
I ran into this issue while adding DoQ support to tokio-quiche. In DoQ we get a DNS query and a FIN in a stream, which triggers this bug.

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -844,6 +844,52 @@ mod server_side_driver {
         assert_eq!(audit_stats.downstream_bytes_sent(), 0);
     }
 
+    /// Test the case where the server's outbound channel closes and
+    /// `audit_stats` records the STOP_SENDING error code from the client
+    /// sent before the server has written any response to a `fin=true` request
+    #[test]
+    fn client_sends_stop_sending_before_first_write() {
```
ghedo, line 1234: I don't think this tests actually reproduces the behaviour of `IoWorker`?

AI seems to agree:

> advance_and_run_loop() calls Pipe::advance() before running the driver. Pipe::advance() exchanges packets 
>
> until idle, producing:
> receive STOP_SENDING
> send RESET_STREAM
> receive ACK for RESET_STREAM
> run process_writes()
>
> The real worker at tokio-quiche/src/quic/io/worker.rs:238-265 instead does:
>
> receive STOP_SENDING
> run process_reads()
> run process_writes()
> send RESET_STREAM
> receive ACK later
>
> During that write pass, stream_writable_next() removes the PR’s new writable link at quiche/src/lib.rs:6413-6424. Since no response is queued, process_write_frame() returns without calling a transport API at tokio-quiche/src/http3/driver/mod.rs:727-733. Nothing observes StreamStopped.
>
> When the RESET_STREAM ACK arrives, the stream is no longer linked. The ACK handler at quiche/src/lib.rs:3677-3715 therefore collects the completed stream. A later application response sees an invalid/missing stream, and the STOP_SENDING error code remains unrecorded.

So I think we'd need a test that follows the actual `IoWorker` behaviour more closely.

Would also be good to add a targeted test to quiche itself using stream_writable_next() to cover the actual transport bug.

vendemiat: You were right. The old fix was not enough and the test didn't correctly capture H3 and DoQ issue. Can you review the current proposal. Also added 3 tests.

quiche/src/lib.rs
```diff
@@ -3624,35 +3624,39 @@ impl<F: BufFactory> Connection<F> {
...
+                        // already unlinked from the writable queue
+                        // regardless of whether the app observed it as
+                        // writable.
                         let is_complete = stream.is_complete();
                         let is_readable = stream.is_readable();
 
-                        if is_complete && !is_readable && !is_writable {
+                        if is_complete &&
+                            !is_readable &&
+                            !stream.send.is_stopped()
```
antoniovicente: Is the polarity of this check correct?

Seems that if the stream is stopped, we can never collect.  Original code is also odd: I would have thought that `send.is_stopped()` should imply that the stream is not writable; you shouldn't write on a stopped stream.

antoniovicente: I think I get it now.

PR #2542, opened 2026-07-16
Customizable variadic length connection id for client

### LINKED ISSUE #2541: feature: Customizable client-side connection id and non-20-byte connection id support for shorter connection id for smaller overhead
This issue is technically consists of two features, but targeting single problem so writing as single issue.

I'm planned to use quiche with unreliable extension for real-time communication and want to reduce packet size overhead.
Therefore I want to use shorter connection id for client side (and reject longer in on_conn_established).

However, currently tokio-quiche doesn't support changing length of connection id in `InboundPacketRouter` and doesn't allow specifying scid or cid generator.

Therefore I want  to add support changing length of connection id to tokio-quiche.

anatawa12:
Closes #2541


zaczh:
Verified this against my use case — a MASQUE proxy on tokio-quiche 0.19.1 with an 8-byte server CID generator. With `scid_len = 8` CONNECT-UDP round-trips fine including near-MTU datagrams; without it short-header packets get misrouted at the hardcoded `MAX_CONN_ID_LEN`, so the handshake completes but no data flows.

One API thought: `scid_len` duplicates info the router can already reach. It holds the `SharedConnectionIdGenerator`, so the length could be derived from the generator instead of a separate field the caller has to keep in sync:

```rust
// defaulted method on ConnectionIdGenerator
fn cid_len(&self) -> usize { MAX_CONN_ID_LEN }
```

then `short_dcid(buf, self.cid_generator.cid_len())`. Drops the mismatch footgun the field doc already warns about, and stays non-breaking. Needs to be a method rather than an assoc const since the generator is used behind `dyn`.

anatawa12:
@zaczh Thank you for feedback! My first use case was reducing client scid so directly using cid_len from `ConnectionIdGenerator` was not possible, but we have ConnectionId itself so we changed the code to derive scid_size from actual ConnectionId for client and `ConnectionIdGenerator` for server endpoint instead providing as separated parameter

PR #2545, opened 2026-07-16
tokio-quiche: make logging calls statements

ghedo:
Nightly Rust now denies logging macros in expression position because their expansions contain trailing semicolons. Wrap the affected calls in blocks and terminate each macro explicitly.

PR #2546, opened 2026-07-16
Tokio-quiche: lazily allocate H3 body_recv_buf

LynnLi100:
H3Driver eagerly allocated a 64 KiB body receive buffer per connection at construction. Make it Option<Limit<BytesMut>>, allocate it lazily on the first body read, and release it once no streams or flows remain, so idle connections hold no receive buffer.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -350,9 +350,11 @@ pub struct H3Driver<H: DriverHooks> {
     dgram_recv: OutboundFrameStream,
     /// Keeps the datagram channel open such that datagram flows can be created.
     dgram_send: OutboundFrameSender,
-    /// A buffer to receive H3 body data from quiche. We initialize a large
-    /// buffer and then `split()` off filled parts until we need to reallocate.
-    body_recv_buf: bytes::buf::Limit<BytesMut>,
+    /// A buffer to receive H3 body data from quiche. Lazily allocated on the
+    /// first body read and released once no streams remain, so idle
+    /// connections hold no receive buffer. We `split()` off filled parts until
+    /// we need to reallocate.
+    body_recv_buf: Option<bytes::buf::Limit<BytesMut>>,
 
     /// The buffer used to interact with the underlying IoWorker.
     io_worker_buf: Vec<u8>,
@@ -393,8 +395,7 @@ impl<H: DriverHooks> H3Driver<H> {
                 dgram_recv,
                 dgram_send: PollSender::new(dgram_send),
                 max_stream_seen: 0,
-                body_recv_buf: BytesMut::with_capacity(BufFactory::MAX_BUF_SIZE)
-                    .limit(BufFactory::MAX_BUF_SIZE),
+                body_recv_buf: None,
                 io_worker_buf: vec![0u8; BufFactory::MAX_BUF_SIZE],
 
                 waiting_streams: FuturesUnordered::new(),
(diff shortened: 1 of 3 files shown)


LynnLi100:
the nightly failure is the known fix-nightly issue (#2545), after #2545 is merged I'll rebase the pr then it'll be fixed

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1000,6 +1007,13 @@ impl<H: DriverHooks> H3Driver<H> {
 
         self.close_if_idle(qconn);
 
+        // Release the body receive buffer once no streams or flows remain; it
+        // is reallocated lazily on the next body read. Body bytes only flow on
+        // active streams, so an empty stream map means it is unused.
+        if self.stream_map.is_empty() && self.flow_map.is_empty() {
```
dineshadhi: I think this cleanup should also happen [here](https://github.com/LynnLi100/quiche/blob/5b151344758753c7e1df1b6095c77136d3d41310/tokio-quiche/src/http3/driver/mod.rs#L933)

Perhaps, can we plug this cleanup logic inside `close_if_idle`?

LynnLi100: Yes! Thank you for pointing that out. I moved the buffer-release logic into close_if_idle so it runs on every idle-transition path and removed the duplicated block from cleanup_stream. It's done in [10e665c](https://github.com/cloudflare/quiche/pull/2546/changes/10e665c297a6274f325ced960a028b94b893e67e)

dineshadhi (approved):

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -243,6 +243,149 @@ mod client_side_driver {
...
+            .driver_send_request(make_request_headers("GET"), true)
+            .unwrap();
+
+        // server reads request and sends response headers (no body)
+        helper.advance_and_run_loop().unwrap();
+        assert_matches!(
+            helper.peer_server_poll().unwrap(),
+            (0, h3::Event::Headers { .. })
+        );
+        helper.peer_server_send_response(0, false).unwrap();
```
inamdarminaz: I think this test/comment seems misleading?
`fin=false` means the response is not headers-only and the stream can still receive body data. Can we use fin true and then assert? 

I think it'd also be good to add the server side no body equivalent to the tests, since body_recv_buf is `H3Driver` state used by both `ClientHooks` and `ServerHooks`.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -521,17 +522,23 @@ impl<H: DriverHooks> H3Driver<H> {
                 };
             }
 
-            // NOTE: `self.body_recv_buf` is `Limit<BytesMut>` so
+            // Lazily allocate the receive buffer on first use; idle
+            // connections never receive body bytes and never allocate it.
+            let body_recv_buf = self.body_recv_buf.get_or_insert_with(|| {
+                BytesMut::with_capacity(BufFactory::MAX_BUF_SIZE)
+                    .limit(BufFactory::MAX_BUF_SIZE)
```
inamdarminaz, line 529: Wasn't the idea to change the MAX_BUF_SIZE down to 16KB?  its still sits at 64KB https://github.com/cloudflare/quiche/blob/7afd499cd32227a1bc3cad4f877df1c07aede901/tokio-quiche/src/buf_factory.rs#L39 
I'm probably missing the context but was this change discarded ?

LynnLi100: Yes, I was trying to make each PR small so #2546 is only about lazy allocation, #2547 is about sizes the allocation to the stream's readable length and #2550 is adds Http3Settings::max_recv_body_buf_size, defaulting to DEFAULT_MAX_BODY_RECV_BUF_SIZE = 16 KiB

tokio-quiche/src/http3/driver/tests.rs
LynnLi100: Yes, thank you for catching this.  Changed the idle test to `peer_server_send_response(0, true)` and add a `DriverTestHelper::<ServerHooks> `no-body test asserting `body_recv_buf.is_none()`.

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1003,15 +1010,23 @@ impl<H: DriverHooks> H3Driver<H> {
...
-                qconn.close(true, quiche::h3::WireErrorCode::NoError as u64, &[]);
+    /// Handles connection cleanup once no streams or flows remain.
+    ///
+    /// Releases the body receive buffer (it is reallocated lazily on the next
+    /// body read; body bytes only flow on active streams, so an empty stream
+    /// map means it is unused) and closes the connection with `NoError` if the
+    /// H3 event receiver has been dropped.
+    fn close_if_idle(&mut self, qconn: &mut QuicheConnection) {
+        if self.stream_map.is_empty() && self.flow_map.is_empty() {
+            self.body_recv_buf = None;
```
jrouviere, line 1021: I'm wondering if that's a good idea to drop the buffer here. If I followed the logic correctly the connection might still get re-used and we would need to re-allocate the buffer.

- Before the change, we were allocating exactly once per connection
- After the change, we improved the best case scenario as we might not allocate at all, but we degraded the worst case scenario by allocating once per request

To be on the safe side I think I would remove that line (and update the tests), so this change is strictly an improvement.

antoniovicente: It is a tradeoff.  Having the memory leaked until connection is closed if there are any requests with body isn't great.  Dropping the buffer in cases where the request count reaches 0 is good, but there's room for further improvement if we consider that the buffer in question is only needed in cases where the connection is waiting for at least 1 request body.  In cases where the request body has been received and only sending of the response body remains, this buffer can be dropped.

jrouviere: Ok the tradeoff seems fair. I suppose the busiest connections will have many requests in parallel so won't be impacted.

jrouviere (approved):

inamdarminaz (approved):

PR #2547, opened 2026-07-20
tokio-quiche: size H3 body_recv_buf to readable length

LynnLi100:
## Summary

Follow-up to the lazy `body_recv_buf` allocation in #2546. Previously, once a body buffer was needed it always allocated the fixed 64 KiB `BufFactory::MAX_BUF_SIZE`, so even a 10-byte body paid for 64 KiB on its first read.

This PR sizes the allocation to data that is immediately readable from the stream, with a 1 KiB floor and the existing 64 KiB cap.

## Changes

- Add `quiche::Connection::stream_readable_len(stream_id)`, backed by `RecvBuf::readable_len()`. It returns the contiguous, in-order unread bytes starting at the current read offset, or 0 when the stream is absent or has no readable data. Bytes buffered behind a gap are not counted.
- Bound that scan at 64 KiB. Its cost is proportional to contiguous front chunks only up to the maximum body buffer size, and it copies no data.
- In `H3Driver::process_h3_data`, set the lazy allocation target to `body_recv_buf_size(stream_readable_len(stream_id))`, clamped to `[1 KiB, MAX_BUF_SIZE]`.
- Reallocate whenever `remaining_mut() < want`, rather than only after exhaustion. A buffer previously sized for a small read can therefore grow before a later larger read instead of throttling it. Capacity remains equal to the limit, preserving the `split()` invariant that spare capacity equals `remaining_mut()`.

## Rationale

`stream_readable_len` reports exactly the contiguous QUIC stream bytes a read can consume now. It includes HTTP/3 frame bytes, which can make the allocation conservatively larger than the body payload, but it never includes out-of-order data behind a gap. The 64 KiB cap bounds both the allocation and the scan.

`body_recv_buf` is shared per connection rather than per stream. The 1 KiB floor preserves allocation amortization for trickle traffic: after each `split()`, a 1-byte-at-a-time body can reuse the allocation for roughly 1024 reads instead of reallocating on every read.

## Testing

- `Connection::stream_readable_len` test covers an absent stream, growth after receives, shrinkage after reads, and draining to zero.
- `RecvBuf::readable_len` test verifies that data behind a gap is ignored and that the scan stops at 64 KiB.
- `body_recv_buf_size` tests cover the floor, exact tracking between the bounds, and the cap.
- Driver tests verify lazy small-body allocation at the floor, reallocation, and release on stream cleanup.

## Validation

Bastion continuous-profiling flamegraphs for heap `inuse_space` and CPU are available at https://drive.google.com/drive/folders/1zvP2zze7xui1pAC_wU-QhMd6vMfIGuks.

- For 10k connections with 1-byte POST bodies (`stalled_active_requests`), total heap fell from **3.25 GB** to **2.49 GB** (-23.5%).
- With the lazy-allocation prerequisite (#2546), 10k idle connections fell from **1.93 GB** to **1.49 GB** (-23%); `H3Driver::new` allocations fell from **1.25 GB (64.6%)** to **0.75 GB (50.3%)**.
- On 1000 connections with 8 chunked streams, heap remained approximately flat and the CPU shares of `recv_body_buf` (~2.8-2.9%) and `process_h3_data` (~4.5%) were unchanged.

## Residual Risk

A large body delivered in small increments can require more reallocations than the previous fixed 64 KiB buffer, but the 1 KiB floor bounds tiny-read churn. The new readable-length scan is not O(1), but its work is bounded by 64 KiB of contiguous data.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -6385,6 +6385,33 @@ impl<F: BufFactory> Connection<F> {
         stream.is_readable()
     }
 
+    /// Returns the number of bytes that can currently be read from a stream
+    /// without gaps.
+    ///
+    /// This is the length of the contiguous, in-order data buffered at the
+    /// stream's current read offset, up to 64 KiB, i.e. the bytes a call to
+    /// [`stream_recv`] would return right now. Data received out of order that
+    /// sits behind a gap is not counted, so this never reports bytes that are
+    /// not yet readable.
+    ///
+    /// This is a companion to [`stream_readable`], which only reports *whether*
+    /// data is available; this reports *how much*. It is intended for sizing a
+    /// receive buffer. The cost is proportional to the number of contiguous
+    /// buffered chunks at the front of the stream, up to 64 KiB, and no data is
+    /// copied.
+    ///
+    /// Returns 0 if the stream does not exist or has no data ready to read.
+    ///
+    /// [`stream_recv`]: struct.Connection.html#method.stream_recv
+    /// [`stream_readable`]: struct.Connection.html#method.stream_readable
+    pub fn stream_readable_len(&self, stream_id: u64) -> usize {
+        match self.streams.get(stream_id) {
+            Some(s) => s.recv.readable_len(),
+
+            None => 0,
+        }
+    }
+
     /// Returns the next stream that can be written to.
     ///
     /// Note that once returned by this method, a stream ID will not be returned
--- a/quiche/src/stream/recv_buf.rs
+++ b/quiche/src/stream/recv_buf.rs
@@ -419,6 +419,39 @@ impl RecvBuf {
         buf.off() == self.off
     }
 
+    /// Returns the number of bytes that can be read contiguously from the
+    /// current read offset.
+    ///
+    /// This is the amount of in-order data available to read right now, up to
+    /// 64 KiB. Data buffered behind a gap (received out of order) is not
+    /// counted, so this never reports bytes that are not yet readable. The cost
+    /// is proportional to the number of contiguous buffered chunks at the front
+    /// of the buffer, up to 64 KiB; no data is copied.
+    pub fn readable_len(&self) -> usize {
+        const MAX_READABLE_LEN: usize = 64 * 1024;
+
+        let mut contiguous = 0;
+        let mut next_off = self.off;
+
+        // `data` is ordered by offset, so walk from the front and stop at the
+        // first gap (a chunk that does not start where the contiguous run so
+        // far leaves off).
+        for buf in self.data.values() {
+            if buf.off() != next_off {
+                break;
+            }
+
+            contiguous = (contiguous + buf.len()).min(MAX_READABLE_LEN);
+            next_off = buf.max_off();
+
+            if contiguous == MAX_READABLE_LEN {
+                break;
+            }
+        }
+
+        contiguous
+    }
+
     #[cfg(test)]
     pub(crate) fn flow_control_for_tests(&self) -> &flowcontrol::FlowControl {
         &self.flow_control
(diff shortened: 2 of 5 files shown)


tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -522,21 +538,31 @@ impl<H: DriverHooks> H3Driver<H> {
...
+            //
+            // When the buffer is exhausted, reallocate it sized to the amount
+            // now readable (rather than always a fixed 64 KiB). Capacity is kept
+            // equal to the limit so the `split()` invariant asserted below (spare
+            // capacity == remaining_mut) continues to hold.
             if !body_recv_buf.has_remaining_mut() {
-                *body_recv_buf =
-                    BytesMut::with_capacity(BufFactory::MAX_BUF_SIZE)
-                        .limit(BufFactory::MAX_BUF_SIZE);
+                *body_recv_buf = BytesMut::with_capacity(want).limit(want);
```
jrouviere, line 575: A bit similar to https://github.com/cloudflare/quiche/pull/2546/changes#r3630179577.

IIUC we are now only allocating enough for the unread bytes in the stream. That means we are not amortizing the allocations anymore, if we receive 1000x one byte, we allocate 1000 times, while previously we would have allocated once.

Should we reduce the impact by setting the `MIN_BODY_RECV_BUF_SIZE` to a larger value (eg: 1024 bytes)?

jrouviere: Alternatively, perhaps we could set the `.limit` to a larger value? The point of the BytesMut is that it can grow automatically like a vector.

So for example:
```
*body_recv_buf = BytesMut::with_capacity(want).limit(MAX_BUF_SIZE);
```
with MIN_BODY_RECV_BUF_SIZE set to 1024

LynnLi100: Yes you are right, I agree. I will raise `MIN_BODY_RECV_BUF_SIZE` to 1024, so a single allocation now absorbs ~1024 small reads before reallocating.
One thing the floor bump exposed: readable.clamp(MIN, max) panics when max < MIN, and max is now configurable via `Http3Settings::max_recv_body_buf_size` (in #2550). So I will make the floor yield to the cap

inamdarminaz (approved):
lgtm

tokio-quiche/src/http3/driver/mod.rs
hargut: Would 1024 be enough for a full QUIC data frame?

quiche/src/lib.rs
```diff
@@ -6385,6 +6385,32 @@ impl<F: BufFactory> Connection<F> {
...
+    ///
+    /// `max_off()` is always `>=` `off_front()` (the lowest unread offset can
+    /// never exceed the largest offset ever received), so the subtraction never
+    /// underflows; `saturating_sub` upholds that invariant defensively rather
+    /// than relying on it to avoid a panic. The `usize` cast may truncate on
+    /// 32-bit platforms, but the sole caller clamps the result to a small
+    /// receive-buffer bound, so a truncated value is harmless.
+    ///
+    /// Returns 0 if the stream does not exist.
+    pub fn stream_readable_len(&self, stream_id: u64) -> usize {
```
antoniovicente, line 6407: I think "readable" is misleading in this function name, since there could be unreceived gaps in the stream's recv buffers.

Could this PR operate on stream bytes that are currently available to read immediately (no gaps)

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -522,21 +546,33 @@ impl<H: DriverHooks> H3Driver<H> {
...
             // `has_remaining_mut()` will indicate if the buffer
             // has space available until the *limit* is
             //  reached. (A plain `BytesMut` can reallocate and would always
             // return true)
+            //
+            // When the buffer is exhausted, reallocate it sized to the amount
+            // now readable (rather than always a fixed 64 KiB). Capacity is kept
+            // equal to the limit so the `split()` invariant asserted below (spare
+            // capacity == remaining_mut) continues to hold.
             if !body_recv_buf.has_remaining_mut() {
```
antoniovicente: Should this line be changed to:
`if self.remaining_mut() < want {`

tokio-quiche/src/http3/driver/mod.rs
LynnLi100: The floor only applies when fewer than 1 KiB is readable. Whenever a full frame's worth of data is buffered, `stream_readable_len` report it and the buffer is sized up to 64KB, so a full data frame is accommodated. It's just to make sure situation like "we receive 1000x one byte, we allocate 1000 times" doesn't happen. Does that make sense to you?

tokio-quiche/src/http3/driver/mod.rs
LynnLi100: Done, changed to `if body_recv_buf.remaining_mut() < want`.

quiche/src/lib.rs
LynnLi100: Good point, `stream_readable_len` now returns the contiguous, in-order bytes available to read right now

quiche/src/stream/recv_buf.rs
```diff
@@ -419,6 +419,33 @@ impl RecvBuf {
...
+    /// to the number of contiguous buffered chunks at the front of the buffer;
+    /// no data is copied.
+    pub fn readable_len(&self) -> usize {
+        let mut contiguous = 0;
+        let mut next_off = self.off;
+
+        // `data` is ordered by offset, so walk from the front and stop at the
+        // first gap (a chunk that does not start where the contiguous run so
+        // far leaves off).
+        for buf in self.data.values() {
```
jrouviere, line 439: This is not O(1) anymore, I'm not sure how the buffers are structured currently, so that might end up being an expensive operations with a lot of small buffers.

To be safe, we should probably also stop iterating once we've reached the max (64k).

LynnLi100: You are right, the fix is done. Now scan stop once it reaches the max (64KB). Also changed in #2550 that scan stop once it reaches effective configured cap

jrouviere (approved):

jrouviere:
the PR description will need to be updated after the recent changes

LynnLi100:
Updated the PR description to reflect the current behavior: contiguous readable length rather than an O(1) upper bound, the 64 KiB scan cap, the 1 KiB floor, and grow-before-read reallocation.

antoniovicente (approved):

tokio-quiche/src/http3/driver/mod.rs
antoniovicente, line 593: Wondering, is there value in setting `self.body_recv_buf` to None if `!has_remaining_mut()`

LynnLi100: fixed in https://github.com/cloudflare/quiche/pull/2558

PR #2550, opened 2026-07-20
 tokio-quiche: make body_recv_buf max size configurable, default 16 KiB

LynnLi100:
### Summary

Follow-up to !2547 (size `body_recv_buf` to readable length). This change makes the upper bound of the H3 body receive buffer configurable instead of hardcoding it to `BufFactory::MAX_BUF_SIZE` (64 KiB), and lowers the default when unset to 16 KiB.

### Motivation

Previously the driver clamped the body receive buffer to a fixed 64 KiB ceiling. A request that carries a body could pin a large allocation for the entire life of the stream. By making the cap configurable and lowering the default, idle/small bodies
no longer pay for a large allocation; a larger streamed body simply reallocates once per driver read-cycle.

### Changes

- Add `Http3Settings::max_recv_body_buf_size: Option<usize>` to let downstream users override the cap.
- Introduce `DEFAULT_MAX_BODY_RECV_BUF_SIZE` (16 KiB), used when the setting is unset.
- `body_recv_buf_size(readable)` now takes an explicit `max` argument; the buffer is clamped to `[MIN_BODY_RECV_BUF_SIZE, max]`.
- `H3Driver` stores `max_recv_body_buf_size`, resolved from `Http3Settings` in the constructor.
- The buffer remains sized to the amount currently readable on the stream (an upper bound that includes H3 framing), so we never under-allocate; the cap only bounds the worst case.

### Behavior

- **Default (unset):** body receive buffer capped at 16 KiB (down from 64 KiB).
- **Configured:** cap may be set below or above `MAX_BUF_SIZE` as needed.
- Floor of `MIN_BODY_RECV_BUF_SIZE` still guarantees non-zero capacity.

### Testing

- Updated `body_recv_buf_size` unit tests to pass an explicit `max`.
- Added `respects_configured_max`: smaller cap bounds the buffer below `MAX_BUF_SIZE`; larger cap allows growth past it.
- Added `default_cap_is_16kib`: verifies the default and that a larger body is capped at 16 KiB.

### Assumptions & Validation

Data (Bastion continuous-profiling flamegraphs): https://drive.google.com/drive/folders/1zvP2zze7xui1pAC_wU-QhMd6vMfIGuks

This change lowers the default cap (64 KiB -> 16 KiB) and makes it configurable. The primary memory win is already delivered by sizing-to-readable (#2547): heap **-23.5%** under 10k conns x 1-byte POST bodies, and **-23%** for 10k idle conns (`H3Driver::new` allocations -40%). Because most reads are well below 16 KiB, the lower default rarely changes the sized allocation; it only tightens the worst case a single (potentially adversarial) read can force, from 64 KiB down to 16 KiB.

**Assumptions**
- 16 KiB is a conservative default: large enough that typical bodies still fill one allocation per read-cycle, small enough to bound per-connection worst-case memory. It is overridable via `Http3Settings::max_recv_body_buf_size` for services that stream large bodies and prefer fewer reallocations.
- `Some(0)` is normalized to the default so a misconfiguration cannot create a zero-limit (no-op) buffer.

**No CPU regression**
- Active-streaming profile (1000 conns x 8 chunked streams) shows the `recv_body_buf` CPU share flat (~2.8-2.9%); the extra reallocations from a smaller cap did not appear as a CPU regression in the profiled workloads.

**Residual risk / mitigation**
- The main theoretical cost is more frequent reallocation on large streamed bodies (cap-sized chunks). The configurable cap lets latency/throughput-sensitive services raise it, and the 1 KiB floor bounds churn on the small-read end.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -74,7 +74,6 @@ use self::streams::StreamReady;
 use self::streams::WaitForDownstreamData;
 use self::streams::WaitForStream;
 use self::streams::WaitForUpstreamCapacity;
-use crate::buf_factory::BufFactory;
 use crate::http3::settings::Http3Settings;
 use crate::http3::H3AuditStats;
 use crate::metrics::Metrics;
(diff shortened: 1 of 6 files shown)


jrouviere (commented):
Looks good on principle, reviewed the second commit only since the first one is from https://github.com/cloudflare/quiche/pull/2547

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -114,6 +114,48 @@ const STREAM_CAPACITY: usize = 1; // Set to 1 to stress write_pending under test
...
+///
+/// The result is clamped to `[floor, max]`, where `floor` is
+/// [`MIN_BODY_RECV_BUF_SIZE`] capped by `max`. Reads below the floor still
+/// allocate the floor (so a trickle of tiny reads reuses one allocation instead
+/// of reallocating each time), while a single (potentially adversarial) read
+/// never allocates more than `max`. `max` is derived from
+/// [`Http3Settings::max_recv_body_buf_size`], defaulting to
+/// [`DEFAULT_MAX_BODY_RECV_BUF_SIZE`] when the setting is unset.
+///
+/// `max` is the hard upper bound: when it is smaller than the floor it wins, so
```
inamdarminaz, line 151: so if callers can configure this to 0, that'd create a zero limit buffer. IIUC doesn't this work against us? should there be some validation & tests for the zero limit buffer case?

LynnLi100: Yes you are right, thank you so much for catching this! Some(0) collapses the floor into a zero-limit buffer. Fixed  it by normalizing Some(0) to the 16 KiB default in the constructor

inamdarminaz: thanks! don't forget to squash commits before you merge.

inamdarminaz (approved):

jrouviere:
> Looks good on principle, reviewed the second commit only since the first one is from #2547

@LynnLi100 now that the first PR is merged you can rebase this branch, and I'll take another look.

jrouviere (approved):
LGTM

PR #2552, opened 2026-07-22
tokio-quiche: don't set UDP_SEGMENT cmsg for single-segment sends

### LINKED ISSUE #2551: tokio-quiche: server handshake fails on NICs without GSO — acceptor Retry/version-negotiation reply always sets UDP_SEGMENT cmsg (EIO)
## Summary

On a host whose NIC does not support GSO, a tokio-quiche **server** cannot complete any QUIC handshake. It receives the client's Initial fine, but every reply the acceptor sends (stateless Retry / version negotiation) fails with `EIO`, so the handshake never progresses and no connection is ever established. This affects IPv4 and IPv6 equally.

The crate already avoids GSO on the **client** path for exactly this reason, but the **server** acceptor path has no such guard.

## Root cause

`ConnectionAcceptor::handshake_reply` (used by `stateless_retry` and version negotiation) unconditionally calls `gso::send_to(...)` on Linux:

```rust
// tokio-quiche/src/quic/router/acceptor.rs — handshake_reply()
#[cfg(target_os = "linux")]
{
    let from = with_pktinfo.then_some(incoming.local_addr);
    let _ = crate::quic::io::gso::send_to(udp, to, from, send_buf, send_buf.len(), ...);
}
```

`gso::send_to` **always** attaches a `UDP_SEGMENT` cmsg (made unconditional in #2060), even for a single-segment packet:

```rust
// tokio-quiche/src/quic/io/gso.rs
cmsgs.push(ControlMessage::UdpGsoSegments(&segment_size_u16));
```

On a NIC that can't do GSO, `sendmsg()` with a `UDP_SEGMENT` cmsg returns `EIO`. So the Retry reply never leaves the box, the client keeps retransmitting its Initial, and nothing happens.

The crate is already aware that some NICs lack GSO and deliberately avoids it on the client path:

```rust
// tokio-quiche/src/quic/mod.rs
// Don't apply_max_capabilities(): some NICs don't support GSO
```

…but the server acceptor's `handshake_reply` has no equivalent guard. The `has_gso` capability only gates the io-worker's connection sends (`worker.rs::flush_buffer_to_socket` checks `self.cfg.with_gso`); it does **not** gate the acceptor's Retry/version-negotiation path. Consequently, even constructing the `QuicListener` with GSO disabled (default `SocketCapabilities`, `has_gso = false`) does not help — the acceptor still calls `gso::send_to` and still fails.

## Environment / repro

- Linux 5.15, `virtio_net`.
- The vNIC reports `scatter-gather: off [fixed]` and `tx-checksum-ip-generic: off [fixed]`, so `generic-segmentation-offload` can't be enabled (`ethtool -K eth0 gso on` → `Could not change any device features`). This feature set is negotiated by the hypervisor and is seen on some lightweight/budget cloud VPS instances.
- The capability probe is a false positive here: `setsockopt(UDP_SEGMENT, …)` **succeeds** (the kernel supports the UDP GSO API regardless of NIC), so `apply_max_capabilities()` reports `has_gso = true`. The failure only surfaces at `sendmsg()` time.

`strace` of the server while a client tries to connect — it receives:

```
recvmsg(9, {msg_name={sa_family=AF_INET6, ... "::ffff:<client-ipv4>" ...}, ...}) = 1200
```

…but every reply errors:

```
sendmsg(9, {..., msg_control=[{cmsg_len=18, cmsg_level=SOL_UDP, cmsg_type=UDP_SEGMENT}], msg_controllen=24, ...}, 0) = -1 EIO (Input/output error)
```

(Identical for IPv6 clients.)

## Impact

A server on such a NIC is completely unreachable over QUIC while address validation (stateless Retry) is enabled — which is the default. Current workarounds are limited to setting `disable_client_ip_validation` **and** building the listener without GSO, or moving to a NIC that supports GSO.

## Suggested fix

Make the acceptor's `handshake_reply` (Retry / version negotiation) respect GSO availability the same way the client path does — fall back to a plain `send_to` (no `UDP_SEGMENT` cmsg) when GSO isn't in use. Because the capability probe gives a false positive on these NICs, a fully robust fix would additionally detect `EIO` on a GSO send and disable GSO for the socket (or perform a real test-send during capability detection instead of relying on the `setsockopt` succeeding).

zaczh:
## Problem

`gso::send_to` attaches a `UDP_SEGMENT` control message to every `sendmsg()`, even single-packet sends. On a NIC that accepts the `UDP_SEGMENT` sockopt but cannot actually segment — e.g. `virtio-net` with `scatter-gather` fixed off, as found on some lightweight cloud VPS instances — such a send fails with `EIO`.

The server acceptor's `handshake_reply()` routes stateless-Retry and version-negotiation replies through `gso::send_to`, so on such a NIC the server can never send them and **no handshake completes** (IPv4 and IPv6 alike). Full analysis, `strace` evidence, and how it regressed in #2060 are in #2551.

Notably the client path already avoids GSO for this reason:

```rust
// quic/mod.rs
// Don't apply_max_capabilities(): some NICs don't support GSO
```

…but the acceptor path does not, and the `has_gso` capability only gates the io-worker's sends — not the acceptor's — so building the listener without GSO doesn't help either.

## Fix

Only attach the `UDP_SEGMENT` cmsg when the buffer actually spans more than one segment (`send_buf.len() > segment_size`). A single-segment send needs no segmentation. This restores the single-segment fast-path that existed before #2060 (which passed `num_pkts` and only set the cmsg when `num_pkts > 1`).

It does **not** reintroduce the truncation footgun #2060 fixed: GSO is now initialized with a `u16::MAX` segment size, so a single packet sent without the cmsg uses that size as the fallback and is never truncated. Multi-segment sends are unaffected — they still carry the cmsg.

`cargo check -p tokio-quiche` passes.

Fixes #2551.

### DIFF
--- a/tokio-quiche/src/quic/io/gso.rs
+++ b/tokio-quiche/src/quic/io/gso.rs
@@ -138,8 +138,17 @@ pub async fn send_to(
 
         let mut cmsgs: SmallVec<[ControlMessage; 3]> = SmallVec::new();
 
-        // Create cmsg for UDP_SEGMENT.
-        cmsgs.push(ControlMessage::UdpGsoSegments(&segment_size_u16));
+        // Only attach the UDP_SEGMENT cmsg when the buffer actually spans more
+        // than one segment. A single-segment send needs no segmentation, and
+        // attaching the cmsg anyway makes sendmsg() return EIO on NICs that
+        // accept the UDP_SEGMENT sockopt but cannot segment (e.g. virtio-net
+        // with scatter-gather off). Since GSO is initialized with a u16::MAX
+        // segment size, omitting the cmsg here cannot cause a truncating
+        // fallback to a smaller socket-level segment size.
+        if send_buf.len() > segment_size {
+            // Create cmsg for UDP_SEGMENT.
+            cmsgs.push(ControlMessage::UdpGsoSegments(&segment_size_u16));
+        }
 
         if tx_time.is_some() {
             // Create cmsg for TXTIME.


tokio-quiche/src/quic/io/gso.rs
```diff
@@ -138,8 +138,17 @@ pub async fn send_to(
...
-        // Create cmsg for UDP_SEGMENT.
-        cmsgs.push(ControlMessage::UdpGsoSegments(&segment_size_u16));
+        // Only attach the UDP_SEGMENT cmsg when the buffer actually spans more
+        // than one segment. A single-segment send needs no segmentation, and
+        // attaching the cmsg anyway makes sendmsg() return EIO on NICs that
+        // accept the UDP_SEGMENT sockopt but cannot segment (e.g. virtio-net
+        // with scatter-gather off). Since GSO is initialized with a u16::MAX
+        // segment size, omitting the cmsg here cannot cause a truncating
+        // fallback to a smaller socket-level segment size.
+        if send_buf.len() > segment_size {
```
antoniovicente, line 148: I think this PR would break the code that motivated https://github.com/cloudflare/quiche/pull/2060 which I think was related to pmtud.  Could we switch to setting UdpGsoSegments when has_gso / with_gso is true?

Longer version:

A connection on a network interface that does not support UDP_SEGMENT is likely to make forward progress as long as all send attempts involve a single segment.  But at some point later in the connection, we may attempt to do a send for multiple segments which would bring UDP_SEGMENT CMSG back.  Would the connection stall and fail to make any further progress once that happens?

If the answer is yes, I think that the current PR is not the right fix for the issue of UDP_SEGMENT not working.  We should instead have code that detects during startup that UDP_SEGMENT API doesn't work at all, and when operating in that mode it should disable GSO entirely, and force use of 1 sendmsg call per UDP packet send.

I'm curious about the value of has_gso / with_gso in the setup in question.  An alternate solution could be to only set UDP_SEGMENT CMSG if has_gso is true after the setsockopt to enable UdpGsoSegment succeeded.
https://github.com/cloudflare/quiche/blob/fc7382d/tokio-quiche/src/socket/capabilities.rs#L181

zaczh: You're right that this isn't the fix — I tested it, and it doesn't fix the problem at all.

Running the stock `async_http3_server` example against an egress device that cannot produce `CHECKSUM_PARTIAL`, this branch behaves identically to stock: every `sendmsg()` returns `EIO`, no connection is ever accepted, and the client times out. `strace` on the first send confirms the branch did take effect and still failed:

```
stock     …, msg_controllen=24, msg_flags=0}, 0) = -1 EIO (Input/output error)
this PR   …, msg_controllen=0,  msg_flags=0}, 0) = -1 EIO (Input/output error)
```

The reason is that the cmsg was never the only source of `gso_size`. `SocketCapabilitiesBuilder::gso()` sets the socket-level `UDP_SEGMENT` option to `u16::MAX`, and the kernel seeds `cork->gso_size` from that before the cmsg can override it. In `udp_send_skb()`:

```c
if (cork->gso_size) {
        ...
        if (skb->ip_summed != CHECKSUM_PARTIAL || is_udplite ||
            dst_xfrm(skb_dst(skb))) {
                kfree_skb(skb);
                return -EIO;
        }

        if (datalen > cork->gso_size) {
                skb_shinfo(skb)->gso_size = cork->gso_size;
```

The `-EIO` is gated on `cork->gso_size != 0`, not on whether segmentation is actually needed — so it fires even for a single segment, and it fires whether the value came from the sockopt or the cmsg. Once the sockopt is set, **every** send on that socket fails on such a device, and omitting the cmsg changes nothing.

To your question about `has_gso`: it is **true** here. `setsockopt(SOL_UDP, UDP_SEGMENT, [65535], 4) = 0` succeeds even on the failing device, because the sockopt only checks kernel support and never the device. So gating the cmsg on `has_gso` wouldn't help either — and the sockopt alone is already fatal regardless.

PR #2554, opened 2026-07-23
tokio-quiche: pool the IO egress buffer per worker

LynnLi100:
Before, each `H3Driver` held a persistent 64 KiB buffer `io_worker_buf` for the connection's entire lifetime, so every idle connection pinned 64 KiB and total egress scratch memory scaled with the connection count.

What this PR does is removing the per-connection buffer and instead borrowing a transient scratch buffer from a per-worker-thread pool for the duration of each send burst, returning it before the worker sleeps. Idle connections now hold no egress buffer, while the pooled buffers' pages stay resident across bursts so steady-state sending avoids per-burst page faults and kernel zero-fill.

- `SEND_BUF_POOL` is a `thread_local` free-list; `PooledSendBuf::acquire()` pops a buffer (allocating only on a miss) and its `Drop` returns it.
- The pool is capped at `SEND_BUF_POOL_CAP = 16` buffers per worker thread (<= 1 MiB per runtime thread), independent of the connection count.
- Buffers are not re-zeroed on reuse: only `send_buf[..bytes_written]` is ever transmitted, so stale bytes are never sent.
- Work-stealing migration across the flush `.await` is benign -- a buffer acquired on one thread may be returned to another; the per-thread cap keeps the total bounded by `cap * workers`.
- `ApplicationOverQuic::buffer()` now returns an empty slice (kept only to satisfy the trait); the unused `ConnectionStageContext::buffer()` helper is removed.

Adds unit tests for `PooledSendBuf` and driver tests asserting `buffer()` stays empty before/after the handshake and with an active stream.

From doing A/B testing for stalled active requests, this PR deduct heap usage by ~35%.

### DIFF
--- a/tokio-quiche/src/http3/driver/mod.rs
+++ b/tokio-quiche/src/http3/driver/mod.rs
@@ -380,8 +380,6 @@ pub struct H3Driver<H: DriverHooks> {
     /// we need to reallocate.
     body_recv_buf: Option<bytes::buf::Limit<BytesMut>>,
 
-    /// The buffer used to interact with the underlying IoWorker.
-    io_worker_buf: Vec<u8>,
     /// The maximum HTTP/3 stream ID seen on this connection.
     max_stream_seen: u64,
 
@@ -420,7 +418,6 @@ impl<H: DriverHooks> H3Driver<H> {
                 dgram_send: PollSender::new(dgram_send),
                 max_stream_seen: 0,
                 body_recv_buf: None,
-                io_worker_buf: vec![0u8; BufFactory::MAX_BUF_SIZE],
 
                 waiting_streams: FuturesUnordered::new(),
 
@@ -1302,11 +1299,6 @@ impl<H: DriverHooks> ApplicationOverQuic for H3Driver<H> {
     #[inline]
     fn should_act(&self) -> bool {
         self.conn.is_some()
-    }
-
-    #[inline]
-    fn buffer(&mut self) -> &mut [u8] {
-        &mut self.io_worker_buf
     }
 
     /// Poll the underlying [`quiche::h3::Connection`] for
--- a/tokio-quiche/src/quic/io/connection_stage.rs
+++ b/tokio-quiche/src/quic/io/connection_stage.rs
@@ -84,17 +84,6 @@ pub struct ConnectionStageContext<A> {
     pub stats: QuicConnectionStatsShared,
 }
 
-impl<A> ConnectionStageContext<A>
-where
-    A: ApplicationOverQuic,
-{
-    // TODO: remove when AOQ::buffer() situation is sorted - that method shouldn't
-    // exist
-    pub fn buffer(&mut self) -> &mut [u8] {
-        self.application.buffer()
-    }
-}
-
 #[derive(Debug)]
 pub struct Handshake {
     pub handshake_info: HandshakeInfo,
--- a/tokio-quiche/src/quic/io/worker.rs
+++ b/tokio-quiche/src/quic/io/worker.rs
@@ -77,6 +77,163 @@ const RELEASE_TIMER_THRESHOLD: Duration = Duration::from_micros(250);
 /// Stop queuing GSO packets, if packet size is below this threshold.
 const GSO_THRESHOLD: usize = 1_000;
 
+/// Size of each full egress buffer borrowed for a send burst.
+///
+/// Matches the maximum quiche buffer size so GSO batching is unaffected while
+/// a connection is actively sending. Unlike a persistent per-connection
+/// buffer, this memory is returned to the per-worker [`SEND_BUF_POOL`] before
+/// the worker sleeps. The free list retains it for reuse (rather than truly
+/// freeing it) but no idle connection owns an egress buffer.
+const SEND_BUFFER_SIZE: usize = crate::buf_factory::BufFactory::MAX_BUF_SIZE;
+
+/// Size of a temporary egress buffer when pooling is disabled.
+///
+/// The cold handshake and connection-close paths generate a single datagram at
+/// a time. When pooling is enabled, they borrow a full-size buffer from
+/// [`SEND_BUF_POOL`]; otherwise a one-MTU buffer is enough.
+const TRANSIENT_SEND_BUFFER_SIZE: usize = 1500;
+
+/// Allocates a zero-initialized egress buffer on the heap.
+///
+/// This is the cold path that fills [`SEND_BUF_POOL`] on a miss; steady-state
+/// bursts borrow a recycled buffer via [`PooledSendBuf::acquire`] and never
+/// hit this. The buffer is boxed (never a stack array) so that holding it
+/// across the `.await` in [`IoWorker::flush_buffer_to_socket`] keeps the
+/// worker's futures small and avoids uncontrolled stack growth.
+fn alloc_send_buffer() -> Box<[u8]> {
+    vec![0u8; SEND_BUFFER_SIZE].into_boxed_slice()
+}
+
+thread_local! {
+    /// Per-runtime-worker free-list of egress scratch buffers.
+    ///
+    /// A buffer is borrowed for a single send burst and returned on drop, so
+    /// its pages stay resident across bursts (no per-burst page fault or
+    /// kernel zero-fill) while no idle connection retains a buffer: the pool
+    /// holds at most [`SEND_BUF_POOL_CAP`] buffers per worker thread,
+    /// independent of the connection count.
+    static SEND_BUF_POOL: std::cell::RefCell<Vec<Box<[u8]>>> =
+        const { std::cell::RefCell::new(Vec::new()) };
+}
+
+/// Upper bound on egress buffers parked per worker thread. The natural
+/// high-water mark is the number of connection tasks simultaneously suspended
+/// at a flush `.await` on one runtime thread; returns beyond the cap are freed
+/// so a burst spike cannot pin unbounded memory to a thread.
+///
+/// This is a fixed per-worker-thread reservation, independent of the
+/// connection count: up to `SEND_BUF_POOL_CAP * SEND_BUFFER_SIZE`
+/// (16 * 64 KiB = 1 MiB) per runtime worker thread. The pool is not shrunk
+/// once grown, so after a burst it stays at its high-water mark for the
+/// process lifetime.
+const SEND_BUF_POOL_CAP: usize = 16;
+
+/// Egress scratch buffer borrowed from the per-thread [`SEND_BUF_POOL`] and
+/// returned to it on drop.
+///
+/// Behaves like the `Box<[u8]>` it replaces via `Deref`/`DerefMut`, so idle
+/// connections still retain no egress buffer, but the backing pages are
+/// recycled instead of re-faulted (and re-zeroed by the kernel) every burst.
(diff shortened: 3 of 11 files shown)


gregor-cf:
I think it should be possible to get same / similar saving by tuning jemalloc (lg_tcache_max:16). That sets the max allocation size for which jemalloc will use the thread-local cache to 64KB (default is 32KB)

jannes:
if we can lean on jemalloc to achieve the same, then i strongly prefer that. have only skimmed this MR so far, but it's non-trivial and introduces further complexity. maybe you can try the specific jemalloc tuning Gregor suggested in your testing?

tokio-quiche/src/http3/driver/mod.rs
```diff
@@ -1270,7 +1252,14 @@ impl<H: DriverHooks> ApplicationOverQuic for H3Driver<H> {
 
     #[inline]
     fn buffer(&mut self) -> &mut [u8] {
-        &mut self.io_worker_buf
+        // The worker no longer gathers packets into a persistent
```
gregor-cf: nit: remove the whole function

tokio-quiche/src/http3/driver/tests.rs
```diff
@@ -2404,3 +2198,67 @@ mod server_side_driver {
...
     }
 }
+
+/// Tests for the transient IO egress buffer strategy. The `H3Driver` holds no
+/// persistent egress buffer: the QUIC IO worker borrows a transient buffer for
+/// the duration of each send burst, so [`ApplicationOverQuic::buffer`] always
+/// returns an empty slice. End-to-end coverage of the worker's transient
+/// buffer (handshake, steady-state, and close paths) lives in
+/// `tokio-quiche/tests/integration_tests`, which drive a real `IoWorker`.
+mod io_buffer {
```
gregor-cf: nit: I don't think you need these tests. It's fine to just get rid of the `buffer()` function

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -227,6 +325,12 @@ where
 
             self.write_state.has_pending_data = true;
 
+            // Egress buffer for this wakeup's send burst. Borrowed from the
+            // per-worker pool on demand (see below) and returned after the
+            // burst, before the worker sleeps in the `select!` further down,
+            // so idle connections still hold no egress buffer.
+            let mut send_buf: Option<PooledSendBuf> = None;
```
gregor-cf: I think we might want a way to enable/disable pooled buffers at runtime. When disabled, we should use a persistent buffer per connection.

(Such a persistent buffer should still live in the IO worker, so we can get rid of ConnectionStageContext and the buffer in the driver)

tokio-quiche/src/quic/io/connection_stage.rs
```diff
@@ -84,17 +84,6 @@ pub struct ConnectionStageContext<A> {
     pub stats: QuicConnectionStatsShared,
 }
 
-impl<A> ConnectionStageContext<A>
```
gregor-cf, line 87: finally, it's gone!

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -329,8 +445,12 @@ where
                 Some(pkt) = incoming_recv.recv() => ctx.in_pkt = Some(pkt),
                 directive = self.wait_for_data_or_handshake(qconn, application) => {
```
gregor-cf, line 526: nit: it's not your change and the previous code had the same "issue": It might be good to add comments to `wait_for_data_or_handshake()` and `wait_for_data()` that make it clear that these two async function must be cancellation safe. It's not necessarily obvious and the risk is somebody accidentally making these function not cancel-safe in the future

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -975,8 +1100,9 @@ where
         // send (ignoring flow/congestion control constraints). We should
         // guarantee that it gets sent by doublechecking the
         // gathered/flushed byte totals and retry if they don't match.
-        let _ = self.gather_data_from_quiche_conn(qconn, ctx.buffer());
-        self.flush_buffer_to_socket(ctx.buffer()).await;
+        let mut send_buf = PooledSendBuf::acquire();
```
gregor-cf: You don't strictly need PooledSendBuf here, since this is only called once per connection on close. You can either use an array on the stack or allocate a Vec. I think you can also use a smaller buffer here (1500) since I think we won't have enough data to send to GSO multiple packets.
(Not using PooledSendBuf here makes it easier to enable/disable the pooled buffers at runtime)

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -717,10 +837,12 @@ where
...
             Ok(WaitForDataOrHandshakeDirective::Noop)
         } else {
-            // Poll quiche to make progress on handshake callbacks.
-            self.wait_for_quiche(qconn, quic_application.buffer())
-                .await?;
-            Ok(WaitForDataOrHandshakeDirective::Flush)
+            // Poll quiche to make progress on handshake callbacks, gathering
+            // any handshake packets into a pooled buffer that the caller
+            // flushes and then returns to the pool.
+            let mut send_buf = PooledSendBuf::acquire();
```
gregor-cf: similar to my comment on line 1130, you can use a newly allocated 1500 byte Vec here, since this is only called during the handshake.

gregor-cf (commented):
Very nice. 
It would be cool if we could have metrics to track how many total allocations we've made and how many buffers where discarded because the TL cache is full. So one counter increment in `alloc_send_buf` and one on `Drop` (only if TL cache is full). We want to avoid counting on the hot-path (re-using a buffer) I think

tokio-quiche/src/http3/driver/tests.rs
LynnLi100: Yes, dropped the `whole io_buffer` test module

tokio-quiche/src/http3/driver/mod.rs
LynnLi100: Done, removed `buffer()` from `ApplicationOverQuic`

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Added `QuicSettings::pool_send_buffer `(default set to true). When it's disabled, the IO worker holds a persistent `Box<[u8]>` for the connection instead of borrowing from the pool

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Yes, added `# Cancel` safety doc sections to `wait_for_data_or_handshake()` and `wait_for_quiche()`

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Agree! Switched to a one-off 1500 byte Vec `TRANSIENT_SEND_BUFFER_SIZE` here, runs once per connection and only sends `CONNECTION_CLOSE`. Also keeps the close path off the pool so the runtime toggle needs no special-casing.

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Yes, use a 1500 byte Vec here for the mid-handshake drive path

LynnLi100:
Thank you for the suggestion, added them. There are now two counters, both on the cold paths. The first one is `send_buffer_pool_allocated`, bumps in `acquire()` only when the pool is empty and we actually have to allocate a new buffer. The second, `send_buffer_pool_discarded`, bumps in Drop, but only when the pool is already full and we throw the buffer away instead of parking it. The hot path stays untouched, we reuse a parked buffer or hand one back to a pool that still has room, so we're not counting on every burst. I wired them up as global foundations `quic:: counters` rather than going through the Metrics trait, that way nothing has to be threaded through `PooledSendBuf` itself.

jannes:
the approach lgtm, need to give it a second pass to grok all the details

jannes (approved):
had another pass and looks all solid to me

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -77,13 +77,137 @@ const RELEASE_TIMER_THRESHOLD: Duration = Duration::from_micros(250);
 /// Stop queuing GSO packets, if packet size is below this threshold.
 const GSO_THRESHOLD: usize = 1_000;
 
+/// Size of the transient egress buffer allocated on demand in the write loop.
+///
+/// Matches the maximum quiche buffer size so GSO batching is unaffected while
+/// a connection is actively sending. Unlike a persistent per-connection
+/// buffer, this memory is only held for the duration of a send burst and is
+/// freed before the worker sleeps, so idle connections retain no egress
```
antoniovicente: Consider mentioning freelist behavior in here.  Buffers are not truly freed.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -717,10 +890,14 @@ where
...
-            // Poll quiche to make progress on handshake callbacks.
-            self.wait_for_quiche(qconn, quic_application.buffer())
-                .await?;
-            Ok(WaitForDataOrHandshakeDirective::Flush)
+            // Poll quiche to make progress on handshake callbacks, gathering
+            // any handshake packets into a one-off buffer that the caller
+            // flushes. This runs only during the handshake and sends a single
+            // datagram at a time, so we allocate a small transient buffer here
+            // rather than borrowing a full-size one from the per-worker pool.
+            let mut send_buf = vec![0u8; TRANSIENT_SEND_BUFFER_SIZE];
```
antoniovicente: I wonder if we could also use freelisted buffers in this branch by changing the API to have wait_for_quiche to allocate the buffer internally when it has to and return it in the result.  But it does seem messy while we allow for the option to disable buffer pooling.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -717,10 +890,14 @@ where
...
+            self.wait_for_quiche(qconn, &mut send_buf[..]).await?;
```
antoniovicente: We should make sure that this doesn't result in a 1200 byte packet followed by a 300 byte packet when trying to do a larger write.  We want a single packet to be generated in this case, and be called again to generate the second packet.  We do have plenty of handshake cases where the generated handshake data does not fit in a single QUIC packet.

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Updated the wording to clarify that buffers are returned to the freelist before sleep, not immediately freed.

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Yes, implemented this in `wait_for_quiche()`.It now acquires the buffer internally and returns it only when it generated data. With pooling enabled it uses the freelist and with pooling disabled it uses the small one-off buffer.

tokio-quiche/src/quic/io/worker.rs
LynnLi100: Yes, added a single-packet gather mode for `wait_for_quiche()`.

antoniovicente (approved):

PR #2555, opened 2026-07-23
count persistent RTT jumps - global min detector

eli-lebeau:
Goal: Add a detector for counting persistent jumps in RTT. 

Detector Logic: This detector defines a jump in RTT as an RTT sample that is >3x the minimum RTT encountered (min_rtt). The detector moves between three states: idle, active, and persistent. The detector begins in idle and then moves into active if it encounters an RTT sample that is >3*min_rtt. For an RTT jump to move to persistent, at least three RTT samples that are >3*min_rtt must occur in a row, and no sample below this threshold must occur for 3*min_rtt seconds after the first elevated RTT sample. The detector returns to idle from active or persistent when it encounters a sample that is  <=3*min_rtt.

Structure: The detector lives in a dedicated file, and there is a selector to activate the model or keep it disabled, with disabled as the default.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -9477,6 +9477,8 @@ pub use crate::path::SocketAddrIter;
 
 pub use crate::recovery::BbrBwLoReductionStrategy;
 pub use crate::recovery::BbrParams;
+#[cfg(feature = "internal")]
+pub use crate::recovery::BbrRttJumpDetector;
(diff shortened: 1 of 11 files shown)


jrouviere (approved):
Already reviewed internally

dineshadhi (approved):

hargut (approved):

PR #2556, opened 2026-07-24
Rtt jump detection HMM approach

eli-lebeau:
Goal: Add a detector for counting persistent jumps in RTT using a Hidden Markov Model

Implementation details:

Internal `BbrRttJumpDetector` setting with three modes:

- `Disabled`, which remains the default.
- `GlobalMin`, which detects sustained RTT increases above `3x` the
  connection-lifetime minimum RTT.
- `Hmm`, a 3-state Hidden Markov Model detector using forward filtering,
  warmup, adaptive per-connection elevation, and RTT-scaled dwell gates.

The detector runs from the BBRv2 network model after full bandwidth has been
reached, so startup/self-queueing RTT changes do not immediately count as
persistent jumps.

Confirmed persistent jump episodes are counted in
`rtt_persistent_jump_count`

Each RTT sample is evaluated as such:
1. Convert the RTT sample into a discrete elevation bin.
2. Predict the next state distribution with the transition matrix.
3. Weight each predicted state by how likely that bin is in the emission matrix.
4. Normalize the result back to a probability distribution.
5. Use the most likely state as the raw HMM state, then apply dwell/sample gates before committing any detector state transition.

The bin boundaries and emission matrix use predetermined values derived from offline training.


inamdarminaz (approved):

jrouviere (approved):
Already reviewed

PR #2563, opened 2026-07-30
octets: reject oversized fixed-length varints

### LINKED ISSUE #2560: octets: `put_varint_with_len` accepts values that do not fit the requested length
`OctetsMut::put_varint_with_len()` in octets/src/lib.rs can return `Ok` when the requested
length `len` cannot represent the supplied value `v`. The resulting bytes can decode to
a different value.

https://github.com/cloudflare/quiche/blob/039d2a96126b00c61c409877ab456cfb0b437aea/octets/src/lib.rs#L519-L553

For example, a two-byte QUIC variable-length integer can represent at most
`16_383`, but `put_varint_with_len(16_384, 2)` succeeds and produces `40 00`, which `get_varint` reads back as 0. `put_varint_with_len(256, 1)` writes `00`, which is worse.

It seems that whoever wrote this assumes all callers should give valid inputs, and all the callers in this repo do seem fine.

However, this API is public and other users could unknowingly call it with an unrepresentable value if they ever forget the first 2 bytes get stripped.

So, the fix would be to either:

- Add the assumption in the docstring, or
- Add bounds checks and reject values too big

If you truly want end-to-end safety, maybe put the bounds check in `put_u`.

BTW, the `if self.cap() < len` check is redundant because `put_u` already checks the same thing.

Pybsama:
Fixes #2560.

## Summary

- reject values that do not fit the requested 1, 2, 4, or 8-byte QUIC
  variable-length integer before mutating the buffer;
- preserve valid non-minimal encodings and the existing public error type; and
- add boundary, round-trip, and non-zero-offset atomicity regressions for all
  four widths, including the examples from the issue.

## Validation

- `cargo test --all-targets --features=async,ffi,qlog --workspace`
- `cargo test --doc --features=async,ffi,qlog --workspace`
- `cargo clippy --features=async,ffi,qlog --workspace -- -D warnings`
- `cargo +nightly fmt -p octets -- --check`

### DIFF
--- a/octets/src/lib.rs
+++ b/octets/src/lib.rs
@@ -518,10 +518,25 @@ impl<'a> OctetsMut<'a> {
 
     /// Writes an unsigned variable-length integer of the specified length, in
     /// network byte-order at the current offset and advances the buffer.
+    ///
+    /// Returns [`BufferTooShortError`] when the buffer is too short or the
+    /// value cannot be represented in the specified length.
     pub fn put_varint_with_len(
         &mut self, v: u64, len: usize,
     ) -> Result<&mut [u8]> {
         if self.cap() < len {
+            return Err(BufferTooShortError);
+        }
+
+        let max = match len {
+            1 => 63,
+            2 => 16_383,
+            4 => 1_073_741_823,
+            8 => MAX_VAR_INT,
+            _ => panic!("value is too large for varint"),
+        };
+
+        if v > max {
             return Err(BufferTooShortError);
         }
 
@@ -546,7 +561,7 @@ impl<'a> OctetsMut<'a> {
                 buf
             },
 
-            _ => panic!("value is too large for varint"),
+            _ => unreachable!(),
         };
 
         Ok(buf)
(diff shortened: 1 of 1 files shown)


Pybsama:
Revalidated this change against current `cloudflare/quiche` `master` (`0b1a89af341b7e42c4e75fedf7bdda246d8319af`) before marking it ready for review.

- Rebased cleanly over the two new upstream commits. The changed `octets/src/lib.rs` blob and stable patch ID are identical to the original patch. Published head: `4bf22ec7a65e790f313bead3547d89c0f0ce6b10`; tree: `a845ae7d09601020c9066353a571ed1889128873`.
- Red/green check on current `master`: the two new regressions produced 1 pass / 1 expected failure before the fix, then 2/2 passes with the fix. The focused pair also passed 50 consecutive runs.
- Stable: full workspace all-target tests passed 1322/1322; doc tests passed 66 with 1 ignored; workspace and examples Clippy passed with `-D warnings`.
- Nightly: all-target tests passed 1322/1322 with `async,ffi,qlog` and 1325/1325 with `custom-client-dcid`; doc tests passed 66 with 1 ignored.
- `octets`: default tests 24/24, all-feature tests 25/25, strict Clippy and strict rustdoc passed.
- Nightly rustfmt checks passed for the root workspace, `tools/http3_test`, and `fuzz`. `git diff --check` passed. Public Rust Semgrep rules reported 0 findings across 11 rules on the changed file.

The exact Stable and Nightly full-workspace `cargo doc` commands also complete; they emit 15 existing warnings in untouched crates, while strict documentation checks for the changed `octets` crate are clean. The C example build and Cloudflare's private Semgrep rules remain covered by repository CI because the local environment does not have the `uthash` header or the private Semgrep token.

PR #2570, opened 2026-08-05
Don't collect a stream before its fin has been sent

### LINKED ISSUE #2525: zero-length STREAM FIN can be dropped before it is sent
`stream_send(id, b"", true)` can set `fin_off` at an already ACKed data
offset, then the stream can be collected before any STREAM frame carrying
`FIN` is emitted.

The peer receives all bytes but never receives final size / EOF.

## Relevant code

`src/stream/send_buf.rs`

```rust
if fin {
    self.fin_off = Some(max_off);
}

if self.ack_off() >= max_off {
    return Ok(SendReserve {
        inner: self,
        reserved: 0,
        fin,
    });
}
```

```rust
pub fn is_complete(&self) -> bool {
    if let Some(fin_off) = self.fin_off {
        if self.acked == (0..fin_off) {
            return true;
        }
    }

    false
}
```

`src/lib.rs`

```rust
stream.send.ack_and_drop(offset, length)
```

The ACK path ignores whether the ACKed STREAM frame carried `FIN`.

## RFC 9000

Exact text:

- Section 3.1: "STREAM frame containing the FIN bit is sent"
- Section 3.1: "all stream data has been successfully acknowledged"
- Section 4.5: "A sender always communicates the final size"
- Section 13.3: "sending ceases when a packet containing that information is acknowledged"
- Section 19.8: "length of 0"
- Section 19.8: "offset of the next byte that would be sent"

For `STREAM offset=N length=0 FIN=true`, the FIN-bearing STREAM frame is the
packet information that communicates final size `N`. ACKing an earlier
data-only STREAM frame for `0..N` is not an ACK of that later FIN frame.

## Reproducer

Paste into `src/tests.rs`:

```rust
#[rstest]
fn stream_empty_fin_is_not_complete_until_fin_acked(
) {
    let mut buf = [0; 65535];

    let mut pipe = test_utils::Pipe::new(cc_algorithm_name).unwrap();

    assert_eq!(pipe.client.stream_send(4, b"hello", true), Ok(5));
    assert_eq!(pipe.server.stream_recv(4, &mut buf), Ok((5, true)));

    assert_eq!(pipe.server.stream_send(4, b"response", false), Ok(8));

    let flight = test_utils::emit_flight(&mut pipe.server).unwrap();
    test_utils::process_flight(&mut pipe.client, flight).unwrap();

    assert_eq!(pipe.client.stream_recv(4, &mut buf), Ok((8, false)));

    assert_eq!(pipe.server.stream_send(4, b"", true), Ok(0));

    let flight = test_utils::emit_flight(&mut pipe.client).unwrap();
    test_utils::process_flight(&mut pipe.server, flight).unwrap();

    assert!(!pipe.server.streams.is_collected(4));

    let flight = test_utils::emit_flight(&mut pipe.server).unwrap();
    let mut frames = Vec::new();

    for (mut pkt, _) in flight {
        frames.extend(
            test_utils::decode_pkt(&mut pipe.client, &mut pkt).unwrap(),
        );
    }

    assert!(frames.iter().any(|f| matches!(
        f,
        frame::Frame::Stream { stream_id: 4, data }
            if data.off() == 8 && data.len() == 0 && data.fin()
    )));
}
```

Result on `0.29.2`:

```text
assertion failed: !pipe.server.streams.is_collected(4)
```

## Fix suggestion

Track final-size delivery separately from byte-range ACKs. In particular,
an empty FIN should remain flushable/retransmittable until a FIN-bearing
STREAM frame at `fin_off` has been emitted and ACKed.

kixelated:
(AI generated but reviewed)

A `fin` carries no bytes of its own, so acking the stream data up to the final offset says nothing about whether the peer learned the final size. SendBuf::is_complete() checked only the acked ranges, which meant a zero-length `fin` written at an already-acked offset reported the send side complete the moment it was written.

That is enough for the caller to collect the stream, and collecting it discards the queued `fin` before any frame carrying it is built. The peer reads every byte and never sees the end of the stream, while the sender believes it delivered one. stream_send() returns Ok, so nothing surfaces the loss.

Reaching it needs the ack for the stream data to land between the `fin` being written and the next send(), which is routine for an application that writes its payload and finishes the stream on a later turn of its event loop.

Track the `fin` ack separately and require it. RESET_STREAM carries the final size in place of a `fin`, so SendBuf::reset() sets the flag too; without that a reset stream would never be collected.

Fixes #2525

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -3593,7 +3593,7 @@ impl<F: BufFactory> Connection<F> {
                         stream_id,
                         offset,
                         length,
-                        ..
+                        fin,
                     } => {
                         // Emit qlog before checking if the stream still exists.
                         // The client does need to ACK frames that were received
@@ -3624,6 +3624,11 @@ impl<F: BufFactory> Connection<F> {
                         };
 
                         let dropped = stream.send.ack_and_drop(offset, length);
+
+                        if fin {
+                            stream.send.ack_fin();
+                        }
+
                         let priority_key = Arc::clone(&stream.priority_key);
 
                         // Only collect the stream if it is complete and not
--- a/quiche/src/stream/mod.rs
+++ b/quiche/src/stream/mod.rs
@@ -1452,6 +1452,11 @@ mod tests {
         assert!(!stream.send.is_complete());
 
         stream.send.ack(0, 1);
+        // Every byte is acked, but the peer has not acked the fin that tells it
+        // where the stream ends.
+        assert!(!stream.send.is_complete());
+
+        stream.send.ack_fin();
         assert!(stream.send.is_complete());
 
         assert!(!stream.is_complete());
--- a/quiche/src/stream/send_buf.rs
+++ b/quiche/src/stream/send_buf.rs
@@ -122,6 +122,13 @@ where
     /// The final stream offset written to the stream, if any.
     fin_off: Option<u64>,
 
+    /// Whether the peer acked the `fin` itself, rather than just the stream
+    /// data up to the final offset.
+    ///
+    /// Tracked separately because a `fin` carries no bytes of its own, so it
+    /// leaves no trace in `acked`.
+    fin_acked: bool,
+
     /// Whether the stream's send-side has been shut down.
     shutdown: bool,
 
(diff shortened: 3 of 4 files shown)


ghedo:
Thanks for the contribution. #2665 addresses the same problem with more complete FIN-acknowledgement state handling and regression coverage. Closing this PR in favor of #2665.

PR #2574, opened 2026-08-07
fix: validate QUIC fixed bit on received packets

### LINKED ISSUE #2557: Validate the QUIC fixed bit (0x40) on incoming packets
`Header::from_bytes()` does not validate the QUIC fixed bit (0x40) on received packets. The bit is currently only set when encoding outgoing packets.

[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-17.2-3.4.1) requires packets with a cleared fixed bit to be discarded. Encrypted packets will generally fail authentication later, but this unnecessarily reaches the crypto path.

Soundcreates:
Fixes #2557.

RFC 9000 requires the fixed bit (0x40) to be set in every QUIC packet. Validate it while parsing the unprotected packet header and reject malformed packets before they reach the cryptographic path.

Tests: `cargo test -p quiche packet::tests::fixed_bit_must_be_set`

### DIFF
--- a/quiche/src/packet.rs
+++ b/quiche/src/packet.rs
@@ -357,10 +357,11 @@ impl<'a> Header<'a> {
     ) -> Result<Header<'a>> {
         let first = b.get_u8()?;
 
-        // RFC 9000 requires the fixed bit to be set in every QUIC packet.
-        // Header protection does not cover this bit, so it can be validated
-        // before decoding the rest of the header.
-        if first & 0x40 == 0 {
+        // RFC 9000 requires the fixed bit to be set in every QUIC packet
+        // except Version Negotiation packets. Header protection does not
+        // cover this bit, so it can be validated before decoding the rest of
+        // the header.
+        if !Header::is_long(first) && first & 0x40 == 0 {
             return Err(Error::InvalidPacket);
         }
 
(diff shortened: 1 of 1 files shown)


quiche/src/packet.rs
```diff
@@ -357,6 +357,13 @@ impl<'a> Header<'a> {
     ) -> Result<Header<'a>> {
         let first = b.get_u8()?;
 
+        // RFC 9000 requires the fixed bit to be set in every QUIC packet.
+        // Header protection does not cover this bit, so it can be validated
+        // before decoding the rest of the header.
+        if first & 0x40 == 0 {
```
tuplecats: For Long Header packets with the Version field set to 0, this check is not required. 
RFC9000 (17.2): The next bit (0x40) of byte 0 is set to 1, unless the packet is a **Version Negotiation** packet.

For Short Header packets this check is always required.

tuplecats (changes requested):

quiche/src/packet.rs
Soundcreates: Thanks for catching this. I updated the check so short headers always require the fixed bit, while long-header Version Negotiation packets (version 0) are exempt. I also added regression coverage for the exception and pushed commit 6e913f62.

tuplecats (approved):

feymartynov (approved):

PR #2577, opened 2026-08-11
Expose TLS exporter keying material

davidtorcivia:
Adds Connection::export_keying_material, backed by SSL_export_keying_material, with caller-owned output, an explicit label, and optional context. This lets application protocols derive channel-binding material without exposing the TLS session or allocating in quiche.\n\nThe public operation returns InvalidState until quiche marks the TLS handshake established, preventing access during the server interval after its Finished message but before the peer Finished is authenticated.\n\nTests verify pre-handshake and partial-handshake rejection, matching client/server output, label and non-empty-context separation, and peer agreement for an explicitly empty context.\n\nValidated with nightly rustfmt, 1,090 quiche tests, 45 doctests, and deny-warnings Clippy with only the current upstream uninlined_format_args baseline allowed. The change introduces no new Clippy warning.

### DIFF
--- a/quiche/src/lib.rs
+++ b/quiche/src/lib.rs
@@ -7614,6 +7614,18 @@ impl<F: BufFactory> Connection<F> {
         self.handshake.peer_cert()
     }
 
+    /// Fills the provided buffer with TLS exporter keying material.
+    ///
+    /// The exporter is available after the TLS handshake completes. The label
+    /// identifies the protocol usage, while the optional context can separate
+    /// exporter instances within that usage.
+    #[inline]
+    pub fn export_keying_material(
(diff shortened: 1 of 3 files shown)


davidtorcivia:
Rebased onto current master and removed the unrelated recovery commit from the branch. Review identified that BoringSSL can derive the TLS 1.3 exporter secret before quiche marks the connection established; the public method now returns InvalidState in that interval, with a server-side partial-handshake regression. Final validation: 1,090 crate tests, 45 doctests, focused exporter and handshake-state tests, and Clippy clean with only the current upstream uninlined_format_args baseline allowed.

PR #2584, opened 2026-08-13
Add support for QUIC stateless reset

csujedihy:
This PR introduces RFC 9000 stateless reset support to quiche and tokio-quiche. Before this change, quiche already has an incomplete implementation of stateless reset where the stateless reset on-wire token is directly exposed in control interfaces (`quiche::Config::set_stateless_reset_token` and `tokio-quiche::QuicSettings::stateless_reset_token`). Basically, with raw quiche, the app would have to implement stateless reset on their own, like deriving a token from a static key, constructing a stateless reset packet from scratch etc. With tokio-quiche,  its completely broken because it sends the same token for all incoming connections and it can only generate randomized token for NEW_CONNECTION_ID frames.

This PR:
1. Introduces new control interfaces (`quiche::Config::set_stateless_reset_key` and `QuicSettings::stateless_reset_key`) where the app can supply the static key for token generation.
2. Adds helpers for building the stateless reset packets for raw quiche applications.
3. Adds support in tokio-quiche to send stateless reset token (derived from the static key) in TPs and issue new CIDs with reset token derived from the static key. Basically, the app just needs to set the key and everything works.
4. Adds client support for handling stateless reset.
5. Adds reset issuance rate-limiting, which only allows 1 reset outstanding per-2-tuple (IP:port).

Things to note:
1. The old  `tokio-quiche::QuicSettings::stateless_reset_token` isn't removed because we want to take this change in before the upstream change is merged and we don't wanna break internal cloudflare ZT stack.
2. No server side reset receipt logic for simplicity (not practical in reality for a client to send resets).


quiche/src/crypto/mod.rs
```diff
@@ -505,6 +506,67 @@ fn make_nonce(iv: &[u8], counter: u64) -> [u8; MAX_NONCE_LEN] {
...
+    let token = derive_stateless_reset_wire_token(static_key, cid).ok()?;
+
+    // Make the packet indistinguishable from a short-header packet: randomize
+    // its unpredictable bytes, clear the header form bit, and set the fixed bit.
+    let mut reset = vec![0; len];
+    rand::rand_bytes(&mut reset);
+    reset[0] = (reset[0] & 0x3f) | 0x40;
+
+    // A stateless reset is identified by its token in the final 16 bytes.
+    reset[len - 16..].copy_from_slice(&token.to_be_bytes());
```
gregor-cf: nit: maybe use `to_ne_bytes()` since pretty much all modern architectures are little endian. Not that it really matters though,

quiche/src/tls/mod.rs
```diff
@@ -462,7 +462,7 @@ impl Handshake {
     pub fn set_quic_transport_params(
         &mut self, params: &crate::TransportParams, is_server: bool,
     ) -> Result<()> {
-        let mut raw_params = [0; 128];
+        let mut raw_params = [0; 256];
```
gregor-cf: is this needed?

quiche/src/lib.rs
```diff
@@ -1213,14 +1221,24 @@ impl Config {
...
+        self.stateless_reset_key = v;
+    }
+
+    /// Sets the static key used to derive stateless reset tokens.
+    ///
+    /// This is an alias for [`set_stateless_reset_key()`] retained for API
+    /// compatibility.
+    ///
+    /// [`set_stateless_reset_key()`]: Config::set_stateless_reset_key
     pub fn set_stateless_reset_token(&mut self, v: Option<u128>) {
```
gregor-cf, line 1246: nit: add a `#[deprecated]` annotation?

quiche/src/crypto/mod.rs
```diff
@@ -505,6 +506,67 @@ fn make_nonce(iv: &[u8], counter: u64) -> [u8; MAX_NONCE_LEN] {
...
+    Ok(u128::from_be_bytes(token_bytes))
+}
+
+/// Builds a QUIC stateless reset packet for `cid`.
+///
+/// Length is [`RECOMMENDED_STATELESS_RESET_LEN`] plus 0..=7 random bytes,
+/// then clamped to `triggering_len - 1` so the reset cannot loop. This
+/// randomization idea is inspired by msquic's implementation.
+/// Returns `None` when `triggering_len` is too small to produce a valid reset.
+pub fn build_stateless_reset(
```
gregor-cf: nit: `build_stateless_reset_packet`? to make the difference to `derive_stateless_reset_wire_token` clearer?

quiche/src/crypto/mod.rs
```diff
@@ -505,6 +506,67 @@ fn make_nonce(iv: &[u8], counter: u64) -> [u8; MAX_NONCE_LEN] {
...
+/// Maximum length of a stateless reset packet generated by quiche
+/// (`RECOMMENDED` + 7).
+pub const MAX_STATELESS_RESET_LEN: usize = RECOMMENDED_STATELESS_RESET_LEN + 7;
+
+/// Derives a stateless reset token from a static key and connection ID.
+///
+/// This computes HMAC-SHA256(static_key, cid) and returns the first 16 bytes
+/// as a u128 token.
+pub fn derive_stateless_reset_wire_token(
+    static_key: &[u8], cid: &[u8],
```
gregor-cf: nit: would be nice to use a different type for either the static_key or the cid parameter. E.g., a newtype wrapper `struct ResetTokenKey(&u8)` or similar so that caller's can't get the order of the two wrong. 
Or change `static_key` to `u128`, since that's what we use in the config.

quiche/src/lib.rs
gregor-cf: nit: reuse the const we defined earlier for the min length

quiche/src/lib.rs
```diff
@@ -2926,21 +2960,13 @@ impl<F: BufFactory> Connection<F> {
             return false;
         }
 
-        // TODO: we should iterate over all active destination connection IDs
-        // and check against their reset token.
-        match self.peer_transport_params.stateless_reset_token {
-            Some(token) => {
-                let token_len = 16;
+        let token_len = 16;
```
gregor-cf: nit: lets make this a constant (and add comment to constant clarifying that this is an RFC mandated value)

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -77,6 +80,138 @@ const PACKET_RX_YIELD_AFTER: usize = 30;
...
+    Tx: DatagramSocketSend + Send + 'static,
+    M: Metrics,
+{
+    let (tx, mut rx) =
+        mpsc::channel::<StatelessReset>(STATELESS_RESET_QUEUE_CAPACITY);
+
+    spawn_with_killswitch(async move {
+        #[cfg(target_os = "linux")]
+        let would_block_metric =
+            _metrics.write_errors(labels::QuicWriteError::WouldBlock);
```
gregor-cf: nit: remove the underscore and annotate `start_stateless_reset_writer` to allow unused arguments

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -243,11 +398,30 @@ where
         #[cfg(feature = "perf-quic-listener-metrics")]
         let start = std::time::Instant::now();
 
-        if let Some(dcid) = short_dcid(&incoming.buf) {
+        if let Some(dcid) =
+            short_dcid(&incoming.buf).map(ConnectionId::into_owned)
```
gregor-cf: IIRC ConnectionId::into_owned does an allocation, so probably makes sense to remove the into_owned here and only call it when we call `enqueue_stateless_reset()`

tokio-quiche/src/quic/router/mod.rs
```diff
@@ -391,6 +566,52 @@ where
...
+            crate::metrics::quic::stateless_reset_dropped_count().inc();
+            return;
+        }
+        let datagram_len = incoming
+            .gro
+            .and_then(|len| usize::try_from(len).ok())
+            .unwrap_or(incoming.buf.len())
+            .min(incoming.buf.len());
+        let Some(payload) = quiche::build_stateless_reset(
+            &ctx.key.to_be_bytes(),
```
gregor-cf: I think it would be good to have a concrete type for key. Having to remember to call `to_be_bytes()` in different places it otherwise error prone. If the key has a concrete type, that can all be encapsulated in the impl for the type.

gregor-cf (commented):
This is great. I just had a couple of smaller nits.

gregor-cf:
When sending stateless resets, I'm wondering if we want to do more aggressive limiting than the current "in-flight" logic. E.g., if a connection with a lot of packets in flight arrives at a new server and it needs to send state-less resets, we could potentially send one stateless reset per incoming packet. Not the end of the world and TCP would do the same IIRC, but we could consider limiting to one reset per 2-tuple every couple 100ms to every second. But it's not clear that the complexity would be worth it.

quiche/src/tls/mod.rs
csujedihy: It is. Without this change, the tests fail the next line due to not having enough space. Its not a new bug but its not been exercised.

quiche/src/crypto/mod.rs
csujedihy: I am wondering if I should make derive_stateless_reset_wire_token return `[u8]`. The order really doesn't matter here as we are not parsing the token.

csujedihy:
> When sending stateless resets, I'm wondering if we want to do more aggressive limiting than the current "in-flight" logic. E.g., if a connection with a lot of packets in flight arrives at a new server and it needs to send state-less resets, we could potentially send one stateless reset per incoming packet. Not the end of the world and TCP would do the same IIRC, but we could consider limiting to one reset per 2-tuple every couple 100ms to every second. But it's not clear that the complexity would be worth it.

msquic exactly does this and I originally wanted to do the same but simplified it. I don't think it's super complicated. We'd need a hashmap instead of a hashset to store the last_sent timestamp and a heap to evict expired entries opportunistically. I can add it.

tokio-quiche/src/quic/router/mod.rs
csujedihy: Added

tokio-quiche/src/quic/router/mod.rs
csujedihy: Fixed

tokio-quiche/src/quic/router/mod.rs
csujedihy: fixed

quiche/src/crypto/mod.rs
csujedihy: We now have concrete types for `StatelessResetToken` and `StatelessResetKey`

PR #2585, opened 2026-08-14
Porting boringssl bindings over to use bssl-sys with bindgen instead of hardcoding them

emarteca:
1. Install `bindgen`: ```bash cargo install bindgen-cli

2. Build BoringSSL with Rust bindings: cmake -B /path/to/boringssl/build -S /path/to/boringssl \ -DRUST_BINDINGS=$(rustc -vV | sed -n 's/host: //p') cmake --build /path/to/boringssl/build -j$(nproc)
Note: you can use the path to boringssl that is included through cargo

3. Build / test quiche with --no-default-features --features boringssl-bssl-sys: BORINGSSL_BUILD_DIR="/path/to/boringssl/build" \ cargo test --no-default-features --features boringssl-bssl-sys -p quiche

### DIFF
--- a/quiche/src/ffi_slice.rs
+++ b/quiche/src/ffi_slice.rs
@@ -1,0 +1,85 @@
+// Copyright (C) 2026, Cloudflare, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//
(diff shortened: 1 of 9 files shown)


quiche/src/ffi_slice.rs
```diff
@@ -0,0 +1,85 @@
...
+        if N == 0 {
+            core::ptr::null()
+        } else {
+            self.as_ptr()
+        }
+    }
+}
+
+/// See [`FfiSlice`].
+pub(crate) trait FfiMutSlice {
```
dingxiangfei2009, line 63: Ah I suppose this trait is published from `bssl-crypto` crate now. We could import it eventually.

Issue #2631, opened 2026-08-20
Client migration is reachable before handshake confirmation


LiD0209:
# Client migration is reachable before handshake confirmation

## Standard Requirement

- RFC 9000 Section 9: [Connection Migration](https://www.rfc-editor.org/rfc/rfc9000.html#section-9)
- RFC 9000 Section 9.1: [Probing a New Path](https://www.rfc-editor.org/rfc/rfc9000.html#section-9.1)
- RFC 9001 Section 4.1.2: [Handshake Confirmed](https://www.rfc-editor.org/rfc/rfc9001.html#section-4.1.2)

RFC 9000 Section 9 says endpoints `MUST NOT initiate connection migration` before the handshake is confirmed. RFC 9001 Section 4.1.2 further defines client-side handshake confirmation: the client must either receive `HANDSHAKE_DONE` or receive an ACK for one of its 1-RTT packets.

RFC 9000 Section 9.1 permits path validation before migrating, but it does not relax the Section 9 precondition. A client with `handshake_confirmed == false` must not move non-probing traffic to a new address.

## Relevant Source Code

- `quiche/src/lib.rs:7230-7289`: `migrate()` does not check `handshake_confirmed`.
- `quiche/src/lib.rs:9173-9211`: `create_path_on_client()` checks only CID availability, not handshake confirmation.
- `quiche/src/lib.rs:8220-8256`: once `is_established()` is true, the scheduler can emit 1-RTT short packets.
- `quiche/src/lib.rs:8805-8815`: the client sets `handshake_confirmed = true` only when processing `HANDSHAKE_DONE`.

The implementation reduces the migration gate to CID availability instead of enforcing the required protocol state.

## Runtime Evidence

I ran the focused reproducer:

```text
cargo test pre_handshake_client_migration_reachable_with_split_new_cid_and_handshake_done -- --nocapture
```

The test observed:

```text
observed pre-handshake migration: out_size=70, client_confirmed=false, send_from=127.0.0.1:5678
```

The reproducer asserted all of the following:

1. The client still had `handshake_confirmed == false`.
2. The client switched to the new address `127.0.0.1:5678` through `new_scid()` and `migrate()`.
3. The next packet was not a probing packet; it carried `"ping"` stream data.
4. The server received that stream data on the new path.

## Reasoning

[...]

Issue #1489, opened 2023-04-24
Client side `quiche::Connection`  `is_established()` returns true before handshake completes.


tegefaulkes:
For reference https://github.com/MatrixAI/js-quic/issues/9#issuecomment-1519298781.

For context, I have a test in our code that checks if a connection fails if the server fails to authenticate the client. My expectation here is that the server will end up with a `TlsFail` error and close the connection. The client should see the closing frame with the `TLS` error BEFORE the handshake has completed and `is_established()` returns true. 

What I am seeing is that the client's `is_established()` is returning true very early in the handshake procedure.

```
1	0.000000000	127.0.0.1	127.0.0.1	34238	QUIC	1242	Initial, DCID=494077f00d584ca236388136a42f9a6d, SCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, PKN: 0, CRYPTO
2	0.002128503	127.0.0.1	127.0.0.1	34148	QUIC	250	Retry, DCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, SCID=4adf076468fb432462e51fb1018c79749b21e0dc
3	0.003273495	127.0.0.1	127.0.0.1	34238	QUIC	1242	Initial, DCID=4adf076468fb432462e51fb1018c79749b21e0dc, SCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, PKN: 1, CRYPTO
4	0.006133042	127.0.0.1	127.0.0.1	34148	QUIC	1242	Handshake, DCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, SCID=4adf076468fb432462e51fb1018c79749b21e0dc, PKN: 0, CRYPTO
// Client's `is_established()` returns true here.
5	0.039734489	127.0.0.1	127.0.0.1	34238	QUIC	1242	Handshake, DCID=4adf076468fb432462e51fb1018c79749b21e0dc, SCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, PKN: 0, ACK, CRYPTO
6	0.040731996	127.0.0.1	127.0.0.1	34238	QUIC	162	Handshake, DCID=4adf076468fb432462e51fb1018c79749b21e0dc, SCID=fa93ec24c633d1f91662cb586d0b57c46cca0dec, PKN: 1, ACK, CRYPTO
// Server rejects client's TLS cert here
[...]

PR #2664, opened 2026-08-22
quic: expose handshake confirmation state

dillona:
## Summary

Expose QUIC handshake confirmation through the public Rust and C APIs.

This adds:

- `Connection::is_handshake_confirmed()` to the Rust API;
- `quiche_conn_is_handshake_confirmed()` to the C API; and
- documentation describing the difference between handshake completion and confirmation.

For servers, the handshake is confirmed immediately after it completes. For clients, it is confirmed after receiving `HANDSHAKE_DONE`.

## Motivation

`Connection::is_established()` reports that the TLS handshake is complete, but applications may need to know when the QUIC handshake is confirmed. In particular, clients may need to wait for confirmation before performing operations such as connection migration.

The confirmation state already exists internally, but callers currently cannot inspect it. This addresses the API request in #1489 and supports use cases related to the migration requirements discussed in #2631.

This PR only exposes the existing state; it does not change migration enforcement or handshake behavior.

## Testing

- Updated the existing handshake-confirmation test to use the public Rust accessor.
- Added an FFI test verifying that the C and Rust APIs report the same state before and after the handshake.

### DIFF
--- a/quiche/include/quiche.h
+++ b/quiche/include/quiche.h
@@ -519,6 +519,11 @@ void quiche_conn_server_name(const quiche_conn *conn, const uint8_t **out, size_
 // Returns true if the connection handshake is complete.
 bool quiche_conn_is_established(const quiche_conn *conn);
 
+// Returns true if the connection handshake is confirmed. The handshake is
+// confirmed immediately after it completes on the server. On the client, it is
+// confirmed after receiving a HANDSHAKE_DONE frame from the server.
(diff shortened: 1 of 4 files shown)


dillona:
@ghedo, given your earlier explanation in #1489, would you be the right person to review this accessor, or could you point me to someone on the Protocols team? The rebased PR’s CI runs also need maintainer approval.

Issue #2696, opened 2026-09-01
Empty non-FIN STREAM frames cause connection flow-control double-counting


jbaldassari:
## Background

This was discovered during load/soak testing with 32 parallel `iperf3` streams transferring approximately 700 Mb/s over QUIC through Netty’s quiche integration, with a lossy WAN path between client and server.

The receiver occasionally closed the QUIC connection with `FLOW_CONTROL_ERROR`, typically within 60 minutes under this workload. Instrumentation added to a local quiche build showed `rx_data` exceeding the sum of the streams’ high-water marks by exact whole-frame multiples.

Affected revision tested: [0b1a89af](https://github.com/cloudflare/quiche/tree/0b1a89af341b7e42c4e75fedf7bdda246d8319af).

## Root cause

A zero-length non-`FIN` `STREAM` frame advances the receiver’s largest received offset for that stream (RFC 9000 Section 19.8) and must therefore be counted against connection-level flow control.

The receive path calculates:

```rust
data.max_off().saturating_sub(stream.recv.max_off())
```

before calling `RecvBuf::write()`. It subsequently adds that delta to `rx_data`.

However, `RecvBuf::write()` discards an empty non-`FIN` buffer without advancing the stream’s high-water mark. Consequently:

1. An empty frame advances from offset `X` to `X + N`.
2. The connection-level ledger charges `N` bytes.
3. `RecvBuf::write()` discards the frame while leaving the stream high-water mark at `X`.
4. When real data through `X + N` arrives, the receive path calculates and charges the same `N` bytes again.

This creates permanent connection-level ledger drift. Repetition eventually causes a legal `STREAM` frame to appear to exceed `MAX_DATA`, and the receiver closes the connection with `FLOW_CONTROL_ERROR`, even though the sender has not exceeded its granted credit.

## Why the sender emits these frames

`send_single()` can emit an empty non-`FIN` `STREAM` frame when the remaining packet capacity fits the encoded `STREAM` header exactly, leaving zero bytes for payload.

The stream-packing block is entered only when more than `MAX_STREAM_OVERHEAD` bytes remain. Therefore, this requires an actual header larger than that constant. For example, a stream ID of at least 64 and an offset of at least `2^30` produce a 13-byte header. This explains why the problem primarily appears late in long-running bulk transfers.

[...]

PR #2697, opened 2026-09-01
Fix for issue 2696: connection flow-control double-counting from zero-length STREAM frames

jbaldassari:
Fix for https://github.com/cloudflare/quiche/issues/2696

A zero-length non-`FIN` `STREAM` frame advances the peer's largest received offset (RFC 9000 Section 19.8). The receive path charges connection flow control from that offset, but `RecvBuf::write()` discards empty non-`FIN` buffers without recording the new high-water mark. When in-flight data covering the gap later arrives, the gap is charged a second time.

`send_single()` can produce such frames when the remaining packet capacity fits the encoded `STREAM` header but no payload. This occurs at large stream IDs and offsets, where the actual header exceeds `MAX_STREAM_OVERHEAD`. The bytes preceding the frame's offset have already been charged to the sender's `tx_data`, so the frame is legal but redundant and exposes the receiver accounting bug.

Between two quiche endpoints, the inflation accumulates silently until a legal frame appears to exceed `MAX_DATA`. The receiver then closes the connection with `FLOW_CONTROL_ERROR`, even though the sender has consumed exactly the credit it was granted. In long-running bulk transfers over a lossy path, this caused 1–3 spurious connection closures per hour. Instrumented captures showed `rx_data` exceeding the sum of the stream high-water marks by exact whole-frame multiples.

This change:
1. Advances the `RecvBuf` high-water mark before returning for an empty non-`FIN` buffer, keeping the connection flow-control charge and stream state in lockstep.
2. Defensively suppresses zero-length non-`FIN` `STREAM` frames in `send_single()`. When packet capacity prevents stream data from being emitted, the sender remains marked as size-limited rather than app-limited.

The `empty_stream_frame` expectations are updated accordingly. The previous sequence also depended on accepting a final size below the largest received offset, which RFC 9000 Section 4.5 forbids.

New regression coverage:
* Injects an empty forward `STREAM` frame and verifies that its offset is charged exactly once. Without the receiver fix, this fails with a spurious `Error::FlowControl`.
* Exercises the sender at a large stream offset across the header-only packet-capacity window, verifies that no zero-length non-`FIN` frame is emitted, and confirms that the queued data is subsequently delivered intact.

### DIFF
--- a/quiche/src/stream/recv_buf.rs
+++ b/quiche/src/stream/recv_buf.rs
@@ -121,8 +121,16 @@ impl RecvBuf {
             self.fin_off = Some(buf.max_off());
         }
 
-        // No need to store empty buffer that doesn't carry the fin flag.
+        // No need to store empty buffer that doesn't carry the fin flag, but
+        // its offset still advances the largest received offset (RFC 9000
+        // Section 19.8). The caller charges connection-level flow control
+        // from that offset, so the high-water mark must advance in lockstep:
+        // otherwise the gap up to the frame's offset is charged a second
+        // time when the in-flight data covering it arrives, and the
+        // accumulated double-counting eventually makes a perfectly legal
+        // frame appear to exceed MAX_DATA.
         if !buf.fin() && buf.is_empty() {
+            self.len = cmp::max(self.len, buf.max_off());
             return Ok(());
         }
 
@@ -542,10 +550,12 @@ mod tests {
 
         assert_emit_discard(&mut recv, emit, 32, 5, false, None);
 
-        // Don't store non-fin empty buffer.
+        // A non-fin empty buffer is not stored, but its offset advances the
+        // largest received offset: connection-level flow control is charged
+        // from that offset, so the high-water mark must track it.
         let buf = RangeBuf::from(b"", 10, false);
         assert!(recv.write(buf).is_ok());
-        assert_eq!(recv.len, 5);
+        assert_eq!(recv.len, 10);
         assert_eq!(recv.off, 5);
         assert_eq!(recv.data.len(), 0);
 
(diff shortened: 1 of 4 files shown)


normanmaurer:
@ghedo any chance we can get this one reviewed and merged ? People do hit this when using netty.

quiche/src/stream/recv_buf.rs
```diff
@@ -121,8 +121,16 @@ impl RecvBuf {
...
+        // No need to store empty buffer that doesn't carry the fin flag, but
+        // its offset still advances the largest received offset (RFC 9000
+        // Section 19.8). The caller charges connection-level flow control
+        // from that offset, so the high-water mark must advance in lockstep:
+        // otherwise the gap up to the frame's offset is charged a second
+        // time when the in-flight data covering it arrives, and the
+        // accumulated double-counting eventually makes a perfectly legal
+        // frame appear to exceed MAX_DATA.
         if !buf.fin() && buf.is_empty() {
+            self.len = cmp::max(self.len, buf.max_off());
```
ghedo, line 128: This early return breaks the drained-stream invariant that `off == len`. `Connection` already marks `max_off_delta` as consumed for a draining STREAM frame, so a later `RESET_STREAM` credits `final_size - self.off` again. Could we update `off` when `self.drain` is true and add a regression test covering shutdown, an empty non-FIN STREAM frame, then `RESET_STREAM`?

ghedo:
Also, the comments are quite wordy and could probably be made more concise.

jbaldassari:
Thanks for taking a look, @ghedo .  I appreciate the feedback as I'm still new to this code.  I believe I've addressed your comments, but please let me know if there is anything else you'd like me to change.

ghedo (approved):

inamdarminaz (approved):

PR #2709, opened 2026-09-05
h3: free stream state for streams reset before local finish

### LINKED ISSUE #2524: Connection leaks state for streams reset before the local side finishes
Hello

## Summary

Android has been carrying [a memleak patch](https://android.googlesource.com/platform/external/rust/android-crates-io/+/bf11ee061e51cf69a8289a0a1b321e8526329722/crates/quiche/patches/0001-Remove-memleak.patch) for some time on top of v0.17.1.

Current master appears to handle normal finish via the [local_finished](https://github.com/cloudflare/quiche/blob/c4c0b978461aa153399a90217d85bebd1800f84d/quiche/src/h3/mod.rs#L2919) path, but still leaks if a peer resets the stream before the sending is finished.

Each such stream leaves a dangling entry in the `streams` map for the lifetime of the connection. On a long-lived server facing clients that abort request, the map grows without bound, causing a memory leak proportional to the number of aborted requests.

## Root cause

From the above-linked local_finished snippet,, we can observe the cleanup of `h3::Connection::streams` is gated on `local_finished()`:
```rust
fn remove_local_finished_stream(&mut self, stream_id: u64) {
    if let hash_map::Entry::Occupied(stream) = self.streams.entry(stream_id) {
        if stream.get().local_finished() {     // <-- only if WE finished sending
            stream.remove();
        }
    }
}
```

The reset path in `poll()` calls the above helper:

```rust
Err(Error::TransportError(crate::Error::StreamReset(e))) => {
    self.remove_local_finished_stream(s);      // <-- no-op when !local_finished
    return Ok((s, Event::Reset(e)));
},
```

So when a stream is reset while the local side is still open, the `Event::Reset` is delivered correctly, but the `streams` entry is never removed. The duplicate-reset dedup is handled by the transport layer (a second RESET_STREAM yields `Done`), so the leak is silent with no symptom other than unbounded memory growth.

The poll loop even carries an acknowledging TODO at the point the free should happen:

```rust
if conn.stream_finished(s) {
    self.process_finished_stream(s);
}
// TODO: check if stream is completed so it can be freed
```

The same gap exists in `pop_finished_stream()`, which can also surface a pending reset (via `Event::Reset`), after calling the gated `remove_local_finished_stream()`.

## Why existing tests miss it

`priority_update_request_collected_stopped` exercises precisely this scenario (request sent with `fin=false`, then client shuts down both directions), but only asserts on the **transport** map (`s.pipe.server.streams.is_collected(0)`); it never checks the **HTTP/3** map (`s.server.streams.len()`), so the dangling H3 entry goes unnoticed.

## Reproduction

Request without fin, no response, then peer reset, then assert the H3 map drains:

```rust
let init = s.server.streams.len();
let (stream, _req) = s.send_request(false).unwrap();   // local side NOT finished
// ... server reads headers ...
s.pipe.client.stream_shutdown(stream, Shutdown::Write, 0x100).unwrap();
s.pipe.client.stream_shutdown(stream, Shutdown::Read, 0x100).unwrap();
s.advance().ok();
assert_eq!(s.poll_server(), Ok((stream, Event::Reset(0x100))));
assert_eq!(s.server.streams.len(), init);              // FAILS
```

## Potential fix

I used Claude to create [a fix which generalizes the cleanup helper](https://github.com/10ne1/quiche/commit/21610fb788a7b20f2f2f380eec51f864c4d40e7f) (`remove_local_finished_stream()`) on the latest `master` branch, adding a transport-level `conn.streams.is_collected()` check, so streams that are dead in both directions get freed even when local_finished() never became true. 

I'm unsure how good this is, so I will let others review it or come up with other suggestions before opening a PR.

PranshulSoni:
Fixes #2524

## Summary

When a peer resets a request stream before the local side finishes sending, the HTTP/3 layer kept its `streams` entry for the lifetime of the connection. On a long-lived server facing clients that abort requests, the map grows without bound. Android has been carrying a downstream patch for this on v0.17.1 (linked in the issue).

## Problem

`remove_local_finished_stream()` only freed the h3 entry when `local_finished()` was true. The reset paths in `poll()` and `pop_finished_stream()` surface `Event::Reset` and then call that helper, which is a no-op whenever the local send direction is still open — e.g. a request received without FIN whose response was never sent.

The transport layer has no such gap: when both directions are dead it removes the stream from its own map and marks it collected (`streams::StreamMap::collect()`), which is also why this leak was silent — the dangling entry lives only in `h3::Connection::streams`.

The existing test `priority_update_request_collected_stopped` covers this scenario but only asserts on the transport map (`s.pipe.server.streams.is_collected(0)`), so the dangling h3 entry went unnoticed.

## Solution

Generalize the cleanup helper to `remove_finished_stream()`, which frees the entry when either the local side finished sending or the transport already collected the stream:

```rust
if stream.get().local_finished() || conn.streams.is_collected(stream_id) {
    stream.remove();
}
```

`is_collected()` is only true once the stream is complete in both directions (`Stream::is_complete()`), so this cannot drop state for a stream that still has work in either direction. It is applied at both reset-surfacing sites: the `StreamReset` arm in `poll()` and `pop_finished_stream()`.

Safety of the additional arm for each h3 stream type:

- Request/push streams: this is exactly the leak being fixed. After collection, h3 send paths (`do_send_body`, `send_headers`) already treat a missing entry as `FrameUnexpected`, and the PRIORITY_UPDATE receive path skips updates for collected streams via `conn.stream_closed()` without re-inserting.
- Control/QPACK streams: reset of a registered critical stream surfaces `ClosedCriticalStream` from `process_control_stream()`/`close_conn_if_critical_stream_finished()` before the reset arm, and a second critical stream is rejected as a protocol error at type detection, so this path cannot run for them.

The `local_finished()` disjunct keeps the existing behavior for streams that finish locally while the transport still buffers state (transport-owned collection stays on its own triggers).

## Testing

- New regression test `reset_stream_before_local_finish_frees_h3_state` reproduces the issue scenario (request without FIN, peer resets both directions) and asserts the h3 map drains. Verified it fails on unpatched code (`left: 4, right: 3`) and passes with the fix.
- `cargo test --all-targets --features=async,ffi,qlog --workspace`: all pass (quiche lib: 1099, including both CC algorithm parameterizations).
- `cargo test --doc --features=async,ffi,qlog --workspace`: all pass.
- `cargo clippy --features=async,ffi,qlog --workspace --all-targets -- -D warnings`: clean.
- `cargo +nightly fmt -- --check`: clean for the touched file (two pre-existing fmt diffs exist on upstream `master` in `lib.rs`/`tests.rs`, unrelated to this change).

### DIFF
--- a/quiche/src/h3/mod.rs
+++ b/quiche/src/h3/mod.rs
@@ -2165,7 +2165,7 @@ impl Connection {
                 // Return early if the stream was reset, to avoid returning
                 // a Finished event later as well.
                 Err(Error::TransportError(crate::Error::StreamReset(e))) => {
-                    self.remove_local_finished_stream(s);
+                    self.remove_finished_stream(conn, s);
 
                     return Ok((s, Event::Reset(e)));
                 },
@@ -2985,9 +2985,18 @@ impl Connection {
         }
     }
 
-    fn remove_local_finished_stream(&mut self, stream_id: u64) {
+    // Frees the HTTP/3 state for a stream that is done in both directions:
+    // either the local side finished sending, or the transport already
+    // collected the stream (e.g. the peer reset it before the local side
+    // finished). Otherwise the entry would leak for the lifetime of the
+    // connection.
+    fn remove_finished_stream<F: BufFactory>(
+        &mut self, conn: &super::Connection<F>, stream_id: u64,
+    ) {
         if let hash_map::Entry::Occupied(stream) = self.streams.entry(stream_id) {
-            if stream.get().local_finished() {
+            if stream.get().local_finished() ||
+                conn.streams.is_collected(stream_id)
+            {
                 stream.remove();
             }
         }
@@ -2998,7 +3007,7 @@ impl Connection {
     ) -> Option<(u64, Event)> {
         let finished = self.finished_streams.pop_front()?;
 
-        self.remove_local_finished_stream(finished);
+        self.remove_finished_stream(conn, finished);
 
         if conn.stream_readable(finished) {
             // The stream is finished, but is still readable, it may indicate
@@ -8541,6 +8550,50 @@ mod tests {
         // The server stream should be gone now
         assert_eq!(s.client.streams.len(), init_streams_client);
         assert_eq!(s.server.streams.len(), init_streams_server);
+    }
+
+    #[test]
+    /// A stream reset by the peer before the local side finished sending must
+    /// not leak its HTTP/3 state: the entry should be freed once the stream is
+    /// done in both directions, even though `local_finished()` never became
+    /// true.
+    fn reset_stream_before_local_finish_frees_h3_state() {
+        let mut s = Session::new().unwrap();
+        s.handshake().unwrap();
+
+        let init_streams_server = s.server.streams.len();
+
+        // Client sends a request without fin. The server hasn't sent a
+        // response, so its local side is not finished.
+        let (stream, req) = s.send_request(false).unwrap();
+        let ev_headers = Event::Headers {
+            list: req,
+            more_frames: true,
+        };
+
+        assert_eq!(s.poll_server(), Ok((stream, ev_headers)));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        assert_eq!(s.server.streams.len(), init_streams_server + 1);
+
+        // Client resets the stream in both directions.
+        s.pipe
+            .client
+            .stream_shutdown(stream, crate::Shutdown::Write, 0x100)
+            .unwrap();
+        s.pipe
+            .client
+            .stream_shutdown(stream, crate::Shutdown::Read, 0x100)
+            .unwrap();
+
+        s.advance().ok();
+
+        assert_eq!(s.poll_server(), Ok((stream, Event::Reset(0x100))));
+        assert_eq!(s.poll_server(), Err(Error::Done));
+
+        // The HTTP/3 stream state must have been freed.
+        assert_eq!(s.server.streams.len(), init_streams_server);
+        assert!(s.pipe.server.streams.is_collected(stream));
     }
 }
 


quiche/src/h3/mod.rs
```diff
@@ -2165,7 +2165,7 @@ impl Connection {
                 // Return early if the stream was reset, to avoid returning
                 // a Finished event later as well.
                 Err(Error::TransportError(crate::Error::StreamReset(e))) => {
-                    self.remove_local_finished_stream(s);
+                    self.remove_finished_stream(conn, s);
```
ghedo, line 2168: This cleanup also needs to cover resets consumed through `recv_body_buf()`. After `poll()` returns `Data`, the peer can reset both directions and `recv_body()` will consume `StreamReset`, causing QUIC to collect the stream. Subsequent `poll()` calls return `Done`, so neither cleanup site added here runs and the H3 entry remains indefinitely. I reproduced this with the H3 stream count staying one above baseline. Please handle `StreamReset` from both `try_consume_data()` and `process_readable_stream()` in `recv_body_buf()`.

quiche/src/h3/mod.rs
```diff
@@ -8542,6 +8551,50 @@ mod tests {
         assert_eq!(s.client.streams.len(), init_streams_client);
         assert_eq!(s.server.streams.len(), init_streams_server);
     }
+
+    #[test]
+    /// A stream reset by the peer before the local side finished sending must
+    /// not leak its HTTP/3 state: the entry should be freed once the stream is
+    /// done in both directions, even though `local_finished()` never became
+    /// true.
+    fn reset_stream_before_local_finish_frees_h3_state() {
```
ghedo, line 8560: This test duplicates `priority_update_request_collected_stopped`, which already creates an unfinished request, shuts down both directions, polls the reset, and verifies transport collection. Could we add the initial H3 stream count and cleanup assertion to that existing test instead, then remove this standalone test? That would cover this regression with only two additional assertions. The separate `recv_body()` path can similarly be covered by extending `reset_finished_at_server_with_data_pending_2`.

ghedo (changes requested):
The reset cleanup remains incomplete for resets consumed through `recv_body_buf()`. The regression coverage can also reuse existing tests.

PR #2715, opened 2026-09-07
Explicitly marking Connection as Sync with safety comment

emarteca:
The `Connection` struct is required to be `Sync` for the test `connection_must_be_sync` but this is not necessarily guaranteed to be sync-safe if the API is expanded. This PR moves the impl of Sync from the Handshake onto the Connection and adds a safety comment to make this more explicit.


dingxiangfei2009:
I would like to fill in some background context. @emarteca and I have be deliberating on where the thread-safety guarantee should be rooted and where the safety boundary should be drawn.

I believe that the safety assertion should be moved onto the public type `Connection` because it is ultimately the API surface of `Connection` that guarantees thread-safety, not `Handshake` alone. `Handshake` is almost transient and no instances or borrows of `Handshake` are reachable from the public `Connection` API as far as I could tell.

However, I would like to advocate for dropping `unsafe impl Sync for Handshake` because it contains a `SSL*` connection handle. At no point of time, not even during handshake, is `SSL*` ever thread-safe. For the sanity, I think it is better to drop this assertion so that we could catch data race bugs more easily.

PR #2719, opened 2026-09-09
stream: compress collected stream tracking into ranges

ghedo:
Store collected stream sequences in separate range sets for each stream type so adjacent completed streams share a range instead of separate hash set entries. This reduces retained memory for sequential stream usage without changing stream lifecycle or credit behavior.

Add exact RangeSet membership checks and cover type separation, out-of-order collection, implicit gaps, and stream limit rejection.


hargut (approved):
Looks good.

quiche/src/stream/mod.rs
```diff
@@ -111,6 +112,26 @@ type BuildStreamIdHasher = std::hash::BuildHasherDefault<StreamIdHasher>;
 pub type StreamIdHashMap<V> = HashMap<u64, V, BuildStreamIdHasher>;
 pub type StreamIdHashSet = HashSet<u64, BuildStreamIdHasher>;
 
+/// Tracks collected stream sequences separately for each stream type.
+#[derive(Default)]
+struct CollectedStreams {
+    // The default capacity is unlimited because evicting a tombstone would
+    // allow a collected stream to be recreated.
+    ranges: [RangeSet; 4],
```
jrouviere: Might be worth boxing this array to allocate lazily. The `RangeSet` enum contains a `SmallVec<4>` with two `u64`, so that starts to add up.

Alternatively we could change the SmallVec to a Vec so everything benefits from reduced memory usage.

jrouviere (approved):

quiche/src/stream/mod.rs
ghedo: Addressed in df16f4538.

CollectedStreams now uses Option<Box<[RangeSet; 4]>>, so the 352-byte array is allocated only when the first stream is collected. Empty lookups, successful stream creation, and stream-limit rejection all leave it uninitialized. RangeSet itself is unchanged.

The layout probe reports CollectedStreams at 8 bytes and RangeSet at 88 bytes. The tests cover lazy initialization while retaining type separation, duplicate handling, merging, fragmented and large IDs, implicit gaps, and stream-credit behavior.

Verified with:
- cargo +nightly fmt -- --check
- cargo test --offline -p quiche --lib --features=ffi,qlog (1,126 passed)
- cargo clippy --offline -p quiche --all-targets --features=ffi,qlog -- -D warnings

PR #2720, opened 2026-09-10
tokio-quiche: Add path event callback

kkelkar:
Forward existing path events from each tokio-quiche worker to the configured connection hook.


tokio-quiche/src/quic/io/worker.rs
```diff
@@ -276,6 +277,20 @@ pub(crate) struct IoWorkerParams<Tx, M> {
...
     pub(crate) metrics: M,
+    pub(crate) connection_hook:
+        Option<Arc<dyn ConnectionHook + Send + Sync + 'static>>,
+}
+
+fn notify_path_events(
+    connection_hook: Option<&(dyn ConnectionHook + Send + Sync + 'static)>,
+    qconn: &mut QuicheConnection,
+) {
+    while let Some(path_event) = qconn.path_event_next() {
```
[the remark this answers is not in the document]
kkelkar: fixed now and added tests

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -416,6 +433,8 @@ where
                     did_recv = true;
                 }
 
+                notify_path_events(self.connection_hook.as_deref(), qconn);
```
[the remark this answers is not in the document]
kkelkar: updated

tokio-quiche/src/quic/io/worker.rs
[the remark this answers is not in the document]
ghedo: I think the original implementation was better actually, because nothing else drains these events so if there's no hook they just end up accumulating for no reason. I'd rather drain the queue regardless of the hook.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -276,6 +277,35 @@ pub(crate) struct IoWorkerParams<Tx, M> {
...
+        return;
+    }
+
+    while let Some(path_event) = qconn.path_event_next() {
+        log::info!(ratelimit=100/m; "Quiche path event {:?}", path_event);
+        hook.on_path_event(qconn, &path_event);
+    }
+}
+
+fn run_and_notify_path_events<T>(
```
ghedo: This wrapper kind of makes things harder to follow... since it's only used in a couple of places I'd rather just inline the `notify_path_events()` call and the result propagation.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -412,11 +444,19 @@ where
...
                 {
-                    self.process_incoming(qconn, pkt)?;
+                    let recv_result = self.process_incoming(qconn, pkt);
+                    notify_path_events(self.connection_hook.as_deref(), qconn);
+                    recv_result?;
                     did_recv = true;
                 }
 
-                self.conn_stage.on_read(did_recv, qconn, ctx)?;
+                notify_path_events(self.connection_hook.as_deref(), qconn);
```
ghedo: This drain immediately follows a drain after every processed packet, and similar extra drains occur after `gather_data_from_quiche_conn()` and `on_flush()`. I think many of these could be consolidated.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -412,11 +444,19 @@ where
                     .take()
                     .or_else(|| ctx.incoming_pkt_receiver.try_recv().ok())
                 {
-                    self.process_incoming(qconn, pkt)?;
+                    let recv_result = self.process_incoming(qconn, pkt);
```
ghedo: If the `process_incoming()` call fails the connection is going to be closed anyway, do we actually care about events being reported in that case? It's likely there wouldn't be any events anyway, so this could probably just be consolidated into the `notify_path_events()` call after the loop.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -292,6 +322,7 @@ pub(crate) struct IoWorker<Tx, M, S> {
     #[cfg(feature = "perf-quic-listener-metrics")]
     init_rx_time: Option<SystemTime>,
     metrics: M,
+    connection_hook: Option<Arc<dyn ConnectionHook + Send + Sync + 'static>>,
```
ghedo: The connection hook is shared per-connection state rather than worker-stage state. Could this live in `ConnectionStageContext` instead, which already persists across handshake, running, and close stages?

ghedo (changes requested):
Thanks for the PR, it generally looks good but added a few comments that could simplify things a bit. It also seems like the AI reviewer took you down the wrong path, sorry about that ☹️

tokio-quiche/src/quic/io/worker.rs
kkelkar: The only thing I can think of is `PathEvent::Closed`. Should we keep this in case a new `PathEvent` gets added in future that could be useful (can't think of any right now). Otherwise I can remove this call

tokio-quiche/src/quic/io/worker.rs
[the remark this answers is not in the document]
kkelkar: I've put the logic back

tokio-quiche/src/quic/io/worker.rs
kkelkar: removed the wrapper

tokio-quiche/src/quic/io/worker.rs
kkelkar: Nevermind, updated per your suggestion

tokio-quiche/src/quic/io/worker.rs
kkelkar: Moved to `ConnectionStageContext`

tokio-quiche/src/quic/io/worker.rs
kkelkar: should be done now

kkelkar:
@ghedo thanks for reviewing! I hope I've got everything right now...

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -416,7 +429,12 @@ where
                     did_recv = true;
                 }
 
-                self.conn_stage.on_read(did_recv, qconn, ctx)?;
+                notify_path_events(ctx.connection_hook.as_deref(), qconn);
+
+                let on_read_result =
+                    self.conn_stage.on_read(did_recv, qconn, ctx);
+                notify_path_events(ctx.connection_hook.as_deref(), qconn);
```
ghedo, line 438: Thinking about this a bit more, I realized that rather than notify in every place where we might return early, we could simply do it where `work_loop()` returns:
* https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/quic/io/worker.rs#L1059
* https://github.com/cloudflare/quiche/blob/master/tokio-quiche/src/quic/io/worker.rs#L1059

so like:
```
let mut work_loop_result = self.work_loop(&mut qconn, &mut ctx).await;
notify_path_events(ctx.connection_hook.as_deref(), &mut qconn);
```

which will simplify things a bit as many of the calls will be superfluous, and cover all error paths automatically now and in the future, rather than having to do it manually (which is error-prone).

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -1308,9 +1337,119 @@ fn random_u128() -> u128 {
...
+    }
+
+    #[test]
+    fn path_events_are_forwarded_to_connection_hook() {
+        let (mut pipe, server_addr, client_addr) = connection_with_path_events();
+        let hook = Arc::new(RecordingHook(Mutex::new(Vec::new())));
+        let connection_hook: Arc<dyn ConnectionHook + Send + Sync + 'static> =
+            hook.clone();
+
+        notify_path_events(Some(connection_hook.as_ref()), &mut pipe.server);
```
ghedo: Looking at these tests more, it seems like it'd be better to exercise the behaviour thropugh an actual `ApplicationOverQuic` implementation rather than calling `notify_path_events()`, to ensure that the notify points in the worker actually work. I _think_ this would also allow moving these tests into `tests/integration_tests/` along the other integration tests.

ghedo (changes requested):
Thanks for the update, getting there... Added a couple more comments but overall approach seems good.

tokio-quiche/src/quic/io/worker.rs
kkelkar: Added new integration test in `tokio-quiche/tests/integration_tests/path_events.rs` that verifies notification from worker is received.

tokio-quiche/src/quic/io/worker.rs
kkelkar: Makes sense. Made the change.

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -278,6 +279,18 @@ pub(crate) struct IoWorkerParams<Tx, M> {
     pub(crate) metrics: M,
 }
 
+fn notify_path_events(
+    connection_hook: Option<&(dyn ConnectionHook + Send + Sync + 'static)>,
+    qconn: &mut QuicheConnection,
+) {
+    while let Some(path_event) = qconn.path_event_next() {
+        log::info!(ratelimit=100/m; "Quiche path event {:?}", path_event);
```
jrouviere: Probably a left-over log? I think we only use info logs in examples and tools.

kkelkar: will remove

ghedo (approved):

tokio-quiche/src/quic/io/worker.rs
```diff
@@ -416,7 +428,10 @@ where
                     did_recv = true;
                 }
 
+                notify_path_events(ctx.connection_hook.as_deref(), qconn);
+
                 self.conn_stage.on_read(did_recv, qconn, ctx)?;
+                notify_path_events(ctx.connection_hook.as_deref(), qconn);
```
jrouviere, line 438: I see that has been discussed already, but I don't fully understand why the calls to `notify_path_events` are where they are, what is the logic here?

jrouviere (approved):
LGTM overall, left a small question,  might be worth explaining the logic in a comment.

tokio-quiche/src/quic/io/worker.rs
ghedo: I think the practical effect is that application-defined `process_reads()` and `process_writes()` hooks (called by `on_read()`) won't see path events generated up to that point, which seems like a reasonable thing given we are saying we will expose the events though the new callback, but could go either way TBH.

kkelkar: I added a comment to explain the logic

kkelkar:
Sorry @ghedo, the latest push seems to have disabled auto-merging.

PR #2721, opened 2026-09-10
quiche: Add PMTU path event

kkelkar:
Notify applications when PMTU discovery changes the validated packet size limit for a network path.


quiche/src/lib.rs
```diff
@@ -3531,24 +3534,29 @@ impl<F: BufFactory> Connection<F> {
 
         // Process acked frames. Note that several packets from several paths
         // might have been acked by the received packet.
+        let mut pmtu_events = Vec::new();
```
ghedo: Rather than having to use an intermediate `Vec` to make borrow checker happy, I think you could split the `PathMap` reference with something like:
```
    /// Returns a mutable iterator over all existing paths and the path event
    /// queue.
    #[inline]
    pub(crate) fn iter_mut_and_events(
        &mut self,
    ) -> (slab::IterMut<'_, Path>, &mut VecDeque<PathEvent>) {
        (self.paths.iter_mut(), &mut self.events)
    }
```
then you can use the iterator for the loop:
```
        let (paths, path_events) = self.paths.iter_mut_and_events();
        for (_, p) in paths {
            ...
        }
```
and just push the events to the VecDeque directly (`path_events.push_back(event)`).

ghedo (changes requested):

quiche/src/lib.rs
kkelkar: Nice idea, let me try this

kkelkar: updated

ghedo (approved):

quiche/src/path.rs
```diff
@@ -1073,6 +1129,140 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn pmtu_event_only_reports_limit_changes() {
```
ghedo: Could we drop these first two tests? `pmtud_probe_success` already covers an increase, decrease, and unchanged value. `successful_probe_notifies_current_pmtu` also duplicates PMTUD algorithm tests and never exercises the event queue.

quiche/src/path.rs
```diff
@@ -1073,6 +1129,140 @@ mod tests {
...
+                    peer,
+                    pmtu: 1300,
+                });
+                break;
+            }
+        }
+    }
+
+    #[test]
+    fn disabling_pmtud_does_not_notify() {
```
ghedo: Could these enable/disable cases be folded into the existing `enable_pmtud_mid_handshake` and `disable_pmtud_mid_handshake` tests? The following reinitialization test already covers disabling without an event, making this test largely redundant.

quiche/src/lib.rs
```diff
@@ -4301,9 +4316,27 @@ impl<F: BufFactory> Connection<F> {
...
+                            trace!("pmtud probe dropped: {failed_probe}");
+
+                            let local = p.local_addr();
+                            let peer = p.peer_addr();
                             if let Some(pmtud) = p.pmtud.as_mut() {
-                                trace!("pmtud probe dropped: {failed_probe}");
+                                let old_pmtu = pmtud.get_current_mtu();
                                 pmtud.failed_probe(failed_probe);
+                                let new_pmtu = pmtud.get_current_mtu();
+                                let pmtu_validated = pmtud.get_pmtu().is_some();
```
ghedo: Is `pmtu_validated` necessary here? When no event exists, the validated limit did not change and recovery was already updated when that limit was validated. This could update recovery only inside the `if let Some(event)` branch.

quiche/src/tests.rs
```diff
@@ -12287,28 +12312,40 @@ fn pmtud_probe_retry_after_loss(
...
+    let active_path = pipe.client.paths.get_active_mut().unwrap();
+    let pmtud = active_path.pmtud.as_mut().unwrap();
 
     // MTU should finally update
     let current_mtu = pmtud.get_current_mtu();
     assert_eq!(current_mtu, 1299);
 
     // Verify should_probe gets reset
     assert!(!pmtud.should_probe());
+    assert_eq!(active_path.recovery.max_datagram_size(), current_mtu);
```
ghedo: This assertion appears unrelated to path-event behavior and successful probes already update recovery. It also does not exercise the newly added lost-probe update, so I suggest removing it unless that separate behavior change is intentional.

quiche/src/lib.rs
```diff
@@ -7702,12 +7735,29 @@ impl<F: BufFactory> Connection<F> {
...
+                None
             }
+        } else {
+            None
+        };
+
+        if let Some(event) = event {
+            self.paths.notify_event(event);
         }
     }
```
ghedo, line 7759: Could we simplify the nested conditionals with early returns while preserving the behavior?

```suggestion
    pub fn revalidate_pmtu(&mut self) {
        let Ok(path) = self.paths.get_active_mut() else {
            return;
        };

        let local = path.local_addr();
        let peer = path.peer_addr();
        let Some(pmtud) = path.pmtud.as_mut() else {
            return;
        };

        let old_pmtu = pmtud.get_current_mtu();
        pmtud.revalidate_pmtu();

        let Some(event) =
            path::pmtu_event(local, peer, old_pmtu, pmtud.get_current_mtu())
        else {
            return;
        };

        self.paths.notify_event(event);
    }
```

ghedo (approved):
Added a few comments to suggest some simplifications, but lgtm overall.

quiche/src/lib.rs
kkelkar: Good catch - `pmtu_validated` is not needed. Will fix

quiche/src/path.rs
kkelkar: will do

quiche/src/path.rs
kkelkar: I think so. will update

quiche/src/tests.rs
kkelkar: okay, I will remove

quiche/src/lib.rs
kkelkar: yes

quiche/src/path.rs
```diff
@@ -117,6 +118,34 @@ pub enum PathEvent {
...
+
+        /// The current validated PMTU limit for normal application traffic.
+        pmtu: usize,
+    },
+}
+
+pub(crate) fn pmtu_event(
+    local: SocketAddr, peer: SocketAddr, old: usize, new: usize,
+) -> Option<PathEvent> {
+    (old != new).then_some(PathEvent::PmtuUpdated {
```
jrouviere, line 144: Never seen that syntax before!

quiche/src/lib.rs
```diff
@@ -7702,13 +7730,32 @@ impl<F: BufFactory> Connection<F> {
...
+        let old_pmtu = pmtud.get_current_mtu();
+        pmtud.revalidate_pmtu();
+
+        let Some(event) =
+            path::pmtu_event(local, peer, old_pmtu, pmtud.get_current_mtu())
+        else {
+            return;
+        };
+
+        self.paths.notify_event(event);
```
jrouviere, line 7758: I'm not sure I completely follow what this function is doing. But I ran an AI review which complains that there is an inconsistency here:

> revalidate_pmtu() clears the previously validated PMTU, so get_current_mtu() falls back to 1200 and the PR emits PmtuUpdated { pmtu: 1200 }. However, it never lowers recovery’s max_datagram_size, which is what max_send_udp_payload_size() and normal packet generation use.
A 1400-byte revalidation probe is correct. The problem is that recovery retains 1400 while that probe is unacknowledged, allowing ordinary application packets of that size too—even though the event says only 1200 is validated.
The tests miss this because both the event and PathStats read from Pmtud; they never check the independent recovery/send limit.

kkelkar: recovery is kept at 1400 to allow construction of the probe. 
`send_single()` caps ordinary packets at get_current_mtu() all the same:
```
let mut left = if let Some(pmtud) = path.pmtud.as_mut() {
    cmp::min(pmtud.get_current_mtu(), b.cap())
} else {
    b.cap()
};
```
So I think only the probe is going to use 1400 size. Is that okay?

quiche/src/path.rs
kkelkar: https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

Want me to change?

jrouviere: no that's pretty readable actually, I was just surprised

quiche/src/lib.rs
jrouviere: Ah that makes sense, I got confused by the scope of the size change.

jrouviere (approved):
LGTM

PR #2731, opened 2026-09-14
Count QUIC connection migrations

jrouviere:
Add two u8 to QuicAuditStats to count how many active and passive connection migrations we have observed.


ghedo:
https://github.com/cloudflare/quiche/pull/2720 should allow you to do this I think.

jrouviere:
> #2720 should allow you to do this I think.

Thanks I'll have a look, sounds like a cleaner approach.

jrouviere:
Closing this one, will try to re-use https://github.com/cloudflare/quiche/pull/2720 as a base.

PR #2744, opened 2026-09-18
ffi: populate max_rtt in quiche_conn_path_stats

MelodicAlbuild:
`quiche_conn_path_stats()` fills every field of `quiche_path_stats` except `max_rtt`, so C callers read whatever the memory held before the call. Convert it like `min_rtt`, reporting zero when no RTT sample was taken.


ghedo:
Thanks for the contribution. This PR duplicates #2531, which contains the same max_rtt fix. To keep review and discussion in one place, I am closing this PR in favor of #2531.

PR #2745, opened 2026-09-18
tls: add Config::verify_peer_optional()

MelodicAlbuild:
Requesting the peer's certificate currently requires `verify_peer(true)`, which also makes any chain validation failure fatal. An application that authenticates peers itself, for example by checking the certificate returned by `peer_cert()` against a set of known certificates, has no way to receive it without also configuring a CA able to validate it.

Add `Config::verify_peer_optional()` and the matching FFI function `quiche_config_verify_peer_optional()`. They request the peer's certificate but install a verification callback that accepts any chain, so the handshake succeeds and the application decides what to do with the certificate.

The `verify_client_optional` test mirrors `verify_client_invalid` the server has no CA, yet the handshake succeeds and the client's certificate is available.


quiche/src/lib.rs
```diff
@@ -780,6 +780,23 @@ impl Config {
...
+    /// This overrides any previous call to [`verify_peer()`].
+    ///
+    /// [`peer_cert()`]: struct.Connection.html#method.peer_cert
+    /// [`verify_peer()`]: struct.Config.html#method.verify_peer
+    pub fn verify_peer_optional(&mut self) {
+        self.tls_ctx.set_verify_optional();
+    }
+
     /// Configures whether to do path MTU discovery.
     ///
```
jawadout164-lgtm, line 801: Salam

jawadout164-lgtm: میں

ghedo:
Closing because applications can already configure this through `Config::with_boring_ssl_ctx_builder()`, using `SslContextBuilder::set_custom_verify_callback()` with `SslVerifyMode::PEER` and an accept-all callback.

PR #2753, opened 2026-09-20
Cap the anti-amplification factor at the limit the RFC allows

### LINKED ISSUE #2679: Anti-amplification factor can exceed the 3x limit
# Anti-amplification factor can exceed the 3x limit

## Summary

Confirmed issue, but only for non-default configuration. quiche defaults to the RFC-required factor `3`, yet its public API allows applications to set a larger server anti-amplification factor. With factor `4`, runtime evidence shows the server sends more than three times the bytes received before address validation.

## Standard Requirement

Standards: [RFC 9001 §9.3](https://www.rfc-editor.org/rfc/rfc9001.html#section-9.3), [RFC 9000 §8.1](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1).

```text
if responding to an unverified source address, the server is
forbidden to send more than three times as many bytes as the number
of bytes it has received
```

```text
Prior to validating the client address, servers MUST NOT send more
than three times as many bytes as the number of bytes they have
received.
```

The limit is a protocol maximum, not a recommended default. Implementations may send less, but not more than `3 * received_bytes` before validation.

## Relevant Source Code

`quiche/src/lib.rs:453,680`

```rust
const MAX_AMPLIFICATION_FACTOR: usize = 3;
max_amplification_factor: MAX_AMPLIFICATION_FACTOR,
```

The default configuration is compliant.

`quiche/src/lib.rs:883-887`

```rust
/// Sets the anti-amplification limit factor.
///
/// The default value is `3`.
pub fn set_max_amplification_factor(&mut self, v: usize) {
    self.max_amplification_factor = v;
}
```

The public setter accepts any `usize`; it does not reject or clamp values above `3`.

`quiche/src/lib.rs:2832-2844`

```rust
if self.is_server && !recv_path.verified_peer_address {
    recv_path.max_send_bytes += len * self.max_amplification_factor;
}
```

Before address validation, received bytes are multiplied by the configured factor.

`quiche/src/lib.rs:4024-4028`

```rust
if !send_path.verified_peer_address && self.is_server {
    left = cmp::min(left, send_path.max_send_bytes);
}
```

The send path enforces only the computed budget, so a configured factor above `3` raises the actual send allowance above the RFC limit.

## Implementation Behavior

Default behavior uses factor `3`. However, `quiche_config_set_max_amplification_factor()` and the Rust setter expose this value without a maximum. A caller can configure `4`, and the server will use that larger budget while the peer address is still unverified.

## Inconsistency Reason

[RFC 9001 Section 9.3](https://www.rfc-editor.org/rfc/rfc9001.html#section-9.3) and [RFC 9000 Section 8.1](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1) forbid servers from exceeding `3 * received_bytes` before client address validation. quiche's public configuration can produce a larger pre-validation budget, and runtime evidence confirms that the server can actually send above the 3x bound.

## Runtime Evidence

Record: `tasks/amplification_factor_0068/result.md`.

Focused temporary test:

```rust
config.set_max_amplification_factor(4);
assert!(server_sent > client_sent * 3);
assert!(server_sent <= client_sent * 4);
```

Command:

```text
cargo test -p quiche --release --lib amplification_factor_above_three_probe -- --nocapture
```

Result:

```text
running 1 test
client_sent=1200 server_sent=4579 three_x=3600 configured_x=4800
test tests::amplification_factor_above_three_probe ... ok
test result: ok. 1 passed; 0 failed
```

Existing positive controls:

```text
limit_handshake_data ... 4 passed
custom_limit_handshake_data ... 2 passed
amplification_limited_stat ... 1 passed
```

Pipeline checks:

```text
python scripts/run_all_self_tests.py -> ok: true
python scripts/run_total_test.py -> result: reproduced
```

## Impact

Applications that leave the default untouched stay within 3x. Applications that set a larger factor can make the server exceed the RFC anti-amplification limit before address validation, increasing reflection-amplification risk.

## Fix Direction

Reject or clamp `set_max_amplification_factor(v)` when `v > 3`. If larger internal budgets are needed for tests, keep them behind test-only code rather than public runtime configuration.

Shubham-Padkonde:
Fixes #2679

## Problem

[RFC 9000, Section 8.1](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1):

> Prior to validating the client address, servers MUST NOT send more than three times as many bytes as the number of bytes they have received.

quiche's default is compliant — `MAX_AMPLIFICATION_FACTOR` is 3 — but `Config::set_max_amplification_factor` (and `quiche_config_set_max_amplification_factor` through the FFI) stored whatever it was given, and the send path multiplies received bytes by that value. An application that sets 4 gets a server that answers a 1200-byte Initial with 4579 bytes, which is a limit the RFC states as a MUST NOT, not a default.

## Fix

Cap the configured value at `MAX_AMPLIFICATION_FACTOR`. Smaller values are untouched, so applications that want to be more conservative than the RFC still can — the existing `custom_limit_handshake_data` test, which configures a factor of 2, is unaffected.

## Test

`max_amplification_factor_is_capped` configures factor 4 with `cert-big.crt`, whose handshake flight is larger than three times the client's Initial, so the budget is what bounds the flight.

On `master`:

```
thread 'tests::max_amplification_factor_is_capped' panicked at quiche/src/tests.rs:
server sent 4579 bytes for 1200 bytes received
```

That is a ratio of 3.8. With the change the server stays within the limit.

```
$ cargo test -p quiche --lib
test result: ok. 1141 passed; 0 failed

$ cargo +nightly fmt -p quiche -- --check   # clean
```

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test, lint and reproduction output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ghedo:
Thanks for the report and patch.

The anti-amplification factor is intentionally configurable. quiche’s default remains 3, while applications can explicitly choose a different deployment tradeoff. Silently clamping the public setter would change existing behavior and break configurations that depend on higher values without notifying the caller.

Applications requiring the RFC 9000 limit should retain the default or configure a value no greater than 3.

Closing this PR as not planned.

PR #2755, opened 2026-09-21
Probe every in-flight packet number space on PTO

### LINKED ISSUE #2685: PTO probes only the selected packet number space
# PTO probes only the selected packet number space

## Problem Description

When PTO expires, [RFC 9002 Section 6.2.4](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.4) recommends probing not only the expired packet number space, but also other packet number spaces that still have in-flight data. quiche selects one PTO space and sets `loss_probes` only there, so the other in-flight spaces are not probed on that PTO.

## Standard Requirement

[RFC 9002 Section 6.2.4, Sending Probe Packets](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.4):

```text
In addition to sending data in the packet number space for which the
timer expired, the sender SHOULD send ack-eliciting packets from
other packet number spaces with in-flight data, coalescing packets if
possible.
```

This is a `SHOULD`, but the RFC highlights exactly this case: server Initial+Handshake or client Handshake+Application data can both be in flight, and the peer might have keys for only one space.

## Relevant Source Code

`implementions/quiche/quiche/src/recovery/congestion/recovery.rs:780-802`

```rust
let epoch = if self.bytes_in_flight.get() > 0 {
    let (_, e) = self.pto_time_and_space(handshake_status, now);
    e
} else if handshake_status.has_handshake_keys {
    Epoch::Handshake
} else {
    Epoch::Initial
};

self.pto_count += 1;

let epoch = &mut self.epochs[epoch];
epoch.loss_probes =
    cmp::min(self.pto_count as usize, MAX_PTO_PROBES_COUNT);
```

Legacy recovery chooses one PTO epoch and sets `loss_probes` only for that epoch.

`implementions/quiche/quiche/src/recovery/gcongestion/recovery.rs:933-954`

```rust
let epoch = if self.bytes_in_flight.get() > 0 {
    let (_, e) = self.pto_time_and_space(handshake_status, now);
    e
} else if handshake_status.has_handshake_keys {
    packet::Epoch::Handshake
} else {
    packet::Epoch::Initial
};

self.pto_count += 1;

let epoch = &mut self.epochs[epoch];
epoch.loss_probes = MAX_PTO_PROBES_COUNT.min(self.pto_count as usize);
```

gcongestion uses the same single-space PTO probe selection.

`implementions/quiche/quiche/src/lib.rs:8200-8211`

```rust
if crypto_ctx.data_available() || pkt_space.ready() {
    return Ok(Type::from_epoch(epoch));
}

for (_, p) in self.paths.iter() {
    if p.recovery.has_lost_frames(epoch) {
        return Ok(Type::from_epoch(epoch));
    }

    if p.recovery.loss_probes(epoch) > 0 {
        return Ok(Type::from_epoch(epoch));
    }
}
```

The sender can only select another packet number space for PTO if that space has `loss_probes > 0`; recovery does not set it for the other in-flight spaces.

## Runtime Evidence

Temporary test added to `implementions/quiche/quiche/src/recovery/mod.rs` and removed after execution:

```rust
#[rstest]
fn pto_probes_other_in_flight_packet_number_spaces(
) {

    let mut r = Recovery::new(&cfg);
    let now = Instant::now();
    let handshake_status = HandshakeStatus {
        has_handshake_keys: true,
        peer_verified_address: true,
        completed: false,
    };

    for (pkt_num, epoch) in
        [packet::Epoch::Initial, packet::Epoch::Handshake]
            .into_iter()
            .enumerate()
    {
        let p = Sent {
            pkt_num: pkt_num as u64,
            frames: smallvec![],
            time_sent: now,
            time_acked: None,
            time_lost: None,
            size: 1000,
            ack_eliciting: true,
            in_flight: true,
            delivered: 0,
            delivered_time: now,
            first_sent_time: now,
            is_app_limited: false,
            tx_in_flight: 0,
            lost: 0,
            has_data: true,
            is_pmtud_probe: false,
        };

        r.on_packet_sent(p, epoch, handshake_status, now, "");
    }

    let timeout = r.loss_detection_timer().unwrap();
    r.on_loss_detection_timeout(handshake_status, timeout, "");

    assert!(r.loss_probes(packet::Epoch::Initial) > 0);
    assert!(r.loss_probes(packet::Epoch::Handshake) > 0);
}
```

Command:

```text
cd implementions/quiche/quiche
cargo test --release pto_probes_other_in_flight_packet_number_spaces -- --nocapture
```

Result:

```text
running 2 tests
cc_algorithm_name_1___cubic__ ... FAILED
cc_algorithm_name_2___bbr2_gcongestion__ ... FAILED

assertion failed: r.loss_probes(packet::Epoch::Handshake) > 0
assertion failed: r.loss_probes(packet::Epoch::Handshake) > 0
```

The test puts both Initial and Handshake data in flight, then fires PTO. Both backends leave the Handshake space without a probe.

## Inconsistency Reason

[RFC 9002 Section 6.2.4](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.4) says a sender should also send ack-eliciting probes from other packet number spaces with in-flight data. quiche computes a single PTO space and records probes only in that space. The runtime test confirms the other in-flight space keeps `loss_probes == 0`.

## Impact

During handshake loss recovery, probing only one packet number space can delay progress when the peer can decrypt only the other in-flight space. This is the situation [RFC 9002 Section 6.2.4](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.4) calls out as particularly valuable for multi-space PTO probing.

## Fix Direction

On PTO expiration, after selecting the expired PTO space, also mark other packet number spaces with in-flight data for an ack-eliciting probe when keys and packet construction allow it. If multiple probes can fit, coalesce them where possible; otherwise send them as separate PTO probe datagrams.

Shubham-Padkonde:
Fixes #2685

## Problem

[RFC 9002, Section 6.2.4](https://www.rfc-editor.org/rfc/rfc9002.html#section-6.2.4):

> In addition to sending data in the packet number space for which the timer expired, the sender SHOULD send ack-eliciting packets from other packet number spaces with in-flight data, coalescing packets if possible.

Both recovery implementations pick one space via `pto_time_and_space()` and set `loss_probes` only there:

```rust
let epoch = &mut self.epochs[epoch];
epoch.loss_probes = cmp::min(self.pto_count as usize, MAX_PTO_PROBES_COUNT);
```

and the send path only picks up a space whose `loss_probes` is non-zero, so the other in-flight space is never probed. The RFC calls this out for exactly the handshake case the issue describes: a server with Initial and Handshake data in flight, or a client with Handshake and Application data, where the peer may hold the keys for only one of them.

## Fix

Set the probe count on the other spaces that still have packets in flight, keeping the expired space's own assignment unchanged. Application data stays excluded until the handshake completes, which matches the existing rule in `pto_time_and_space`, and `max` is used so an existing higher probe count is never lowered.

Both the legacy and the gcongestion paths carry the same selection, so both are changed.

## Test

`pto_probes_every_in_flight_packet_number_space` puts one packet in flight in the Initial space and one in the Handshake space, fires the PTO, and asserts both spaces are probed. It runs for `cubic` and `bbr2_gcongestion`.

On `master` both fail:

```
the Handshake space has data in flight and was not probed
test result: FAILED. 0 passed; 2 failed
```

With the change:

```
$ cargo test -p quiche --lib
test result: ok. 1142 passed; 0 failed

$ cargo +nightly fmt -p quiche -- --check   # clean
```

No existing test changed behaviour.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jawadout164-lgtm (commented):
Ok

jawadout164-lgtm (changes requested):
Salam

